PR #35, opened 2022-12-14
Triggers Proof of Concept

abuaboud:


src/worker/webpack.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   },
   entry: path.resolve(__dirname, 'src/index.ts'),
   output: {
+    libraryTarget: 'commonjs2',
```
khaledmashaly: what does this do?

PR #69, opened 2022-12-18
Moving gmail and google sheets connectors

AbdulTheActivePiecer:
google sheet + google email + hubspot + send back to login on authorized response


abuaboud:
Disabled Google Sheets due that interface is needed to represent an array.

PR #86, opened 2022-12-21
Add TypeORM Entities with their relations,indices

abuaboud:


docs/contributing/worker.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+The repeatable tasks are required for flows with schedule trigger.
+
+### Flow Worker
+
+Consumes the flow queue and do the following in this order:
+
+**1. Create Sandbox:**
+
+First we obtain an available sandbox number from queue, we will call isolate binary file to clean and initialize this sandbox number.
```
khaledmashaly, line 38: - What is meant by "obtain an available sandbox number from queue"?
- What is the range of available numbers?
- How do you know that a number is available?
- What is the number of boxes available?
- How is this number calculated?

src/worker/packages/server/src/database/database.connection.ts
```diff
@@ -1,13 +1,13 @@
...
-    password: 'mysecretpassword',
-    database: 'test-typescript-example',
+    password: 'A79Vm5D4p2VQHOp2gd5',
+    database: 'activepieces',
     synchronize: true,
 });
 
-export default AppDataSource;
\ No newline at end of file
+export default DatabaseSource;
```
khaledmashaly: Can we please agree on using named exports instead of default exports? Named exports are better:
- Default exports can be named anything when imported, i.e. `DatabaseSource` may well be imported as `Cat`
- IDEs can suggest or auto-import named imports
- Linters warn when a named import doesn't exist

I think we should add this rule to our conventions doc.

khaledmashaly (commented):
Great! please see my comments. Also, can you please explain why use `KSUID` for IDs instead of the industry standard `UUID`? This seems like an unnecessary complexity which needs justification.

src/worker/packages/server/src/entity/user-entity.ts
```diff
@@ -1,8 +1,12 @@
 import {EntitySchema} from "typeorm"
-import {User} from "shared/dist";
+import {Project, User} from "shared/dist";
```
khaledmashaly, line 2: Seems like importing from `shared` directly doesn't work. Also importing from `shared/dist` seems to need a rebuild for every change. Can you please address this?

src/worker/packages/server/src/database/database.connection.ts
abuaboud: Done

docs/contributing/worker.mdx
abuaboud: Updated on main branch, I commit it here by mistake

src/worker/packages/server/src/entity/user-entity.ts
abuaboud: I edited the README in root directory, for build + install packages command line, I will add detailed instruction once we finish the backend.

PR #109, opened 2022-12-25
Add PR template

mabumusa1:
This PR will be the template to be used in new PRs, so when someone creates a new PR the content of this PR will be rendered for the PR issuer. 

The content of this document is to be agreed on, as it will affect the versioning system of the product


abuaboud:
Thank you, done and pushed to the main :dancers:

PR #110, opened 2022-12-25
Add issue template

mabumusa1:
I like PR #96, this template is making the bug reporting as form that the reporter fills, I suggest we modify this PR and PR #96 to provide a seamless bug reporting experience

For the file `.github/config.yml`, it provides and experience like this https://github.com/mautic/mautic/issues/new/choose

I think it is a good idea to separate bugs, from feature requests, to new ideas. etc


abuaboud:
Thank you for suggestion, I created simple templates, closing the PR :dancers:

PR #113, opened 2022-12-25
WIP: Security Policy is required 

mabumusa1:
This is a copy of the security policy of what we have in Mautic, this document needs to be updated to match your polices


abuaboud:
Thank you, Done in https://github.com/activepieces/activepieces/pull/399

PR #130, opened 2022-12-28
Feature piece triggers

abuaboud:


src/worker/packages/backend/src/instance/instance-side-effects.ts
```diff
@@ -0,0 +1,48 @@
...
+            return;
+        }
+
+        const flowVersionIds = Object.values(instance.flowIdToVersionId);
+
+        const flowVersions = await flowVersionRepo.findBy({
+            id: In(flowVersionIds),
+        });
+
+        const disableTriggers = flowVersions.map(triggerUtils.enable);
```
abuaboud: I think this is need to be disable

src/worker/packages/backend/src/instance/instance-side-effects.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        await Promise.all(disableTriggers);
+    }
+};
+
+const lockFlowVersions = async (flowVersions: FlowVersion[]): Promise<void> => {
```
abuaboud: can we please lock collection version as well

khaledmashaly: yes see 6e0b91b

src/worker/packages/backend/src/instance/instance-side-effects.ts
khaledmashaly: oops good catch see 4f45c19

john-jova:
Phase 3 complete — merged to main. All VM tests pass.

PR #140, opened 2022-12-30
Moved frontend package into packages

abuaboud:


john-jova:
Fixed in f175459a4c. Shadow rows for project-scoped + onRequest hooks for platform-scoped sets. Verified on clean VM.

PR #241, opened 2023-01-09
App auth

abuaboud:


packages/backend/src/app-connection/app-connection-entity.ts
```diff
@@ -0,0 +1,46 @@
...
+    {
+      name: "idx_app_connection_project_id_and_app_name_and_name",
+      columns: ["projectId", "appName", "name"],
+      unique: true,
+    },
+    {
+      name: "idx_app_connection_project_id_and_name",
+      columns: ["projectId", "name"],
+      unique: true,
+    },
```
khaledmashaly, line 32: seems like the unique constraint on `["projectId", "appName", "name"]` is redundant.

packages/backend/src/app-connection/app-connection-service.ts
```diff
@@ -0,0 +1,162 @@
...
+import { paginationHelper } from "../helper/pagination/pagination-utils";
+import { AppConnectionEntity } from "./app-connection-entity";
+import axios, { AxiosError } from "axios";
+import qs from "qs";
+
+const appConnectionRepo = databaseConnection.getRepository(AppConnectionEntity);
+
+export const appConnectionService = {
+    async upsert(request: UpsertConnectionRequest): Promise<AppConnection> {
+        await appConnectionRepo.upsert({ ...request, id: apId() }, ["name", "projectId"]);
```
khaledmashaly, line 12: FYI, upsert docs says: `Unlike save method executes a primitive operation without cascades, relations and other operations included.`

packages/backend/src/app-connection/app-connection-service.ts
```diff
@@ -0,0 +1,162 @@
...
+    const response = (
+        await axios.post(
+            settings.token_url,
+            qs.stringify({
+                client_id: settings.client_id,
+                client_secret: settings.client_secret,
+                redirect_uri: settings.redirect_url,
+                grant_type: "refresh_token",
+                refresh_token: appConnection.refresh_token,
+            }),
```
khaledmashaly, line 131: FYI, `qs` is deprecated, `URLSearchParams` is the new API.

packages/backend/src/app.ts
```diff
@@ -61,6 +64,7 @@ const start = async () => {
       host: "0.0.0.0",
       port: 3000,
     });
+    await databaseConnection.runMigrations();
```
khaledmashaly: I think migrations should run before starting the application, right? What happens if a query is performed while migrations are running?

packages/backend/src/oauth2/oauth2.service.ts
```diff
@@ -3,12 +3,8 @@ import qs from "qs";
...
   ClaimTokenWithSecretRequest,
-  CloudOAuth2Config,
-  ConfigType,
-  OAuth2Config,
-  OAuth2Response,
-  RefreshTokenFromCloudRequest,
 } from "shared";
+import { formatOAuth2Response } from "../app-connection/app-connection-service";
 
 export const oauth2Service = {
```
khaledmashaly, line 8: Is this service used anymore?

abuaboud: Yes It's used in oauth2 controller

packages/backend/src/app-connection/app-connection-service.ts
abuaboud: Removed thanks

packages/backend/src/app.ts
abuaboud: Changed to databaseModule, thank you for catching up this

packages/backend/src/app-connection/app-connection-entity.ts
abuaboud: This is needed as index rather constraint, since we list per these three variables.

PR #247, opened 2023-01-10
Release 0.1.1

abuaboud:
Hot fix:

- Redirect URL is correct (there used to be extra /api)
- Webhook used to running latest collection version instead of published one.


abuaboud:
Merged hot fixes

PR #264, opened 2023-01-11
docs(docker): upgrading from previous version

abuaboud:


ashrafsamhouri (approved):
It'd be even better if you write "Docker" instead of "docker" in the future!

PR #295, opened 2023-01-12
[Pieces] [Feature] Secret text props

khaledmashaly:


packages/pieces/src/apps/cloud-vision/actions/get-text-from-pdf.ts
```diff
@@ -34,6 +34,7 @@ export const convertPdfToText = createAction({
     serviceAccountCredentials: Property.LongText({
       displayName: 'Service Account Credentials',
       required: true,
+      secret: true,
```
abuaboud, line 37: I do believe the use case for this is only connection secrets, and so the connections should have separate props and they are rendered secrets by default.

What do you think?

khaledmashaly: Yes, fair enough. Will close this PR.

PR #298, opened 2023-01-12
Feature: custom auth

khaledmashaly:


packages/pieces/src/framework/property/oauth-prop.ts
```diff
@@ -0,0 +1,21 @@
+import { BasePropertySchema, PropertyType, TPropertyValue } from "./base-prop";
+import { AuthProp } from "./custom-auth-prop";
+
+type OAuthUrlResolver = (propsValue: Record<string, AuthProp>) => Promise<string>;
```
abuaboud: The counter use case for this is salesforces.

Salesforces requires a static dropdown where you select sandbox or production and depend on that calculate the url value, I guess this should be any prop value.

packages/pieces/src/framework/property/property.ts
```diff
@@ -0,0 +1,54 @@
...
+	Number(request: BasePropertySchema): NumberProperty {
+		return {...request, valueSchema: undefined, type: PropertyType.NUMBER};
+	},
+	OAuth2(request: OAuth2PropertySchema): OAuth2Property {
+		return {...request, valueSchema: undefined, type: PropertyType.OAUTH2};
+	},
+	Dropdown<T>(request: DropdownPropertySchema<T>): DropdownProperty<T> {
+		return {...request, valueSchema: undefined, type: PropertyType.DROPDOWN};
+	},
+	ApiKey(request: BasePropertySchema): SecretTextProperty {
```
abuaboud: Is this intended? to be api key not secret key?

abuaboud (approved):

packages/pieces/src/framework/property/property.ts
khaledmashaly: ouch

PR #327, opened 2023-01-19
Pipedrive piece

ashrafsamhouri:
Here is how I tested the piece:

- Add Person action by Test Flow, you can see the output:
Screenshot 2023-01-20 at 00 20 16

- Created flows using all the triggers and published them
Screenshot 2023-01-20 at 00 20 33

- Went to Pipedrive and performed actions to trigger the flows
Screenshot 2023-01-20 at 00 20 45

- This is an example flow ran successfully with its output
Screenshot 2023-01-20 at 00 21 06


ashrafsamhouri:
@abuaboud It looks like apps folder has gone inside lib? Can you help with the conflicts?

PR #330, opened 2023-01-21
Pipedrive

abuaboud:
On behalf of @ashrafsamhouri 

Here is how I tested the piece:

- Add Person action by Test Flow, you can see the output:
Screenshot 2023-01-20 at 00 20 16

- Created flows using all the triggers and published them
Screenshot 2023-01-20 at 00 20 33

- Went to Pipedrive and performed actions to trigger the flows
Screenshot 2023-01-20 at 00 20 45

- This is an example flow ran successfully with its output
Screenshot 2023-01-20 at 00 21 06


abuaboud:
@ashrafsamhouri Could you please add sample data for all actions and triggers, you can do that by filling `sample_data` field in object or action, this would help us provide auto complete for people.

ashrafsamhouri:
@abuaboud I added all sample data and tested them with the autocomplete UI.

PR #346, opened 2023-01-23
Feature: Mailchimp new subscriber trigger and others

khaledmashaly:
- Backend and engine now work when using docker compose dev
- Solve the webpack global issue
- enabled form body parsing for fastify
- created mail chimp subscribe trigger


packages/pieces/src/lib/apps/mailchimp/common/index.ts
```diff
@@ -1,24 +1,76 @@
...
+  const access_token = authProp.access_token;
+  const mailChimpServerPrefix = await getMailChimpServerPrefix(access_token!);
+  mailchimp.setConfig({
+    accessToken: access_token,
+    server: mailChimpServerPrefix
+  });
+
+  console.log(`token:${access_token}`);
+  console.log(`server ${mailChimpServerPrefix}`);
+
```
abuaboud, line 53: Let's not log an access token there (as it's sensitive data), We will start encrypting the one in the database later.

abuaboud (approved):

packages/pieces/src/lib/apps/mailchimp/common/index.ts
khaledmashaly: lol copy pasting is never a good idea 😂

PR #349, opened 2023-01-24
fixed all comments in pr

AbdulTheActivePiecer:
1)Now the auto complete only closes on the click of the close button and clicking on another control that uses auto complete.
2)Added search bar for steps list.
3)Hidden the replace trigger/remove step in readonly mode (Runs).
4)The tooltip for fullscreen on the code editor now shows once the code step is clicked so people can notice it.
5)The trigger sample data gets sent to test flow when it is an app trigger, if no sample data is provided we send an empty object.

### DIFF
--- a/packages/backend/src/app/webhooks/webhook-service.ts
+++ b/packages/backend/src/app/webhooks/webhook-service.ts
@@ -1,31 +1,39 @@
-import { CollectionId, FlowId, Instance, RunEnvironment, TriggerType } from "@activepieces/shared";
-import { collectionService } from "../collections/collection.service";
-import { flowRunService } from "../flow-run/flow-run-service";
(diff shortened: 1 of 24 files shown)


packages/backend/src/app/webhooks/webhook-service.ts
```diff
@@ -1,31 +1,39 @@
...
+    );
+    const payloads: any[] = await triggerUtils.executeTrigger({
       projectId: collection.projectId,
       collectionVersion: collectionVersion,
       flowVersion: flowVersion,
       payload: payload,
     });
 
-
-    console.log(`Triggers returned a ${payloads.length} payloads`);
```
abuaboud, line 27: Get this back and use logger.info

abuaboud (approved):

PR #393, opened 2023-01-27
feat: authorization for project resources

abuaboud:


packages/backend/src/app/authentication/token-verify-middleware.ts
```diff
@@ -26,6 +26,10 @@ export const tokenVerifyMiddleware = async (request: FastifyRequest, _reply: Fas
     try {
       const token = rawToken.substring(HEADER_PREFIX.length);
       const principal = await tokenUtils.decode(token);
+      // Todo remove in the future, as the old JWT token doesn't contain project id.
```
khaledmashaly, line 29: What is meant by this comment?

khaledmashaly (dismissed):

packages/backend/src/app/authentication/token-verify-middleware.ts
abuaboud: Currently JWT token doesn't have project id in claims, So this make them 401 so the frontend force them to logout.

PR #402, opened 2023-01-29
fix: run instance side effect on flow deletion

abuaboud:
## What does this PR do?

If the flow is deleted, the published flow side effects will not run for their triggers, the expected behavior to run the side effects on deleted flow.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [X] Create Flow with Schedule -> Publish -> Delete the Flow -> Schedule should be deleted.

### DIFF
--- a/packages/backend/src/app/workers/flow-worker/flow-queue.ts
+++ b/packages/backend/src/app/workers/flow-worker/flow-queue.ts
@@ -3,6 +3,7 @@ import { ApId } from "@activepieces/shared";
 import { createRedisClient } from "../../database/redis-connection";
 import { ActivepiecesError, ErrorCode } from "@activepieces/shared";
 import { OneTimeJobData, RepeatableJobData } from "./job-data";
+import { logger } from "packages/backend/src/main";
 
 interface BaseAddParams {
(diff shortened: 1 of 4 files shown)


packages/backend/src/app/workers/flow-worker/flow-queue.ts
```diff
@@ -70,24 +71,22 @@ export const flowQueue = {
...
+        await client.del(repeatableJobKey(id));
+
+        if (!result) {
+          throw new ActivepiecesError({
+            code: ErrorCode.JOB_REMOVAL_FAILURE,
+            params: {
+              jobId: id,
+            },
+          });
+        }
```
khaledmashaly, line 89: Could this possibly leave the system with a lot of zombie jobs?

Throwing when encountering anomalies is a safe action to ensure the system is not left in an inconsistent state.

abuaboud: Let's keep the discussion there: 
https://github.com/activepieces/activepieces/issues/423

Issue #68, opened 2022-12-18
Google Calendar New Event or Updated Trigger


khaledmashaly:
Trigger on Google Calendar when new event or updated (except deleted) is created.

Property:
- Google OAuth2

Acceptance Criteria:
- Prefer Webhooks over Polling
- Sample Data is provided.

OsamaHaikal:
test

abuaboud:
@OsamaHaikal Now I can assigned it to you finally, Thank you for contributing for activepieces

PR #452, opened 2023-02-02
Piece/i 68 add google calendar piece

OsamaHaikal:
## What does this PR do?

# Issue:  https://github.com/activepieces/activepieces/issues/68

A new piece has been added:
- Name: google calendar
- Actions: Quick add event
- triggers: Event updated or created or  deleted
- Trigger type is webhook
- on-enable: a watch request to google resource is sent
notes: with each publish, a new channel is generated to watch the event so to avoid having multiple watchers at the same time, we stop the old channel and then create a new watcher
-run: Google APIs don't return the events in descending order so the only way was for me to get the last event modified to do multiple requests till getting to the last page and then get the last element

## Type of change

## How should this be tested?

1. In the Google Cloud console, enable the Google Calendar API.
2. Click In the Google Cloud console, and go to Menu menu > APIs & Services > Credentials.
3. Click Create Credentials > OAuth client ID.
4. In the Name field, type a name for the credential. This name is only shown in the Google Cloud console.
5. Click Create. The OAuth client created screen appears, showing your new Client ID and Client secret.
6. Click OK. The newly created credential appears under OAuth 2.0 Client IDs.
7. ngrok http 300
8. setup new google calendar trigger/action 
9. publish

### DIFF
--- a/docs/pieces/apps/google_calendar.mdx
+++ b/docs/pieces/apps/google_calendar.mdx
@@ -1,0 +1,36 @@
+---
+title: "Google Calendar"
+description: ""
+---
+
+## Set up and run an app that calls a Google calendar API.
+
+1. In the Google Cloud console, enable the Google Calendar API.
+2. Click In the Google Cloud console, and go to Menu menu > APIs & Services > Credentials.
+3. Click Create Credentials > OAuth client ID.
+4. In the Name field, type a name for the credential. This name is only shown in the Google Cloud console.
+5. Click Create. The OAuth client created screen appears, showing your new Client ID and Client secret.
+6. Click OK. The newly created credential appears under OAuth 2.0 Client IDs.
+---
+
+
+## Triggers
+
+
+    <CardGroup cols={2}>
+      <Card title="New or Updated or Deleted Event">
+        Triggers when there is a new or updated or deleted Event for selected calendar.
+      </Card>
+    </CardGroup>
+
+
+---
+
+## Actions
+
+
+    <CardGroup cols={2}>
+      <Card title="Add Quick Event">
+        Add a Quick Event to a Google calendar
+      </Card>
+    </CardGroup>
--- a/packages/pieces/src/lib/apps/google-calendar/actions/create-quick-event.ts
+++ b/packages/pieces/src/lib/apps/google-calendar/actions/create-quick-event.ts
@@ -12,13 +12,13 @@ export const createQuickCalendarEvent = createAction({
   displayName: 'Quick Event',
   props: {
     authentication: googleCalendarCommon.authentication,
-    calendarId: googleCalendarCommon.calendarDropdown,
+    calendar_id: googleCalendarCommon.calendarDropdown,
     text: Property.LongText({
       displayName: 'Summary',
       description: 'The text describing the event to be created',
       required: true,
     }),
-    sendUpdates: Property.Dropdown<string>({
+    send_updates: Property.Dropdown<string>({
       displayName: "Send Updates",
       description: "Guests who should receive notifications about the creation of the new event.",
       refreshers: [],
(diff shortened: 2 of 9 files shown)


packages/pieces/src/lib/apps/google-calendar/triggers/calendar-event.ts
```diff
@@ -0,0 +1,71 @@
...
+  TriggerStrategy,
+} from '../../../framework/trigger/trigger';
+import { googleCalendarCommon } from '../common';
+import { getLatestEvent, stopWatchEvent, watchEvent } from '../common/helper';
+import { GoogleWatchResponse } from '../common/types';
+
+export const calendarEventUpdatedOrCreatedOrDeleted = createTrigger({
+  // docs: https://developers.google.com/calendar/api/guides/push
+  name: 'new_event_added',
+  displayName: 'New Event',
```
OsamaHaikal: I think wording should be changed

packages/pieces/src/lib/apps/google-calendar/triggers/calendar-event.ts
```diff
@@ -0,0 +1,71 @@
...
+    const googleChannel = await context.store?.get<GoogleWatchResponse>('_trigger');
+    if (googleChannel?.id) await stopWatchEvent(googleChannel, authProp);
+    await context.store?.put<GoogleWatchResponse | object>('_trigger', {});
+  },
+  async run(context) {
+    const authProp: OAuth2PropertyValue = context.propsValue[
+      'authentication'
+    ] as OAuth2PropertyValue;
+    const event = await getLatestEvent(
+      context.propsValue['calendarId']!,
```
abuaboud: I think there is better way to do it, Google should at least return the event id in the webhook payload, we can use it to fetch the information for the event by id.

What do you think?

packages/pieces/src/lib/apps/google-calendar/actions/create-quick-event.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Property } from '../../../framework/property';
+import { googleCalendarCommon } from '../common';
+
+export const createQuickCalendarEvent = createAction({
+  name: 'create_quick_event',
+  description: 'Add Quick Calendar Event',
+  displayName: 'Quick Event',
+  props: {
+    authentication: googleCalendarCommon.authentication,
+    calendarId: googleCalendarCommon.calendarDropdown,
```
abuaboud: Let use the snake case in the props because this will show up in the input and output part in the frontend, so we try to keep it consistent.

docs/pieces/apps/google_calendar.mdx
```diff
@@ -0,0 +1,36 @@
...
+---
+
+## Set up and run an app that calls a Google calendar API.
+
+1. In the Google Cloud console, enable the Google Calendar API.
+2. Click In the Google Cloud console, and go to Menu menu > APIs & Services > Credentials.
+3. Click Create Credentials > OAuth client ID.
+4. In the Name field, type a name for the credential. This name is only shown in the Google Cloud console.
+5. Click Create. The OAuth client created screen appears, showing your new Client ID and Client secret.
+6. Click OK. The newly created credential appears under OAuth 2.0 Client IDs.
```
abuaboud: I just noticed this is not mentioned in the docs, pieces docs get auto generated form the code.

But since you have this custom part for authentication check discord folder (discord.mdx) and create similar file and we can remove this file

abuaboud (dismissed):

packages/pieces/src/lib/apps/google-calendar/triggers/calendar-event.ts
abuaboud: Oh I see the issue there, google sent the information in the headers rather than body.

I think we need to modify on our core, I will leave this comment there as I will be working on fix for that.

packages/pieces/src/lib/apps/google-calendar/triggers/calendar-event.ts
abuaboud: I think it should be changed or `new_or_updated_event` , since it capture everything from add / update and deleting

abuaboud (approved):

abuaboud:
@all-contributors please add @OsamaHaikal for plugin

PR #455, opened 2023-02-02
feat: expose headers in the webhook payload as well

abuaboud:
## What does this PR do?
separated the context in Action Context and Trigger Context.
Payload for webhook should include headers

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/packages/backend/src/app/webhooks/webhook-controller.ts
+++ b/packages/backend/src/app/webhooks/webhook-controller.ts
@@ -15,7 +15,10 @@ export const webhookController: FastifyPluginCallback = (app, _opts, done): void
     async (request: FastifyRequest<{ Querystring: WebhookQueryParams }>, reply) => {
       await webhookService.callback({
         flowId: request.query.flowId,
-        payload: request.body,
+        payload: {
+          headers: request.headers,
+          body: request.body
+        },
       });
 
       await reply.status(StatusCodes.OK).send();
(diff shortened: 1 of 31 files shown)


packages/pieces/src/lib/framework/context.ts
```diff
@@ -1,13 +1,17 @@
 
+export interface TriggerContext<T> {
+    payload: any;
```
khaledmashaly: I think it's better to use `unknown` instead of `any`

packages/backend/src/app/webhooks/webhook-controller.ts
```diff
@@ -15,7 +15,10 @@ export const webhookController: FastifyPluginCallback = (app, _opts, done): void
     async (request: FastifyRequest<{ Querystring: WebhookQueryParams }>, reply) => {
       await webhookService.callback({
         flowId: request.query.flowId,
-        payload: request.body,
+        payload: {
+          headers: request.headers,
+          body: request.body
+        },
```
khaledmashaly, line 22: why not pass query params as well?

khaledmashaly (dismissed):

packages/backend/src/app/webhooks/webhook-controller.ts
abuaboud: Done

PR #492, opened 2023-02-07
Issue 444 Basic Auth

AbdulTheActivePiecer:
## What does this PR do?

Added Basic Auth for frontend

## How should this be tested?

Add a basic auth property in a piece action and test it in the front end for both creating and editing.


packages/frontend/src/app/modules/common/components/configs-form/connector-action-or-config.ts
```diff
@@ -25,6 +26,10 @@ export interface PieceConfig {
 	required: boolean;
 	extra?: Record<string, unknown>;
 	refreshers?: string[];
+	basicAuthConfigs?: {
+		password: Pick<PieceProperty, "required" | "displayName" | "description">,
```
abuaboud: Please remove required, I did remove it in this pr because it has always to be true.

packages/frontend/.eslintrc.json
```diff
@@ -9,6 +9,7 @@
         "rxjs/no-nested-subscribe": "error",
         "rxjs-angular/prefer-async-pipe": "error",
         "@typescript-eslint/adjacent-overload-signatures": "off",
+        "@typescript-eslint/no-non-null-assertion": "off",
```
abuaboud, line 12: Why this must be off?

abuaboud (changes requested):

abuaboud (approved):

PR #493, opened 2023-02-07
fix: limit sandbox run time

### LINKED ISSUE #470: Separate .env from docker-compose
When pull the new changes, it won't cause conflicts with docker-compose, since we change versions.

khaledmashaly:
## What does this PR do?

- feat: Move sandbox run time to an env var
- feat: Set sandbox run time to 2 minutes if env var is not defined
- chore: sort env vars alphabetically
- chore: use `.env` file for docker compose instead of inline environment (closes #470)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)


docker-compose.yml
```diff
@@ -11,18 +11,7 @@ services:
     depends_on:
       - postgres
       - redis
-    environment:
```
abuaboud, line 14: Are you sure this works?

I thought you would have to do something like
AP_FRONTEND_URL=${AP_FRONTEND_URL}

abuaboud (approved):

docker-compose.yml
khaledmashaly: yes this works, there's an `env_file` argument that points to the folder, you can test with a quick `docker compose up`

Issue #360, opened 2023-01-25
Delegating Property Values (Dropdown, Connection, Object, Array)


abuaboud:
# Use Case

Some inputs type in apps cannot taken from variable or previous step, (Dropdown, Connection, Object and Array).

# Solution

To have option near each input to take it directly from auto complete menu.

PR #547, opened 2023-02-12
Delegating Property Values

AbdulTheActivePiecer:
## What does this PR do?

Allows people to assign dropdown, object, array and JSON piece properties to dynamic values.
Fixes # (issue)
1) Addresses #360 
2) Disables piece trigger forms and full screen editor in read only mode.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

## How should this be tested?
Add a piece action -> toggle an input to a dynamic value -> input your value -> test flow and make sure the value passed was correct.


packages/pieces/framework/src/lib/apps/asana/common/index.ts
```diff
@@ -21,6 +21,7 @@ export const asanaCommon = {
         required: true,
         refreshers: ['authentication'],
         options: async (value) => {
+            console.debug("test");
```
abuaboud: Remove this

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -43,10 +42,10 @@ export type PieceActionSettings = {
   pieceName: string;
   actionName: string | undefined;
   input: Record<string, unknown>;
+  customizedInputs: Record<string, boolean>
```
abuaboud: Can you make it something llike `renderInfo` and then place customize inside it's object, that way you can add more render information in the future.

tsconfig.json
```diff
@@ -0,0 +1,38 @@
+{
```
abuaboud: This is just suppose to be template tsconfig.base.json, why it's added there?

packages/shared/src/lib/flows/actions/action.ts
abuaboud: I think this should be added to piece triggers as well

abuaboud (changes requested):

abuaboud (approved):

PR #555, opened 2023-02-13
chore: refactor properties

abuaboud:
## What does this PR do?
1- context.propsValue[KEY] could be only undefined, If required is false.

Fixes # (issue)

## Type of change

### DIFF
--- a/packages/pieces/framework/src/lib/apps/google-calendar/actions/create-quick-event.ts
+++ b/packages/pieces/framework/src/lib/apps/google-calendar/actions/create-quick-event.ts
@@ -18,7 +18,7 @@ export const createQuickCalendarEvent = createAction({
       description: 'The text describing the event to be created',
       required: true,
     }),
-    send_updates: Property.Dropdown<string, false>({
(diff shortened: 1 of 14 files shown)


packages/pieces/framework/src/lib/apps/google-calendar/actions/create-quick-event.ts
```diff
@@ -18,7 +18,7 @@ export const createQuickCalendarEvent = createAction({
       description: 'The text describing the event to be created',
       required: true,
     }),
-    send_updates: Property.Dropdown<string>({
+    send_updates: Property.Dropdown<string, false>({
```
abuaboud: Optional Dropdown will require to specficy the second argument now, @khaledmashaly not sure If we should consider other alternative like

Property.Optional({})

What do you think?

abuaboud: I just found this could be avoided by making default value is boolean instead of true :) :) :)

khaledmashaly (approved):

PR #573, opened 2023-02-14
chore: customize nginx backend url

abuaboud:
## What does this PR do?

To allow users to decouple the frontend from the backend, you can provide them with the ability to customize the backend URL within the frontend. This can be accomplished by creating a configuration or settings file within the frontend that allows users to enter the URL of the backend server.

## Type of change

## How should this be tested?

- [X] Production, Frontend URL is undefined -> use ip instead and append /api
- [X] Production, Frontend URL is localhost -> use ip instead and append /api
- [X] Development, Backend url is loccalost -> keep it as localhost


docker-compose.yml
```diff
@@ -23,6 +23,9 @@ services:
     container_name: activepieces-frontend
     restart: unless-stopped
     depends_on: [ backend ]
+    environment:
+      - AP_BACKEND_URL=${AP_BACKEND_URL}
+      - NGINX_ENVSUBST_OUTPUT_DIR=/etc/nginx
```
khaledmashaly, line 28: This environment variable doesn't seem to be used anywhere, no?

packages/backend/src/app/flags/flag.service.ts
```diff
@@ -14,81 +16,97 @@ export const flagService = {
...
   },
   async getAll(): Promise<Flag[]> {
     const flags = await flagRepo.find({});
-
     const now = new Date().toISOString();
     const created = now;
     const updated = now;
-
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const currentVersion = require('../../../../../package.json').version;
```
khaledmashaly: I think it's a bad idea to use `import` and `require` in the same file, btw there's a way to import JSON files using `import`, see https://www.typescriptlang.org/tsconfig#resolveJsonModule

packages/backend/src/app/flags/flag.service.ts
```diff
@@ -14,81 +16,97 @@ export const flagService = {
...
     return flags;
   },
+  async getLatestPackageDotJson() {
+    try {
+      const pkgJson = (await axios.get("https://raw.githubusercontent.com/activepieces/activepieces/main/package.json")).data;
+      return pkgJson;
+    } catch (ex) {
+      return { version: '0.0.0' }
+    }
+  }
```
khaledmashaly, line 98: I would move all `version` related stuff to another service.

packages/frontend/nginx.conf
```diff
@@ -8,7 +8,7 @@ http {
         index index.html;
 
         location /api/ {
-          proxy_pass http://backend:3000/;
+          proxy_pass ${AP_BACKEND_URL}/;
```
khaledmashaly, line 11: How does this environment variable get replaced?

docker-compose.yml
abuaboud: It's used inside the nginx config

packages/frontend/nginx.conf
abuaboud: That's already built-in inside nginx, anything in template each time the docker run it will substitute and create new configs.

khaledmashaly (approved):

PR #582, opened 2023-02-15
feat: flow worker concurrency limit

abuaboud:
## What does this PR do?

One slow flow willl slow down all other flows execution, increasing the concurrency for consuming queue, the default is 1.

Fixes # (issue)

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)


packages/backend/src/app/helper/system/system.ts
```diff
@@ -6,6 +6,14 @@ export const system = {
     return getEnvVar(prop);
   },
 
+  getNumber(prop: SystemProp): number | undefined {
+    const number = getEnvVar(prop);
+    if(number === undefined) {
+      return undefined;
+    }
+    return Number(number);
```
khaledmashaly: Please note that this line could return a lot of unexpected results, for example:
- `Number('ab')` returns `NaN`
- `Number('0x10')` returns `16`, i.e. parsed as hexadecimal
- `Number('')` returns `0`

Here's one way how this could be handled
```ts
getNumber(prop: SystemProp): number | null {
  const stringNumber = getEnvVar(prop);

  if (!stringNumber) {
    return null;
  }

  const parsedNumber = Number.parseInt(stringNumber, 10);

  if (Number.isNaN(parsedNumber)) {
    return null;
  }

  return parsedNumber;
}
```

sorry, I couldn't resist implementing an example!

khaledmashaly (approved):

PR #587, opened 2023-02-16
chore: credentials

abuaboud:
## What does this PR do?

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Run in development mode with existing credentials
- [ ] Run in production mode with existing credentials


packages/backend/src/app/helper/encryption.ts
```diff
@@ -0,0 +1,32 @@
...
+const encryptionKey = system.getOrThrow(SystemProp.ENCRYPTION_KEY);
+
+export interface EncryptedObject {
+    iv: string;
+    data: string;
+}
+
+export function encryptObject(object: unknown): EncryptedObject {
+    const iv = crypto.randomBytes(16); // Generate a random initialization vector
+    const key = Buffer.from(encryptionKey.padEnd(32, '\0'), 'binary'); // Create a key from the encryption key
```
khaledmashaly: I think padding the encryption key with nulls is a bad idea. This could allow the key to be an empty string which kind of defeats the purpose of encryption.

A better option in my opinion is to validate that the key's length is 256 bits at the start and throw if this is not the case.

packages/backend/src/app/helper/encryption.ts
```diff
@@ -0,0 +1,32 @@
...
+        iv: iv.toString('hex'),
+        data: encrypted
+    };
+}
+
+
+export function decryptObject<T>(encryptedObject: EncryptedObject): T {
+    const iv = Buffer.from(encryptedObject.iv, 'hex');
+    const key = Buffer.from(encryptionKey.padEnd(32, '\0'), 'binary'); // Create a key from the encryption key
+    const decipher = crypto.createDecipheriv('aes-256-cbc', key, iv);
```
khaledmashaly: Moving common constant values like algorithm name and iv length to module-level variables will be a good idea.

packages/backend/src/main.ts
```diff
@@ -23,6 +23,7 @@ import { system } from "./app/helper/system/system";
 import { SystemProp } from "./app/helper/system/system-prop";
 import { databaseConnection } from "./app/database/database-connection";
 import { logger } from './app/helper/logger';
+import { encryptObject } from "./app/helper/encryption";
```
khaledmashaly: An unused import!

packages/backend/src/app/database/database-connection.ts
```diff
@@ -50,10 +51,10 @@ const getMigrations = () => {
     return [
       FlowAndFileProjectId1674788714498,
       initializeSchema1676238396411,
+      encryptCredentials1676505294811
     ];
   }
-
-  return [];
+  return [encryptCredentials1676505294811];
```
khaledmashaly: what is the reason for running this migration in development?

khaledmashaly (commented):
I think it will be a good idea to add a command that auto-generates the encryption key to the `deploy.sh` script.

packages/backend/src/app/database/database-connection.ts
abuaboud: Data migration, there is no schema changes.

PR #636, opened 2023-02-20
feat: added multiselect dropdown

AbdulTheActivePiecer:
## What does this PR do?

Makes a multiselect dropdown available as a property for pieces

## Type of change

## How should this be tested?
Have an action within a piece that's of this type, make sure refreshers work and values are saved as an array.


AbdulTheActivePiecer:
@abuaboud I can't get the multipleDropdown property value to be of type array it is always a single value, it takes the value's type from the dropdown state I think.

AbdulTheActivePiecer:
also for some reason changes made on pieces don't get detected unless I restart the server by exiting the terminal and re running nx serve backend

packages/pieces/framework/src/lib/framework/property/dropdown-prop.ts
```diff
@@ -14,11 +14,16 @@ export type DropdownOption<T> = {
 	value: T;
 };
 
-export type DropdownProperty<T, R extends boolean> =  BasePropertySchema & {
+export type DropdownProperty<T, R extends boolean> = BasePropertySchema & {
 	refreshers: string[];
-	options: (propsValue: Record<string, OAuth2PropertyValue | number | string | DropdownState<any> | BasicAuthPropertyValue>) => Promise<DropdownState<T>>
+	options: (propsValue: Record<string, OAuth2PropertyValue | number | string | object | BasicAuthPropertyValue | any[]>) => Promise<DropdownState<T>>
```
abuaboud: Revert back DropdownState<any> changes, and You need MultiDropdownState<any>

packages/pieces/framework/src/lib/framework/property/dropdown-prop.ts
```diff
@@ -14,11 +14,16 @@ export type DropdownOption<T> = {
...
 } & TPropertyValue<T, PropertyType.DROPDOWN, R>;
 
 export type StaticDropdownProperty<T, R extends boolean> = BasePropertySchema & {
 	options: DropdownState<T>;
 } & TPropertyValue<T, PropertyType.STATIC_DROPDOWN, R>;
+
+export type MultipleDropdownProperty<T, R extends boolean> = BasePropertySchema & {
+	refreshers: string[];
+	options: (propsValue: Record<string, OAuth2PropertyValue | number | string | object | BasicAuthPropertyValue | any[]>) => Promise<DropdownState<T>>
```
abuaboud: You need to create DropdownState<T> since it's return different type

abuaboud: Remove object and any[] and replace it with right props

PR #664, opened 2023-02-22
Configure Redis using a URL

fomojola:
Support for the AP_REDIS_URL parameter to allow configuration of Redis via a single environment variable.

## What does this PR do?

Enables the use of the AP_REDIS_URL variable to allow configuration of the Redis connection using a single environmental variable. 

- Added the REDIS_URL system property.
- Updated the Redis connection library to look for the REDIS_URL system property, and use that if present.

The primary motivation was to allow the use of Redis databases other than the default 1. 

The existing Redis connection assumes a dedicated Redis server where ActivePieces is connected to database 1. However when running on shared infrastructure or using an external Redis server there may be multiple databases and database 1 may already be in use. 

ioredis supports passing in a db parameter to the constructor, but it also supports a single URL that configures all the parameters. This is formatted as **redis(s)://:<password>@<hostname>:<port number>/<database number>** (with the extra s in the protocol indicating that TLS is required).

From a deployment perspective this is quite clean: instead of 5 variables we now have 1 that captures everything required for Redis.

Dependencies: none.

Fixes # (issue)

## Type of change

Enhancement

- [x] This change requires a documentation update

## How should this be tested?

Replace the AP_REDIS_HOST env parameter in the docker-compose.dev.yml file with the following value:

AP_REDIS_URL=redis://redis/1


abuaboud:
@allcontributors please add @fomojola for code

abuaboud:
I appreciate it greatly. This is good quality, I'm not sure how I overlooked it.

abuaboud (approved):

abuaboud:
Updated the environment variables in the documentation and merged.

fomojola:
Happy to contribute: ActivePieces works great. I'm using the webhook trigger to send Slack messages from a Google Cloud uptime check: its been silky smooth.

abuaboud:
@fomojola I am very happy to hear that 

I forget to mention this changes will be released (docker) with the next release (usually every Wednesday)

PR #676, opened 2023-02-25
feat: add billing

abuaboud:
## What does this PR do?

This PR integrates Stripe Billing and adds a webhook to handle the downgrade/upgrade of subscriptions. It also integrates Sentry for error monitoring.

Fixes # (issue)

## Type of change


packages/ee/billing/backend/billing.module.ts
```diff
@@ -0,0 +1,61 @@
...
+      const payloadString = request.rawBody;
+      const sig = request.headers["stripe-signature"] as string;
+      try {
+        const event = stripe.webhooks.constructEvent(
+          payloadString,
+          sig,
+          stripeSecret
+        );
+        await billingService.handleWebhook({ webhook: event });
+        reply.status(StatusCodes.OK).send(`Invalid webhook signature`);
```
khaledmashaly: I think you meant `valid signature`? Also, it doesn't make sense to reply to a webhook with a response body.

packages/ee/billing/backend/usage.service.ts
```diff
@@ -0,0 +1,77 @@
...
+                throw new ActivepiecesError({
+                    code: ErrorCode.TASK_QUOTA_EXCEEDED,
+                    params: { id: request.projectId },
+                });
+            }
+            projectUsage.consumedTasks += numberOfSteps;
+            await projectUsageRepo.save(projectUsage);
+        } catch (e) {
+            // TODO add sentry, Don't block for internal error, just continue and log it for now.
+            logger.error(e);
```
khaledmashaly: This catch block renders the `limit` method useless. you probably meant to check if the error is an instance of `ActivepiecesError` or not and act accordingly.

packages/ee/billing/backend/usage.service.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function countSteps(flowVersion: FlowVersion): number {
+    let steps = 0;
+    let currentStep: Trigger | Action | undefined = flowVersion.trigger;
+    while (currentStep !== undefined) {
+        currentStep = currentStep.nextAction;
+        steps++;
+    }
+    return steps;
+}
```
khaledmashaly, line 66: Shouldn't step count be calculated one time on flow creation?

packages/ee/billing/frontend/runs-left-snackbar/runs-left-snackbar.component.ts
```diff
@@ -1,30 +1,31 @@
...
   ) {
-    this.runsStats$ = this.usageService.getUsage().pipe(map(res => {
+    this.runsStats$ = this.billingService.getUsage().pipe(map(res => {
       return {
-        runsCap: res.metrics.steps.remaining + res.metrics.steps.consumed,
-        runsExecuted: res.metrics.steps.consumed,
-        daysLeftBeforeReset: Math.ceil(res.metrics.steps.nextResetInMs / 24 / 60 / 60 / 1000)
+        runsCap: res.plan.tasks,
+        runsExecuted: res.usage.consumedTasks,
+        daysLeftBeforeReset: Math.round((new Date(res.usage.nextResetDatetime ).getTime() - Date.now()) / 1000 / 60 / 60 / 24),
```
khaledmashaly: Using plain javascript dates is a recipe for disasters, we already have `dayjs` as a dependency, why not use it?

khaledmashaly (commented):
See comments, if everything is as you intended please merge.

packages/ee/billing/backend/usage.service.ts
abuaboud: Well it's better to not ruin the experience for the users, but I have integrated sentry, so we will be alerted if there is any issues there.

packages/ee/billing/backend/usage.service.ts
```diff
@@ -0,0 +1,77 @@
...
+        } catch (e) {
+            // Ignore quota errors for sake of user experience and log them instead
+            captureException(e);
+        } finally {
+            quotaLock.release();
+        };
+        return {
+            perform: true,
+        }
+    },
```
khaledmashaly: The `limit` method always returns `perform: true`, is this the intended behaviour?

abuaboud: Fixed, Thanks!

packages/ee/billing/backend/usage.service.ts
abuaboud: Yes, I will follow up on this on branching branch.

khaledmashaly (dismissed):

khaledmashaly (dismissed):

PR #617, opened 2023-02-17
feat: Implemented Trello Piece

ShayPunter:
## What does this PR do?

Implements a 'create card' action piece for Trello. 
Implements a 'create card' action piece for Trello
Implements documentation update effecting:
- Apps.mdx
- [new] trello.mdx
- quickstart piece guide (updated to reflect new import paths)

Fixes #278 (partial fix)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## How should this be tested?

Tested in a codespaces environment, please use the following test flow outlined in the documentation:

Obtain API Key & Token:
1. Go to https://trello.com/app-key
2. Copy 'personal key' and enter it into the Trello API Key connection
3. Click 'generate a token' in trello
4. Copy the token and enter it into the Trello Token connection
5. Your connection should now work!

Creating your first Trello card using Activepiece:
1. Create a new collection
2. Select trigger as 'schedule' and we will use its default value of 5 minutes
3. Click on + to add a new piece
4. Select the 'Trello' piece
5. Select the 'Create Card' Action on the right-hand sidebar
6. Select your Trello API Key connection you created above
7. Select your Trello Token Key connection above
8. Head over to the Trello board you wish to create a new card in
9. Create a new card, lets call it 'test'
10. Click into that card
11. In the URL, add '.json', this should return a wall of JSON text
12. Look for a value called 'idList', it should look something like this: `"idList":"5aea17e911c71fee96375d94"`
13. Copy the idList value, in our case `5aea17e911c71fee96375d94` and enter it into the idList field
14. Set the card name you'd like to create
15. Set a description to enter into the card (optional)
16. Lastly, lets test our flow by clicking 'Test Flow'
17. A new card should have been created in the selected list
18. Now click publish, and thats it!

Fetching a card via Activepieces

1. Create a new collection
2. Select trigger as 'schedule' and we will use its default value of 5 minutes
3. Click on + to add a new piece
4. Select the 'Trello' piece
5. Select the 'Get Card' Action on the right-hand sidebar
6. Select your Trello API Key connection you created above
7. Select your Trello Token Key connection above
8. Head over to the Trello board you wish to create a new card in
9. Create a new card, lets call it 'test'
10. Click into that card
11. In the URL, add '.json', this should return a wall of JSON text
12. Look for a value called 'id', it should look something like this: `"id":"5aea17e911c71fee96375d94"`
[...]

PR #677, opened 2023-02-25
Piece: Trello

ShayPunter:
Please see closed PR for changes: https://github.com/activepieces/activepieces/pull/617

Created a new PR since the previous branch fell too far behind commits to keep it up to date without a rebase.

Pending changes:
- Implementing a dropdown for lists 
- Implementing a dropdown for boards


sabuto:
I have managed to get the webhook working, in [here](https://github.com/sabuto/activepieces/tree/add-trello) I will push the code and open a PR when I have polished it. I have kept your name in it as most of the code has come from you. I am planning to add more things. I currently have the problem however that i cannot use context.webhookUrl as it just links to localhost when trying to send to trello which will not work. Manually entering the url into the code works however will only work for one flow. If you have any idea how to change this within codespace i would be appreciatative.

abuaboud:
> I have managed to get the webhook working, in [here](https://github.com/sabuto/activepieces/tree/add-trello) I will push the code and open a PR when I have polished it. I have kept your name in it as most of the code has come from you. I am planning to add more things. I currently have the problem however that i cannot use context.webhookUrl as it just links to localhost when trying to send to trello which will not work. Manually entering the url into the code works however will only work for one flow. If you have any idea how to change this within codespace i would be appreciatative.

Please go to `packages/backend/.env` and change `AP_WEBHOOK_URL` to the backend url (the one with port 3000), that should fix it.

sabuto:
I'll open a new pr after I've merge the changes from this pr :) in regards the the Triggers, trello allows you to add a modelid and then listens to any changes. Do you want different Triggers for cards etc or just a global trigger and then the end user to be able to filter @abuaboud

abuaboud:
@sabuto Great questions, users can be non technical and might be not aware how Trello API works.

It's better to create multiple triggers (New Card, New Activity , etc) and you can share the underlying code, at least this the language the user know.

For example in cal.com piece, there is multiple triggers where each trigger map to webhook event type but they share same underlying code.

PR #698, opened 2023-02-28
chore: support webhooks with path variables only

### LINKED ISSUE #697: Support alternative URL
## Is your feature request related to a problem? Please describe.

Currently, when using the webhook feature in Active Pieces, the URL for the webhook only supports query parameters for the flow id parameter. This can be limiting in certain use cases where alternative URL formats are needed.
for example currently https://cloud.activepieces.com/api/v1/webhooks?flowId=xxxxxxxx 
that could be  https://cloud.activepieces.com/api/v1/webhooks/flows/xxxxxxxx 

## Describe the solution you'd like

I would like Active Pieces to support alternative URL formats in the webhook feature, such as using a path instead of query parameters. This would provide greater flexibility and compatibility with other systems that may not support query parameters in their webhook URLs.

## Describe alternatives you've considered

As an alternative, I have attempted to use URL shorteners to convert the webhook URL with query parameters into a shorter URL with a path format. However, this approach has not been successful as the URL shorteners do not recognize the URL format used by Active Pieces webhooks.

I have also tried to use URL encoding and decoding to replace special characters with their encoded equivalents (e.g. replacing "=" with "%3D"), but this approach has also not been successful.

Despite attempting these alternatives, I have not been able to find a solution that allows me to use an alternative URL format with Active Pieces webhooks.

## Additional context

I believe that supporting alternative URL formats in the webhook feature would greatly benefit Active Pieces users and make it easier to integrate with other systems. Thank you for considering this feature request.

abuaboud:
## What does this PR do?

Support both types of url.

Fixes #697

## Type of change

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/packages/backend/src/app/webhooks/webhook-controller.ts
+++ b/packages/backend/src/app/webhooks/webhook-controller.ts
@@ -5,6 +5,21 @@ import { ApId } from "@activepieces/shared";
 import { webhookService } from "./webhook-service";
 
 export const webhookController: FastifyPluginCallback = (app, _opts, done): void => {
+
+    app.post(
+        "/:flowId",
+        {
+            schema: {
+                params: WebhookQueryParams,
+            },
+        },
+        async (request: FastifyRequest<{ Params: WebhookQueryParams }>, reply) => {
+            handler(request, request.params.flowId);
+            await reply.status(StatusCodes.OK).send();
+        }
+    );
+
+
     app.post(
         "/",
         {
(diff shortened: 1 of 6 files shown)


packages/backend/src/main.ts
```diff
@@ -28,6 +28,7 @@ import { firebaseAuthenticationModule } from "@ee/firebase-auth/backend/firebase
 import { billingModule } from "@ee/billing/backend/billing.module";
 import { getEdition } from "./app/helper/license-helper";
 import { ApEdition } from "@activepieces/shared";
+import { HttpMethod } from "@mailchimp/mailchimp_marketing";
```
khaledmashaly: why import something from MailChimp's SDK?

packages/backend/src/app/webhooks/webhook-controller.ts
```diff
@@ -5,6 +5,21 @@ import { ApId } from "@activepieces/shared";
 import { webhookService } from "./webhook-service";
 
 export const webhookController: FastifyPluginCallback = (app, _opts, done): void => {
+
+    app.post(
+        "/:flowId",
+        {
+            schema: {
+                params: WebhookQueryParams,
```
khaledmashaly: path params?

aboudzein (approved):

PR #700, opened 2023-02-28
fix: support atom feeds

abuaboud:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/pieces/apps/src/lib/rss/triggers/new-item-trigger.ts
+++ b/packages/pieces/apps/src/lib/rss/triggers/new-item-trigger.ts
@@ -1,6 +1,7 @@
 import { XMLParser } from 'fast-xml-parser';
 import { createTrigger, TriggerStrategy, httpClient, HttpMethod } from '@activepieces/framework';
 import { rssFeedUrl } from '../common/props';
+import crypto from "crypto";
 
 export const rssNewItemTrigger = createTrigger({
     name: 'new-item',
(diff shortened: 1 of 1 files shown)


packages/pieces/apps/src/lib/rss/triggers/new-item-trigger.ts
```diff
@@ -24,7 +25,10 @@ export const rssNewItemTrigger = createTrigger({
 
     async onEnable({ propsValue, store }): Promise<void> {
         const data = await getRss(propsValue.rss_feed_url);
-        store.put('lastFetchedRssItem', data?.rss?.channel?.item?.[0]?.guid);
+        const items = data?.rss?.channel?.item || data.feed?.entry || [];
+        store.put('lastFetchedRssItem', getId(items?.[0]));
+        console.log("HELLO");
+        console.log(getId(items?.[0]));
```
khaledmashaly: seems like you forgot to remove these debug logs?

packages/pieces/apps/src/lib/rss/triggers/new-item-trigger.ts
```diff
@@ -39,18 +43,34 @@ export const rssNewItemTrigger = createTrigger({
...
+    if (item === undefined) {
+        return undefined;
+    }
+    if (item.guid) {
+        return item.guid
+    }
+    if (item.id) {
+        return item.id;
+    }
+    return crypto.createHash('sha256').update(JSON.stringify(item)).digest('hex');
```
khaledmashaly: sync hashing, please see the web crypto API for an async alternative: https://nodejs.org/docs/latest-v18.x/api/webcrypto.html#subtledigestalgorithm-data

PR #708, opened 2023-03-01
chore(billing): update plans

abuaboud:
## What does this PR do?

Fixes # (issue)

## Type of change

### DIFF
--- a/packages/ee/billing/backend/billing.service.ts
+++ b/packages/ee/billing/backend/billing.service.ts
@@ -22,13 +22,23 @@ function getDefaultPlanId(): string {
     if (stripeSecret.startsWith('sk_test')) {
         return "price_1MeoK3KZ0dZRqLEKXIsGoguO";
     } else {
-        return "price_1MfMlJKZ0dZRqLEKWHrxY97A";
+        return "price_1MgpQ4KZ0dZRqLEKqMiS8vrf";
     }
 }
 
 function parsePlanFromId(id: string | null): { name: string, tasks: number } {
     switch (id) {
         // Stripe production plans
+        case 'price_1MgpQ4KZ0dZRqLEKqMiS8vrf':
+            return {
+                name: 'free-1',
+                tasks: 1000
+            }
+        case 'price_1MgpSTKZ0dZRqLEKLejnL5GW':
+            return {
+                name: 'growth-1',
+                tasks: 10000
+            }
         case 'price_1MfMlJKZ0dZRqLEKWHrxY97A':
             return {
                 name: 'free-0',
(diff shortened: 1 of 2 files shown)


packages/ee/billing/backend/billing.service.ts
```diff
@@ -22,13 +22,23 @@ function getDefaultPlanId(): string {
     if (stripeSecret.startsWith('sk_test')) {
         return "price_1MeoK3KZ0dZRqLEKXIsGoguO";
     } else {
-        return "price_1MfMlJKZ0dZRqLEKWHrxY97A";
+        return "price_1MgpQ4KZ0dZRqLEKqMiS8vrf";
```
khaledmashaly, line 25: I think It will be better to give these ids some names and save them into constant vars instead of repeating them mysteriously everywhere

packages/backend/src/app/database/migration/1677286751592-billing.ts
```diff
@@ -20,7 +20,7 @@ export class billing1677286751592 implements MigrationInterface {
         console.log("rolling back migration billing1677286751592");
         await queryRunner.query(`ALTER TABLE "project_usage" DROP CONSTRAINT "fk_project_usage_project_id"`);
         await queryRunner.query(`ALTER TABLE "project_plan" DROP CONSTRAINT "fk_project_plan_project_id"`);
-        await queryRunner.query(`DROP INDEX "public"."idx_project_usage_project_id"`);
+        await queryRunner.query(`DROP UNIQUE INDEX "public"."idx_project_usage_project_id_nextResetDatetime"`);
```
khaledmashaly: I think we should avoid editing old migrations, even if edits are trivial, it would be better to create a new migration file.

packages/ee/billing/backend/usage.entity.ts
```diff
@@ -21,9 +21,9 @@ export const ProjectUsageEntity = new EntitySchema<ProjectUsageSchema>({
     },
     indices: [
         {
-            name: "idx_project_usage_project_id",
-            columns: ["projectId"],
-            unique: false,
+            name: "idx_project_usage_project_id_nextResetDatetime",
+            columns: ["projectId", "nextResetDatetime"],
+            unique: true,
```
khaledmashaly: I think moving historic data to another entity altogether is safer to avoid doing complex queries when dealing with the current usage cycle

packages/ee/billing/backend/usage.service.ts
```diff
@@ -40,14 +41,22 @@ export const usageService = {
...
+        });
+        if (projectUsage === undefined || projectUsage === null) {
             const plan = await billingService.getPlan({ projectId });
-            projectUsage = await projectUsageRepo.save({
+            await projectUsageRepo.upsert({
                 id: apId(),
                 projectId,
                 consumedTasks: 0,
                 nextResetDatetime: nextResetDatetime(plan.subscriptionStartDatetime),
+            }, ['projectId', 'nextResetDatetime']);
```
khaledmashaly: I think this is prone to errors.

Issue #524, opened 2023-02-10
Dynamic list of properties


tpatel:
# **Is your feature request related to a problem? Please describe.**

The integration of apps with variable parameters is not possible.

# **Describe the solution you'd like**

I would like to create a dynamic list (created asynchronously) of text properties based on one static property.

This is a similar idea that the option list that updates its values based on the `refreshers` property, except it's about creating dynamically other properties (instead of the value of the own property).

# **Describe alternatives you've considered**

An alternative would be to create a type of property that is a list of key/values. This would put some work on the end user who would have to put the right keys though.

# **Additional context**

There are two use cases that I will describe here: an image-automation api, a spreadsheet.

## **Image automation api**

The user should be able to enter a string (template ID). This would trigger an API call that would retrieve the list of available properties.

Example 1: When entering the id for this template, the UI should offer the option for the user to enter the properties `content_text`, `avatar_background_image`...
<img width="400" alt="Screenshot 2023-02-10 at 17 29 57" src="https://user-images.githubusercontent.com/494686/218145132-4a1d5b21-a6a7-47b7-a692-71ceaa1b0b21.png">

Example 2: When entering the id for this template, the UI should offer the option for the user to enter the properties `image_background_image`, `button_text`...
<img width="400" alt="Screenshot 2023-02-10 at 17 34 33" src="https://user-images.githubusercontent.com/494686/218145506-c99b16c3-d08d-4465-ab75-aa592a7130b9.png">

## **spreadsheet**

The user should be able to select a spreadsheet sheet with a static property. This would trigger an API call that would retrieve the list of columns that could be made available as properties.

abuaboud:
@allcontributors please add @tpatel for ideas

abuaboud:
The backend part is merged, we need to finish the frontend part.

Will mark this closed once it's merged

abuaboud:
Now it's merged to main

PR #722, opened 2023-03-03
feat: added the GenerateBanners piece

tpatel:
## What does this PR do?

Adds a new piece to make the GenerateBanners service available to Activepiece users. GenerateBanners is a solution to generate images, pdf documents and videos from visual templates.

Unlocked thanks to https://github.com/activepieces/activepieces/issues/524.

## Type of change

New feature (non-breaking change which adds functionality)

## How should this be tested?

- [ ] Make sure the documentation is good enough to get the api keys from GenerateBanners
- [ ] Copy a template from the GenerateBanners library and use it in Activepieces to create an image
- [ ] Fill some variable fields to see them reflected in the generated image.

### DIFF
--- a/packages/pieces/apps/src/lib/generatebanners/actions/renderTemplate.action.ts
+++ b/packages/pieces/apps/src/lib/generatebanners/actions/renderTemplate.action.ts
@@ -1,0 +1,191 @@
+import {
+  createAction,
+  Property,
+  AuthenticationType,
+  HttpRequest,
+  HttpMethod,
+  httpClient,
+  BasicAuthPropertyValue,
+} from '@activepieces/framework';
+
+export const renderTemplate = createAction({
+  name: 'render_template',
+  description: 'Render a GenerateBanners template',
+  displayName: 'Render Template',
+  props: {
+    authentication: Property.BasicAuth({
+      displayName: 'API Key',
+      description: 'Find your api key in your GenerateBanners account page.',
+      required: true,
+      username: {
+        displayName: 'Public API Key',
+      },
+      password: {
+        displayName: 'Secret API Key',
+      },
+    }),
+    template_id: Property.Dropdown({
+      displayName: 'Template',
+      required: true,
+      refreshers: ['authentication'],
+      options: async ({ authentication }) => {
+        const auth = authentication as BasicAuthPropertyValue;
+        if (!auth) {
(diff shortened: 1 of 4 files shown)


packages/pieces/apps/src/lib/generatebanners/index.ts
```diff
@@ -0,0 +1,12 @@
+import { createPiece } from '@activepieces/framework';
+import { renderTemplate } from './actions/renderTemplate.action';
+
+export const generatebanners = createPiece({
+  name: 'generatebanners',
+  logoUrl: 'https://www.generatebanners.com/logo-transparent-512.png',
```
abuaboud: Uploaded, can you please change it to

`https://cdn.activepieces.com/pieces/generatebanners.png`

packages/pieces/apps/src/lib/generatebanners/actions/renderTemplate.action.ts
```diff
@@ -0,0 +1,191 @@
...
+      },
+    }),
+    audio_url: Property.ShortText({
+      displayName: 'Audio url',
+      description: 'Link to an audio file. Only used for videos.',
+      required: false,
+    }),
+    variables: Property.DynamicProperties({
+      displayName: 'Variables',
+      required: false,
```
abuaboud: I think this should be required for better user experience, what do you think?

abuaboud (commented):
Thank you, This operates flawlessly; I  can merge this pull request on Sunday.

abuaboud:
@allcontributors please add @tpatel for plugin

packages/pieces/apps/src/lib/generatebanners/index.ts
tpatel: Fixed, thanks!

packages/pieces/apps/src/lib/generatebanners/actions/renderTemplate.action.ts
tpatel: I was confused, I thought "required" meant the user had to fill all the inputs shown.
After testing, I agree it's better to always show these fields instead of hiding them in the optional section 👌

I've updated it to required. Good catch, thanks!

abuaboud (approved):

PR #732, opened 2023-03-05
[Trigger]: Gmail new mail trigger

### LINKED ISSUE #429: Gmail New Email Trigger
Trigger on when there is new email on google gmail

### Property:
- Google OAuth2

### Acceptance Criteria:
- Prefer Webhooks over Polling
- Sample Data is provided.

kanarelo:
## What does this PR do?

Fixes #429 

## Type of change

## Checklist

- I haven't checked if my PR needs changes to the documentation


packages/pieces/apps/src/lib/gmail/triggers/new-email.ts
```diff
@@ -0,0 +1,62 @@
...
+  description: 'Triggers when new mail is found in your Gmail inbox',
+  props: {
+    authentication: GmailProps.authentication,
+    subject: GmailProps.subject(true),
+    from: GmailProps.from(),
+    to: GmailProps.to(),
+    label: GmailProps.label(),
+    category: GmailProps.category()
+  },
+  sampleData: {
```
abuaboud: Lets use get mail and resolve the id and thread it to the email object in this trigger.

packages/pieces/apps/src/lib/gmail/triggers/new-email.ts
```diff
@@ -0,0 +1,62 @@
...
+import { GmailLabel } from '../common/models';
+import { GmailProps } from '../common/props';
+
+export const gmailNewEmailTrigger = createTrigger({
+  name: 'gmail_new_email_received',
+  displayName: 'New Email',
+  description: 'Triggers when new mail is found in your Gmail inbox',
+  props: {
+    authentication: GmailProps.authentication,
+    subject: GmailProps.subject(true),
```
abuaboud: Is it possible for subject to be optional? that would allow more usecases.

packages/pieces/apps/src/lib/gmail/common/props.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }),
+  label: (required = false) => Property.Dropdown<GmailLabel>({
+    displayName: "Label",
+    description: "The label tagged to the mail",
+    required,
+    defaultValue: "",
+    refreshers: ["authentication"],
+    options: async ({ authentication }) => {
+      if (authentication === undefined) {
```
abuaboud: I updated the docs today, but let's use `!authentication` that would make it work in case it was null as well.

abuaboud (changes requested):

packages/pieces/apps/src/lib/gmail/triggers/new-email.ts
kanarelo: Ok

kanarelo: fixed

abuaboud: image

I believe the issue persists as the data has not been enriched and it seems to be returning two emails instead of one. We need to loop through the messages, enrich them, and return an array.

abuaboud (approved):

PR #779, opened 2023-03-09
chore: sed: 1: ".env": invalid command code

astorozhevsky:
## What does this PR do?

On Mac OS, executing the command `sh tools/deploy.sh` causes the errors:

```sh
sed: 1: ".env": invalid command code .
sed: 1: ".env": invalid command code .
sed: 1: ".env": invalid command code .
A .env file containing random passwords and secrets has been successfully generated.
```

This change performs checking that the operating system is [macOS](https://wikipedia.org/wiki/Mac OS) and executes `sed ...` in a different way as described [here](https://stackoverflow.com/a/19457213).

## Type of change

## How should this be tested?

- use Mac OS :)
- clone the repo
- `sh tools/deploy.sh`

### DIFF
--- a/tools/deploy.sh
+++ b/tools/deploy.sh
@@ -1,6 +1,13 @@
 cp .env.example .env
-sed -i 's|AP_POSTGRES_PASSWORD=.*|AP_POSTGRES_PASSWORD='"$(openssl rand -hex 32)"'|g' .env
-sed -i 's|AP_JWT_SECRET=.*|AP_JWT_SECRET='"$(openssl rand -hex 32)"'|g' .env
-sed -i 's|ENCRYPTION_KEY=.*|ENCRYPTION_KEY='"$(openssl rand -hex 16)"'|g' .env
+
+if [[ $OSTYPE == 'darwin'* ]]; then
+  sed -i '' -e 's|AP_POSTGRES_PASSWORD=.*|AP_POSTGRES_PASSWORD='"$(openssl rand -hex 32)"'|g' .env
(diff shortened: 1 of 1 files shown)


tools/deploy.sh
```diff
@@ -1,6 +1,13 @@
 cp .env.example .env
-sed -i 's|AP_POSTGRES_PASSWORD=.*|AP_POSTGRES_PASSWORD='"$(openssl rand -hex 32)"'|g' .env
-sed -i 's|AP_JWT_SECRET=.*|AP_JWT_SECRET='"$(openssl rand -hex 32)"'|g' .env
-sed -i 's|ENCRYPTION_KEY=.*|ENCRYPTION_KEY='"$(openssl rand -hex 16)"'|g' .env
 
-echo "A .env file containing random passwords and secrets has been successfully generated."
\ No newline at end of file
+if [[ $OSTYPE == 'darwin'* ]]; then
```
abuaboud: This is causing `run.sh: 3: [[: not found`

We can replace it with `if test "$OSTYPE" = "darwin"*; then` but can you please check if it works on your MacOS?

astorozhevsky: @abuaboud thanks for your feedback! 

I tried your code and unfortunately it didn't work, so I googled again and found another solution: https://stackoverflow.com/a/17072017

Could you please check it on your side?

abuaboud: This works with bash `bash tools/deploy.sh`

abuaboud: Replacing the if with single equals make it work for me, can you check?

`if [ "$(uname)" = "Darwin" ]; then`

abuaboud: Tested it on mac with single quote, It works!

Merging!

abuaboud (approved):

PR #785, opened 2023-03-11
chore: clean action/trigger schema

abuaboud:
## What does this PR do?

Fixes # (issue)

## Type of change

### DIFF
--- a/packages/shared/src/lib/flows/flow-helper.ts
+++ b/packages/shared/src/lib/flows/flow-helper.ts
@@ -184,7 +184,8 @@ function createAction(
     case ActionType.LOOP_ON_ITEMS:
       action = {
         ...baseProperties,
-        // firstLoopAction: request.firstLoopAction,
+        // Required to do when loops is supported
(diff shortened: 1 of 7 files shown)


packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -184,7 +184,7 @@ function createAction(
     case ActionType.LOOP_ON_ITEMS:
       action = {
         ...baseProperties,
-        firstLoopAction: request.firstLoopAction,
+        // firstLoopAction: request.firstLoopAction,
```
khaledmashaly: Why not just remove this instead of commenting out?

khaledmashaly (approved):

packages/shared/src/lib/flows/flow-helper.ts
abuaboud: We haven't done loops yet, I want it to be there to serve as reminder.
Let me add comment

PR #787, opened 2023-03-12
chore: added url encoding for code parameter

J0LGER:
## Description

The `code` parameter shall be URL encoded using `encodeURIComponent()` before being reflected in the `reply.send()` call, in order to escape special characters that can be rendered as malicious JS.

## PoC 
Sending `code` parameter with value `test' },'*')} else alert(1);//` will cause cross-site scripting.

### DIFF
--- a/packages/backend/src/main.ts
+++ b/packages/backend/src/main.ts
@@ -109,7 +109,7 @@ app.get(
             reply.send("The code is missing in url");
         }
         else {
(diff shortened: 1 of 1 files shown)


abuaboud:
@allcontributors please add @J0LGER for security

abuaboud (approved):
Good Catch, Thank you!

PR #797, opened 2023-03-13
chore: docker compose dev

astorozhevsky:
## What does this PR do?

- formatted docker-compose.dev.yml
- added running `npm i` inside a container 

The backend does not start on the Mac OS due to the error:

image

I've found the answer here - https://stackoverflow.com/a/20590261, that **bcrypt compiled on OSX will not quite work on Linux**:

> I've found that bcrypt compiled on OSX will not quite work on Linux. In other words, if you check in the bcrypt compiled on your local OSX workstation, and try to run the node app on your linux servers, you will see the error above.

So, if the dependencies installed locally do not match the dependencies required in the container, then it is a good practice to run `npm i` inside the container to ensure that the correct dependencies are installed.

## Type of change

- [x] Chore (refactoring code, technical debt, workflow improvements)

## How should this be tested?

- `sh tools/deploy.sh`
- `docker compose -f ./docker-compose.dev.yml up`

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,4 +1,4 @@
-version: '3.0'
+version: '3'
 
 services:
   backend:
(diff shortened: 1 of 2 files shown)


khaledmashaly:
Thanks for the PR!

Some thoughts:

- There are a lot of issues with installing npm packages from within the docker container, for e.g.
    - Husky git hooks fail to register
    - the `node_module` directory becomes owned by the root user, which in turn makes running any npm commands outside the docker container fails, e.g. `npx nx refresh-docs pieces-apps`
- npm packages are installed when both the backend and the frontend containers start (i.e. npm packages are installed two times simultaneously!)

There are a lot of issues with the docker compose dev setup right now, that's why the dev container is the officially supported method for running the local development environment right now.

Please let me know if you are willing to solve the docker compose dev issues

astorozhevsky:
> Thanks for the PR!
> 
> Some thoughts:
> 
> * There are a lot of issues with installing npm packages from within the docker container, for e.g.
>   
>   * Husky git hooks fail to register
>   * the `node_module` directory becomes owned by the root user, which in turn makes running any npm commands outside the docker container fails, e.g. `npx nx refresh-docs pieces-apps`
> * npm packages are installed when both the backend and the frontend containers start (i.e. npm packages are installed two times simultaneously!)
> 
> There are a lot of issues with the docker compose dev setup right now, that's why the dev container is the officially supported method for running the local development environment right now.
> 
> Please let me know if you are willing to solve the docker compose dev issues

I would like to express my sincere gratitude for taking the time to review my pull request!

After careful consideration, I have decided to close this request as your concerns are reasonable and I believe that it is important to address them before moving forward.

Thanks!

PR #836, opened 2023-03-17
chore: move webhooks secrets to local file

abuaboud:
## What does this PR do?

[ ]:  Shows use own credentials / own connections in self hosted version if there trigger is app webhook
[ ]:  Show Triggers that are app webhooks on self hosted, only if piece is in `APP_SECRETS` from flags service.


abuaboud:
Closing will reopen when developing more app webhooks.

PR #850, opened 2023-03-22
feat: multiple piece versions per flow #819

### LINKED ISSUE #819: [BUG]: use one version for all piece steps in a single flow

khaledmashaly:
## What does this PR do?

Multiple piece versions per flow using [pnpm aliases](https://github.com/pnpm/pnpm.github.io/blob/main/docs/aliases.md)

Fixes #819

## Type of change


packages/shared/src/lib/pieces/utils.ts
```diff
@@ -0,0 +1,16 @@
+type PackageInfoParams = {
+  pieceName: string;
+  pieceVersion: string;
+}
+
+export const getPackageNameForPiece = (params: PackageInfoParams): string => {
```
abuaboud: I assume this is the alias inside activepieces right? can you we rename it to getPackageInternalNameForPiece / or getPackageAliasForPiece?

abuaboud (changes requested):

abuaboud (dismissed):

packages/shared/src/lib/pieces/utils.ts
khaledmashaly: fixed in 4b91624

abuaboud (approved):

Issue #267, opened 2023-01-11
Shopify Piece


ashrafsamhouri:
- Triggers:
    - New Paid Order.
    - New Customer.

Action:
    - Create Customer

sabuto:
Out of interest why is the shopify piece blocked? I wouldn't mind trying to make this?

abuaboud:
@ashrafsamhouri Please split them into triggers and actions.

Only the triggers is blocked because they are app webhooks, which means we need a routing logic to know which users to route the events to.

PR #871, opened 2023-03-24
feat: add shopify piece

abuaboud:
## What does this PR do?

Fixes https://github.com/activepieces/activepieces/issues/267


packages/frontend/src/app/modules/flow-builder/service/action-meta.service.ts
```diff
@@ -80,6 +82,9 @@ export class ActionMetaService {
     const { piecesManifest, release } = params;
 
     return piecesManifest.filter((piece) => {
+      if (params.environment === ApEnvironment.DEVELOPMENT) {
+        return true;
+      }
```
khaledmashaly, line 87: any reason for this?

packages/pieces/shopify/src/lib/triggers/new-cancelled-order.ts
```diff
@@ -0,0 +1,755 @@
+import { createShopifyWebhookTrigger } from "../common/register-webhook";
+
+
+export const newCancelledOrder = createShopifyWebhookTrigger({
+    name: "new_cancelled_order",
+    description: "Triggered when order is cancelled",
+    topic: "orders/cancelled",
+    displayName: "New Cancelled Order",
+    sampleData: {
```
khaledmashaly, line 9: This is an overkill for "sampleData", right?

khaledmashaly (approved):

packages/frontend/src/app/modules/flow-builder/service/action-meta.service.ts
abuaboud: Yes, If i put minimum release for next version, It will stop showing up.

It's either we bump package.json after each release, which cause the notification appears for everyone or just stop the filter in local environment.

packages/pieces/shopify/src/lib/triggers/new-cancelled-order.ts
abuaboud: Yes :/ Not sure btter alternative right now

PR #895, opened 2023-03-26
chore: seed a dev user

khaledmashaly:
## What does this PR do?

Creating a user on everytime after running `docker compose down` was getting on my nerves

## Type of change

- [x] Chore (refactoring code, technical debt, workflow improvements)


packages/backend/src/app/database/seeds/dev-seeds.ts
```diff
@@ -0,0 +1,32 @@
...
+import { logger } from "../../helper/logger";
+import { system } from "../../helper/system/system";
+import { SystemProp } from "../../helper/system/system-prop";
+
+const seedDevUser = async (): Promise<void> => {
+    logger.debug("[seedDevUser]");
+
+    await authenticationService.signUp({
+        email: "dev@ap.com",
+        password: "12345678",
```
abuaboud: If people skipped reading the docs, I am not sure how they would find this.

I am thinking of following options:
1- Allow sign up on development `.env` in case someone couldn't find it docs.
2- Print it in console (which might be not clear)
3- Don't merge this PR at all

Btw why this case is often for you? when running docker compose it shouldn't delete the volumes or am i wrong?

abuaboud (changes requested):

packages/backend/src/app/database/seeds/dev-seeds.ts
khaledmashaly: oops! good catch!
enabled sign-up on dev `.env` in 155f3c2
logged credentials in 6554377

packages/backend/src/app/database/seeds/dev-seeds.ts
```diff
@@ -0,0 +1,39 @@
...
+            email: devEmail,
+            password: devPassword,
+            firstName: "firstName",
+            lastName: "lastName",
+            trackEvents: false,
+            newsLetter: false,
+        });
+    }
+
+    logger.info(`[seedDevUser] email=${devEmail} pass=${devPassword}`);
```
abuaboud, line 24: make it human readable, like You can use the following account to sign in

khaledmashaly: added a warning message in cbf594b

PR #918, opened 2023-03-28
feat: add collection status in the dashboad

AbdulTheActivePiecer:
## What does this PR do?

Adds collection status to collection table as a toggle, also added a new endpoint for instances to just update the instance status without having to publish the collection with the latest version.


packages/backend/src/app/instance/instance.service.ts
```diff
@@ -78,6 +81,29 @@ export const instanceService = {
...
+            throw new ActivepiecesError({
+                code: ErrorCode.COLLECTION_NOT_FOUND,
+                params: {
+                    id: request.collectionId,
+                },
+            });
+        }
+        const instance: Instance | null = await instanceRepo.findOneBy({ projectId, collectionId: request.collectionId });
+        if(instance) {
+            instance.status = request.status;
```
abuaboud, line 97: You also need to handle that the state where old state is not equal to new state.

packages/backend/src/app/instance/instance.controller.ts
```diff
@@ -50,4 +50,17 @@ export const instanceController = async (app: FastifyInstance) => {
         });
         reply.status(StatusCodes.OK).send();
     });
+    // update status
+    app.post(
+        "/updateStatus",
```
abuaboud: The naming convention for url is update-status, but need to think of the naming.

packages/backend/src/app/collections/collection.service.ts
```diff
@@ -66,6 +71,23 @@ export const collectionService = {
...
+            previous:collections.previous,
+            data:[]
+        };
+    
+        for(const c of collections.data) {
+            const instance = await instanceService.getByCollectionId({projectId:c.projectId,collectionId: c.id});
+            const dto= {...c,
+                status: instance?
+                 instance.status === InstanceStatus.DISABLED ? CollectionStatus.DISABLED: CollectionStatus.ENABLED :
+                  CollectionStatus.UNPUBLISHED } as unknown as  CollectionListDto;
```
abuaboud: It's better to move the state decision to function.

abuaboud (changes requested):
Let me think about it and get back to you.

abuaboud (approved):

PR #922, opened 2023-03-29
Action: Added delete row action to Google Sheets

### LINKED ISSUE #920: Action: Delete row from Google Sheets
**Describe your usecase.**
The use case I am currently using involves polling a spreadsheet for new rows, then an IF statement is put in place, if the value does not match the if statement, then it will remove the row from the spreadsheet, otherwise it'll summarise the response using ChatGPT and send a notification into Discord.

tl;dr: Need a delete row since a flow depends on it

ShayPunter:
## What does this PR do?

- Adds a new action called 'delete-row' to google sheets which will delete a row from a spreadsheet
- Return a Row ID value from new row action

Fixes #920

### DIFF
--- a/packages/pieces/google-sheets/src/lib/common/common.ts
+++ b/packages/pieces/google-sheets/src/lib/common/common.ts
@@ -72,7 +72,8 @@ export const googleSheetsCommon = {
     }),
     getValues: getValues,
     appendGoogleSheetValues: appendGoogleSheetValues,
-    findSheetName:findSheetName
+    findSheetName:findSheetName,
+    deleteRow: deleteRow,
 }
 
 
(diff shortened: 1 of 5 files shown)


packages/pieces/google-sheets/src/lib/common/common.ts
```diff
@@ -140,6 +141,33 @@ async function getValues(spreadsheetId: string, accessToken: string, sheetId: nu
     return response.body.values;
 }
 
+async function deleteRow(spreadsheetId: string, sheetId: number, rowIndex: number, accessToken: string) {
+    const sheetName = await findSheetName(accessToken, spreadsheetId, sheetId);
```
abuaboud: Remove it's unsued.

packages/pieces/google-sheets/src/lib/common/common.ts
```diff
@@ -140,6 +141,33 @@ async function getValues(spreadsheetId: string, accessToken: string, sheetId: nu
...
+        },
+        body: {
+            requests: [
+                {
+                    deleteDimension: {
+                        range: {
+                            sheetId: sheetId,
+                            dimension: "ROWS",
+                            startIndex: rowIndex,
+                            endIndex: rowIndex + 1,
```
abuaboud, line 160: There is weird bug, try make sheet with 4 rows, and delete the second row.

It will delete rows (2,3,4)

PR #966, opened 2023-04-01
feat: add notification url

abuaboud:
## What does this PR do?

Fixes # (issue)


khaledmashaly (dismissed):
I think we should document webhook payload for users

PR #974, opened 2023-04-02
feat: adds box app and triggers

kanarelo:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/pieces/box/src/lib/triggers/index.ts
+++ b/packages/pieces/box/src/lib/triggers/index.ts
@@ -1,0 +1,74 @@
+import { Property } from '@activepieces/framework'
+import { boxRegisterTrigger } from "./register"
+
+const itemProps = {
+  id: Property.ShortText({
+    displayName: 'Item ID',
+    description: 'The ID of the item to trigger a webhook',
+    required: true
+  }),
+  type: Property.StaticDropdown({
+    displayName: 'Item Type',
+    description: 'The type of the item to trigger a webhook',
+    required: true,
+    options: {
+      options: [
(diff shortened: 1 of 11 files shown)


packages/pieces/box/src/lib/triggers/index.ts
```diff
@@ -0,0 +1,48 @@
+import { boxRegisterTrigger } from "./register"
+
+const triggerData = [
+  {
```
abuaboud, line 6: Let's focus on top usecases, I suggest:

- New File
- New Folder
- New Comment

For thew new file can we make the default root directory.

Rename Item Id -> Folder Id and make it optional.
and Item type in that case will always be Folder.

packages/pieces/box/src/lib/triggers/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    description: 'Triggers on `FILE.DELETED`',
+    event: 'FILE.DELETED',
+    sampleData: {}
+  },
+  {
+    name: 'folder_created',
+    displayName: "Folder Created",
+    description: 'Triggers on `FOLDER.CREATED`',
+    event: 'FOLDER.CREATED',
+    sampleData: {}
```
abuaboud: Can you provide sample data? that helps a lot in auto complete UI

packages/pieces/box/src/lib/triggers/index.ts
kanarelo: defaulting to root folder `Access denied - insufficient permission` 

Any other id works.... This s for new file upload trigger

PR #995, opened 2023-04-03
feat: test action

khaledmashaly:
## What does this PR do?

This PR creates a test step endpoint.
the test endpoint just tests a single action in isolation
next pr will introduce flow context to test step 

relates to #828

### DIFF
--- a/packages/shared/src/lib/flows/step-run/dto/test-step-request.ts
+++ b/packages/shared/src/lib/flows/step-run/dto/test-step-request.ts
@@ -1,0 +1,15 @@
+import { Type, Static } from "@sinclair/typebox";
+import { SemVerType } from "../../../pieces";
+
+export const TestStepRequestBody = Type.Object({
+    actionName: Type.String(),
+    pieceName: Type.String(),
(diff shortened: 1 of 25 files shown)


packages/shared/src/lib/flows/step-run/dto/test-step-request.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const TestStepRequestBody = Type.Object({
+    actionName: Type.String(),
+    pieceName: Type.String(),
+    pieceVersion: SemVerType,
+    input: Type.Record(
+        Type.String(),
+        Type.Any(),
+    ),
+    collectionId: Type.String(),
```
abuaboud, line 12: I don't this this is needed

abuaboud (approved):

packages/shared/src/lib/flows/step-run/dto/test-step-request.ts
khaledmashaly: This DTO will be completely changed in next PR isA

PR #1007, opened 2023-04-05
chore: replace redis lock library

abuaboud:
## What does this PR do?

All retries should be at least 2x of timeout, and replaced the library with redlock.

Fixes # (issue)


khaledmashaly (commented):
I think we should wrap the lock logic inside a helper class, which has two methods: acquire, and release. To avoid coupling the code to a specific library, because this is the third time we changed the locking library

PR #1029, opened 2023-04-07
feat: cache sandbox and make them reuseable.

abuaboud:
## What does this PR do?

Reuse the sandbox for same flow, to avoid reinstalling even using pnpm it's still slow.

Execution time overhead goes down to 300 ms after this change.


packages/backend/src/app/workers/sandbox.ts
```diff
@@ -22,33 +23,65 @@ export class Sandbox {
...
+        const filesToDelete = [
+            '_functionOutput.txt',
+            '_standardOutput.txt',
+            '_standardError.txt',
+            'output.json',
+            'meta.txt',
+        ]
+        const promises = filesToDelete.map((file) => {
+            const filePath = path.join(__dirname, this.getSandboxFilePath(file))
+            return fs.unlink(filePath).catch((error) => {
```
khaledmashaly, line 57: Any reason to use `fs#unlink` instead of the more obvious `fs#rm`? Also why use `Promise#catch` instead of a `try/catch` block?

khaledmashaly, line 57: Any reason to use `fs#unlink` instead of the more obvious `fs#rm`?

packages/backend/src/app/helper/engine-helper.ts
```diff
@@ -65,30 +65,29 @@ export const engineHelper = {
         }) as ExecutionOutput
     },
     async executeParseEvent(operation: ExecuteEventParserOperation): Promise<ParseEventResponse> {
-        const sandbox = sandboxManager.obtainSandbox()
+        const sandbox = await sandboxManager.obtainSandbox(apId())
```
khaledmashaly, line 68: Any reason not to use `${pieceName}:${pieceVersion}` for `resourceId` in `executeParseEvent` as well?

packages/backend/src/app/workers/sandbox.ts
```diff
@@ -135,31 +168,79 @@ export class Sandbox {
...
-    obtainSandbox(): Sandbox {
-        const sandboxId = this.queue.pop()
-        if (sandboxId === undefined) {
-            throw new Error('Unexpected error, ran out of sandboxes')
+    async obtainSandbox(key: string): Promise<Sandbox> {
+        // Acquire the lock
+        const release = await this.mutex.acquire()
+
+        // Find sandbox with resourceId equal to key and not used
+        const sandbox = Array.from(this.sandboxes.values()).find(s => s.resourceId === key && !s.used)
```
khaledmashaly, line 199: suggestion: another map can be used to store `keyId` to `sandboxId` mapping instead of looping through all sandboxes

packages/backend/src/app/workers/sandbox.ts
```diff
@@ -135,31 +168,79 @@ export class Sandbox {
...
+        // Find oldest sandbox not used
+        const oldestSandbox = Array.from(this.sandboxes.values()).reduce((oldest, current) => {
+            if (oldest === null) {
+                return current
+            }
+            if (current.lastUsed < oldest.lastUsed) {
+                return current
+            }
+            return oldest
+        }, null)
```
khaledmashaly, line 220: tip: if no initial value is provided for the accumelator, then the first element in the array is used and reduction starts from the second element in the array

khaledmashaly (approved):
Good overall! did you benchmark the performance gain?

abuaboud:
@khaledmashaly is not exactly the benchmark but preparing sandbox time went down from 700-1200ms to 300ms.

Is there a certain concern you are worried about?

khaledmashaly:
> @khaledmashaly is not exactly the benchmark but preparing sandbox time went down from 700-1200ms to 300ms due to engine size being 4mb+
> 
> Is there a certain concern you are worried about?

gr8, just wanted to know how much differrence caching makes

PR #1045, opened 2023-04-10
Piece: XML 

Willianwg:
## What does this PR do?

 Convert JSON to XML


packages/pieces/xml/src/lib/actions/convert-json-to-xml.ts
```diff
@@ -0,0 +1,29 @@
...
+            displayName: 'Label',
+            description: 'Label of the data content',
+            required: true,
+        }),
+    },
+    async run(context) {
+        const { json, label } = context.propsValue;
+
+        const convertedToXml = js2xml({
+            [label]: JSON.parse(JSON.stringify(json))
```
abuaboud, line 23: Why there is label, I don't get it can you explain and add comment?

PR #1049, opened 2023-04-11
Fix: pino-pretty works only in dev mode

### LINKED ISSUE #910: [BUG]: Logs
**Describe the bug**
Standard logger is probably not used

**Expected behavior**
- Not able to set log levels - debug, info, error etc
- All the logs seems like console logs which are spread across multiple lines in the cloud logging enviornment. Filtering logs and batching becomes difficult

**Screenshots**
Screenshot 2023-03-28 at 7 45 26 PM

abdullahranginwala:
## What does this PR do?

This PR removes the pino-pretty transport from the logger configuration when the app is in production mode. This is to ensure that the logs are in the default JSON format.

Fixes #910

### DIFF
--- a/packages/backend/src/app/helper/logger.ts
+++ b/packages/backend/src/app/helper/logger.ts
@@ -30,17 +30,14 @@ const initLogger = () => {
         ? 'debug'
         : 'info'
 
+    const transports = env === ApEnvironment.DEVELOPMENT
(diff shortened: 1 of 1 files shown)


abuaboud:
That you for picking this up.

@allcontributors please add @abdullahranginwala for code

PR #1053, opened 2023-04-11
Make Message Input Required in Telegram Piece Action

### LINKED ISSUE #1038: Message Input not optional in Telegram piece action.
The message input in Telegram piece action is under the  "Optional Inputs" field in the builder, but it's not actually optional because the flow wouldn't run successfully if there is no message inputted.

Screenshot (3)

The message input should not be under the "Optional Inputs" field as it is not optional.

abdullahranginwala:
## What does this PR do?

Since the telegram piece action doesn't run successfully without the message field, I've changed it to 'required'

Fixes #1038

### DIFF
--- a/packages/pieces/telegram-bot/package.json
+++ b/packages/pieces/telegram-bot/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-telegram-bot",
-  "version": "0.2.3"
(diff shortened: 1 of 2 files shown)


abuaboud:
@abdullahranginwala can you please bump the version in package.json

abuaboud (approved):

PR #1081, opened 2023-04-17
feat: return success boolean with action test response

khaledmashaly:


abuaboud (approved):
Nit: always prefer enum over boolean in api respones

PR #1088, opened 2023-04-18
feat: add playwright tests

abuaboud:
## What does this PR do?

Fixes # (issue)


khaledmashaly (approved):
should we run this test on a build also?

PR #1089, opened 2023-04-18
chore: enable strict ts for backend

### LINKED ISSUE #1015: [Chore]: enable strict typescript compiler option on backend
For some reason, `strict` is enabled for all projects except the backend.
`strict` helps catch a lot of bugs in compile time.
Right now the `120` errors are reported when `strict` is enabled on the backend.

khaledmashaly:
## What does this PR do?

Fixes #1015


packages/backend/src/app/authentication/authentication.service.ts
```diff
@@ -37,7 +37,7 @@ export const authenticationService = {
                     projectId: project.id,
                 },
             })
-            user.password = undefined
+            user.password = ''
```
abuaboud: Please add todo to make it undefined so api response doesn't return empty string

packages/backend/src/app/flows/flow-version/flow-version-entity.ts
```diff
@@ -2,7 +2,7 @@ import { EntitySchema } from 'typeorm'
 import { Flow, FlowVersion } from '@activepieces/shared'
 import { ApIdSchema, BaseColumnSchemaPart } from '../../helper/base-entity'
 
-type FlowVersionSchema = {
+export type FlowVersionSchema = {
```
abuaboud: You can use the FlowVersion interface in the repo and keeps this one private

packages/backend/src/app/authentication/authentication.service.ts
khaledmashaly: fixed in f9e85866

packages/backend/src/app/flows/flow-version/flow-version-entity.ts
khaledmashaly: fixed in b7bed4a7

PR #1100, opened 2023-04-19
feat: add postgres piece

Willianwg:
## What does this PR do?

Fixes # (issue)


packages/pieces/postgres/src/lib/actions/run-query.ts
```diff
@@ -0,0 +1,95 @@
...
+                    required: true,
+                })
+            },
+            required: true
+        }),
+        query: Property.ShortText({
+            displayName: 'Query',
+            required: true,
+        }),
+        rejectUnauthorized: Property.Checkbox({
```
abuaboud: Please use snake case for all props, to make them consistent.

packages/pieces/postgres/src/lib/actions/run-query.ts
```diff
@@ -0,0 +1,95 @@
...
+            displayName: 'Reject Unauthorized',
+            required: true,
+            defaultValue: true
+        }),
+        keepAlive: Property.Checkbox({
```
abuaboud: I think this option should always be false, we can remove it from props.

packages/pieces/postgres/src/lib/actions/run-query.ts
```diff
@@ -0,0 +1,95 @@
...
+            displayName: 'Reject Unauthorized',
+            required: false,
+        }),
+        query_timeout: Property.Number({
+            displayName: 'Query Timeout',
+            required: false,
+        }),
+        statement_timeout: Property.Number({
```
abuaboud: I think this can be same as query timeout in this case, so you can remove it.

packages/pieces/postgres/src/lib/actions/run-query.ts
```diff
@@ -0,0 +1,95 @@
...
+            displayName: 'Statement Timeout',
+            required: false,
+        }),
+        connectionTimeoutMillis: Property.Number({
```
abuaboud: add default value for example 60 seconds.

packages/pieces/postgres/src/lib/actions/run-query.ts
```diff
@@ -0,0 +1,95 @@
...
+        rejectUnauthorized: Property.Checkbox({
+            displayName: 'Reject Unauthorized',
+            required: true,
+            defaultValue: true
+        }),
+        keepAlive: Property.Checkbox({
+            displayName: 'Reject Unauthorized',
+            required: false,
+        }),
+        query_timeout: Property.Number({
```
abuaboud, line 49: Please add description for the props.

- `host`: A string indicating the hostname of the PostgreSQL server to connect to. If omitted, the default is `localhost`.

- `user`: A string indicating the user to authenticate as when connecting to the PostgreSQL server. This parameter is required.

- `password`: A string indicating the password to use for authentication. If omitted, the client will attempt to connect to the server using other authentication methods (such as an authentication token or the user's operating system credentials).

- `database`: A string indicating the name of the database to connect to. This parameter is required.

- `query_timeout`: An integer indicating the maximum number of milliseconds to wait for a query to complete before timing out. If omitted, there is no query timeout.

- `application_name`: A string indicating the name of the client application connecting to the server. This can be useful for identifying which client applications are using the server's resources. If omitted, the default is to use the name of the calling Node.js process.

- `connectionTimeoutMillis`: An integer indicating the maximum number of milliseconds to wait for a connection to be established before timing out. If omitted, the default is to wait indefinitely.

packages/pieces/postgres/src/lib/actions/run-query.ts
```diff
@@ -0,0 +1,95 @@
...
+    description: "Run Query",
+    props: {
+        authentication: Property.CustomAuth({
+            displayName: "Authentication",
+            props: {
+                host: Property.ShortText({
+                    displayName: 'Host',
+                    required: true,
+                }),
+                port: Property.ShortText({
```
abuaboud, line 17: please add default port

abuaboud (changes requested):

PR #1109, opened 2023-04-20
Chore/folders and flows actions

AbdulTheActivePiecer:
## What does this PR do?

Continuation of work


packages/ui/common/src/lib/ui-common.module.ts
```diff
@@ -129,7 +133,7 @@ export function markedOptionsFactory(): MarkedOptions {
   providers: [
     { provide: MAT_TOOLTIP_DEFAULT_OPTIONS, useValue: materialTooltipDefaults },
   ],
-  declarations: [...exportedDeclarations],
+  declarations: [...exportedDeclarations, GenericSnackbarTemplateComponent],
```
abuaboud: Duplicate `GenericSnackbarTemplateComponent`

packages/backend/src/app/flows/flow/flow.service.ts
```diff
@@ -110,7 +112,7 @@ export const flowService = {
         })
         return paginationHelper.createPage<Flow>(paginationResult.data, paginationResult.cursor)
     },
-    async getOne({ projectId, id, versionId, includeArtifacts = true }: { projectId: ProjectId, id: FlowId, versionId: FlowVersionId | undefined, includeArtifacts: boolean }): Promise<Flow | null> {
+    async getOne({ projectId, id, versionId, includeArtifacts = true }: { projectId: ProjectId, id: FlowId, versionId: FlowVersionId | undefined, includeArtifacts: boolean }): Promise<FlowBuilderDto | null> {
```
abuaboud: `FlowBuilderDto` revert to `Flow`, the folder display name be obtained by the frontend and shouldn't be part of API.

packages/backend/src/app/flows/folder/folder.controller.ts
```diff
@@ -19,31 +19,31 @@ export const folderController = async (fastify: FastifyInstance) => {
...
+                Body: CreateOrRenameFolderRequest
             }>,
         ) => {
             return await folderService.create({ projectId: request.principal.projectId, request: request.body })
         },
     )
 
 
-    fastify.post(
+    fastify.put(
```
abuaboud: Why did you change to `put`? for consistency we use post in all other endpoints.

Also PUT usually is used when you replace the whole object

abuaboud (changes requested):

packages/shared/src/lib/flows/flow.ts
```diff
@@ -15,4 +15,8 @@ export interface Flow extends BaseModel<FlowId> {
 
 export interface FlowTableDto extends Flow {
     status: FlowInstanceStatus
+}
+
+export interface FlowBuilderDto extends Flow {
+    folderDisplayName:string;
 }
```
abuaboud: Remove this and place them in origianl interface

abuaboud (changes requested):

PR #1128, opened 2023-04-22
Fixed issue #781

### LINKED ISSUE #781: Fix Static Dropdown Structure & Remove Piece Config Class in Frontend
**Describe the bug**

Static dropdown today has structure of `options` inside `options`.

```typescript

export type DropdownState<T> = {
	disabled?: boolean;
	placeholder?: string;
	options: DropdownOption<T>[];
}

export type StaticDropdownProperty<T, R extends boolean> = BasePropertySchema & {
	options: DropdownState<T>;
} & TPropertyValue<T, PropertyType.STATIC_DROPDOWN, R>;
```

It should be 

```typescript
export type StaticDropdownProperty<T, R extends boolean> = BasePropertySchema & {
	placeholder?: string;
	options: DropdownOption<T>[];
} & TPropertyValue<T, PropertyType.STATIC_DROPDOWN, R>;
```

Same for Multi select static dropdown.

Look Into this class: https://github.com/activepieces/activepieces/blob/main/packages/pieces/framework/src/lib/property/dropdown-prop.ts

Digitalguyco:
Title: Update type definition for StaticDropdownProperty and MultiSelectDropdownProperty

Description:
This pull request updates the type definition for `StaticDropdownProperty` and `MultiSelectDropdownProperty` to fix an issue where the `options` property was nested inside an object in the `DropdownState` type. The updated definition separates `options` from `DropdownState` and includes a `placeholder` property. 

I have also updated the return statement in the piece-helper.ts file to use the new `MainDropdownState<T>` type, which extends `DropdownState<T>` and includes the `placeholder` property.

Please let me know if you have any questions or feedback on this change.

Fixes #781

### DIFF
--- a/packages/engine/src/lib/helper/piece-helper.ts
+++ b/packages/engine/src/lib/helper/piece-helper.ts
@@ -3,7 +3,7 @@ import {
     Action,
     ActionContext,
     DropdownProperty,
-    DropdownState,
+    MainDropdownState,
     DynamicProperties,
     DynamicPropsValue,
     MultiSelectDropdownProperty,
@@ -216,7 +216,7 @@ export const pieceHelper = {
                 disabled: true,
                 options: [],
                 placeholder: "Throws an error, reconnect or refresh the page"
-            } as DropdownState<unknown>;
+            } as MainDropdownState<unknown>; //MainDropdownState extends DropdownState;
         }
     },
 
--- a/packages/engine/src/lib/helper/piece-helper.ts
+++ b/packages/engine/src/lib/helper/piece-helper.ts
@@ -3,7 +3,8 @@ import {
     Action,
     ActionContext,
     DropdownProperty,
-    MainDropdownState,
+    DropdownOption,
+    DropdownState,
     DynamicProperties,
     DynamicPropsValue,
     MultiSelectDropdownProperty,
@@ -216,7 +217,10 @@ export const pieceHelper = {
                 disabled: true,
                 options: [],
                 placeholder: "Throws an error, reconnect or refresh the page"
-            } as MainDropdownState<unknown>; //MainDropdownState extends DropdownState;
+            } as DropdownState<unknown> & {
+                placeholder?: string,
+                options: DropdownOption<unknown>[],
+            }
         }
     },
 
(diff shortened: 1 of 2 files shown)


packages/pieces/framework/src/lib/property/dropdown-prop.ts
```diff
@@ -5,29 +5,35 @@ import { OAuth2PropertyValue } from "./oauth2-prop";
...
+
 export type DropdownProperty<T, R extends boolean> = BasePropertySchema & {
 	refreshers: string[];
-	options: (propsValue: Record<string, undefined | OAuth2PropertyValue | number | string | object | BasicAuthPropertyValue | unknown[]>) => Promise<DropdownState<T>>
+	options: (propsValue: Record<string, undefined | OAuth2PropertyValue | number | string | object | BasicAuthPropertyValue | unknown[]>) => Promise<MainDropdownState<T>>
 } & TPropertyValue<T, PropertyType.DROPDOWN, R>;
 
 export type StaticDropdownProperty<T, R extends boolean> = BasePropertySchema & {
-	options: DropdownState<T>;
+	options: MainDropdownState<T>;
```
abuaboud: This still has the problem in the issue, where the user have to type
options inside options in static dropdown.

I am not sure why did you create MainDropdownState in that case

Digitalguyco: Turns out I didn't look at the problem that way, MainDropdownState was supposed to make more sense but ended up replicating the problem, however I've come up with a better solution.

packages/pieces/framework/src/lib/property/dropdown-prop.ts
```diff
@@ -5,22 +5,23 @@ import { OAuth2PropertyValue } from "./oauth2-prop";
 
 export type DropdownState<T> = {
 	disabled?: boolean;
-	placeholder?: string;
```
abuaboud, line 8: Can you revert this?

packages/engine/src/lib/helper/piece-helper.ts
```diff
@@ -216,7 +217,10 @@ export const pieceHelper = {
                 disabled: true,
                 options: [],
                 placeholder: "Throws an error, reconnect or refresh the page"
-            } as DropdownState<unknown>;
+            } as DropdownState<unknown> & {
```
abuaboud, line 220: casting shouldn't happen, can you handle the static case differently. that would make it work.

abuaboud (changes requested):

PR #1140, opened 2023-04-24
feat(piece/postgres): add checkbox to toggle SSL on or off

dentych:
## What does this PR do?

This PR adds a new checkbox to enable or disable SSL connection. Previously the connection would always attempt an SSL connection.

The "Enforce Encryption" has also been changed to "Verify server certificate", as that is more appropriate.


abuaboud (approved):

abuaboud:
@allcontributors please add @dentych for plugin

PR #1143, opened 2023-04-25
chore: added folder name validation frontend and backend wise

AbdulTheActivePiecer:
## What does this PR do?
Added validation when renaming/creating folder for its name.

### DIFF
--- a/packages/backend/src/app/flows/folder/folder.service.ts
+++ b/packages/backend/src/app/flows/folder/folder.service.ts
@@ -20,6 +20,14 @@ export const flowFolderService = {
             projectId,
             id: folderId,
         })
+        const folderWithDisplayName = await folderRepo.findOneBy({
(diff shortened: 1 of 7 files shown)


packages/backend/src/app/flows/folder/folder.service.ts
```diff
@@ -20,6 +20,14 @@ export const flowFolderService = {
             projectId,
             id: folderId,
         })
+        const folderWithDisplayName = await folderRepo.findOneBy({
+            projectId,
+            displayName:request.displayName
+        })
+        if(folderWithDisplayName && folderWithDisplayName.id !== folderId)
```
abuaboud, line 27: Make it database constraint unique on the folder name and project id

abuaboud (changes requested):

PR #1192, opened 2023-05-01
fix: return code test result as JSON

khaledmashaly:


packages/backend/src/assets/code-executor.js
```diff
@@ -1,23 +1,38 @@
...
+
+        const output = {
+            status: 'OK',
+            response: response ? response : ''
+        }
+
+        await writeOutput(output)
+    }
+    catch (e) {
+        // fill _standardError.txt
```
abuaboud, line 27: Add TODO

abuaboud (changes requested):

packages/backend/src/assets/code-executor.js
khaledmashaly: todo for what? please see the original comment on this line

PR #1205, opened 2023-05-02
feat: add StabilityAI piece (Text to Image)

Willianwg:
## What does this PR do?

Fixes # (issue)


abuaboud (approved):

packages/pieces/stability-ai/src/lib/actions/text-to-image.ts
```diff
@@ -0,0 +1,280 @@
...
+      clip_guidance_preset: clip_guidance_preset || 'NONE',
+      height: Number(height) || 512,
+      width: Number(width) || 512,
+      samples: Number(samples) || 1,
+      steps: Number(steps) || 50,
+      style_preset,
+    };
+
+    try {
+      const response = await axios.post(
```
abuaboud: can you use our http client, It's wrapper for axios and it also provide betters errors, so you don't need to catch it and reformat it.

You can remove catch part after changing http client

abuaboud (approved):

PR #1223, opened 2023-05-05
fix: save static sample data with app-webhook trigger

AbdulTheActivePiecer:
## What does this PR do?

Static sample data wasn't being saved for app-webhook triggers, this PR fixes that.

Tested this by stopping filterAppWebhooks method inside ActionMetaService, then added a slack trigger, selected "new message"   and switched to another step where I saw that the sample data was now being fetched in the autocomplete dropdown.

### DIFF
--- a/packages/ui/feature-builder-right-sidebar/src/lib/edit-step-sidebar/edit-step-form-container/edit-step-form-container.component.ts
+++ b/packages/ui/feature-builder-right-sidebar/src/lib/edit-step-sidebar/edit-step-form-container/edit-step-form-container.component.ts
@@ -123,9 +123,7 @@ export class EditStepFormContainerComponent {
           if (this._selectedStep.type === TriggerType.WEBHOOK) {
             this.updateNonAppWebhookTrigger(res.step!);
           } else {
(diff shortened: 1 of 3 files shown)


abuaboud (approved):

packages/ui/feature-builder-right-sidebar/src/lib/edit-step-sidebar/edit-step-form-container/edit-step-form-container.component.ts
```diff
@@ -99,23 +103,42 @@ export class EditStepFormContainerComponent {
...
-                res!
-              ) as UpdateTriggerRequest,
-            })
-          );
+          if (this._selectedStep.type === TriggerType.WEBHOOK) {
+            this.updateNonAppWebhookTrigger(res.step!);
+          } else {
+            const newTriggerSettings = this.createNewStepSettings(
+              res.step!
+            ) as PieceTriggerSettings;
```
abuaboud: You can avoid this by checking for type

abuaboud (approved):

PR #1236, opened 2023-05-08
[PIECE] Gotify

MyWay:
## What does this PR do?

This PR adds a new piece, Gotify. The piece has a single action, send notification, which allows users to easily send notifications using their Gotify instance. An app token is required.

### Parameters
Auth uses an app token which the user can generate from their web panel, or using Rest API.

Configurable parameters are:
- title (required)
- message (required)
- priority (optional, a number from 0 to 10)

### Official logo
Can be found at https://github.com/gotify/logo


abuaboud:
@allcontributors please add @MyWay  for plugin

Issue #1111, opened 2023-04-20
[BUG]: minimumSupportedRelease


abuaboud:
**Describe the bug**
Releasing new version of piece with minimumSupportedRelease, will make the frontend assume the old versions of that piece doesn't exist as well.

khaledmashaly:
Fixed in
- #1395 
- activepieces/activepieces-cloud#31
- #1400

PR #1244, opened 2023-05-10
feat: add private pieces

abuaboud:
## What does this PR do?

Implement private pieces regsitry

How to use

1. Use npm run help.

Tests:
1- Ran `publish hackernews` and it listed under lists & in the builder with test flow working fine.
2- Ran 'list` and delete `hackernews`


packages/backend/src/app/flows/common/piece-installer.ts
```diff
@@ -1,11 +1,17 @@
...
 import { system } from '../../helper/system/system'
 import { SystemProp } from '../../helper/system/system-prop'
 import { logger } from '../../helper/logger'
 import { packageManager } from '../../helper/package-manager'
+import { pieceMetadataLoader } from '../../pieces/piece-metadata-loader'
+import { ApEnvironment, ProjectId, getPackageAliasForPiece } from '@activepieces/shared'
+import { PieceType } from '@activepieces/pieces-framework'
+import { fileService } from '../../file/file.service'
+import * as fs from 'fs/promises'
+import * as path from 'path'
```
khaledmashaly: (a)
please use the following format for node builtins, just for the sake of consistency
```ts
import xyz from 'node:xyz'
```

packages/backend/src/app/pieces/piece-metadata-loader.ts
```diff
@@ -117,7 +124,7 @@ const filePieceMetadataLoader = (): PieceMetadataLoader => {
     }
 }
 
-const getPieceMetadataLoader = (): PieceMetadataLoader => {
+export const getPieceMetadataLoader = (): PieceMetadataLoader => {
```
khaledmashaly: I think this was exported by mistake.

packages/backend/src/app/flows/common/piece-installer.ts
```diff
@@ -69,3 +88,48 @@ export const pieceManager = {
...
+    try {
+        await fs.access(filePath)
+        return true
+    }
+    catch (error) {
+        return false
+    }
+}
+
+const getPackageLocalVersionForPiece = async (params: {
```
khaledmashaly: very confusing name for this method, the method does all sorts of stuff besides getting the piece version.

packages/backend/src/app/flows/common/piece-installer.ts
```diff
@@ -69,3 +88,48 @@ export const pieceManager = {
...
+    projectId: ProjectId
+}): Promise<string> => {
+    const tarFilePath = path.join('/', 'tmp', params.projectId, 'pieces', params.pieceName, params.pieceVersion, `${params.pieceName}.tar.gz`)
+    const fileAlreadyExists = await fileExists(tarFilePath)
+    const alias = 'file:' + tarFilePath
+    if (!fileAlreadyExists) {
+        const tarFileBuffer = await fileService.getOneOrThrow({
+            fileId: params.tarFileId,
+            projectId: params.projectId,
+        })
```
khaledmashaly: Should we just store pieces on disk unarchived? Is there any advantage to storing pieces as archives in db? what about storing them in something like `/private-pieces/${projectId}/${pieceName}/${pieceVersion}/package.json` instead of extracting the archive everytime a piece is used?

packages/cli/src/main.ts
```diff
@@ -0,0 +1,216 @@
...
+      deletePiece({
+        pieceName,
+      });
+    }
+    break;
+  }
+  default:
+    logger.warn(chalk.red("Unknown command."));
+    break;
+}
```
khaledmashaly: FYI, there's [yargs](https://github.com/yargs/yargs)

packages/pieces/framework/src/lib/piece.ts
```diff
@@ -46,11 +46,12 @@ export class Piece implements PieceBase {
     return this._triggers[triggerName];
   }
 
-  metadata(): PieceMetadata {
+  metadata(type = PieceType.PUBLIC): PieceMetadata {
```
khaledmashaly: why is `type` being passed by the caller?!
Shouldn't the piece contain the `type` information?!

khaledmashaly (commented):
I think all `PieceMetadata` stuff should be in `backend` not `ee` since they will be used to store pieces to solve issues like #1111 . we can add a separate entity like `PrivatePiece` for example that contains logic for private pieces which will be `ee` only

packages/backend/src/app/flows/common/piece-installer.ts
abuaboud: Done

packages/backend/src/app/pieces/piece-metadata-loader.ts
abuaboud: Done

packages/backend/src/app/flows/common/piece-installer.ts
abuaboud: It doesn't matter, pnpm accept both formats and the generated file from npm pack is tar file.

What do you think?

packages/backend/src/app/flows/common/piece-installer.ts
abuaboud: Check now

packages/pieces/framework/src/lib/piece.ts
abuaboud: Done moved it up

nileshtrivedi:
Would be good to see some documentation for adding private pieces.

PR #1248, opened 2023-05-12
feat: inline functions

### LINKED ISSUE #1018: Inline Functions
It could come really handy if I can convert Stripe's cents into dollars just in-line. Stripe returns the amount with 00 more, and I don't feel like adding a Code step to divide into 100. I'm also writing tutorials about this so it will make it harder for people to follow and understand why they need to use a code step for it.

Screenshot 2023-04-05 at 17 58 17

abuaboud:
## What does this PR do?

- Migrate everything from ${} to {{}} format
- Allow JS Execution inside {{}}

Fixes #1018

### DIFF
--- a/packages/engine/src/lib/services/variable-service.ts
+++ b/packages/engine/src/lib/services/variable-service.ts
@@ -3,7 +3,7 @@ import { ExecutionState } from '@activepieces/shared';
 import { connectionService } from './connections.service';
 
 export class VariableService {
-  private VARIABLE_TOKEN = RegExp('\\$\\{(.*?)\\}', 'g');
+  private VARIABLE_TOKEN = RegExp('\\{\\{(.*?)\\}\\}', 'g');
   private static CONNECTIONS = 'connections';
   private async resolveInput(input: string, valuesMap: Record<string, unknown>, censorConnections: boolean): Promise<any> {
     // If input contains only a variable token, return the value of the variable while maintaining the variable type.
@@ -13,7 +13,7 @@ export class VariableService {
       matchedTokens.length === 1 &&
       matchedTokens[0] === input
     ) {
-      const variableName = input.substring(2, input.length - 1);
+      const variableName = input.substring(2, input.length - 2);
       if (variableName.startsWith(VariableService.CONNECTIONS)) {
         return this.handleTypeAndResolving(variableName, censorConnections);
       }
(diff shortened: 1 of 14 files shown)


packages/engine/test/services/variable-service.test.ts
```diff
@@ -17,29 +18,35 @@ executionState.insertStep(
...
-    expect(await variableService.resolve('${configs.msg}', executionState)).toEqual(
-      'Hello World'
-    );
-  });
-
-  test('Test resolve text with configs variables', async () => {
+  test('Test resolve text with double variables', async () => {
     expect(
-      await variableService.resolve('Price is ${configs.price}', executionState)
+      await variableService.resolve('Price is ${trigger.price}', executionState)
```
khaledmashaly: Do these tests pass? Since the variable syntax was changed to `{{ }}` instead of `${ }`, right?

Also, It would be excellent if those tests can be added to the GitHub Actions pipeline.

packages/engine/src/lib/services/variable-service.ts
```diff
@@ -86,11 +56,26 @@ export class VariableService {
...
+      const values = Object.values(contextAsScope);
+      const functionBody = `return (${js})`;
+      const evaluatedFn = new Function(...keys, functionBody);
+      const result = evaluatedFn(...values);
+      return result ?? "";
+    } catch (exception) {
+      console.warn('Error evaluating expression', exception);
+      return "";
+    }
+  }
```
khaledmashaly, line 72: Please note that this could be used to execute any JavaScript code, right? I mean this can resolve `{{process.env}}` for example or any other JavaScript code for that matter. I know this is executed inside `isolate` but for unsandboxed mode, this could be potentially harmful, I think we should work on restricting code that can be run next isA.

khaledmashaly (commented):
Please see comments and merge if you think everything is ok

packages/engine/src/lib/services/variable-service.ts
abuaboud: Valid point, but let's revisit the isolation in the next sprint.

Running in isolated binary should be fine, as it's give the same privileged as js code.

packages/engine/test/services/variable-service.test.ts
abuaboud: Sorry fixed, no they dont and added to github actions.

PR #1283, opened 2023-05-17
PIECE: Mautic

bibhuty-did-this:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/pieces/mautic/src/lib/actions/create-contact.ts
+++ b/packages/pieces/mautic/src/lib/actions/create-contact.ts
@@ -1,0 +1,66 @@
+import { createAction } from '@activepieces/pieces-framework';
+import {
+    httpClient,
+    HttpHeaders,
+    HttpMethod,
+    HttpRequest
+} from "@activepieces/pieces-common";
+import {appender, mauticCommon} from "../common";
+
+export const createContact = createAction({
+    description: 'Creates a new contact in Mautic CRM', // Must be a unique across the piece, this shouldn't be changed.
+    displayName: 'Create Contact',
+    name: 'create_mautic_contact',
+    sampleData:{
+        "firstname":"firstname",
+        "lastname":"lastname",
+        "email":"email@email.com",
+        "tags":"danger",
+        "company":"company",
+    },
+    props: {
+        authentication: mauticCommon.authentication,
+        firstname: mauticCommon.firstname,
+        lastname: mauticCommon.lastname,
+        email: mauticCommon.email,
+        tags: mauticCommon.tags,
+        company: mauticCommon.company,
+    },
+    run: async function (context) {
+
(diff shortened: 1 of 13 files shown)


packages/pieces/mautic/src/lib/common/index.ts
```diff
@@ -0,0 +1,64 @@
...
+        'If this `https://mautic.ddev.site/s/dashboard` is your dashboard link \n' +
+        'then your base_url should be `https://mautic.ddev.site/`',
+    access_token: 'To authorize a request for basic authentication,\n' +
+        'first enable Basic Authentication by navigating to \n' +
+        'Settings -> Configuration -> API Settings, set an Authorization header.\n' +
+        '1. Combine the username and password of a Mautic user with a colon :. \n' +
+        'For example,if your username is `user` and password is `password` then , user:password.\n' +
+        '2. Base64 encode the string from above. You can go to this website `https://www.base64encode.org/.` for the same.  \n' +
+        'After auth it\'ll look like. dXNlcjpwYXNzd29yZA==.\n' +
+        '3. Copy and paste the same.',
```
abuaboud: If Mauatic users is more familiar with user and password, It might be better for the end user to add them directly and in the authentication we can create two proprieties `user` and `password` and just encode them in the http request, what do you think? 

I.E shifting the encoding and merging to the piece level.

packages/pieces/mautic/src/lib/actions/create-contact.ts
```diff
@@ -0,0 +1,71 @@
...
+            authentication,
+            firstname,
+            lastname,
+            email,
+            tags,
+            company,
+        } = context.propsValue;
+
+        const {base_url: BASE_URL, access_token} = authentication;
+
```
abuaboud, line 43: A common mistake by the user to forget adding /, maybe we can handle this here.
```
base_url = base_url.endsWith('/') ? base_url : base_url + '/';
```

packages/pieces/mautic/src/lib/actions/create-contact.ts
```diff
@@ -0,0 +1,71 @@
...
+        const request: HttpRequest = {
+            method: HttpMethod.POST,
+            url: BASE_URL + appender.CONTACT_NEW,
+            body,
+            headers
+        }
+        try {
+            return await httpClient.sendRequest(request);
+        }catch (error) {
+            console.log('mautic-create-contact: body\n', body);
```
abuaboud: Would there be any issues if it is thrown directly? I believe that the engine captures these HTTP wrapper errors and presents them in a user-friendly manner.

abuaboud (changes requested):

abuaboud (approved):

PR #1285, opened 2023-05-17
chore: enable strict mode & switch check lint rule

abuaboud:
## What does this PR do?

1- Enabled Check on missing cases in switch
2- Enable Strict mode
3- Fix Validate Props (Props has to be not null and not undefined but unknown)


khaledmashaly (approved):
Excellente!

PR #1291, opened 2023-05-18
feat: resume branches loops

khaledmashaly:
## What does this PR do?

Fixes # (issue)


packages/engine/src/lib/action/branch-action-handler.ts
```diff
@@ -28,52 +35,72 @@ export class BranchActionHandler extends BaseActionHandler<BranchAction> {
...
-        }
       }
+
+      executionState.insertStep(stepOutput, this.currentAction.name, ancestors);
+
+      const firstStep = condition
+        ? this.onSuccessAction
+        : this.onFailureAction
+
+      if (isNil(firstStep)) {
```
abuaboud: OnSuccessAction / onFailureAction is optional field, so throwing an error would make the flow stop.

packages/shared/src/lib/flow-run/execution/step-output.ts
```diff
@@ -11,6 +14,7 @@ export class StepOutput<T = any>{
   errorMessage?: unknown;
   standardOutput?: unknown;
   status?: StepOutputStatus;
+  pauseMetadata?: Omit<PauseMetadata, 'executionState'>
```
abuaboud, line 19: Why this should exists inside StepOutput? Maybe we can use the global in engine if we want to only the one that paused?

abuaboud (changes requested):

packages/engine/src/lib/action/branch-action-handler.ts
khaledmashaly: fixed in 0cf89088

packages/shared/src/lib/flow-run/execution/step-output.ts
khaledmashaly: because it needs to be recursive

Issue #931, opened 2023-03-29
Enhancement: HTTP Request Timeout & Return Response Code


ShayPunter:
**Is your feature request related to a problem? Please describe.**
Not a problem, just looking to build an uptime monitoring service from within Activepieces

**Describe the solution you'd like**
Adding the following to HTTP Request:
- Timeout (if it doesn't get a response from the site within X seconds)
- Returning status code from the site by default

**Describe alternatives you've considered**
An alternative solution I can implement would be some code which would do the request and parse it accordingly

bibhuty-did-this:
@abuaboud I would like to take this up.

bibhuty-did-this:
> http piece to not fail on error

@abuaboud 
Just confirming my understanding here. You want me to send an empty response in case the request fails right?

abuaboud:
> > http piece to not fail on error
> 
> @abuaboud Just confirming my understanding here. You want me to send an empty response in case the request fails right?

I think instead of throwing an error, if this option is turned on, we can try and catch it and return it as response to prevent http piece from failing.

bibhuty-did-this:
Adding [document](https://axios-http.com/docs/req_config) for future reference if needed

PR #1299, opened 2023-05-20
feat: HTTP -> Enhancement : adding timeout feature and failsafe option

bibhuty-did-this:
## What does this PR do?

Fixes https://github.com/activepieces/activepieces/issues/931

### DIFF
--- a/packages/pieces/http/src/lib/actions/send-http-request-action.ts
+++ b/packages/pieces/http/src/lib/actions/send-http-request-action.ts
@@ -5,6 +5,7 @@ import {
   HttpHeaders,
   QueryParams,
   httpClient,
+  HttpError,
 } from '@activepieces/pieces-common';
 import { httpMethodDropdown } from '../common/props';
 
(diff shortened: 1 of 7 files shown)


abuaboud:
@bibhuty-did-this This is great!

Can you please bump the versions i common package.json and http package.json.

packages/pieces/http/src/lib/actions/send-http-request-action.ts
```diff
@@ -24,10 +30,19 @@ export const httpSendRequestAction = createAction({
       displayName: 'Body',
       required: false,
     }),
+    failsafe: Property.Checkbox({
+      displayName: "On error don't abort",
```
abuaboud: I am suggesting to change display Name to be something like this:

`No Error on Failure`

What do you think?

bibhuty-did-this:
> @bibhuty-did-this This is great!
> 
> 
> 
> Can you please bump the versions i common package.json and http package.json.
> 
> 
> 
> 

@abuaboud I didn't test it. I'll test and bump the version. Will ping you post that 😊

packages/pieces/http/src/lib/actions/send-http-request-action.ts
ghost: Works. Will do the necessary change.

packages/pieces/http/src/lib/actions/send-http-request-action.ts
```diff
@@ -38,8 +53,23 @@ export const httpSendRequestAction = createAction({
...
       queryParams: queryParams as QueryParams,
       body,
+      timeout: timeout ? timeout * 1000 : 0,
     };
 
-    return await httpClient.sendRequest(request);
+    try {
+        const response = await httpClient.sendRequest(request);
+        return {
+            success: response.status && response.status == 200,
```
abuaboud: 2xx is considered success, for example some services might return 201 which means okay.

This will fail in that case.

ghost: Taken care of.

PR #1300, opened 2023-05-20
feat: add validation for number

abuaboud:
## What does this PR do?

1- Remove Property Number from OAuth2
2- Remove  Property Number from Dynamic Props (No Way to validate it currently)
3- Property Number get validated runtime, otherwise it throw an error.


packages/engine/src/lib/services/variable-service.ts
```diff
@@ -105,4 +106,28 @@ export class VariableService {
       censorConnections
     );
   }
+
+  validateAndCast(resolvedInput: any, props: PiecePropertyMap): { result: any, errors: Record<string, any> } {
+    const errors: Record<string, string | Record<string, string>> = {};
+    const clonedInput = JSON.parse(JSON.stringify(resolvedInput));
+    for (const [key, value] of Object.entries(resolvedInput)) {
+      const property = props[key];
+      if (property.type === PropertyType.NUMBER && value) {
```
AbdulTheActivePiecer: This will not parse zeros '0'

AbdulTheActivePiecer: check if value is not equal to undefined or null or empty-string

abuaboud: Added tests

PR #1301, opened 2023-05-21
fix: handle number as number not as string

bibhuty-did-this:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/pieces/framework/package.json
+++ b/packages/pieces/framework/package.json
@@ -1,5 +1,5 @@
 {
   "name": "@activepieces/pieces-framework",
-  "version": "0.3.19",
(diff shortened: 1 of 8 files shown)


abuaboud:
Thank you for taking this up, Sadly this change is more complex and I did open pr for it yesterday 

https://github.com/activepieces/activepieces/pull/1300

The whole issue is there is a need to be casting to number and runtime validation in the engine.


For example, you can use a number from the previous step, initially, that is a string (the name of the variable), and later at running flow the engine will resolve the real value and do the validation.


I am closing this pr as the above pr should solve it 😁

PR #1346, opened 2023-05-24
chore: add flow tests

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/backend/src/app/flows/flow-version/flow-version.service.ts
```diff
@@ -194,14 +194,51 @@ function getImportOperations(step: Action | Trigger | undefined): (FlowOperation
...
+            }
+        case ActionType.MISSING:
+            return {
+                type: action.type,
+                name: action.name,
+                displayName: action.displayName,
+                settings: action.settings,
+                valid: action.valid,
+            }
+    }
```
khaledmashaly, line 238: correct me if I'm wrong, but this function can be reduced to this, no?
```ts
function keepBaseAction(action: Action): Action {
    return {
        type: action.type,
        name: action.name,
        displayName: action.displayName,
        settings: action.settings,
        valid: action.valid,
    } as Action
}
```

packages/backend/src/app/flows/flow-version/flow-version.service.ts
```diff
@@ -288,47 +325,53 @@ async function prepareRequest(projectId: ProjectId, flowVersion: FlowVersion, re
...
+                    await uploadArtifact(projectId, codeSettings)
+                    const previousStep = flowHelper.getStep(flowVersion, clonedRequest.request.name)
+                    if (
+                        previousStep !== undefined &&
+                        previousStep.type === ActionType.CODE &&
+                        codeSettings.artifactSourceId !== previousStep.settings.artifactSourceId
+                    ) {
+                        await deleteArtifact(projectId, previousStep.settings)
+                    }
+                    break
```
khaledmashaly, line 371: The `prepareRequest` method is gigantic, I think it will be better to split it to multiple methods or even files

khaledmashaly (approved):
Approved, see non blocking comments

packages/backend/src/app/flows/flow-version/flow-version.service.ts
abuaboud: I will do this in follow up PR.

packages/backend/src/app/flows/flow-version/flow-version.service.ts
abuaboud: The reason behind this, you won't get compilation error when something is missing.

PR #1371, opened 2023-05-29
fix: unify code testing & step testing

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/backend/src/app/pieces/pieces.controller.ts
```diff
@@ -113,7 +113,10 @@ async function stats(): Promise<Record<string, PieceStats>> {
...
     for (const flowWithoutVersion of flows) {
         const flow = await flowService.getOneOrThrow({ id: flowWithoutVersion.id, projectId: flowWithoutVersion.projectId })
-        const trigger = flow?.version?.trigger
+        if(isNil(flow.version)) {
+            continue
+        }
+        const trigger = flow.version.trigger
         if (isNil(trigger)) {
             continue
         }
```
AbdulTheActivePiecer, line 122: These cases shouldn't exist and if they do you might want to throw an error instead to be knowledgeable of these cases.

packages/shared/src/lib/flows/step-run/dto/test-step-request.ts
```diff
@@ -1,6 +1,15 @@
-import { Type } from "@sinclair/typebox";
+import { Static, Type } from "@sinclair/typebox";
 
 export const CreateStepRunRequestBody = Type.Object({
     flowVersionId: Type.String(),
     stepName: Type.String(),
 })
+
+export const CreateStepRunResponse = Type.Object({
```
AbdulTheActivePiecer: Why not rename it to stepRunResponse ?

packages/backend/src/app/helper/engine-helper.ts
```diff
@@ -240,6 +248,28 @@ export const engineHelper = {
         }
     },
 
+    async executeCode(operation: ExecuteCodeOperation): Promise<EngineHelperResponse<EngineHelperCodeResult>> {
+        logger.debug(operation, '[EngineHelper#executeAction] operation')
+
+        const sandbox = await sandboxManager.obtainSandbox(`${apId()}`)
```
AbdulTheActivePiecer: Unnecessary string interpolation.

packages/backend/src/app/pieces/pieces.controller.ts
abuaboud: It does exists, I will investigate soon why they are there.

PR #1372, opened 2023-05-29
feat: add action create/update Salesforce contact

HKudria:
Added action for update contact by Id and create new contact for SalesForce crm


abuaboud:
@allcontributors please add @HKudria for plugin

PR #1377, opened 2023-05-29
feat: add webhook response

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/pieces/http/src/index.ts
```diff
@@ -1,14 +1,17 @@
 import packageJson from '../package.json';
 import { createPiece } from '@activepieces/pieces-framework';
 import { httpSendRequestAction } from './lib/actions/send-http-request-action';
+import { httpReturnResponse } from './lib/actions/return-response';
 
 export const http = createPiece({
 	name: 'http',
-	displayName: 'HTTP Request',
+	displayName: 'HTTP',
+	description: 'Sens HTTP requests and return responses',
```
JanHolger:
```suggestion
	description: 'Sends HTTP requests and return responses',
```

packages/pieces/http/src/lib/actions/return-response.ts
```diff
@@ -0,0 +1,33 @@
+import { Property, createAction } from '@activepieces/pieces-framework';
+import { StatusCodes } from 'http-status-codes';
+
+export const httpReturnResponse = createAction({
```
JanHolger, line 4: Does it really make sense that this is an action of the http client? Shouldn't there rather be a webhook piece for this?

JanHolger (commented):
This should also be implemented for webhook triggers. This would allow implementing Facebook webhooks which require a verification by returning a query parameter they sent

JanHolger (commented):
This should also be implemented for webhook triggers. This would allow implementing Facebook webhooks which require a verification by returning a query parameter they sent

packages/pieces/http/src/lib/actions/return-response.ts
abuaboud: We thought since people are http request, they would expect that return is in http piece.

We also modified the description so it's searchable, I think we can go with that and see if people finds it

packages/pieces/http/src/index.ts
abuaboud: Thank you

PR #1386, opened 2023-05-30
PIECE: MySQL

### LINKED ISSUE #956: Mysql Piece
**Describe your usecase.**
Mysql Piece to perform queries, or retrieve data to send to the next piece in the flow.

Bonus points if this can be done over an SSH tunnel to access non-exposed mysql hosts (a common scenario, as most default mysql configs listen on 127.0.0.1 or sockets only)

JanHolger:
## What does this PR do?

Fixes #956

### DIFF
--- a/packages/pieces/mysql/src/lib/actions/delete-row.ts
+++ b/packages/pieces/mysql/src/lib/actions/delete-row.ts
@@ -2,18 +2,13 @@ import { createAction, Property } from "@activepieces/pieces-framework";
 import { mysqlCommon, mysqlConnect } from "../common";
 
 export default createAction({
-    name: 'update_rows',
-    displayName: 'Update Rows',
-    description: 'Updates one or more rows in a table',
+    name: 'delete_row',
+    displayName: 'Delete Row',
+    description: 'Deletes one or more rows from a table',
     props: {
         authentication: mysqlCommon.authentication,
         timezone: mysqlCommon.timezone,
         table: mysqlCommon.table(),
-        values: Property.Object({
-            displayName: 'Values',
-            description: 'Values to be updated',
-            required: true
-        }),
         condition: Property.ShortText({
             displayName: 'Condition',
             required: true
@@ -25,13 +20,10 @@ export default createAction({
         })
     },
     async run(context) {
-        const fields = Object.keys(context.propsValue.values)
-        const values = fields.map(f => context.propsValue.values[f])
-        const qsValues = fields.map(f => "`" + f + "`=?").join(',')
-        const qs = "UPDATE `" + context.propsValue.table + "` SET " + qsValues + " WHERE " + context.propsValue.condition + ";"
+        const qs = "DELETE FROM `" + context.propsValue.table + "` WHERE " + context.propsValue.condition + ";"
         const conn = await mysqlConnect(context.propsValue);
         try {
-            const result = await conn.query(qs, [ ...values, ...(context.propsValue.args || []) ])
+            const result = await conn.query(qs, context.propsValue.args)
             await conn.end()
             return result
         } catch(e) {
--- a/packages/pieces/mysql/src/lib/actions/delete-row.ts
+++ b/packages/pieces/mysql/src/lib/actions/delete-row.ts
@@ -1,5 +1,5 @@
 import { createAction, Property } from "@activepieces/pieces-framework";
-import { mysqlCommon, mysqlConnect } from "../common";
+import { mysqlCommon, mysqlConnect, sanitizeColumnName } from "../common";
 
 export default createAction({
     name: 'delete_row',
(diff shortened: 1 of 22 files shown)


packages/pieces/mysql/src/lib/common/index.ts
```diff
@@ -0,0 +1,92 @@
...
+export const mysqlCommon = {
+    authentication: Property.CustomAuth({
+        displayName: "Authentication",
+            props: {
+                host: Property.ShortText({
+                    displayName: 'Host',
+                    required: true,
+                    description: "The hostname or address of the mysql server"
+                }),
+                port: Property.ShortText({
```
abuaboud: Can you please change the following into `Property.Number`

packages/pieces/mysql/src/lib/common/index.ts
```diff
@@ -0,0 +1,92 @@
...
+                    displayName: 'Port',
+                    defaultValue: '3306',
+                    description: "The port to use for connecting to the mysql server",
+                    required: true,
+                }),
+                user: Property.ShortText({
+                    displayName: 'User',
```
abuaboud: Please displayName change to `Username`

packages/pieces/mysql/src/lib/actions/update-rows.ts
```diff
@@ -0,0 +1,42 @@
...
+    props: {
+        authentication: mysqlCommon.authentication,
+        timezone: mysqlCommon.timezone,
+        table: mysqlCommon.table(),
+        values: Property.Object({
+            displayName: 'Values',
+            description: 'Values to be updated',
+            required: true
+        }),
+        condition: Property.ShortText({
```
abuaboud: This provides high flexibility, but it's hard for tech savvy, I would replace this action with `Update Row`.

This means to replace condition with `Column name` and `Column Value`, my assumption that in most cases the user will update row by id or something unique like email, If not execute query always an option.

packages/pieces/mysql/src/lib/actions/select-rows.ts
```diff
@@ -0,0 +1,41 @@
+import { createAction, Property } from "@activepieces/pieces-framework";
+import { mysqlCommon, mysqlConnect, isSpecialColumn } from "../common";
+
+export default createAction({
+    name: 'select_rows',
```
abuaboud: This is interesting, but it doesn't solve a problem for developers, as most of them will execute a query.

If you aim to make such action no-code friendly, I suggest it be labeled as `Find Row`, and the props Table / Lookup Column / Column Value.

packages/pieces/mysql/src/lib/actions/delete-rows.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export default createAction({
+    name: 'delete_rows',
+    displayName: 'Delete Rows',
+    description: 'Deletes one or more rows from a table',
+    props: {
+        authentication: mysqlCommon.authentication,
+        timezone: mysqlCommon.timezone,
+        table: mysqlCommon.table(),
+        condition: Property.ShortText({
```
abuaboud: The same suggestion regarding `Find Row` / `Update Row` can be used here, replacing it with a simple condition for a no-code user. 

Cover delete by single column easily and for more sophisticated case, execute query always an option.

abuaboud (changes requested):
Let me know what do you think, will try to do E2E tests tomorrow as well.

packages/pieces/mysql/src/lib/actions/update-rows.ts
JanHolger: I thought about this as well, but it would not allow to use other operators than "=". I personally often have queries like "update everything before date x" or "disable all users that have not logged in for x days" but you are probably right, that those helpers should be as user friendly as possible. I'll change them

packages/pieces/mysql/src/lib/actions/delete-row.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+    async run(context) {
+        const qs = "DELETE FROM `" + context.propsValue.table + "` WHERE " + context.propsValue.condition + ";"
+        const conn = await mysqlConnect(context.propsValue);
+        try {
+            const result = await conn.query(qs, context.propsValue.args)
+            await conn.end()
+            return result
+        } catch(e) {
+            await conn.end()
```
abuaboud, line 28: Nit: this can be moved to finally instead of two conn.end()

packages/pieces/mysql/src/lib/actions/select-rows.ts
JanHolger: Renamed it to Find Rows but kept it plural and also kept the condition as is. I added a description to the condition field to make sure it's clear that it's a sql condition.

abuaboud (approved):

PR #1395, opened 2023-05-31
Fix minimum supported release

khaledmashaly:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -33,8 +33,6 @@ services:
       POSTGRES_DB: activepieces
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: A79Vm5D4p2VQHOp2gd5
-    ports:
-     - "5432:5432"
     volumes:
       - postgres_data_dev:/var/lib/postgresql/data
     networks:


docker-compose.dev.yml
```diff
@@ -33,6 +33,8 @@ services:
       POSTGRES_DB: activepieces
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: A79Vm5D4p2VQHOp2gd5
+    ports:
```
abuaboud: Please remove as it could cause issues for other developer due to port conflcit

packages/backend/src/app/pieces/piece-metadata-service/cloud-piece-metadata-service.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    if (response.status !== StatusCodes.OK) {
+        throw new Error(await response.text())
+    }
+}
+
+export const CloudPieceMetadataService = (): PieceMetadataService => {
+    return {
+        async list(): Promise<PieceMetadataSummary[]> {
+            const response = await fetch(`${CLOUD_API_URL}/latest.json`)
```
abuaboud, line 28: I assume this will be changed once it's ready, just resolve this comment if that is correct

abuaboud (approved):

packages/backend/src/app/pieces/piece-metadata-service/cloud-piece-metadata-service.ts
khaledmashaly: yup correct, so that things don't break after merging

PR #1402, opened 2023-06-01
feat: add file property

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/engine/src/lib/services/variable-service.ts
```diff
@@ -127,6 +130,66 @@ export class VariableService {
...
+          };
+        }
+
+      }
+      const response = await axios.head(urlOrBase64);
+      const contentType = response.headers['content-type'];
+
+      // Check if content type is file
+      if (!contentType || !(contentType.startsWith('application/') || contentType.startsWith("image") || contentType === 'application/octet-stream')) {
+        return null;
```
khaledmashaly, line 168: quick question, why are file types restricted to these types? we can have for example `text/html` and many others, right?

khaledmashaly (approved):

PR #1409, opened 2023-06-03
feat: support native development and remove docker dev

abuaboud:
## What does this PR do?

Fixes # (issue)


docker-compose.dev.yml
```diff
@@ -1,33 +1,17 @@
+## **************** DOCUMENTATION ****************
+##
+## Please refer to the documentation for more information on how to use this file. 
+## https://www.activepieces.com/docs/contributing/development-setup/docker-compose
+##
+## ***********************************************
 version: '3.0'
 
 services:
-  app:
```
JanHolger, line 4: I would strongly suggest to keep it and use [profiles](https://docs.docker.com/compose/profiles/) instead

abuaboud: Done!

turbowatch.ts
```diff
@@ -11,7 +11,6 @@ export default defineConfig({
             persistent: false,
             onChange: async ({ spawn, first, files } :ChangeEvent) => {
                 if (first) {
-                    await spawn`nx run-many -t build --projects=pieces-* --parallel=10`
                     return
```
khaledmashaly, line 16: why not provide a set of 2-3 pieces that are built by default?

khaledmashaly (approved):

PR #1410, opened 2023-06-03
PIECE: Implement a store append action

### LINKED ISSUE #1354: Add "Append" to Storage
In many cases, I had to get a Storage value and append a new value to it, it was two steps and the input wasn't clean. I suggest adding an "Append" action to storage that concatenates the original value with the new one I add there. If the value didn't exist, it can also create it and in that case it does the Put job.

JanHolger:
## What does this PR do?

Fixes #1354

### DIFF
--- a/packages/pieces/store/package.json
+++ b/packages/pieces/store/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-store",
-  "version": "0.2.6"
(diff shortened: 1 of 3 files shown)


abuaboud (dismissed):
Thank you Jan!

abuaboud (approved):

PR #1411, opened 2023-06-04
feat: enhanced delay step using pause/resume

### LINKED ISSUE #847: Add a delay  - seconds, minutes, days
Currently, we have a feature to add a delay in seconds but with a limit of 5 minutes. This limit is not flexible enough for my use case, where I want to be able to set a delay in seconds, minutes, or days.

I would like to have the ability to set a delay in seconds, minutes, or days, with no limits on the maximum value. This will allow me to send reminders after a specified amount of time, such as sending a follow-up email after 3 days if the original email has not been read.

A user suggested using another flow with a trigger every minute to follow up on unread emails, but this is not user-friendly for the end user. I would prefer a more intuitive and streamlined solution integrated into my current system.

**Additional context**
This feature would be useful for automating follow-up tasks and improving my workflow.

nileshtrivedi:
## What does this PR do?

Fixes #847 

- Supports longer delay durations using pause/resume
- Add a new action `delayTill` which takes a timestamp parameter instead of duration

### DIFF
--- a/packages/pieces/delay/src/index.ts
+++ b/packages/pieces/delay/src/index.ts
@@ -1,7 +1,7 @@
 import { createPiece } from '@activepieces/pieces-framework';
 import packageJson from '../package.json';
 import { delayForAction } from './lib/actions/delay-for-action';
-import { delayTillAction } from './lib/actions/delay-till-action';
+import { delayUntilAction } from './lib/actions/delay-untill-action';
 
 export const delay = createPiece({
   name: 'delay',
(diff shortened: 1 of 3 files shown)


khaledmashaly (changes requested):
Great!

There are two issues that need to be fixed
- Separate each action into its own file
- Use `dayjs()` for date-related logic

Thank you!

abuaboud:
Please bump the version in `package.json`

abuaboud:
Approved and merged. This is decent.

Thank you, @nileshtrivedi. I have added you as an author and renamed "till" to "until."

abuaboud (approved):

abuaboud (approved):

PR #1418, opened 2023-06-05
feat: add pastebin piece

### LINKED ISSUE #1417: [PIECE] Pastebin
**Describe your usecase.**
A pastebin piece for creating and reading pastes

Using the http client is hard (or might not even work) because their "api" consists of php scripts that only accept x-www-form-urlencoded

JanHolger:
## What does this PR do?

Fixes #1417


packages/pieces/pastebin/src/lib/actions/create-paste.ts
```diff
@@ -0,0 +1,69 @@
+import { Property, createAction } from "@activepieces/pieces-framework";
+import { makeClient, pastebinCommon } from "../common";
+import { PasteExpiry, PastePrivacy } from "../common/client";
+
+export default createAction({
+    name: 'create_paste',
+    displayName: 'Create Paste',
+    description: 'Creates a new paste',
+    props: {
+        authentication: pastebinCommon.authentication(true),
```
abuaboud, line 10: image

I added the following, but not sure why username and password is required.

Can you help me with that?

JanHolger: it's not required it's optional and only needed if you want to create private pastes

JanHolger: I also tried to move username and password into a different credential set but it seems like only a single credential set per piece is allowed right now since it will show me the dev token credential in the selection for the username/password credential so I moved it into a single one with optional props

abuaboud: Added note, good point!

abuaboud (approved):

PR #1443, opened 2023-06-08
feat: add community npm (API only)

abuaboud:
## What does this PR do?

The PR is API only where the user can provide npm name and version to import the piece.

Fixes # (issue)


packages/backend/src/app/pieces/piece-metadata-service/cloud-piece-metadata-service.ts
```diff
@@ -24,24 +30,68 @@ const handleHttpErrors = async (response: Response) => {
...
+                return piece
+            }
+            const savedPiece: PieceMetadata = pieceMetadataHelper.toPieceMetadata(await repo.save({
+                id: apId(),
+                minimumSupportedRelease: pieceMetadata.minimumSupportedRelease ?? '0.0.0',
+                maximumSupportedRelease: pieceMetadata.maximumSupportedRelease ?? '99999.99999.99999',
+                projectId: projectId,
+                ...pieceMetadata,
+            }))
+            return savedPiece
```
khaledmashaly: This code is just a clone of the `DbPieceMetadataService` in the cloud edition, It will be more maintainable to move the `DbPieceMetadataService` to the open source repo and build something like an `AggregatePieceMetadataService` in the open source version that _aggregates_ the output from both the `CloudPieceMetadataService` and the `DbPieceMetdataService`, to avoid code duplication.

packages/backend/src/app/pieces/pieces.controller.ts
```diff
@@ -1,20 +1,57 @@
...
+        },
+    )
+
+    app.get('/', {
+        schema: {
+            querystring: Type.Object({
+                release: SemVerType,
+            }),
+        },
+    }, async (req): Promise<PieceMetadataSummary[]> => {
```
khaledmashaly, line 45: Let's use the already existing style for this controller methods, eg:
```ts
app.get('/', ListPiecesRequest, async (req): Promise<PieceMetadataSummary[]> => {
    const { release } = req.query

    return await pieceMetadataService.list({
        release,
        projectId: req.principal.projectId,
    })
})
```
vs
```ts
app.post(
    '/',
    {
        schema: {
            body: InstallPieceRequest,
        },
    },
    async (
        request: FastifyRequest<{
            Body: InstallPieceRequest
        }>,
    ) => {
        const {result} = await engineHelper.extractPieceMetadata({
            pieceName: request.body.pieceName,
            pieceVersion: request.body.pieceVersion,
        })
        return pieceMetadataService.create({
            projectId: request.principal.projectId,
            pieceMetadata: {
                ...result,
                name: request.body.pieceName,
                version: request.body.pieceVersion,
            },
        })
    },
)
```

packages/shared/src/lib/authentication/model/principal-type.ts
```diff
@@ -1,4 +1,5 @@
 export enum PrincipalType {
     USER = 'USER',
-    WORKER = 'WORKER'
+    WORKER = 'WORKER',
+    UNKNOWN = 'UNKNOWN'
```
khaledmashaly, line 4: I think `ANONYMOUS` is more descriptive than `UNKNOWN`

khaledmashaly (commented):
Solid! please see comments

PR #1456, opened 2023-06-10
feat: add validators and processors for property framework

### LINKED ISSUE #771: Properties: Field Validation
Theres a need for validation at property level.

On creation of a property, there needs to be an option to provide, either a callback that can do the validation, or do the validation under the hood. 
The callable is a better option, as it can be easier to implement.

Otherwise, the input depending on type can provide its means of validation;
- for text, regex
- for number, max, min value
- for date/time, date/time format
etc

Validation inside the action/trigger is possible, but repetitive.

OsamaHaikal:
## What does this PR do?

Fixes #771

Add validators and processors support for property framework

The processors currently act as data transformers, for example converting'' to NaN, or URL to a file

The validators will be applied to the processed data to validate them,

New keys have been added to TPropertyValue

1. 	defaultProcessors?: ProcessorFn[]; 
2. 	defaultValidators?: ValidatorFn[];
3. 	processors?: ProcessorFn[];
4. 	validators?: ValidatorFn[];

and for file, datetime, and number a default processor and validator have been added. like in the property.ts file

processors, and validators you can use them as follows:
```
      text: Property.LongText({
        displayName: 'Text',
        required: true,
        validators: [Validators.url, Validators.oneOf(["activepieces.com", "www.activepieces.com"])],
      }),
      value1: Property.Number({
        displayName: 'Age',
        required: true,
        validators: [Validators.maxValue(2), Validators.oneOf(Object.values(choices))],
      }),
      value2: Property.Number({
        displayName: 'Age',
        required: true,
        validators: [Validators.inRange(5, 10)],
      }),
      value3: Property.Number({
        displayName: 'Age',
        required: true,
        validators: [Validators.minValue(10)],
      }),
```

the errors returned for each propety is now array of strings instead of string


packages/pieces/framework/src/lib/validators/validators.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  static url(property: PieceProperty, processedValue: any, userInput: any): string | null {
+    return urlValidator(property, processedValue, userInput)
+  }
+
+  static pattern(regex: string | RegExp): ValidatorFn {
+    return patternValidator(regex);
+  }
+
+  static datetime_iso(property: PieceProperty, processedValue: any, userInput: any): string | null {
```
abuaboud: I think the naming convention should be camel case.

packages/pieces/framework/src/lib/property/base-prop.ts
```diff
@@ -9,6 +11,10 @@ export type TPropertyValue<T, U extends PropertyType, REQUIRED extends boolean>
 	valueSchema: T;
 	type: U;
 	required: REQUIRED;
+	defaultProcessors?: ProcessorFn[];
```
abuaboud: Another option would be explicit types

```ts
export type ProcessorFn<INPUT,OUTPUT> = (value: INPUT) => OUTPUT;

const StringToNumberConvertor: ProcessorFn<string, number> = (value: string) => {
    return 1
}

const StringToStringConvertor: ProcessorFn<string, string> = (value: string) => {
    return "1"
}

interface StringProperty{
    processors: typeof StringToStringConvertor;
}

// Error
const stringProperty: StringProperty = {
    processors: StringToNumberConvertor
}

// Working
const stringTwoProperty: StringProperty = {
    processors: StringToStringConvertor
}
```

abuaboud (changes requested):

PR #1469, opened 2023-06-12
fix: hide-instance-information

### LINKED ISSUE #1209: hide activepieces instance information
**Describe the bug**
Not exactly a bug, but when you are not logged in, you are still able to leak some info of the running instance, because activepieces reports the availability of new updates before logging in, in the login screen and via the flags api page. 

**To Reproduce**
1. Logout if you are logged in 
2. Go to the login page or GET the flags endpoint 

**Expected behavior**
I shouldn't be able to get any info about the installation, such as the current version, or the availability of updates. 

**Screenshots**
No need to. 

**Additional context**
None.

abdullahranginwala:
## What does this PR do?
In the `/v1/flags` route, the sensitive information regarding the AP instance version is only visible for an authenticated request

Fixes #1209

### DIFF
--- a/packages/backend/src/app/flags/flag.controller.ts
+++ b/packages/backend/src/app/flags/flag.controller.ts
@@ -2,8 +2,11 @@ import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify'
 import { flagService } from './flag.service'
 
 export const flagController = async (app: FastifyInstance) => {
+
     app.get('/', async (_request: FastifyRequest, reply: FastifyReply) => {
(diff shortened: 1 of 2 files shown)


abuaboud:
Caching the information while unauthenticated can potentially cause issues in some parts of the frontend (try open it from login page and navigate to the builder without refresh, pieces won't load), it might be better to address this later when there is more capacity available. 

In the meantime, it would be best to close the pull request without making these changes. This will ensure that the current behavior, which may not be ideal but is functional, remains in place.

PR #1473, opened 2023-06-13
fix: support bracket notation

### LINKED ISSUE #1359: [BUG]: Switch to Bracket notion
**Bug Description**
If a JSON key contains a space or dot, it will cause a bug. To resolve this, we need to switch from using this notation: `trigger.queryParams` to `trigger['queryParams']`.

AbdulTheActivePiecer:
## What does this PR do?

Closes #1359.

### DIFF
--- a/packages/ui/feature-builder-form-controls/src/lib/interpolating-text-form-control/utils.ts
+++ b/packages/ui/feature-builder-form-controls/src/lib/interpolating-text-form-control/utils.ts
@@ -186,20 +186,6 @@ export function fromOpsToText(operations: QuillEditorOperationsObject) {
     .join('');
   return result;
 }
-const dotRegex = /\./g;
-export function replaceDotsWithSpaces(str: string) {
-  return str.replace(dotRegex, ' ');
-}
-export const arrayNotationRegex = /\[[0-9]+\]/g;
-export function replaceArrayNotationsWithSpaces(str: string) {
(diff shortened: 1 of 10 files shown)


packages/ui/feature-builder-form-controls/src/lib/interpolating-text-form-control/utils.ts
```diff
@@ -10,6 +10,48 @@ import { DomSanitizer } from '@angular/platform-browser';
...
+  }
+
+  return result.map((w) => {
+    if (w.startsWith(`"`) || w.startsWith(`'`)) {
+      return w.slice(1, w.length - 1);
+    }
+    return w;
+  });
+};
+
```
abuaboud, line 54: Replace with

```ts
const keysWithinPath = (path: string) => {
  const result = [];
  let word = '';
  for (let i = 0; i < path.length; i++) {
    const char = path[i];
    if (char === '.' || char === '[' || char === ']') {
      if (word) {
        result.push(word.replace(/^['"]|['"]$/g, ''));
        word = '';
      }
    }
    if (char !== '.' && char !== '[' && char !== ']') {
      word += char;
    }
  }
  if (word) {
    result.push(word.replace(/^['"]|['"]$/g, ''));
  }
  return result;
};
```

abuaboud (approved):

PR #1489, opened 2023-06-14
feat: add google sheets new find rows & better output for actions 

### LINKED ISSUE #1484: Google Sheets: Lookup Row
I would like to add an Action to the Google Sheets piece to lookup row based on cell content. This is a very frequently-used action in my Zapier automations.

Ideally it would read the entire contents of a row in a specific Google Sheet based on the value of a single cell in a specified content. Properties would be a row (described by the heading) to search and a value to search for. Output would include the values in that row as a hash, identified by the header, as well as the row number, which is useful for the Update Row Action.

A more advanced option would be to create a row if a matching row is not found, with all of the columns described by their headers.

As mentioned, this is one thing I do a lot of in Zapier and would be very helpful!

Abdallah-Alwarawreh:
## What does this PR do?

Adds new action for finding rows in a sheet, and changes all google sheets actions output

Fixes #1484


packages/pieces/google-sheets/src/lib/actions/find-rows.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (!sheetName) {
+            throw Error("Sheet not found in spreadsheet");
+        }
+        const alphabet = 'abcdefghijklmnopqrstuvwxyz';
+        // find the column index
+        const column = alphabet.indexOf(context.propsValue.column_name.toLowerCase());
+        if (column === -1) {
+            throw Error("Column not found in sheet");
+        }else{
+            const values = await googleSheetsCommon.getValues(context.propsValue.spreadsheet_id, context.propsValue['authentication']['access_token'], context.propsValue.sheet_id);
```
abuaboud, line 36: Can you move the conversion logic to getValue so it's the same logic in new row trigger as well.

abuaboud (approved):

PR #1516, opened 2023-06-18
feat: add remove button to optional properties

### LINKED ISSUE #1082: Remove Optional Input Button
It is not possible to remove optional input fields after adding it. So if i don't want to use it anymore i would need to create another step to replace the one that i am using. If i let the input blank, instead of undefined, the property will have a string "". If i am integrating with an update endpoint, this could update fields that i don't want to as blank.

Option to remove optional fields after adding it.

Create a small button somewhere near the field to remove it.

AbdulTheActivePiecer:
## What does this PR do?

Closes #1082.

Property types tested:
1)Any singular property using the textTemplate.
2)All dropdowns including multiselect and static versions.
3)Authentication dropdown.
4)Dictionary.
5)Array. 
6)Checkbox (slide toggle)
7)JSON.

I haven't accounted for optional dynamic properties, as I don't know how to handle removing a refresher atm.
I have this suggestion:
1)Make compilation of pieces throw an error if an optional dynamic property is used.
2)Throw an error if a refresher is an optional property, so basically any optional property should have an empty array of refreshers if it is of type (dropdown, multiselect dropdown or dynamic) .

### DIFF
--- a/packages/ui/feature-builder-form-controls/src/lib/array-form-control/array-form-control.component.ts
+++ b/packages/ui/feature-builder-form-controls/src/lib/array-form-control/array-form-control.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { ChangeDetectionStrategy, Component, ViewChild } from '@angular/core';
 import {
   ControlValueAccessor,
   FormArray,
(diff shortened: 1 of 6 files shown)


AbdulTheActivePiecer:
Don't merge this, because I will need to make the new insert data UX first, then we can merge it, but if you have comments let me know.

abuaboud (approved):

PR #1547, opened 2023-06-25
feat: Disable taking sample data by default from piece triggers

### LINKED ISSUE #1508: [BUG]: Testing experience bugs
**Describe the bug**

The user will try to press Test Flow after importing the flow, while there is no sample data (Disable Test Flow and add tooltip to test trigger)

AbdulTheActivePiecer:
## What does this PR do?

Currently, if the flow's trigger is not tested, it gets sample data from the sample data within the code for the trigger, this PR stops that and asks the user to test the trigger if it doesn't have selected sample data, unless it is a schedule, and fixes minor issues with templates dialog. 

Closes #1508


abuaboud:
Remove webhook dialog no longer used

abuaboud (approved):

PR #1549, opened 2023-06-25
feat: show indicator for add btn

AbdulTheActivePiecer:
## What does this PR do?

Currently if you click on an add button within the flow, you would need to remember the add button you clicked to know where the step will be placed.
To fix this this PR adds an indicator to show where the new step will be placed.
image
image


abuaboud (dismissed):
Good catch

abuaboud (approved):

PR #1560, opened 2023-06-27
feat: support ^ to load latest major version for piece

abuaboud:
## What does this PR do?

Add the ability to include versions with `^` like npm does, which brings the latest major version. I also updated the documentation accordingly, the frontend still doesn't use this functionality.

Please read the documentation.

image

I want to learn about tools we can use to avoid causing unintended widespread problems.

Here are a few examples of tools that might be helpful when we release a faulty version:
1. Gradual migration: Instead of instantly switching to the new version, we can transition gradually, **If needed how it works exactly?**
2. Rollback capability: We should have the ability to revert back to a previous working version if needed, **If needed how it works exactly?**
3. Publishing fixes: We can release a patch version with the necessary fixes and improvements, everything will be auto upgraded.

I'm curious to know what other scenarios and tools you think we should consider.


docs/developers/piece-reference/piece-versioning.mdx
```diff
@@ -0,0 +1,34 @@
...
+## Semantic Versioning
+
+The version number consists of three numbers: `MAJOR.MINOR.PATCH`, where:
+
+- **MAJOR** It should be incremented when there are breaking changes to the piece.
+- **MINOR** It should be incremented for new features or functionality that is compatible with the previous version.
+- **PATCH** It should be incremented for bug fixes and small changes that do not introduce new features or break backward compatibility.
+
+## Engine
+The engine will always try to use latest **major** version when the user selected the piece.
```
khaledmashaly: suggestion:
```
The engine will use the latest compatible version for a given piece version, i.e. latest version with the same **major** version number
```

docs/developers/piece-reference/piece-versioning.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Examples
+<Tip>
+when you make a change, remember to increment the version accordingly.
+</Tip>
+
+Changes that are **not backward compatible** include:
+- Changing the logical behavior of actions or triggers.
```
khaledmashaly: suggestion:
```md
### Breaking changes
- remove an existing action
- add a required action prop
- remove an existing action prop, whether required or optional
- remove an attribute from an action output

### Non-breaking changes
- add a new action
- add an optional action prop
- add an attribute to an action output

i.e. any removal is breaking, any required addition is breaking, everything else is not breaking
```

khaledmashaly (approved):
This only addresses new flows. Any plans for old flows or is this the intended behaviour?

abuaboud:
@khaledmashaly We dont know if the old pieces are following same versioning, it's better not break it at this moment, what do you think?

khaledmashaly:
> @khaledmashaly We dont know if the old pieces are following same versioning, it's better not break it at this moment, what do you think?

 yup, agreed

PR #1575, opened 2023-06-30
feat(invoice-ninja): create task 

buttonsbond:
A piece that creates a task on an instance of Invoice Ninja


abuaboud:
@allcontributors please add @buttonsbond for plugin

PR #1595, opened 2023-07-02
fix: facebook-leadds get lead data

MoShizzle:
## What does this PR do?

Webhook only receives a ping on lead creation that does not contain lead form data, added code to retrieve data for incoming leads


abuaboud (dismissed):

abuaboud:
Please bump the version

Issue #1568, opened 2023-06-29
Google Sheets Enhancement


abuaboud:
**Is your feature request related to a problem? Please describe.**
- Use First Row as Labels in all actions
- Support Clear whole sheets expect first row.

zdaar:
Suggestions : 
Add a range of column to the query, can go from a single column to a range like in any sheets with standard syntax A:Z
Add starting point because many sheets have headers that may be undesirable to parse so you dont need to do housekeeping in post processing
Add a row limit to stop iterating on row retrieval
return data in a tree like
rows
1
A : Debit
B: 23-06-25
C: Uncategorized
2
A: account
B: 23-06-25
C: Bills

or as json

```
{
  "1": {
    "A": "Debit",
    "B": "23-06-25",
    "C": "Uncategorized"
  },
  "2": {
    "A": "account",
    "B": "23-06-25",
    "C": "Bills"
  }
}
```

PR #1605, opened 2023-07-03
feat(google-sheets): enhanced google sheets

### LINKED ISSUE #1574: [BUG]: Google Sheet piece "Find Rows" action unexpected behavior
**Describe the bug**
Google Sheets piece "Find rows" action seems broken

**To Reproduce**
- Have a  spreadsheet containing data
- Go to google sheets piece configuration
- Write a column in the "Column Name" field of the piece which does NOT contain the data you want to look for (i/e if there is no string with the word "cleared" in column E, write E in that field
- write "cleared" in the "Search Value" field
- Matches are returned as values with seemingly unrelated column letters as keys

**Expected behavior**
- If a column does not contain the data from the Search Values, return "No matches"
- Instead of a seemingly random letter as a key in the reponse, write the row number

**Screenshots**
image
msedge_DfR6lmgf84

Abdallah-Alwarawreh:
## What does this PR do?

Enhanced google sheets, fixed find bug & added clear sheet action

Fixes #1574 #1568


packages/pieces/google-sheets/src/lib/actions/find-rows.ts
```diff
@@ -10,11 +10,44 @@ export const findRowsAction = createAction({
...
+
+                if (!sheetName) {
+                    throw Error("Sheet not found in spreadsheet");
+                }
+
+                const values = await googleSheetsCommon.getValues(spreadsheet_id, accessToken, sheet_id);
+
+                const ret = [];
+                
+                const firstRow = values[0].values;
```
abuaboud: What if there is no first row? I think we shouldn't throw an error maybe show letters?

packages/pieces/google-sheets/src/lib/actions/insert-row.action.ts
```diff
@@ -16,11 +15,45 @@ export const insertRowAction = createAction({
...
+                const sheetName = await googleSheetsCommon.findSheetName(accessToken, spreadsheet_id, sheet_id);
+
+                if (!sheetName) {
+                    throw Error("Sheet not found in spreadsheet");
+                }
+
+                const values = await googleSheetsCommon.getValues(spreadsheet_id, accessToken, sheet_id);
+
+                
+                const firstRow = values[0].values;
```
abuaboud: Same Issue here, if there is no rows this will throw error I think we should show default option like columns names.

packages/pieces/google-sheets/src/lib/actions/clear-sheet.ts
```diff
@@ -0,0 +1,45 @@
...
+    props: {
+        authentication: googleSheetsCommon.authentication,
+        spreadsheet_id: googleSheetsCommon.spreadsheet_id,
+        include_team_drives: googleSheetsCommon.include_team_drives,
+        sheet_id: googleSheetsCommon.sheet_id,
+        ignore_first_row: Property.Checkbox({
+            displayName: 'Is First row Headers?',
+            description: 'If the first row is headers',
+            required: false,
+            defaultValue: false,
```
abuaboud: Make it `true` as default

packages/pieces/google-sheets/src/lib/actions/clear-sheet.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    async run(context) {
+        const sheetName = await googleSheetsCommon.findSheetName(context.propsValue['authentication']['access_token'],
+            context.propsValue['spreadsheet_id'],
+            context.propsValue['sheet_id']);
+        if (!sheetName) {
+            throw Error("Sheet not found in spreadsheet");
+        }
+     
+        const RowsToDelete: number[] = [];
```
abuaboud: make sure you unify the name convention, follow camel case everywhere

packages/pieces/google-sheets/src/lib/actions/update-row.ts
```diff
@@ -28,14 +79,21 @@ export const updateRowAction = createAction({
...
+
+        if (formattedValues.length > 0) {
             const res = await googleSheetsCommon.updateGoogleSheetRow({
                 accessToken: context.propsValue['authentication']['access_token'],
                 rowIndex:  Number(context.propsValue.row_id),
                 sheetName: sheetName,
                 spreadSheetId: context.propsValue['spreadsheet_id'],
                 valueInputOption: ValueInputOption.USER_ENTERED,
-                values: values as string[],
+                values: formattedValues as string[],
```
abuaboud: Don't use casting, It's make types useless, instead of change the line above:

`const formattedValues: string[] = [];`

abuaboud (changes requested):

abuaboud (approved):

PR #1647, opened 2023-07-11
fix: find rows should return row instead of single value

Ozak93:
Fixed find-rows action in Google Sheets piece, it should now return the row number instead of the column name.

also changed the value returned, instead of only the key matching the search parameter, it should now return the entire row. this should be helpful if the returned data were to be used in the preceding steps in a flow.


Abdallah-Alwarawreh (approved):

abuaboud:
@allcontributors please add @Ozak93 for plugin

abuaboud (approved):

PR #1650, opened 2023-07-12
chore: added route to read and create combinations

### LINKED ISSUE #1649: Combinations
A combination is basically a short flow with two or more steps, that gets created by entering a link that includes the template in a base64 query string, so it can be shared easily.

We would need to handle the case where users get sent to the builder and they are not signed in.

route: /combinations?flow="....."

AbdulTheActivePiecer:
## What does this PR do?

closes #1649


abuaboud (changes requested):
Why not use the existing import flow template module?

abuaboud (changes requested):
change name to import flow by base64

abuaboud (approved):

abuaboud (approved):

PR #1658, opened 2023-07-13
feat: webhook trigger handshake

### LINKED ISSUE #1657: Configurable Trigger Webhook Handshake
Right now only webhook triggers for apps that do not require a handshake can be supported. This is quite limiting in the amount of apps that can be supported as a piece builder since there are quite a few apps that require a successful handshake. Specifically I've run into this while extending the asana piece.

I've already discussed a little bit on Discord with @abuaboud about how this can be solved especially because webhooks should be queued and ran async by their nature. I came up with the idea of adding an onHandshake hook in the trigger + a handshake strategy that allows to specify when a request is considered a handshake. This could support the following strategies:
- FIRST_REQUEST (the first webhook request is always considered a handshake and will be the only handshake request)
- HEADER_PRESENT (with an option that configures the header name)
- QUERY_PRESENT (with an option that configures the query param name)
- BODY_PARAM_PRESENT (with an option that configures the field name)

If this criteria is matched, the onHandshake hook should be called (instead of run) with the payload and should be allowed to set response status, headers and body in order to correctly respond to the handshake request.

I could set up a proxy in front of AP that handles the handshake but this wouldn't work for cloud users and would kill the user experience.

JanHolger:
## What does this PR do?

Fixes #1657

### DIFF
--- a/packages/backend/src/app/helper/trigger-utils.ts
+++ b/packages/backend/src/app/helper/trigger-utils.ts
@@ -40,7 +40,7 @@ export const triggerUtils = {
             const handshakeConfig = pieceTrigger.handshakeConfiguration
             switch(handshakeConfig.strategy) {
                 case WebhookHandshakeStrategy.HEADER_PRESENT: {
-                    if(handshakeConfig.paramName && (handshakeConfig.paramName in payload.headers)) {
+                    if(handshakeConfig.paramName && (handshakeConfig.paramName.toLowerCase() in payload.headers)) {
                         return await executeHandshake(flowVersion, projectId, payload)
                     }
                     break
--- a/packages/backend/src/app/helper/trigger-utils.ts
+++ b/packages/backend/src/app/helper/trigger-utils.ts
@@ -1,4 +1,4 @@
-import { TriggerBase, TriggerStrategy } from '@activepieces/pieces-framework'
+import { TriggerBase, TriggerStrategy, WebhookHandshakeConfiguration, WebhookHandshakeStrategy, WebhookResponse } from '@activepieces/pieces-framework'
 import {
     ExecutionType,
     EngineResponseStatus,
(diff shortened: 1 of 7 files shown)


packages/backend/src/app/helper/trigger-utils.ts
```diff
@@ -25,6 +25,44 @@ import { SystemProp } from './system/system-prop'
 const POLLING_FREQUENCY_CRON_EXPRESSON = `*/${system.getNumber(SystemProp.TRIGGER_DEFAULT_POLL_INTERVAL ?? 5)} * * * *`
 
 export const triggerUtils = {
+    /* eslint-disable */ async tryHandshake(params: ExecuteTrigger): Promise<WebhookResponse|null> { // disabled eslint because of weird no-redundant-type-constituents error
```
khaledmashaly: This comment is not needed, lint passes

packages/backend/src/app/helper/trigger-utils.ts
```diff
@@ -97,6 +135,28 @@ export const triggerUtils = {
     },
 }
 
+async function executeHandshake(flowVersion: FlowVersion, projectId: ProjectId, payload: TriggerPayload): Promise<WebhookResponse> {
```
khaledmashaly: Please use an object for params

packages/backend/src/app/webhooks/webhook-controller.ts
```diff
@@ -42,6 +42,25 @@ export const webhookController: FastifyPluginAsync = async (app) => {
         },
         async (request: FastifyRequest<{ Params: WebhookUrlParams }>, reply) => {
             const flow = await getFlowOrThrow(request.params.flowId)
+            const handshakeResponse = await webhookService.handshake({
+                flow: flow,
+                payload: {
+                    method: request.method,
+                    headers: request.headers as Record<string, string>,
+                    body: await convertBody(request),
```
khaledmashaly: in case of `handshakeResponse` being `null`, `convertBody` will be called twice, once here and once in the `handler`.

I think we should optimise for one call only since `convertBody` is a very intensive method.

packages/backend/src/app/webhooks/webhook-service.ts
```diff
@@ -23,8 +23,41 @@ import { isNil } from '@activepieces/shared'
 import { logger } from '../helper/logger'
 import { webhookSimulationService } from './webhook-simulation/webhook-simulation-service'
 import { flowInstanceService } from '../flows/flow-instance/flow-instance.service'
+import { WebhookResponse } from '@activepieces/pieces-framework'
 
 export const webhookService = {
+    /* eslint-disable */ async handshake({ flow, payload }: CallbackParams): Promise<WebhookResponse|null> { // disabled eslint because of weird no-redundant-type-constituents error
```
khaledmashaly: comment not needed

khaledmashaly (changes requested):
Solid, please see comments. Also can you please confirm that the functionality was tested end to end?

packages/backend/src/app/webhooks/webhook-controller.ts
JanHolger: Didn't want to touch too much existing code here. Since the raw request is not needed in the handler anyway I've changed it to also consume the EventPayload so it only needs to be converted once. This also removed some duplicated code from the existing code.

packages/backend/src/app/helper/trigger-utils.ts
JanHolger: Don't really see the point here but I've changed it anyway

JanHolger:
> Solid, please see comments. Also can you please confirm that the functionality was tested end to end?

I've done manual E2E testing with all 4 strategy's (NONE, HEADER_PRESENT, QUERY_PRESENT, BODY_PARAM_PRESENT) and every case was successfully working (handshake requests where handled by onHandshake, normal requests by run)

abuaboud (approved):

PR #1660, opened 2023-07-13
fix: import flow lookup

AbdulTheActivePiecer:
## What does this PR do?

Changes when to lookup templates in our cdn.


abuaboud:
The backend should throw 404, this is no longer needed.

PR #1676, opened 2023-07-16
feat: validate connections

### LINKED ISSUE #1403: Validate Connection on Saving
API Key / Custom Authentication should have a validate function in the framework that backend use to validate on saving.

khaledmashaly:
## What does this PR do?

Fixes #1403 

On a side note, performance is horrible because validating the connection takes forever to call the engine and return the validation response to the API caller. Also, the validation method is a user-defined method that could contain any blocking code like HTTP calls, for example.

A better approach is to validate the connection asynchronously and notify the front-end of the result.


packages/pieces/framework/src/lib/piece.ts
```diff
@@ -81,8 +64,47 @@ export const createPiece = <PieceAuth extends PieceAuthProperty>(params: CreateP
     params.actions,
     params.triggers,
     params.auth ?? undefined,
+    params.ValidatePieceAuth,
```
abuaboud: Camel case

abuaboud (dismissed):

PR #1700, opened 2023-07-19
fix: encrypted objects

abuaboud:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/backend/src/app/app-connection/app-connection.service.ts
+++ b/packages/backend/src/app/app-connection/app-connection.service.ts
@@ -67,20 +67,18 @@ export const appConnectionService = {
             projectId,
         })
 
-        updatedConnection.value = decryptObject(updatedConnection.value)
-
-        return updatedConnection
+        return decryptConnection(updatedConnection)
     },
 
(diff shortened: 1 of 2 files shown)


packages/backend/src/app/app-connection/app-connection.service.ts
```diff
@@ -189,6 +186,16 @@ const validateConnectionValue = async (params: ValidateConnectionValueParams): P
...
 
+function decryptConnection(encryptedConnection: AppConnectionSchema): AppConnection {
+    const connection = {
+        ...encryptedConnection,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        value: decryptObject<any>(encryptedConnection.value),
+    }
+    connection.status = getStatus(connection.value)
+    return connection
+}
```
khaledmashaly, line 269: This bit is very confusing, `decryptConnection` returns an `AppConnection` so `connection` type is `AppConnection`, and `getStatus` takes an `AppConnection` so `connection.value` type is also `AppConnection`?!

I think this issue happened in the first place because of the wrong types so let's get types right this time to avoid future pitfalls, what do you think?

PR #1708, opened 2023-07-19
refactor(backend): make a customisable preUpsertHook for connection service

khaledmashaly:
An experiment to avoid conflicts between ee and oss implementations


packages/backend/src/app/app-connection/app-connection-service/oss-app-connection-service.ts
```diff
@@ -0,0 +1,14 @@
+import { ProjectId, UpsertConnectionRequest } from '@activepieces/shared'
+import { BaseAppConnectionService } from './base-app-connection-service'
+import { logger } from '../../helper/logger'
+
+export class OssAppConnectionService extends BaseAppConnectionService {
```
abuaboud: Let's rename this without OSS, we will name the one in the fork to be cloud, as this shouldn't be aware cloud

khaledmashaly: done in [5094f27](https://github.com/activepieces/activepieces/pull/1708/commits/5094f27e7e33e005a1873a181d1d5292b850467e)

PR #1746, opened 2023-07-24
feat: support sqllite3

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/backend/src/app/helper/base-entity.ts
```diff
@@ -1,4 +1,13 @@
 import { EntitySchemaColumnOptions } from 'typeorm'
+import { system } from './system/system'
+import { SystemProp } from './system/system-prop'
+import { ApEnvironment } from '@activepieces/shared'
+
+export const JSON_COLUMN_TYPE = system.get(SystemProp.ENVIRONMENT) === ApEnvironment.DEVELOPMENT ? 'simple-json' : 'json'
+export const JSONB_COLUMN_TYPE = system.get(SystemProp.ENVIRONMENT) === ApEnvironment.DEVELOPMENT ? 'simple-json' : 'jsonb'
+export const BLOB_COLUMN_TYPE = system.get(SystemProp.ENVIRONMENT) === ApEnvironment.DEVELOPMENT ? 'blob' : 'bytea'
+export const TIMESTAMP_COLUMN_TYPE = system.get(SystemProp.ENVIRONMENT) === ApEnvironment.DEVELOPMENT ? 'datetime' : 'timestamp with time zone'
+export const COLLATION = system.get(SystemProp.ENVIRONMENT) === ApEnvironment.DEVELOPMENT ? undefined : 'en_natural'
```
khaledmashaly: It seems like this logic prevents using `sqlite3` in production or testing, right?

Would it be possible to detect the database and adjust data types accordingly?

packages/backend/src/app/database/database-connection.ts
```diff
@@ -132,4 +102,62 @@ export const databaseConnection = new DataSource({
...
+                PieceMetadata1685537054805,
+                AddProjectIdToPieceMetadata1686090319016,
+                UnifyPieceName1686138629812,
+                AddScheduleOptions1687384796637,
+                AddAuthToPiecesMetadata1688922241747,
+                AddUpdatedByInFlowVersion1689292797727,
+                AddTasksToRun1689351564290,
+            ],
+            ...commonProperties,
+        })
```
khaledmashaly: Before this PR: postgres logic was hardcoded
After this PR: postgres, and sqlite3 are hardcoded
Future suggestion: make ap database independent

BTW, Do you think that maintaining two sets of migrations for different databases is a good idea?

abuaboud: @khaledmashaly It's actually the only option since it's has different column types definition

packages/backend/src/app/helper/base-entity.ts
abuaboud: Done

PR #1751, opened 2023-07-24
feat: support simple setup without external dependency  

abuaboud:
## What does this PR do?

Fixes # (issue)


package.json
```diff
@@ -3,10 +3,9 @@
   "version": "0.5.3",
   "scripts": {
     "prepare": "husky install",
-    "docker": "docker compose -f docker-compose.dev.yml up",
     "frontend": "nx serve ui-core",
-    "backend": "wait-on tcp:5432 tcp:6379 && nx serve backend",
-    "dev": "concurrently -p \"[{name}]\" -n \"DKR,GUI,API\" -c \"bgBlue.bold,bgMagenta.bold,bgGreen.bold\" \"npm:docker\" \"npm:frontend\" \"npm:backend\"",
+    "backend": "nx serve backend",
+    "dev": "concurrently -p \"[{name}]\" -n \"DKR,GUI,API\" -c \"bgBlue.bold,bgMagenta.bold,bgGreen.bold\" \"npm:frontend\" \"npm:backend\"",
```
khaledmashaly: Seems like you forget to remove `DKR` from `\"DKR,GUI,API\"`?

packages/backend/src/app/helper/database-common.ts
```diff
@@ -0,0 +1,49 @@
...
+        name: 'created',
+        type: TIMESTAMP_COLUMN_TYPE,
+        createDate: true,
+    } as EntitySchemaColumnOptions,
+    updated: {
+        name: 'updated',
+        type: TIMESTAMP_COLUMN_TYPE,
+        updateDate: true,
+    } as EntitySchemaColumnOptions,
+}
```
khaledmashaly: I think it makes sense to move this file from `helper` to `database`, right?

packages/backend/src/app/helper/system/system.ts
```diff
@@ -1,9 +1,28 @@
 import { ActivepiecesError, ErrorCode } from '@activepieces/shared'
 import { SystemProp } from './system-prop'
 
+const systemPropDefaultValues: Record<string, string> = {
```
khaledmashaly: use `Partial<Record<SystemProp, string>>` instead

packages/ui/core/src/app/app.component.html
```diff
@@ -1,13 +1,3 @@
-<mat-toolbar class="!ap-bg-primary-light" *ngIf="warningMessage$ | async as warning" @fadeInUp>
-  <div class="ap-bg-primary-light ap-typography-body-1 ap-flex ap-flex-col ap-items-center ap-p-1 ap-flex-grow">
-    <div class="ap-typography-body-1 ap-text-primary !ap-font-bold">
-      {{ warning.title }}
-    </div>
-    <div class="ap-typography-body-2 ap-text-title ap-text-center">
-      {{ warning.body }}
-    </div>
-  </div>
-</mat-toolbar>
```
khaledmashaly, line 10: Any reason for removing the warning message?

khaledmashaly (commented):
Good PR, can you please explain why we should maintain the code for an in-memory queue instead of using a library?

Also, did you benchmark the in-memory queue to make sure it's not blocking the main thread? i.e. blocking the API server?

packages/ui/core/src/app/app.component.html
abuaboud: No longer used :dagger:

PR #1776, opened 2023-07-28
docs: add docs for property validators and processors

OsamaHaikal:
## What does this PR do?

Add docs
Fixes # (issue)


docs/developers/piece-reference/properties.mdx
```diff
@@ -334,3 +334,179 @@ Property.DynamicProperties({
...
 ```
+
+## Validators and Processors
+The Validators and Processors feature offers you the means to optimize data quality by efficiently processing user input and validating it accordingly. This section provides comprehensive guidance on how to effectively utilize these functionalities.
+
+### PieceProperty Type Changes
+The following four keys have been added to the `PieceProperty` type:
+
+```typescript
+defaultProcessors?: ProcessorFn[];
```
abuaboud: The target user is a developer creating a piece, this is great for framework contributors.

In that I suggest remove everything related defaultProcessors and defaultValidators

docs/developers/piece-reference/properties.mdx
```diff
@@ -334,3 +334,179 @@ Property.DynamicProperties({
     }
 })
 ```
+
+## Validators and Processors
```
abuaboud: This is under reference, I would suggest remove the whole section until Processors, If that's removed make sure you update the following headers for example Processors should be ## instead of ###

docs/developers/piece-reference/properties.mdx
```diff
@@ -334,3 +334,179 @@ Property.DynamicProperties({
...
+
+```typescript
+text: Property.LongText({
+    displayName: 'Text',
+    required: true,
+    validators: [Validators.url, Validators.oneOf(["activepieces.com", "www.activepieces.com"])],
+})
+```
+
+### Processors
```
abuaboud: I would move everything related to processor under this section even custom processor, so it's self contained

abuaboud: I would suggest the following as well, so it shows the outline in better way

image

1- Change the Header to ## 
2- Remove Predefined Processors Headers
3- Each Predefined Processor should be ### So it shows in the outline.
4- and as well move the custom processor as one of the  ###  here,
### File 
### DateTime
### Number

Same to the validators.

abuaboud (changes requested):

PR #1787, opened 2023-07-31
feat: test branch step

### LINKED ISSUE #1696: [Suggestion] add a test button for Branch piece
A branch piece should have a test button to check if the First/Second value provide the desired true/false output per the condition selected! as of now you must test the entire flow to test 1 branch, which is counterintuitive.

khaledmashaly:
## What does this PR do?

fixes #1696


packages/engine/src/lib/action/piece-action-handler.ts
```diff
@@ -143,7 +143,7 @@ export class PieceActionHandler extends BaseActionHandler<PieceAction> {
 
       const context: ActionContext = {
         executionType: this.executionType,
-        store: createContextStore('', globals.flowId),
+        store: createContextStore('', globals.flowVersionId),
```
abuaboud, line 146: Why is this changed? Is it meant to persist across different publishing? Or u think thats bad?

abuaboud (approved):

packages/engine/src/lib/action/piece-action-handler.ts
khaledmashaly: It's already `flowVersionId` that's assigned to `globals.flowId`, I just renamed it see [here](https://github.com/activepieces/activepieces/blob/5cbd1da9b379b3c0989e383209a8c3b9184bae0f/packages/engine/src/main.ts#L31)

khaledmashaly: even before this it was `flowVersionId`, see here: [here](https://github.com/activepieces/activepieces/blob/2dc69d299e49d5a55c7391fe8f1b4df9343c2bb6/packages/engine/src/lib/executors/flow-executor.ts#L170)

you sure that storage persists across publishing?

PR #1840, opened 2023-08-09
feat: Added actions [Monday.js]

haseebrehmanpc:
## What does this PR do?

This PR will add 3 actions in monday.js piece
1- Change Multiple Columns Value in [monday.com](http://monday.com/) - with this you should be able to input the item id and input new values into the columns for that board.
2- Get column Values: input the item id and retrieve the column data for that item
3- Get item by column value: Search a column for specific data and retrieve the matched items data.

### DIFF
--- a/packages/pieces/monday/package.json
+++ b/packages/pieces/monday/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-monday",
-  "version": "0.1.0"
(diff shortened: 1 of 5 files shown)


abuaboud:
@allcontributors please add @haseebrehmanpc for plugin

abuaboud (approved):

PR #1844, opened 2023-08-10
feat: support multiple slack action buttons

rita-gorokhod:
## What does this PR do?

Allows to send an approval-like message to a user/channel and wait for an action to be selected. The motivation is that the approval piece only supports two actions (approve/deny), whereas this is generic and allows multiple actions.

### DIFF
--- a/packages/pieces/slack/src/lib/common/props.ts
+++ b/packages/pieces/slack/src/lib/common/props.ts
@@ -1,5 +1,5 @@
-import { OAuth2PropertyValue, Property } from "@activepieces/pieces-framework";
-import { AuthenticationType, httpClient, HttpMethod, HttpRequest } from "@activepieces/pieces-common";
+import { OAuth2PropertyValue, Property } from '@activepieces/pieces-framework';
+import { AuthenticationType, httpClient, HttpMethod, HttpRequest } from '@activepieces/pieces-common';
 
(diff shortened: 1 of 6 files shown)


packages/pieces/slack/src/lib/common/props.ts
```diff
@@ -1,5 +1,5 @@
-import { OAuth2PropertyValue, Property } from "@activepieces/pieces-framework";
-import { AuthenticationType, httpClient, HttpMethod, HttpRequest } from "@activepieces/pieces-common";
+import { OAuth2PropertyValue, Property } from '@activepieces/pieces-framework';
+import { AuthenticationType, httpClient, HttpMethod, HttpRequest } from '@activepieces/pieces-common';
```
rita-gorokhod, line 2: prettier did some auto formatting

packages/pieces/slack/src/lib/common/props.ts
```diff
@@ -98,9 +98,14 @@ export const text = Property.LongText({
   required: true,
 });
 
+export const actions = Property.Array({
```
rita-gorokhod, line 101: The new property, rest are linter cosmetics

abuaboud:
@allcontributors please add @rita-gorokhod for plugin

abuaboud (approved):

PR #1877, opened 2023-08-13
fix: build code without webpack

abuaboud:
## What does this PR do?

- Replaced tsc with webpack
- Building actual happens everytime, pnpm should cache the packages
- Sandbox caching should handle this for published flow
- Tested with 4 code pieces, It takes around 3-5 seconds to setup them
- Removed overwrite in flow version as this cause a race condition, for example someone is editing the flow while this test flow is running


khaledmashaly (commented):
Great! Just a couple of comments:
- There's a breaking behaviour: `import dayjs from 'dayjs';` doesn't work, while it used to work with the old code bundling. Also, `import * as dayjs from 'dayjs';`, works now which wasn't the case in the old method, so the behaviour is reversed, which could cause breaking changes for some users editing old code snippets, this is called "synthetic default imports" btw.
- We need a migration that removes the `packagedArtifactId` from old flow versions, and deletes the corresponding files, right?
- [OPTIONAL] The logic for installing deps and transpiling code is duplicated across the flow worker and the engine helper, let's make a follow-up PR that fixes this (I could take it if you are ok with this).

PR #1883, opened 2023-08-14
feat: More flexible connection names

abuaboud:
## What does this PR do?

Allows users to name their connections in a more flexible manner.

They will be allowed to use letters, numbers, "+", "-", "_" ,"@"  and "." .


packages/engine/src/lib/services/variable-service.ts
```diff
@@ -48,29 +48,43 @@ export class VariableService {
...
+  private findConnectionName(path: string): string | null {
+    const paths = path.split('.');
+    // Connections with square brackets
+    if (path.includes('[')) {
+      const matches = path.match(/\['([^']+)'\]/g);
+      if (matches && matches.length >= 1) {
+        const secondPath = matches[0].replace(/\['|'\]/g, '');
+        return secondPath;
+      }
+      return null;
```
AbdulTheActivePiecer: Please write down what each regex does.

AbdulTheActivePiecer (approved):

packages/engine/src/lib/services/variable-service.ts
AbdulTheActivePiecer: I am guessing the first one is checking if the path is using bracket notation and the second one checks for the surrounding brackets and removes them.

AbdulTheActivePiecer (approved):

PR #1894, opened 2023-08-15
feat: support tags in flow runs api

abuaboud:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/backend/src/app/database/database-common.ts
+++ b/packages/backend/src/app/database/database-common.ts
@@ -10,12 +10,18 @@ export const JSONB_COLUMN_TYPE =
   databaseType === DatabaseType.SQLITE3 ? 'simple-json' : 'jsonb'
 export const BLOB_COLUMN_TYPE =
   databaseType === DatabaseType.SQLITE3 ? 'blob' : 'bytea'
+export const ARRAY_COLUMN_TYPE =
+  databaseType === DatabaseType.SQLITE3 ? 'simple-array' : String
 export const TIMESTAMP_COLUMN_TYPE =
   databaseType === DatabaseType.SQLITE3
       ? 'datetime'
       : 'timestamp with time zone'
 export const COLLATION =
   databaseType === DatabaseType.SQLITE3 ? undefined : 'en_natural'
+
+export function isPostgres(): boolean {
+    return databaseType === DatabaseType.POSTGRES
+}
 
 export const ApIdSchema = {
     type: String,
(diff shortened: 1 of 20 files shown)


packages/backend/src/app/database/database-common.ts
```diff
@@ -10,13 +10,19 @@ export const JSONB_COLUMN_TYPE =
   databaseType === DatabaseType.SQLITE3 ? 'simple-json' : 'jsonb'
 export const BLOB_COLUMN_TYPE =
   databaseType === DatabaseType.SQLITE3 ? 'blob' : 'bytea'
+export const ARRAY_COLUMN_TYPE =
+  databaseType === DatabaseType.SQLITE3 ? 'simple-array' : String
```
khaledmashaly, line 14: I think `simple-array` works for `postgres` and `sqlite`, no?

packages/backend/src/app/database/database-connection.ts
```diff
@@ -46,3 +47,25 @@ export const databaseConnection =
...
+                    ...params,
+                    [`tag${index}`]: `%${tag}%`,
+                }
+            }, {})
+            return query.andWhere(likeConditions, likeParams)
+        }
+        default:
+            throw new Error(`Unsupported database type: ${databaseType}`)
+    }
+}
```
khaledmashaly, line 71: wow! the decision to incorporate `sqlite` is making a lot of noise, we are basically doing everything db-related two times, I think we should either drop support for `sqlite` or if adoption numbers justify the use of `sqlite` let's see if we can switch to another ORM that can make things a little bit uniform across multiple DBs, right now TypeORM does little to nothing in this regard.

packages/backend/src/app/database/migration/1692106268660-AddTagsToRun.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export class AddTagsToRun1692106268660 implements MigrationInterface {
+    name = 'AddTagsToRun1692106268660'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query('ALTER TABLE "flow_run" ADD "tags" text')
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query('ALTER TABLE "flow_run" DROP COLUMN "tags"')
```
khaledmashaly, line 11: I think this migration is not needed, right?

khaledmashaly (commented):
Can you please clarify what are tags exactly? Are they just some tags for search purposes? Are they connected to flows? or to connections?

Sorry, I'm kind of confused! Did we discuss this feature?

packages/backend/src/app/database/migration/1692106268660-AddTagsToRun.ts
abuaboud: It's auto generated in case you want to rollback yes.

PR #1900, opened 2023-08-15
fix(piece-ntfy): piece authentication was not set

facferreira:
## What does this PR do?

Piece NTFY had no authentication object set, which made it impossible to execute its action correctly.


abuaboud:
@allcontributors please add @facferreira for plugin

abuaboud:
@allcontributors please add @facferreira for plugin

PR #2005, opened 2023-08-18
feat(contiguity): initial piece - sms platform

Owlcept:
## What does this PR do?

Sets up the initial piece for [Contiguity](https://contiguity.co) to allow texting within steps aside from twilio


abuaboud:
@allcontributors please add @Owlcept for plugin

PR #2021, opened 2023-08-19
feat: add SimplePDF piece

bendersej:
## What does this PR do?

This PR introduces the SimplePDF piece.

I will amend the existing blog post once this PR is merged as it was previously using the webhook piece: https://simplepdf.eu/help/how-to/connect-simplepdf-with-activepieces-to-automate-your-pdf-forms-processing

## Notes

I ran into some issues when developing, namely:
- Setting up my environment via Codespaces didn't work (a call to localhost:3000/flags kept failing and the UI was not displayed)
- My local environment failed initially because of a NX postinstall script: I had to `npm install nx` separately and then run `npm install`
- It appears that when testing the flow, even though I am returning `context.payload.body.data` in the run context, I am getting the full request (including the headers), is that expected?

### DIFF
--- a/packages/pieces/simplepdf/.eslintrc.json
+++ b/packages/pieces/simplepdf/.eslintrc.json
@@ -1,0 +1,18 @@
+{
+  "extends": ["../../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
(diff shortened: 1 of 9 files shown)


abuaboud:
@allcontributors please add @bendersej for plugin

abuaboud:
I'll send the first two points to the team, and I corrected the third point you mentioned :D

I still remember your article - it really brightened my day 

Thanks for contributing once more!

PR #2132, opened 2023-08-22
fix: test branch execution context

### LINKED ISSUE #2016: Testing Branch Condition is Missing Test Context
**Describe the bug**

Test a Branch with resolving variable from previous steps, It won't work

That's because Test Context is missing and in the test step, It creates new flow with empty trigger, We should use same context as exceuteAction

khaledmashaly:
## What does this PR do?

Fixes #2016

### DIFF
--- a/packages/backend/src/app/flows/step-run/step-run-service.ts
+++ b/packages/backend/src/app/flows/step-run/step-run-service.ts
@@ -15,7 +15,6 @@ import {
     EmptyTrigger,
     TriggerType,
     apId,
-    ExecuteFlowOperation,
     ExecutionType,
     EngineResponseStatus,
     ExecutionState,
@@ -23,6 +22,8 @@ import {
     ExecutionOutputStatus,
     UserId,
     FlowOperationType,
+    EngineTestOperation,
+    StepOutputStatus,
 } from '@activepieces/shared'
 import { engineHelper } from '../../helper/engine-helper'
 import { flowVersionService } from '../flow-version/flow-version.service'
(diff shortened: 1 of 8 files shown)


packages/backend/src/app/flows/step-run/step-run-service.ts
```diff
@@ -144,34 +145,44 @@ const executeBranch = async ({ step, flowVersion, projectId }: ExecuteParams<Bra
...
+            nextAction: undefined,
+            onSuccessAction: undefined,
+            onFailureAction: undefined,
+        },
         type: TriggerType.EMPTY,
         settings: {},
     }
 
-    flowVersion.trigger = testTrigger
+    const testFlowVersion: FlowVersion = {
```
abuaboud, line 162: Isn't that the same issue, how it would extract the context? if it doesn't have the flow?

abuaboud: Edit: why we are sending two flows? why not the origianl one only and use flowUtils to find the correct step?

khaledmashaly: The second flow is an empty one containing the test branch only, this is a generic design that can handle multiple cases like: testing branches, testing code pieces, testing piece actions, and so on.

Other parts of the test step service will be migrated to this design after proving its correctness on the branch step. This will benefit the system in, mainly, two points:
- Reduce the number of operations executed by the engine
- Reduce duplication across different engine operations

abuaboud (approved):

PR #2162, opened 2023-08-23
fix: properties form cleanup

AbdulTheActivePiecer:
## What does this PR do?

Clean properties form and fix description for dateTime airtable fields showing "undefined".


packages/ui/feature-builder-form-controls/src/lib/piece-properties-form/piece-properties-form.component.ts
```diff
@@ -389,63 +389,137 @@ export class PiecePropertiesFormComponent implements ControlValueAccessor {
...
-      } else if (prop.type === PropertyType.JSON) {
-        if (!this.customizedInputs || !this.customizedInputs[pk]) {
-          validators.push(jsonValidator);
+      switch (prop.type) {
+        case PropertyType.ARRAY: {
+          const controlValue = propValue
+            ? propValue
+            : Array.isArray(prop.defaultValue) && prop.defaultValue.length > 0
+            ? prop.defaultValue
+            : [''];
```
abuaboud, line 398: Let's test wordpress, If i sent empty tags that would be problematic can you please doubel check?

I believe this should be null or empty

PR #2220, opened 2023-08-24
fix(storage): use deep equal for comparing objects 

Salem-Alaa:
## What does this PR do?
This code fix the storge bug for storing duplicate items even if ignore duplicate was selected
Fixes # (issue)

### DIFF
--- a/packages/backend/src/app/workers/sandbox.ts
+++ b/packages/backend/src/app/workers/sandbox.ts
@@ -264,7 +264,7 @@ export default class SandboxManager {
         if (SandboxManager._instance != null) {
             throw new Error('Use Singleton.instance instead of new.')
         }
-        for (let boxId = 0; boxId < 5000; ++boxId) {
+        for (let boxId = 0; boxId < 1000; ++boxId) {
             this.sandboxes.set(boxId, new Sandbox({
                 boxId,
                 cached: false,


packages/pieces/store/src/lib/actions/store-add-to-list.ts
```diff
@@ -1,5 +1,5 @@
 import { createAction, Property, StoreScope } from "@activepieces/pieces-framework";
-
+import isEqual from 'lodash/isEqual';
```
abuaboud: Please use deep-equal package that already installed, loadash is heavy package.

packages/pieces/store/src/lib/actions/store-add-to-list.ts
```diff
@@ -43,8 +43,12 @@ export const storageAddtoList = createAction({
         if (!Array.isArray(items)) {
             throw new Error(`Key ${context.propsValue['key']} is not an array`);
         }
-        if (context.propsValue['ignore_if_exists'] && items.includes(context.propsValue['value'])) {
-            return items;
+        if (context.propsValue['ignore_if_exists'] ) {
```
abuaboud: Can you also fix the remove item from the list, to filter everything that is equal to that object?

PR #2255, opened 2023-08-25
feat(salesforce): add bulk upsert action

landonmoir:
## What does this PR do?

This PR adds an action to the Salesforce piece that uses the Salesforce Bulk API to upsert records. Using the Bulk API bypasses the 200 record limit of the current Salesforce upsert action.


abuaboud:
@allcontributors please add @landonmoir for plugin

PR #2312, opened 2023-08-28
chore: load log config from env

khaledmashaly:
## What does this PR do?

- Add `SystemProp.LOG_LEVEL` to control log level
- Add `SystemProp.LOG_PRETTY` to control whether to output logs in a human-readable format
- Remove environment checks in logger code


packages/backend/src/app/helper/system/system.ts
```diff
@@ -60,8 +61,8 @@ export const system = {
         return getEnvVar(prop) === 'true'
     },
 
-    getOrThrow(prop: SystemProp): string {
-        const value = getEnvVar(prop)
+    getOrThrow<T extends string>(prop: SystemProp): T {
+        const value = getEnvVar(prop) as T | undefined
```
abuaboud, line 65: This is wrong, we need to fix that as far as I know process.env will always return string, I know we casteed but I think we need to fix it in human way

khaledmashaly: yes that's why `T extends string`, i.e. only enums or unions of string literals are qualified, so at the end they are still strings

abuaboud (approved):

PR #2336, opened 2023-08-29
feat(stripe): add create customer / create an invoice / search for a customer by email actions

lldiegon:
## What does this PR do?
This pull request adds 3 actions for the stripe piece:
- Create a customer
- Create an invoice
- Search for a customer by email

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/stripe/src/lib/trigger/new-customer.ts
+++ b/packages/pieces/stripe/src/lib/trigger/new-customer.ts
@@ -42,7 +42,7 @@ export const stripeNewCustomer = createTrigger({
   },
   type: TriggerStrategy.WEBHOOK,
   async onEnable(context) {
-    const webhook = await stripeCommon.subscribeWebhook('customer.created', context.webhookUrl, context.auth);
+    const webhook = await stripeCommon.subscribeWebhook('customer.created', context.webhookUrl, context.auth.publicKey);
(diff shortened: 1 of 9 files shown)


lldiegon:
One thing to note: This might break current connections, because I had to adjust the authentication to be able to add a secret key. Which was required for the requests I added.

packages/pieces/stripe/src/lib/trigger/new-customer.ts
```diff
@@ -42,15 +42,15 @@ export const stripeNewCustomer = createTrigger({
...
+    const webhook = await stripeCommon.subscribeWebhook('customer.created', context.webhookUrl, context.auth.publicKey);
     await context.store.put<WebhookInformation>('_new_customer_trigger', {
       webhookId: webhook.id
     });
   },
   async onDisable(context) {
     const response = await context.store?.get<WebhookInformation>('_new_customer_trigger');
     if (response !== null && response !== undefined) {
-      await stripeCommon.unsubscribeWebhook(response.webhookId, context.auth);
+      await stripeCommon.unsubscribeWebhook(response.webhookId, context.auth.publicKey);
```
abuaboud: We should revert auth to secret key, because I think subscribe webhooks uses secrets key as well because they register endpoint.

abuaboud (approved):

PR #2406, opened 2023-08-31
feat(google-sheet): adding new action and fixing some text

### LINKED ISSUE #1921: [Piece] Google sheets enhancements
1) Add action to get row by row number.
2) Add action to get row iteratively by retrieving the next row from top to bottom every time the action is executed.
3) Make search optional in "Find Rows".
4) Add a filter property for "Find Rows" where to satisfy the query, the search would either be "included" or "exactly matches" the cell value being queried.

Salem-Alaa:
## What does this PR do?
Adding getRow(s) action
adding more descriptive error

Fixes #1921

### DIFF
--- a/packages/backend/.env
+++ b/packages/backend/.env
@@ -1,5 +1,5 @@
 AP_DB_TYPE=SQLITE3
-AP_DEV_PIECES="discord,delay,approval,hackernews,github,slack,store,wordpress"
+AP_DEV_PIECES="google-sheets,store"
 AP_ENVIRONMENT="dev"
 AP_FRONTEND_URL="http://localhost:4200"
 AP_LOG_LEVEL=debug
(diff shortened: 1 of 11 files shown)


packages/pieces/google-sheets/src/lib/actions/get-rows.ts
```diff
@@ -0,0 +1,100 @@
...
+        const row = await getGoogleSheetRows({
+            accessToken: auth['access_token'],
+            sheetName: sheetName,
+            spreadSheetId: propsValue['spreadsheet_id'],
+            rowIndex_s: starting_row,
+            rowIndex_e: end_row - 1
+        });
+
+        return row;
+    }
```
abuaboud, line 68: The logic is duplicated, It's better to have function and boolean that one doesn't save the end_row, and please use camelcase `endRow` and startingRow

packages/backend/.env
abuaboud, line 1: Can you please revert?

PR #2412, opened 2023-08-31
Fix: fasten saving steps and fix failing to save steps with long sample data

AbdulTheActivePiecer:
## What does this PR do?

Remove nextAction,firstLoopAction, onSuccessAction and onFailuireAction from being sent to backend on UPDATE_ACTION.

### DIFF
--- a/packages/ui/feature-builder-store/src/lib/store/flow/flow.effects.ts
+++ b/packages/ui/feature-builder-store/src/lib/store/flow/flow.effects.ts
@@ -40,6 +40,7 @@ import {
 import { canvasActions } from '../builder/canvas/canvas.action';
 import { ViewModeActions } from '../builder/viewmode/view-mode.action';
 import { ViewModeEnum } from '../../model';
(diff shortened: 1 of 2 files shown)


packages/ui/feature-builder-store/src/lib/store/flow/flow.effects.ts
```diff
@@ -220,24 +221,34 @@ export class FlowsEffects {
         const genSavedId = UUID.UUID();
         let flowOperation: FlowOperationRequest;
         switch (action.type) {
-          case FlowsActionType.UPDATE_TRIGGER:
+          case FlowsActionType.UPDATE_TRIGGER: {
+            const op =
+              FlowStructureUtil.removeAnySubequentStepsFromFlowOperationStep(
+                action.operation
+              ) as typeof action.operation;
```
abuaboud: Please don't use as, and removeAnySubequentStepsFromFlowOperationStep make return type update operation and map values there.

abuaboud (approved):

PR #2416, opened 2023-08-31
feat(openai-embeddings): New openai action that creates embeddings

Armangiau3:
## What does this PR do?

This is a new action for the OpenAI piece. It allows you to create an index of a text for similarity search or other purposes. To do this, the action transforms the text into an embedding, which is a numerical representation of the text. However, if your text is too large, it will be difficult to query it efficiently. Therefore, the piece splits the text into different chunks and creates embeddings for each chunk.

### DIFF
--- a/packages/pieces/openai/src/lib/actions/generate-embeddings.ts
+++ b/packages/pieces/openai/src/lib/actions/generate-embeddings.ts
@@ -79,12 +79,12 @@ const splitBychar = (
 export const createEmbeddingsFromText = createAction({
   auth: openaiAuth,
   name: 'create_embeddings',
-  displayName: 'Generate embeddings',
-  description: 'Chuck a text and generate embeddings for it',
+  displayName: 'Generate Embeddings from File',
+  description: 'Chuck a text file and generate embeddings for it',
   props: {
     textInput: Property.File({
-      displayName: 'Text Input File',
-      description: 'Enter the text input you want to tarsform to embeddings',
+      displayName: 'Text File',
+      description: 'The text file you want to transform into embeddings.',
       required: true,
     }),
     title: Property.ShortText({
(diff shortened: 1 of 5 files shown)


packages/pieces/openai/src/lib/actions/generate-embeddings.ts
```diff
@@ -0,0 +1,307 @@
...
+      for (const text of textInput) {
+        splitText(text as string);
+      }
+    } else {
+      splitText(textInput);
+    }
+
+    enc.free();
+
+    console.log('documents: ', textSplited);
```
abuaboud: nit: let's remove this, this would overflow the logs

packages/pieces/openai/src/lib/actions/generate-embeddings.ts
```diff
@@ -0,0 +1,307 @@
...
+};
+
+export const createEmbeddingsFromText = createAction({
+  auth: openaiAuth,
+  name: 'create_embeddings',
+  displayName: 'Generate embeddings',
+  description: 'Chuck a text and generate embeddings for it',
+  props: {
+    textInput: Property.File({
+      displayName: 'Text Input File',
```
abuaboud: rename to `Text File`

packages/pieces/openai/src/lib/actions/generate-embeddings.ts
```diff
@@ -0,0 +1,307 @@
...
+      description: 'Enter the text input you want to tarsform to embeddings',
```
abuaboud: Rename to `The text file you want to transform into embeddings.`

packages/pieces/openai/src/lib/actions/generate-embeddings.ts
```diff
@@ -0,0 +1,307 @@
...
+    }
+  }
+
+  return textSplited;
+};
+
+export const createEmbeddingsFromText = createAction({
+  auth: openaiAuth,
+  name: 'create_embeddings',
+  displayName: 'Generate embeddings',
```
abuaboud: Rename this to `Generate Embeddings from File`

abuaboud (changes requested):

PR #2419, opened 2023-08-31
feat(google-drive): add download file action

Armangiau3:
## What does this PR do?

This PR creates a new action to download any file from Google Drive, using its file id. The specific Google formats (Google Docs, Sheets, etc.) will be converted to their equivalent Microsoft 365 (docx, xlsx, etc.) format.

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/google-drive/package.json
+++ b/packages/pieces/google-drive/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-google-drive",
-  "version": "0.6.0"
+  "version": "0.5.1"
 }
--- a/packages/pieces/google-drive/package.json
+++ b/packages/pieces/google-drive/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-google-drive",
-  "version": "0.5.0"
+  "version": "0.6.0"
 }
--- a/packages/pieces/google-drive/src/lib/action/download-file.ts
+++ b/packages/pieces/google-drive/src/lib/action/download-file.ts
@@ -1,0 +1,76 @@
+import { googleDriveAuth } from '../..';
+import { createAction, Property } from '@activepieces/pieces-framework';
+
+export const downloadFile = createAction({
+  auth: googleDriveAuth,
+  name: 'download_ggdrive_file',
+  displayName: 'Download file',
+  description: 'Download a selected file from google drive file',
+  props: {
+    fileId: Property.ShortText({
+      displayName: 'File ID',
+      description: 'The ID of the file to download',
+      required: true,
+    }),
+    mimeType: Property.ShortText({
+      displayName: 'MIME type',
+      description: 'The MIME type of the file to download',
+      required: true,
+    }),
+  },
+  run: async ({ auth, propsValue }) => {
+    const googledlCall = async (url: string) => {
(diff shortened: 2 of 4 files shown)


packages/pieces/google-drive/package.json
```diff
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-google-drive",
-  "version": "0.5.0"
+  "version": "0.6.0"
```
abuaboud: Please only increment the minor version, as this change is backward compatible

packages/pieces/google-drive/src/lib/action/download-file.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const downloadFile = createAction({
+  auth: googleDriveAuth,
+  name: 'download_ggdrive_file',
+  displayName: 'Download file',
+  description: 'Download a selected file from google drive file',
+  props: {
+    fileId: Property.ShortText({
+      displayName: 'File ID',
+      description: 'The ID of the file to download',
```
abuaboud: In description, let's add this can come from new trigger or how to obtain this would be more useful.

packages/pieces/google-drive/src/lib/action/download-file.ts
```diff
@@ -0,0 +1,87 @@
...
+      required: true,
+    }),
+    mimeType: Property.ShortText({
+      displayName: 'MIME type',
+      description: 'The MIME type of the file to download',
```
abuaboud: I think it's good if we remove the description, it would clutter the UI.

packages/pieces/google-drive/src/lib/action/download-file.ts
```diff
@@ -0,0 +1,87 @@
...
+      required: true,
+    }),
+    fileName: Property.ShortText({
+      displayName: 'File name',
+      description: 'The indicative name of the file to download',
```
abuaboud: I think destination file name and the description would be that not the file name to download but rather what name to save it with.

abuaboud (changes requested):

packages/pieces/google-drive/src/lib/action/download-file.ts
```diff
@@ -0,0 +1,87 @@
...
+    fileId: Property.ShortText({
+      displayName: 'File ID',
+      description: 'File ID comming from | New File -> id |',
+      required: true,
+    }),
+    mimeType: Property.ShortText({
+      displayName: 'MIME type',
+      description: 'MIME type comming from | New File -> mimeType |',
+      required: true,
+    }),
```
abuaboud: I think it's possible to know mimetype from google while downloading the file, It would be better than asking the user.

packages/pieces/google-drive/src/lib/action/download-file.ts
```diff
@@ -0,0 +1,87 @@
...
+    fileName: Property.ShortText({
+      displayName: 'File name',
```
abuaboud: Change display Name to 'Destination File name'

abuaboud (changes requested):

abuaboud (approved):

PR #2427, opened 2023-08-31
feat(extract-text): New piece to extract text from any file

Armangiau3:
## What does this PR do?

This PR creates a new piece that contains one action (`Native Extraction`) that converts many types of documents (Microsoft 365 docs, OpenOffice docs, and PDF) to a `.textract.txt` file. This piece does not support any third-party services or OCR functionalities yet.


packages/pieces/extract-text/src/lib/actions/native-text-extraction.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export const nativeTextExtraction = createAction({
+  name: 'native_text_extraction',
+  description:
+    'Extract text from many a file types (pdf, docx, pptx, xlsx, odt, odp, ods)',
+  displayName: 'Native Extraction',
+  props: {
+    file: Property.File({
+      displayName: 'File',
+      description: 'The file to extract text from',
```
abuaboud: We can remove the description here it's redundant

packages/pieces/extract-text/src/lib/actions/native-text-extraction.ts
```diff
@@ -0,0 +1,123 @@
...
+      required: true,
+    }),
+    precisePdfBreakingLines: Property.Checkbox({
+      displayName: 'Precise PDF Breaking Lines',
+      description:
+        'The pdf breaking lines annalisis is expensive but should works for small files',
```
abuaboud: `Checking this option will analyze PDF breaking lines more accurately. This might slow down processing for large files, but it works well for smaller ones.'`

abuaboud (changes requested):

abuaboud (dismissed):

PR #2499, opened 2023-09-04
fix: HTML/XML output not being shown correctly 

### LINKED ISSUE #1699: [BUG]: Result display will not display xml/html correctly
**Describe the bug**
If a piece returns something that contains xml or html tags are stripped in the result display. This is really bad user experience as it's unexpected and a normal user, who doesn't know what xss is and why tags may be stripped from ugc, will be completely confused by this. It also makes debugging way harder as you don't have any useful insight in how the structure looks and attribute values are not visible at all.

**To Reproduce**
1. Create a code step with the following code
```typescript
export const code = async (inputs) => {
    return '<hello>aaa<world>bbb</world>ccc</hello>'
};
```
2. Test the step
3. Look at the result

**Expected behavior**
Especially as scraping websites is a common use-case in no-code engines xml and html should be fully supported. Instead of stripping tags they should be escaped with their html encodings (&amp;lt; and &amp;gt;) to prevent xss.

**Screenshots**
image

AbdulTheActivePiecer:
## What does this PR do?

closes #1699

### DIFF
--- a/packages/ui/feature-builder-test-steps/src/lib/test-code-step/test-code-step.component.ts
+++ b/packages/ui/feature-builder-test-steps/src/lib/test-code-step/test-code-step.component.ts
@@ -86,7 +86,6 @@ export class TestCodeStepComponent extends TestStepCoreComponent {
           take(1),
           tap((step) => {
             if (step && step.type === ActionType.CODE) {
-              console.log(JSON.stringify(result.output));
               const clone: CodeAction = {
                 ...step,
                 settings: {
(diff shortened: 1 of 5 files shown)


packages/ui/feature-builder-test-steps/src/lib/test-code-step/test-code-step.component.ts
```diff
@@ -86,6 +86,7 @@ export class TestCodeStepComponent extends TestStepCoreComponent {
           take(1),
           tap((step) => {
             if (step && step.type === ActionType.CODE) {
+              console.log(JSON.stringify(result.output));
```
abuaboud: Remove

abuaboud (approved):

abuaboud (approved):

PR #2517, opened 2023-09-05
feat(linear): create project, update project, create comment action & new issue trigger

lldiegon:
## What does this PR do?

This combination of triggers and actions allows me to:

- Get one step closer to automating client onboarding by creating a project board.
- Get notifications of new issues in my private discord server.
- Let ChatGPT brainstorm about new issues and place a comment to the issue/task with the output.

Announcement=true

### DIFF
--- a/packages/pieces/linear/src/lib/actions/projects/update-project.ts
+++ b/packages/pieces/linear/src/lib/actions/projects/update-project.ts
@@ -1,0 +1,62 @@
+import { createAction, Property } from '@activepieces/pieces-framework';
+import { linearAuth } from '../../..';
+import { props } from '../../common/props';
+import { makeClient } from '../../common/client';
+import { LinearDocument } from '@linear/sdk';
+
+export const linearUpdateProject = createAction({
+    auth: linearAuth,
+    name: 'linear_update_project',
+    displayName: 'Update Project',
+    description: 'Update a existing project in Linear workspace',
+    props: {
+        team_id: props.team_id(),
+        project_id: props.project_id(),
+        name: Property.ShortText({
+            displayName: 'Project Name',
+            description: 'The name of the new project',
+            required: true,
+        }),
+        description: Property.LongText({
+            displayName: 'Description',
+            description: 'The description of the new project',
+            required: false,
+        }),
+        icon: Property.ShortText({
+            displayName: 'Icon',
+            description: 'The icon for the new project',
+            required: false,
+        }),
+        color: Property.ShortText({
+            displayName: 'Color',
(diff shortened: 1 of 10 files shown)


packages/pieces/linear/src/lib/common/props.ts
```diff
@@ -188,4 +188,36 @@ export const props = {
         };
       },
     }),
+
+  project_id: (required = true) =>
+    Property.Dropdown({
+      displayName: 'Project',
+      required,
+      description: 'ID of Linear Project',
```
abuaboud, line 196: Nit: Please remove this as It would clutter the interface

packages/pieces/linear/src/lib/common/props.ts
```diff
@@ -98,7 +98,7 @@ export const props = {
     }),
   assignee_id: (required = false) =>
     Property.Dropdown({
-      description: 'Assignee of the Issue',
+      description: 'Assignee of the Issue / Comment',
```
abuaboud, line 101: Nit: Let's remove as it's clear

packages/pieces/linear/src/lib/common/props.ts
```diff
@@ -5,7 +5,7 @@ import { LinearDocument } from '@linear/sdk';
 export const props = {
   team_id: (required = true) =>
     Property.Dropdown({
-      description: 'The team for which the issue will be created',
+      description: 'The team for which the issue, project or comment will be created',
```
abuaboud, line 8: Nit: Let's remove as it's clear

packages/pieces/linear/src/lib/actions/projects/update-project.ts
```diff
@@ -0,0 +1,62 @@
...
+            required: false,
+        }),
+        startDate: Property.DateTime({
+            displayName: 'Start Date',
+            description: 'The start date for the new project',
+            required: false,
+        }),
+        targetDate: Property.DateTime({
+            displayName: 'Target Date',
+            description: 'The target date for the new project',
```
abuaboud: I think all props here in general doesn't need the description, we can remove it to make the user interface clear

abuaboud (changes requested):

packages/pieces/linear/src/lib/common/props.ts
lldiegon: Alright done! I agree that it should only be used when it needs additional explanation.

abuaboud:
@allcontributors please add @lldiegon for plugin

abuaboud (approved):

PR #2585, opened 2023-09-09
feat(sessions): Create Session/ Get Transcripts/ Get Takeaways

### LINKED ISSUE #1911: [Piece]: Sessions
Sessions.us can be integrated  for major activities

Owlcept:
## What does this PR do?

This PR does the following:

# New Piece: Sessions

- Create a Session (returns id and link)
- Get Transcripts from Sessions
- Get Takeaways from Sessions

Fixes activepieces/under-review#589


Owlcept:
Ooops, I see, linting issues, let me fix

Owlcept:
one last lint issue, but in one of the files I use let x in <list>, it is not very happy with this, should I convert it to const x ?

abuaboud:
HI @Owlcept I just noticed the issue is not descriptive but I think most important usecases are all related to triggers, I think sessions.us support webhooks, do you think we can do that?

Owlcept:
> HI @Owlcept I just noticed the issue is not descriptive but I think most important usecases are all related to triggers, I think sessions.us support webhooks, do you think we can do that?

Yes! I did not even realize they had webhooks, let me add that in as well!

Owlcept:
[Here](https://resources.sessions.us/resources/help-center/tools-files-integrations/integrations/webhooks#4744bee67526412fb8fb4462a6dc7a2f) is the relevant info just incase

Owlcept:
@abuaboud Webhooks have been added! Still may not pass due to some warnings but they are there and good - they are based on the gravityform type of typehooks as Sessions does it the same way all webhooks are treated the same way from what I can tell

PR #2592, opened 2023-09-10
fix(slack): load more than 100 channel

abuaboud:
## What does this PR do?

ANNOUNCEMENT=true
Fixes # (issue)


im-bravo:
Hi, 
I have same requirement about slack.
My slack workspace have lot of channels.
If the filter outed channel list can be displayed based on my input, it's could be better.

PR #2594, opened 2023-09-10
feat(notion): add update database item action

### LINKED ISSUE #1134: Update a page in a Notion database
Hello 👋

**Describe your usecase.**
Currently there is only a Notion trigger. We should add a Notion action to update a page in a database for example. 

**Example workflow**
When a Notion Page is created in a database, then a request to the Wikipedia API is made, finally the Notion Page is updated with these informations.

I'll try to investigate and maybe make my first PR in this project.
Have a good day!

kishanprmr:
## What does this PR do?

Fixes #1134 

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/notion/package.json
+++ b/packages/pieces/notion/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-notion",
-  "version": "0.3.2"
+  "version": "0.2.3"
 }
--- a/packages/pieces/notion/package.json
+++ b/packages/pieces/notion/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-notion",
-  "version": "0.2.2"
+  "version": "0.3.2"
 }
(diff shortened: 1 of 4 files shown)


packages/pieces/notion/package.json
```diff
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-notion",
-  "version": "0.2.2"
+  "version": "0.3.2"
```
abuaboud: Please bump the minor version as it is backward compatible

abuaboud (changes requested):

packages/pieces/notion/package.json
kishanprmr: Is it should be 0.4.2? I have already increased it to 0.3.2 from 0.2.2

abuaboud: It should be 0.2.3, as it doesn't effect existing versions

abuaboud (approved):

abuaboud (approved):

PR #2618, opened 2023-09-12
fix: unquoted migration parameter

dennis-tra:
Just ran into this when updating from 0.3.9 to 0.8.0.


abuaboud:
@allcontributors please add @dennis-tra for bug

abuaboud:
Thank you for updating that. We had to rewrite the migration using native SQL because it will fail if the entity is no longer used in TypeORM, and this went under the radar.

You are awesome. Thank you!

PR #2619, opened 2023-09-12
feat(open-router): adding openrouter piece

### LINKED ISSUE #2607: Open Router Piece

Salem-Alaa:
## What does this PR do?

adding open router piece

Fixes #2607


packages/pieces/open-router/src/lib/common/index.ts
```diff
@@ -0,0 +1,36 @@
+export const openRouterModels  = [
```
abuaboud, line 1: We can load this dynamically?

abuaboud (dismissed):

PR #2649, opened 2023-09-13
feat(text-helper): add concat texts / find / replace / split actions

### LINKED ISSUE #2304: String utility piece
A string utility piece that does the following :

1)  Concatenate two or more strings.
2)  Split string by delimiter.
3)  Find substring (Regex or string).
4)  Replace in string (Regex or string).

abaza738:
### Commits
- feat(string-utils): add string utils piece

ANNOUNCEMENT=true
Fixes #2304.


abaza738:
Note: the piece icon needs to be changed.

abuaboud (approved):

PR #2709, opened 2023-09-18
feat(openai): adding whisper model actions

### LINKED ISSUE #2072: [Piece]: Whisper
Transcribe my audio

Salem-Alaa:
## What does this PR do?
Fixing whisper model existing with Ask "ChatGPT" action.
Adding transcription action using whisper-1 model
Adding translate action using whisper-1 model

ANNOUNCEMENT=true
Fixes #2072

### DIFF
--- a/packages/pieces/openai/src/lib/actions/translation.ts
+++ b/packages/pieces/openai/src/lib/actions/translation.ts
@@ -15,18 +15,12 @@ export const translateAction = createAction({
             required: true,
             description: 'Audio file to translate',
         }),
-        fileName : Property.ShortText({
-            displayName: 'File name',
-            required: true,
-            description: 'Name of the file',
-        }),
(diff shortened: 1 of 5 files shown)


packages/pieces/openai/src/lib/actions/translation.ts
```diff
@@ -0,0 +1,48 @@
...
+    displayName: 'translate',
+    description: 'translate audio to text using whisper-1 model',
+    auth: openaiAuth,
+    props: {
+        audio : Property.File({
+            displayName: 'Audio',
+            required: true,
+            description: 'Audio file to translate',
+        }),
+        fileName : Property.ShortText({
```
abuaboud: You don't really need this, it won't effect the result, you can just use the name in audio property.

PR #2757, opened 2023-09-20
fix(airtable): fixing attachment field

Salem-Alaa:
## What does this PR do?
Fixing the attachment field not uploading

Fixes # (issue)

### DIFF
--- a/packages/pieces/airtable/src/lib/common/index.ts
+++ b/packages/pieces/airtable/src/lib/common/index.ts
@@ -105,7 +105,7 @@ export const airtableCommon = {
     displayName: 'Table',
     required: true,
     refreshers: ["base", "tableId"],
-
+    
     props: async ({ auth, base, tableId }) => {
       if (!auth) return {}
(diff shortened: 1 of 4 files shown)


packages/pieces/airtable/src/lib/common/index.ts
```diff
@@ -157,8 +157,41 @@ export const airtableCommon = {
...
 
+    try{
+      const airtable: AirtableTable = await airtableCommon.fetchTable({
+        token: auth,
+        baseId: base,
+        tableId: tableId
+      });
 
+      let count = 0;
+      for( let i = 0 ; i < airtable.fields.length ; i++ ){
```
AbdulTheActivePiecer: please use airtable.fields.forEach instead of for() as it is easier to read and evade errors with the index.

AbdulTheActivePiecer (changes requested):

abuaboud (approved):

PR #2794, opened 2023-09-24
fix(google-sheet): Get next row handling out of bound

### LINKED ISSUE #2751: feat(google-sheets): Get next row handling out of bound.
The piece errors out if there are no more rows to be returned, we should just return an empty row.

Salem-Alaa:
## What does this PR do?
Calling the get next row(s) rapidly will cause a race condition in which the newer calls will user the old value of the counter.
Outbound fixed
Fixes #2751 
Announcement=true

### DIFF
--- a/packages/pieces/google-sheets/package.json
+++ b/packages/pieces/google-sheets/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-google-sheets",
-  "version": "0.7.2"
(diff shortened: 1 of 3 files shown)


abuaboud (commented):
Did you test this e2e, I remember google threw an error on http request when its outbound

abuaboud (approved):

PR #2806, opened 2023-09-25
feat(openai): adding API key validation

Salem-Alaa:
## What does this PR do?
Adding Api key validation feature

Fixes # (issue)

### DIFF
--- a/packages/pieces/openai/src/index.ts
+++ b/packages/pieces/openai/src/index.ts
@@ -2,6 +2,7 @@ import { PieceAuth, createPiece } from '@activepieces/pieces-framework';
 import { askOpenAI } from './lib/actions/send-prompt';
 import { transcribeAction } from './lib/actions/transcriptions';
 import { translateAction } from './lib/actions/translation';
+import { AuthenticationType, HttpMethod, httpClient } from '@activepieces/pieces-common';
 const markdownDescription = `
 Follow these instructions to get your OpenAI API Key:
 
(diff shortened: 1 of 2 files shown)


packages/pieces/openai/src/index.ts
```diff
@@ -14,7 +15,32 @@ It is strongly recommended that you add your credit card information to your Ope
 export const openaiAuth = PieceAuth.SecretText({
   description: markdownDescription,
   displayName: 'Api Key',
-  required: true
+  required: true,
+  validate: async (auth) => {
+    try{
+      console.log("-=-=-=-=-=-=-=-");
+      console.log(auth);
+      console.log("-=-=-=-=-=-=-=-");
```
abuaboud: Please remove these console.log

packages/pieces/openai/src/index.ts
```diff
@@ -14,7 +15,32 @@ It is strongly recommended that you add your credit card information to your Ope
...
+          token: auth.auth as string
+        }
+      });
+      return{
+        valid: true,
+      }
+    }catch(e){
+      return{
+        valid: false,
+        error: 'Invalid API token'
```
abuaboud, line 37: I would try to use ask instead of model, just to make sure we avoid issues like 429 too and the error i would make it has more information like at least status code or message so they can debug further.

abuaboud (changes requested):

abuaboud (approved):

PR #2903, opened 2023-10-04
feat(math): division / addition / subtraction /  Multiplication / module / generate random number actions

### LINKED ISSUE #2303: Math utility piece
A utility piece that has the following actions: 

1) Add 2 numbers.
2) Subtract 2 numbers.
3) Multiplies 2 numbers.
4) Divides 2 numbers.
5) Mods 2 numbers.
6) Generate random number.

Make sure you validate the numbers during runtime.

lisander-lopez:
## What does this PR do?
- Added "Math" Piece
  - Division Action
  - Addition Action
  - Subtraction Action
  - Multiplication Action
  - Modulo Action
  - Generate Number between two numbers (inclusive)
 
- Created `nonZero` validator and added to docs

This is a draft PR as this is my first OSS contribution, I know I will have comments, FIRE THEM AWAY!

Some Questions I had:
What about the image for the piece?
Although, 0 !=== "0" I added a note in validator docs that this and Validator.number pairs well together. Is it even worth it to add that note? I added that as a "just in case" other devs try to use it by itself expecting more out of it. 
Is the UI good? (Text labels etc...)

Fixes #2303 

ANNOUNCEMENT=true


packages/pieces/math/src/lib/actions/division.ts
```diff
@@ -0,0 +1,25 @@
...
+            displayName: 'First Number',
+            description: undefined,
+            required: true,
+            validators: [Validators.number]
+        }),
+        second_number: Property.Number({
+            displayName:  'Second Number',
+            description: undefined,
+            required: true,
+            validators: [Validators.number, Validators.nonZero]
```
abuaboud: You can use Validator.minValue(1) instead and then you can remove Validators.NonZero

packages/pieces/math/src/lib/actions/multiplication.ts
```diff
@@ -0,0 +1,25 @@
...
+    name: 'multiplication_math',
+    auth: PieceAuth.None(),
+    displayName: 'Multiplicattion',
+    description: 'Multiply first number by the second number',
+    props: {
+        first_number: Property.Number({
+            displayName: 'First Number',
+            description: undefined,
+            required: true,
+            validators: [Validators.number]
```
abuaboud: You can remove validators: [Validators.number] as it's included by default in Property.Number

abuaboud (changes requested):

packages/pieces/math/src/lib/actions/division.ts
lisander-lopez: What if we want to generate a number between -5 and 5?

abuaboud: @lisander-lopez This is division action, I think we dont need non zero for generate random numbee

packages/pieces/math/src/lib/actions/multiplication.ts
lisander-lopez: Ok, I will remove them from almost all of them then since I used [Validators.number] a lot throughout this.

packages/pieces/math/src/lib/actions/division.ts
lisander-lopez: Oh sorry, I was reading the wrong code. The reason I put non-zero is because you cannot divide by 0. but you can divide by a negative number

packages/pieces/math/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@activepieces/piece-math",
```
AbdulTheActivePiecer: please change this to math-helper

packages/pieces/math/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import { division } from "./lib/actions/division";
+import { modulo } from "./lib/actions/modulo";
+import { generateRandom } from "./lib/actions/generateRandom";
+
+const markdownDescription = `
+Perform mathematical operations.
+`;
+
+export const math = createPiece({
+  displayName: "Math",
```
AbdulTheActivePiecer: Please change this to "Math Helper"

packages/pieces/math/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  description: markdownDescription,
+  auth: PieceAuth.None(),
+  minimumSupportedRelease: '0.9.0',
+  logoUrl: "https://cdn.activepieces.com/pieces/math.png",
```
AbdulTheActivePiecer: the url is this https://cdn.activepieces.com/pieces/math-helper.svg

AbdulTheActivePiecer (changes requested):

abuaboud (approved):

PR #2913, opened 2023-10-05
fix: notion number casting

abuaboud:
## What does this PR do?

Fixes # (issue)

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/notion/package.json
+++ b/packages/pieces/notion/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-notion",
-  "version": "0.2.4"
+  "version": "0.2.5"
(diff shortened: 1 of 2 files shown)


kishanprmr:
@mo Although the number field of Notion is set as the NUMBER field of Activepieces, inside **run** function of **Create Database Item**  number field value comes as a string. Because of that, this error occurred.
Shouldn't it should come as a number as it is NumberProperty?
Am I understanding this DynamicProps wrong?

AbdulTheActivePiecer:
@kishanprmr we currently don't parse numbers to be so, we might need to add that @abuaboud  same thing for JSON properties.

PR #2933, opened 2023-10-10
fix: docs links that are broken

rbnswartz:
## What does this PR do?

Fixes some broken documentation links about piece sharing


abuaboud:
@allcontributors please add @rbnswartz for docs

PR #2934, opened 2023-10-10
fix: branch failure & migrate flow templates

abuaboud:
## What does this PR do?

Fixes # (issue)


khaledmashaly (commented):
Seems like the flow template migration wasn't added to sqlite migrations. Is this on purpose?

PR #2948, opened 2023-10-12
feat(framework): adding more features to pieces

Salem-Alaa:
## What does this PR do?

adding the feature to add the piece type and others

Fixes # (issue)

### DIFF
--- a/packages/pieces/framework/src/lib/piece.ts
+++ b/packages/pieces/framework/src/lib/piece.ts
@@ -69,6 +69,10 @@ export const createPiece = <PieceAuth extends PieceAuthProperty>(params: CreateP
   )
 }
 
+export enum PieceCategory {
+  core = 'core',
+}
+
 type CreatePieceParams<PieceAuth extends PieceAuthProperty = PieceAuthProperty> = {
   displayName: string
   logoUrl: string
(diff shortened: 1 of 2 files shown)


packages/pieces/framework/src/lib/piece.ts
```diff
@@ -69,6 +69,10 @@ export const createPiece = <PieceAuth extends PieceAuthProperty>(params: CreateP
   )
 }
 
+export enum PieceCategory {
+  core = 'core',
+}
+
```
AbdulTheActivePiecer, line 75: in enums, we use ALL CAPS.

packages/pieces/framework/src/lib/piece.ts
```diff
@@ -80,6 +84,8 @@ type CreatePieceParams<PieceAuth extends PieceAuthProperty = PieceAuthProperty>
   maximumSupportedRelease?: string
   actions: Action<PieceAuth>[]
   triggers: Trigger<PieceAuth>[]
+  mentionIconUrl?:string
+  categories?: PieceCategory[]
```
AbdulTheActivePiecer, line 88: You forgot to add these to the Piece class itself.

AbdulTheActivePiecer (commented):
This is a review just so you know what was wrong in it, I am fixing them and merging other prs that depend on it, I apologise for doing so, I am just catching up after being off for a while Salem.

PR #2994, opened 2023-10-18
fix: webhook multipart requests

### LINKED ISSUE #2930: Mutlipart Form Data
The convert body function doesn't work with the multipart form.
request.parts() do not return the required result.

Image

khaledmashaly:
## What does this PR do?

* fixes #2930 
* fixes logger level in dev mode

### DIFF
--- a/packages/backend/src/app/webhooks/webhook-controller.ts
+++ b/packages/backend/src/app/webhooks/webhook-controller.ts
@@ -176,23 +176,17 @@ async function convertRequest(request: FastifyRequest): Promise<EventPayload> {
     return payload
 }
 
-const convertBody = async (request: FastifyRequest) => {
+const convertBody = async (request: FastifyRequest): Promise<unknown> => {
     if (request.isMultipart()) {
         const jsonResult: Record<string, unknown> = {}
-        const parts = request.parts()
-        for await (const part of parts) {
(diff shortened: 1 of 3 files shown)


packages/backend/src/app/webhooks/webhook-controller.ts
```diff
@@ -176,23 +176,17 @@ async function convertRequest(request: FastifyRequest): Promise<EventPayload> {
...
-                const fileBuffer = Buffer.concat(chunks)
-                jsonResult[part.fieldname] = fileBuffer.toString('base64')
-            }
-            else {
-                jsonResult[part.fieldname] = part.value
-            }
+        const requestBodyEntries = Object.entries(request.body as Record<string, unknown>)
+
+        for (const [key, value] of requestBodyEntries) {
+            jsonResult[key] = value instanceof Buffer ? value.toString('base64') : value
```
abuaboud, line 185: Can we use memory file format, its ugly to expose it from engine but whay do you think?

abuaboud (dismissed):

packages/backend/src/app/webhooks/webhook-controller.ts
khaledmashaly: will move it to another task

PR #2996, opened 2023-10-19
fix(common): http doesn't add query params correctly

### LINKED ISSUE #2993: [BUG]: Invalid request URL with parameters
**Describe the bug**
If I use a **httpClient** from **@activepieces/pieces-common** with **queryParams**, I'm getting an incorrect final request URL.
Related to #186

**To Reproduce**
I'm creating a piece with code like this:
```js
const API_URL = 'https://example.com/api/items';
const queryParams: Record<string, string> = {
    api_key: params.apiKey
};
const request: HttpRequest = {
    method: HttpMethod.GET,
    url,
    queryParams
};
const response = await httpClient.sendRequest<ChatDataItem[]>(request);
```
When testing the trigger, I see that the request is sent to: **https://example.com/api/itemsapi_key=123456**.
That is, there is no separator "?".

**Expected behavior**
The final request URL should be: **https://example.com/api/items?api_key=123456**.

**Additional context**
This can be fixed like this:

_/packages/pieces/common/src/lib/http/core/base-http-client.ts_
Remove lines 28 -32:
```js
// if (request.queryParams) {
//    for (const [name, value] of Object.entries(request.queryParams)) {
//      url.searchParams.append(name, value);
//    }
// }
```

_/packages/pieces/common/src/lib/http/axios/axios-http-client.ts_
Add line after 32:
```js
params: request.queryParams,
```

andchir:
## What does this PR do?

When testing the trigger, I'm getting a 404 error because the URL is invalid. HttpClient code fixed.

Fixes #2993

### DIFF
--- a/packages/pieces/common/package.json
+++ b/packages/pieces/common/package.json
@@ -1,5 +1,5 @@
 {
   "name": "@activepieces/pieces-common",
-  "version": "0.2.3",
+  "version": "0.2.4",
   "type": "commonjs"
 }
--- a/packages/pieces/common/src/lib/http/axios/axios-http-client.ts
+++ b/packages/pieces/common/src/lib/http/axios/axios-http-client.ts
@@ -29,6 +29,7 @@ export class AxiosHttpClient extends BaseHttpClient {
             const config:AxiosRequestConfig = {
                 method: axiosRequestMethod,
                 url,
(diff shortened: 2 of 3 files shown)


abuaboud:
@allcontributors please add @andchir for bug

PR #3013, opened 2023-10-22
feat: platform APIs

### LINKED ISSUE #2978: Add Platform Entity
Platform Entity should has the following

- Branding Theme
  - Primary Color
  - Fav icon
  - Logo Icon
  - Full Logo Icon

Add Project
- Type / Platform Id

khaledmashaly:
## What does this PR do?

- platform endpoints
- admin platform endpoints
- tests
- fixes #2978


packages/backend/src/app/ee/platform/platform.controller.ts
```diff
@@ -0,0 +1,37 @@
+import { FastifyPluginCallbackTypebox, Type } from '@fastify/type-provider-typebox'
+import { UpdatePlatformRequestBody } from '@activepieces/ee-shared'
+import { ApId } from '@activepieces/shared'
+import { platformService } from './platform.service'
+
+export const platformController: FastifyPluginCallbackTypebox = (app, _opts, done) => {
+    app.patch('/:id', UpdatePlatformRequest, async (req) => {
```
abuaboud: `patch` and `post` is both valid, usually patch is used when you update certain fields and keeping others same, while post doesn't have this restriction in the convention it can either be means replace or update certian fields but to stay consistent with other APIs let's keep it post.

packages/backend/src/app/ee/platform/platform.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+        return await platformService.update({
+            id: req.params.id,
+            userId: req.principal.id,
+            ...req.body,
+        })
+    })
+
+    app.get('/:id', GetPlatformRequest, async (req) => {
```
abuaboud, line 15: Nit: both APIs required to be restricted by information from JWT token, but i assume this coming in the next pr.

abuaboud: Update: I see you assert by owner Id, usually another fail safe way to make it query by owner id and id, this also prevent mistakes.

packages/backend/src/app/project/project-entity.ts
```diff
@@ -2,14 +2,14 @@ import { EntitySchema } from 'typeorm'
...
     flows: Flow[]
     files: File[]
     folders: Folder[]
     events: TriggerEvent[]
     appConnections: AppConnection[]
-} & Project
+}
 
 export const ProjectEntity = new EntitySchema<ProjectSchema>({
     name: 'project',
```
abuaboud, line 17: I think we need to add platformId and type with their migration, please keep the migration separated as it will be for all editions

abuaboud (changes requested):
This is pure gold!

packages/backend/src/app/ee/platform/platform.module.ts
```diff
@@ -0,0 +1,8 @@
+import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox'
+import { platformController } from './platform.controller'
+import { adminPlatformController } from './admin-platform.controller'
+
+export const platformModule: FastifyPluginAsyncTypebox = async (app) => {
+    await app.register(platformController, { prefix: '/v1/platform' })
+    await app.register(adminPlatformController, { prefix: '/v1/admin/platform' })
```
abuaboud: Please separate this into another module, so we can exclude it from enterprise version.

PR #3047, opened 2023-10-26
fix: improve docker layer caching and ignored files

maor-rozenfeld:
An easy change in the Dockerfile to improve layer caching of subsequent builds, and avoid cache invalidation due to unrelated files.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,11 @@
 .angular
-node_modules
+.dockerignore
 .env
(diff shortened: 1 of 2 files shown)


abuaboud:
@allcontributors please add @maor-rozenfeld for code

abuaboud:
This is interesting!

abuaboud (approved):

PR #3075, opened 2023-10-30
ci: add ImageName as optional deployment paramter and minor fixes

maor-rozenfeld:
1. ImageName can be defined in pulumi config or as an environment variable (unless `deployLocalBuild` is `true`)
2. Communication between the load balancer and fargate is defined through a security group instead of keeping fargate open to the entire internet
3. `apEncryptionKey` and `apJwtSecret` are defined as pulumi secrets

I think we should also change the default rules of the ALB security group, it shouldn't be completely open by default. Not sure what's the best way to configure it.

### DIFF
--- a/deploy/pulumi/index.ts
+++ b/deploy/pulumi/index.ts
@@ -9,8 +9,8 @@ import * as child_process from "child_process";
 const stack = pulumi.getStack();
 const config = new pulumi.Config();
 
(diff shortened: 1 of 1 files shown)


abuaboud (approved):

abuaboud:
Thank you @maor-rozenfeld keep these contributions going <3 We love them!

PR #3083, opened 2023-10-30
feat(backend): Managed Authentication

### LINKED ISSUE #2979: Managed User Authentication Backend
Give the Platform a Public/Private Key, Where they use Public Key to Sign users and redirect them to the frontend so it's exchanged to temporary JWT Token.

khaledmashaly:
## What does this PR do?

fixes #2979


packages/backend/src/app/database/postgres-connection.ts
```diff
@@ -159,6 +160,7 @@ const getMigrations = (): (new () => MigrationInterface)[] => {
                 AddPlatform1697717995884,
                 AddCustomDomain1698077078271,
                 AddSigningKey1698602417745,
+                ManagedAuthnInitial1698700720482,
```
abuaboud: Let's move to all platforms specially the edits on project and user

packages/ee/shared/src/lib/managed-authn/managed-authn-requests.ts
```diff
@@ -0,0 +1,12 @@
+import { Type, Static } from '@sinclair/typebox'
+
+export const ManagedAuthnRequestBody = Type.Object({
```
abuaboud, line 3: Why not ManagedAuth, or ManagedAuthentication?

abuaboud (changes requested):

packages/ee/shared/src/lib/managed-authn/managed-authn-requests.ts
khaledmashaly: authn is the defacto abbreviation for authentication, authz is for authorization

packages/backend/src/app/database/postgres-connection.ts
khaledmashaly: done

PR #3093, opened 2023-11-01
feat(mailchimp): add note to subscriber / add member to list / add tag to subscriber / update subscriber status / remove tag from subscriber actions and new subscriber trigger 

TaskMagicKyle:
## What does this PR do?
Add Mailchimp triggers/actions

ANNOUNCEMENT=true


packages/pieces/mailchimp/src/lib/actions/add-subscriber-to-tag.ts
```diff
@@ -0,0 +1,58 @@
...
+        {
+          headers: {
+            Authorization: `Bearer ${access_token}`,
+          },
+        }
+      );
+
+      return  response.data,
+    } catch (error) {
+      return error;
```
abuaboud: @TaskMagicKyle This will cause the step always to be successful, is this intended in that use-case?

I think it's better to inform the user there is a failure rather than consider it a successful 

Same for other actions, what do you think?

abuaboud (commented):
Great Pr, Thank you!

Can you please bump version in package.json?

packages/pieces/mailchimp/src/lib/actions/add-subscriber-to-tag.ts
KyleBuilds: Removed now. Not sure why axios was used vs standard httpClient but I updated the PR to use that. Getting the hang of things!


Package.json bumped to 0.3.2, I assume this is what you had meant.

abuaboud (approved):

packages/pieces/mailchimp/src/lib/actions/add-subscriber-to-tag.ts
abuaboud: Thank you!

abuaboud (approved):

PR #3097, opened 2023-11-01
feat(airtable): find record / delete record / update record actions

TaskMagicKyle:
## What does this PR do?
Add Find record
Add Delete record
Add Update record

ANNOUNCEMENT=true


packages/pieces/airtable/src/lib/common/index.ts
```diff
@@ -14,254 +22,492 @@ export const airtableCommon = {
...
+        const airtable: AirtableTable = await airtableCommon.fetchTable({
+          token: auth as unknown as string,
+          baseId: base as unknown as string,
+          tableId: tableId as unknown as string,
+        });
 
-    try{
+        fieldNames = airtable.fields.map((field: AirtableField) => field.name);
+      } catch (e) {
+        console.debug(e);
```
abuaboud: I think it's better to make it throw the error, so the user can know in case there is problem instead of blank field names

What do you think?

abuaboud: Also the same thing in above files

abuaboud (changes requested):
Can you please bump the package.json version

packages/pieces/airtable/src/lib/common/index.ts
KyleBuilds: bumped + removed try/catch

abuaboud (approved):

PR #3106, opened 2023-11-02
fix: turbowatch only watch give files and skip-nx-cache

iam-dev0:
## What does this PR do?

https://github.com/gajus/turbowatch/issues/32

This PR introduces optimizations to the file-watching process in our development environment. Previously, turbo watch was monitoring all files across packages, even when specific pieces like http were singled out in AP_DEV_PIECES. This caused performance bottlenecks and ran up against system watcher limits.


abuaboud (approved):

abuaboud:
@allcontributors please add @awais000 for code

PR #3111, opened 2023-11-03
refactor: engine with test coverage

abuaboud:
## What does this PR do?

Fixes # (issue)


AbdulTheActivePiecer:
Once this is merged we can work on #3420

packages/backend/src/app/flows/step-run/step-run-service.ts
```diff
@@ -203,10 +184,8 @@ const executeBranch = async ({ step, flowVersion, projectId }: ExecuteParams<Bra
         }
     }
 
-    const branchStepOutput = new ExecutionState(result.executionState).getStepOutput<BranchStepOutput>({
-        stepName: branchStep.name,
-        ancestors: [],
-    })
+    // TODO FIX
+    const branchStepOutput = result.executionState.steps[branchStep.name]
```
khaledmashaly: Just bringing your attention to make sure everything is ok!

packages/engine/src/lib/handler/context/step-execution-path.ts
```diff
@@ -0,0 +1,20 @@
+export class StepExecutionPath {
+    public path: readonly [string, number][] = []
```
khaledmashaly, line 2: I think it will be more readable to convert this tuple into an object instead.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -0,0 +1,55 @@
...
+        executionState,
+        constants,
+    }: {
+        action: LoopOnItemsAction
+        executionState: FlowExecutorContext
+        constants: EngineConstantData
+    }) {
+        const { resolvedInput, censoredInput } = await constants.variableService.resolve<LoopOnActionResolvedSettings>({
+            unresolvedInput: action.settings,
+            executionState,
```
khaledmashaly, line 25: Just thinking out loud, why not check for completeness to skip execution just like other executers?

abuaboud: So its get back to the correct iteration

PR #3128, opened 2023-11-06
feat: configurable webhook timeout

fdundjer:
## What does this PR do?

Adds the ability to configure the webhook timeout. Before this change, the timeout was hardcoded to 30 seconds.

Fixes # (issue)


packages/ui/feature-builder-right-sidebar/src/lib/edit-step-sidebar/edit-step-form-container/edit-step-form-container.component.html
```diff
@@ -16,7 +16,8 @@
               <ap-markdown [fullWidth]="true"
                 data="If you are expecting a reply from this webhook, append <b>/sync</b> to the URL.<br> <br>
                   In that case, you will also have to add an <b>HTTP step</b>  with return response to the end of your flow.<br> <br>
-                  **If the flow takes more than 30 seconds, it will give a 204 no-content response.**
+                  **If the flow takes more than 30 seconds, it will give a 204 no-content response.**<br>
+                  To change this time, set the **AP_WEBHOOK_TIMEOUT** to desired number.
```
abuaboud: You can use flagService to check If it's not cloud

.env.example
```diff
@@ -13,6 +13,7 @@ AP_JWT_SECRET=
 
 AP_ENVIRONMENT=prod
 AP_FRONTEND_URL=http://localhost:8080
+AP_WEBHOOK_TIMEOUT=30
```
abuaboud: I suggest to rename it to AP_WEBHOOK_TIMEOUT_SECONDS

packages/ui/feature-builder-right-sidebar/src/lib/edit-step-sidebar/edit-step-form-container/edit-step-form-container.component.html
abuaboud: You can check like that

In html:

`edition$ | async === ApEdition.CLOUD`

in TS:
`const edition$ = flagsService.getEdition()`

There might be typos, I am typing it from mobile

abuaboud (changes requested):

.env.example
fdundjer: Done

PR #3132, opened 2023-11-06
feat(piece): Humanbot Piece (triggers)

andchir:
## What does this PR do?

Added 4 triggers for HumanBot app integration.


abuaboud:
Hi @andchir, I apologize for the late reply. I took a look at the website, and it looks great, especially for an early product. I suggest publishing it directly to npm to gain traction. It's often challenging to maintain pieces for early products as they changes alot.

PR #3146, opened 2023-11-09
fix: auto verify on invitation accept

abuaboud:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/backend/src/app/user/user-service.ts
+++ b/packages/backend/src/app/user/user-service.ts
@@ -36,18 +36,17 @@ export const userService = {
         return userRepo.save(user)
     },
     async verify({ userId }: { userId: string }): Promise<User> {
-        const user = await userRepo.findOneByOrFail({
-            id: userId,
-        })
+        const user = await userRepo.findOneByOrFail({ id: userId })
+    
         if (user.status === UserStatus.SHADOW) {
-            await userRepo.update(userId, {
-                status: UserStatus.VERIFIED,
-            })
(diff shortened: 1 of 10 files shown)


packages/backend/src/app/user/user-service.ts
```diff
@@ -35,6 +35,19 @@ export const userService = {
...
+        })
+        if (user.status === UserStatus.SHADOW) {
+            await userRepo.update(userId, {
+                status: UserStatus.VERIFIED,
+            })
+        }
+        return userRepo.findOneByOrFail({
+            id: userId,
+        })
+    },
```
khaledmashaly: I think that this method can be optimised to make less than 3 db calls to get the user verified, right?

packages/shared/src/lib/common/activepieces-error.ts
```diff
@@ -315,6 +323,7 @@ export enum ErrorCode {
     INVALID_CLAIM = 'INVALID_CLAIM',
     INVALID_CLOUD_CLAIM = 'INVALID_CLOUD_CLAIM',
     INVALID_CREDENTIALS = 'INVALID_CREDENTIALS',
+    EMAIL_IS_NOT_VERFIED = 'EMAIL_IS_NOT_VERFIED',
```
khaledmashaly, line 326: shared package version needs to be incremented!

khaledmashaly (commented):
Great! Please see my comments.

PR #3197, opened 2023-11-15
chore: added script to remove platform module from xlf files

AbdulTheActivePiecer:
## What does this PR do?

Removes platform module text from i18n

### DIFF
--- a/tools/scripts/xlf/remove-ee-modules-from-xlf-files.ts
+++ b/tools/scripts/xlf/remove-ee-modules-from-xlf-files.ts
@@ -33,12 +33,10 @@ async function findXlfFiles(directory: string): Promise<string[]> {
 }
 
 // Example usage
-const directoryPath = 'packages/ui/core/src/locale';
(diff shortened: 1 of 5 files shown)


abuaboud (approved):

tools/scripts/xlf/remove-ee-modules-from-xlf-files.ts
```diff
@@ -0,0 +1,64 @@
...
+              }
+              });
+            });
+
+           
+        }
+        });
+   })
+  })
+  .catch((error) => console.error('Error:', error));
```
AbdulTheActivePiecer: Currently this code iterates over all xlf files and removes the context/context-groups that contain files from specified modules, I don't know if instead we are just supposed to modifying the main messages.xlf file and then the other locales will be extracted from it instead, if so please let me know.

AbdulTheActivePiecer: Per Mo, we only generate messages.xlf and then let crowdin generate the other files.

PR #3206, opened 2023-11-16
feat(pieces-form): add separator, title, group

Doskyft:
Hello !

I am currently working on the creation of a piece which contains A LOT of fields, to improve the experience, I added "PropertyStyle", to be able to separate and group the fields together.

A small breaking change, mandatory fields are no longer put first.

Example
pieces-form-group
pieces-form-title


abuaboud:
Nice, @Doskyft!

The first picture seems like a group, and the second one looks like a title with a separator—kind of like creating a form group on the frontend, which makes sense.

I'm curious, though, about why someone would prefer Title + Separator over a Group. Any specific reasons or examples you have in mind? What's your take on this?

One drawback of having both is that it might confuse contributors about which props to use. Eventually, we could end up with a mix of opinions, leading to confusion or inconsistency from the end user's perspective.

In my view, the frontend for activepieces should have a clear opinion on how things are displayed, making it easier to apply new learnings across all pieces. However, if the functionality differs, we should support the prop.

In a group, I see it as more functional, eliminating the need for contributors to spread the object over many properties and assemble them in the code, as in the example you mentioned.

abuaboud:
@Doskyft Great perspective, Let's me try to get other team mates opinions and get back to you

I am being extra careful as these options are not easy to revert back once pieces start using it

PR #3215, opened 2023-11-18
refactor: render engine

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/ui/feature-builder-canvas/src/lib/canvas-utils/drawing/svg-drawer.ts
```diff
@@ -0,0 +1,193 @@
...
+                    break
+                }
+                case 'v': {
+                    currentPositionY += operation.delta
+                    break
+                }
+                case 'arrow': {
+                    break
+                }
+            }
```
AbdulTheActivePiecer, line 61: Please don't forget to rename these (operation.type enum) to something meaningful and add a default case that throws an error which takes a never parameter in it, so we have runtime errors.

AbdulTheActivePiecer:
Closed because it is within #3216

Issue #3164, opened 2023-11-12
In-house Authentication


khaledmashaly:
1. Forget Password / Verify Email
2. Multiple Types of Users (Platform vs Normal) 
3. Login with Google / Github

PR #3222, opened 2023-11-19
feat: reset password and verify email endpoints

khaledmashaly:
## What does this PR do?

in prepare for #3164


packages/backend/src/app/app.ts
```diff
@@ -209,6 +211,8 @@ export const setupApp = async (): Promise<FastifyInstance> => {
             await app.register(signingKeyModule)
             await app.register(managedAuthnModule)
             await app.register(oauthAppModule)
+            await app.register(otpModule)
```
abuaboud, line 214: Let's remove until we migrate firebase? and we can merge this to main

packages/backend/src/app/authentication/authentication-service/hooks/default-authentication-service-hooks.ts
```diff
@@ -24,10 +23,9 @@ export const defaultAuthenticationServiceHooks: AuthenticationServiceHooks = {
             projectType: project.type,
         })
 
-        telemetry.identify(user, project.id)
-            .catch((e) => logger.error(e, '[AuthenticationService#signUp] telemetry.identify'))
+        await telemetry.identify(user, project.id)
```
abuaboud, line 26: I always wanted to be async so it continue if this throws an error

packages/backend/src/app/database/migration/postgres/1700396157624-add-otp-entity.ts
```diff
@@ -0,0 +1,44 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+import { logger } from '../../../helper/logger'
+
+export class AddOtpEntity1700396157624 implements MigrationInterface {
```
abuaboud, line 4: You can avoid this by using JWT signing a token instead of storing it in the database.

packages/backend/src/app/ee/authentication/enterprise-local-authn/enterprise-local-authn-module.ts
```diff
@@ -0,0 +1,6 @@
+import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox'
+import { enterpriseLocalAuthnController } from './enterprise-local-authn-controller'
+
+export const enterpriseLocalAuthnModule: FastifyPluginAsyncTypebox = async (app) => {
+    await app.register(enterpriseLocalAuthnController, { prefix: '/v1/authn/local' })
```
abuaboud, line 8: I think we can omit `local` what do you think?

packages/backend/src/assets/emails/verify-email.html
```diff
@@ -0,0 +1,6 @@
+<html>
+   <body>
+      <h3>Use this code to verify your email!</h3>
+      <h1>{{otp}}</h1>
```
abuaboud, line 4: We can replace this with click on the following URL but this can come later.

abuaboud (changes requested):

packages/backend/src/app/authentication/authentication-service/hooks/default-authentication-service-hooks.ts
khaledmashaly: what about order? does `identify` need to come before `trackProject`?

packages/backend/src/app/ee/authentication/enterprise-local-authn/enterprise-local-authn-module.ts
khaledmashaly: local = username + password
federated = Google or Github (this will still come in the next PR isA)

packages/backend/src/app/authentication/authentication-service/hooks/default-authentication-service-hooks.ts
abuaboud: oh I see identify has to come first

abuaboud (approved):

PR #3281, opened 2023-11-29
feat: support version history

abuaboud:
## What does this PR do?

Fixes # (issue)


abuaboud:
There are to main issues:

- Let's add Icons to View / Move to Draft in the Dropdown
- Let's add a UX way to figure which version is currently being viewed, Let's ask @AshrafSam

PR #3321, opened 2023-12-02
feat(text-helper): markdown to html / html to markdown actions

joeworkman:
## What does this PR do?

This adds 2 new text helpers

* Markdown to HTML
* HTML to Markdown

ANNOUNCEMENT=true


abuaboud:
This is great addition, Let's make sure this is searchable or explorable, let me ask the team too ether we should separate or keep inside text helper

joeworkman:
FYI. The reason I ended up adding it to text helpers and not as it's own piece was that Zapier has markdown processing inside Text Helper. This does not make it right. But just food for thought...

abuaboud (approved):

PR #3326, opened 2023-12-03
feat: Update USER_CREATED flag on sign-up, enable initial sign-up on EE, hide thrid-party authentication from EE, password reset styling improvement

AbdulTheActivePiecer:
## What does this PR do?

1) Turn USER_CREATED flag to true after sign up, regardless of edition.
2) Enable first sign up on ee even if sign ups are disabled.
3) Fixed bug where third party providers sign in wasn't showing for ee after first sign up.
4) Reset password doesn't get stuck on loading if otp expired.
5) EE users get their platform id set after signing up.

### DIFF
--- a/packages/ui/common/src/lib/service/flag.service.ts
+++ b/packages/ui/common/src/lib/service/flag.service.ts
@@ -53,7 +53,7 @@ export class FlagService {
     return this.getAllFlags().pipe(
       map((flags) => {
         const firstUser = flags['USER_CREATED'] as boolean;
-        if (!firstUser && flags['EDITION'] === ApEdition.COMMUNITY) {
+        if (!firstUser && flags['EDITION'] !== ApEdition.CLOUD) {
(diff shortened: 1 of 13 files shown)


packages/ui/common/src/lib/service/flag.service.ts
```diff
@@ -53,7 +53,7 @@ export class FlagService {
     return this.getAllFlags().pipe(
       map((flags) => {
         const firstUser = flags['USER_CREATED'] as boolean;
-        if (!firstUser && flags['EDITION'] === ApEdition.COMMUNITY) {
+        if (!firstUser && flags['EDITION'] !== ApEdition.CLOUD) {
```
abuaboud, line 56: Use ApFlagId

abuaboud (changes requested):

abuaboud (approved):

PR #3329, opened 2023-12-03
feat: add client credentials oauth2

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/ee/product-embed/product-sdk/src/app/enable-integration-modal/enable-integration-modal.component.ts
```diff
@@ -86,6 +87,7 @@ export class EnableIntegrationModalComponent {
           OAuth2Settings.tokenUrl = OAuth2Settings.tokenUrl.replace('login.', 'test.');
         }
       }
+      assertNotEqual(OAuth2Settings.grantType, OAuth2GrantType.CLIENT_CREDENTIALS, 'OAuth2GrantType', 'AUTHORIZATION_CODE');
```
khaledmashaly, line 90: is it ok to throw like this in the frontend?

khaledmashaly (commented):
Great! just one comment to make sure things are ok!

packages/ee/product-embed/product-sdk/src/app/enable-integration-modal/enable-integration-modal.component.ts
abuaboud: This SDK is only used by one customer at this moment, and this case is not required for their use case, adding it here just as reminder for us that it's not supported.

PR #3346, opened 2023-12-05
docs: projects / flows / connections docs

abuaboud:
## What does this PR do?

Fixes # (issue)


khaledmashaly (commented):
Great! Please make sure to recheck values that were changed from `null` to `undefined` or vice versa, especially with regarding to db ops because they carry different meaning for db ops

PR #3371, opened 2023-12-07
Done 1st draft

RaadKloob:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/shared/src/lib/flows/raad-flow-helper.ts
+++ b/packages/shared/src/lib/flows/raad-flow-helper.ts
@@ -1,0 +1,112 @@
+type Step = {
+    name: string;
+    nextAction?: Step,
+};
+
+type StepWithParent = {
+    step: Step;
+    parentStep?: Step;
+}
+
+type SingleStep = Omit<Step, "nextAction">
+
+class StepManager {
+    sourceStep: Step;
+
+    constructor(sourceStep: Step) {
+        this.sourceStep = JSON.parse(JSON.stringify(sourceStep));
+    }
+
+    updateAction(stepName: string, attributes: SingleStep) : Step {
+        let clonedSourceStep = JSON.parse(JSON.stringify(this.sourceStep));
+
+        const stepWithParent = this.findStepByName(clonedSourceStep, stepName);
+
+        if(!stepWithParent)
+            throw new Error('Step can\'t be found!');
+
+        const updatedStep = { ...attributes, nextAction: stepWithParent.step.nextAction } as Step;
+
+        if(stepWithParent.parentStep)
+            stepWithParent.parentStep.nextAction = updatedStep;
+        else
+            clonedSourceStep = updatedStep;
+
+        return clonedSourceStep;
+    }
+    
+    duplicateAction(sourceStep: Step, stepName: string) : Step {
+        let clonedSourceStep = JSON.parse(JSON.stringify(sourceStep));
+    
+        const stepWithParent = this.findStepByName(clonedSourceStep, stepName);
+    
+        if(!stepWithParent)
+            throw new Error('Step can\'t be found!');
+    
+        const stepToDuplicate = stepWithParent.step;
+        
+        // Here we can create a clone of the step and change it's name, in case if the name was a unique identifier
+    
+        clonedSourceStep = this.addAction(clonedSourceStep, stepToDuplicate, stepName);
+    
+        return clonedSourceStep;
+    }
+    
+    moveAction(sourceStep: Step, stepName: string, targetStepName: string) : Step {
+        let clonedSourceStep = JSON.parse(JSON.stringify(sourceStep));
+    
+        if(stepName == targetStepName)
+            return clonedSourceStep;
+    
+        const stepWithParent = this.findStepByName(clonedSourceStep, stepName);
+    
+        if(!stepWithParent)
+            throw new Error('Step can\'t be found!');
+    
+        const stepToBeAdded = stepWithParent.step;
+    
+        clonedSourceStep = this.deleteAction(clonedSourceStep, stepToBeAdded.name);
+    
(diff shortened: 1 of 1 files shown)


packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+            throw new Error('Step can\'t be found!');
+
+        const updatedStep = { ...attributes, nextAction: stepWithParent.step.nextAction } as Step;
+
+        if(stepWithParent.parentStep)
+            stepWithParent.parentStep.nextAction = updatedStep;
+        else
+            clonedSourceStep = updatedStep;
+
+        return clonedSourceStep;
```
abuaboud, line 35: I think it's better to return `StepManager` so you can continue building and in the end you can make function to convert toStep()

packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+    
+        if(stepWithParent.parentStep)
+            stepWithParent.parentStep.nextAction = stepWithParent.step.nextAction;
+        else
+            return stepWithParent.step.nextAction as Step;
+    
+        return clonedSourceStep;
+    }
+    
+    addAction(sourceStep: Step, step: SingleStep, name: string) : Step {
```
abuaboud, line 91: `sourceStep` shouldn't exists here, It's already in the class.

packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+        const stepToBeAdded = stepWithParent.step;
+    
+        clonedSourceStep = this.deleteAction(clonedSourceStep, stepToBeAdded.name);
+    
+        clonedSourceStep = this.addAction(clonedSourceStep, stepToBeAdded, targetStepName);
+    
+        return clonedSourceStep;
+    }
+    
+    deleteAction(sourceStep: Step, stepName: string) : Step {
```
abuaboud, line 75: `sourceStep` shouldn't exists here, read the comment below

packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+        const stepToDuplicate = stepWithParent.step;
+        
+        // Here we can create a clone of the step and change it's name, in case if the name was a unique identifier
+    
+        clonedSourceStep = this.addAction(clonedSourceStep, stepToDuplicate, stepName);
+    
+        return clonedSourceStep;
+    }
+    
+    moveAction(sourceStep: Step, stepName: string, targetStepName: string) : Step {
```
abuaboud, line 55: `sourceStep` shouldn't exists here, read the comment below

packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        if(stepWithParent.parentStep)
+            stepWithParent.parentStep.nextAction = updatedStep;
+        else
+            clonedSourceStep = updatedStep;
+
+        return clonedSourceStep;
+    }
+    
+    duplicateAction(sourceStep: Step, stepName: string) : Step {
```
abuaboud, line 38: `sourceStep` shouldn't exists here, read the comment below

packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+    
+    findStepByName(step: Step, name: string, parentStep?: Step) : StepWithParent | null {
+        if(step.name == name)
+            return { step, parentStep };
+        else if(step.nextAction)
+            return this.findStepByName(step.nextAction, name, step);
+        else 
+            return null;
+    }
```
abuaboud, line 111: NIT: this just preference 


I would write in that way to avoid recursive:

```ts
findStepByName(name: string) : StepWithParent | null {
       const currentStep = this.sourceStep;
      const parentStep = null;
       while(currentStep && currentStep.name !== name){
           parentStep = currentStep;
           currentStep = currentStep.nextAction;
       }
       return {parentStep, currentStep}
    }
    ```

Comment (abuaboud) [reply packages/shared/src/lib/flows/raad-flow-helper.ts:111]: You should remove step from this as it's always `sourceStep`

Comment (abuaboud) [reply packages/shared/src/lib/flows/raad-flow-helper.ts:111]: another way you can name this function `findStepByNameOrThrow` and make it throw an error when it's null so it doesn't have to be handled by each function inside the class.

File: packages/shared/src/lib/flows/raad-flow-helper.ts [hunk 7]
```diff
@@ -0,0 +1,112 @@
...
+
+type StepWithParent = {
+    step: Step;
+    parentStep?: Step;
+}
+
+type SingleStep = Omit<Step, "nextAction">
+
+class StepManager {
+    sourceStep: Step;
```
abuaboud, line 14: I would make this private and readonly

packages/shared/src/lib/flows/raad-flow-helper.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    updateAction(stepName: string, attributes: SingleStep) : Step {
+        let clonedSourceStep = JSON.parse(JSON.stringify(this.sourceStep));
+
+        const stepWithParent = this.findStepByName(clonedSourceStep, stepName);
+
+        if(!stepWithParent)
+            throw new Error('Step can\'t be found!');
+
+        const updatedStep = { ...attributes, nextAction: stepWithParent.step.nextAction } as Step;
```
abuaboud, line 28: `as` usually is sign of bug, why it's needed here?

PR #3396, opened 2023-12-10
feat: viewable link for files when testing steps

### LINKED ISSUE #2643: View Files in Testing
The user doesn't make sense of db://file.png, we should provide the user an clickable protected url so they can view the file on the browser or download it.

MoShizzle:
## What does this PR do?

Fixes #2643


packages/engine/src/lib/services/files.service.ts
```diff
@@ -96,7 +97,16 @@ async function writeDbFile({ stepName, flowId, fileName, data }: { stepName: str
         throw new Error('Failed to store entry ' + response.body)
     }
     const result = await response.json()
-    return DB_PREFIX_URL + `${result.id}`
+
+    const viewToken = await (await fetch(STEP_FILE_URL + result.id + '/generate-view-token', {
```
abuaboud: Can you make this endpoint returns the full url in the same files endpoint? instead of new url for generate token?

abuaboud (approved):

abuaboud (approved):

PR #3397, opened 2023-12-10
feat(Mastodon): add media file property to post status action

denieler:
## What does this PR do?

Adds feature of media upload to the Mastodon piece, Post Status action. Before this piece could post only text messages without an attachment.

| Before | After |
|------|------|
| Screenshot 2023-12-10 at 15 52 07 |  Screenshot 2023-12-10 at 19 17 39 |
| | |

Result:

Screenshot 2023-12-10 at 15 54 05

Post Status action keeps working as is when the Media URL is not specified.
I had to change slightly http-client method and add the `responseType` parameter, so media could be streamed to Mastodon, instead of uploading to memory.

ANNOUNCEMENT=TRUE

### DIFF
--- a/packages/pieces/common/package.json
+++ b/packages/pieces/common/package.json
@@ -1,5 +1,5 @@
 {
   "name": "@activepieces/pieces-common",
-  "version": "0.2.5",
+  "version": "0.2.6",
   "type": "commonjs"
 }
--- a/packages/pieces/common/package.json
+++ b/packages/pieces/common/package.json
@@ -1,5 +1,5 @@
 {
   "name": "@activepieces/pieces-common",
-  "version": "0.2.6",
+  "version": "0.2.5",
   "type": "commonjs"
 }
--- a/packages/pieces/common/src/lib/http/axios/axios-http-client.ts
+++ b/packages/pieces/common/src/lib/http/axios/axios-http-client.ts
@@ -35,8 +35,7 @@ export class AxiosHttpClient extends BaseHttpClient {
 				},
 				headers,
 				data: request.body,
-				timeout,
-				responseType: request.responseType,
+				timeout
 			};
 
 			const response = await axios.request(config)
--- a/packages/pieces/common/src/lib/http/axios/axios-http-client.ts
+++ b/packages/pieces/common/src/lib/http/axios/axios-http-client.ts
@@ -35,7 +35,8 @@ export class AxiosHttpClient extends BaseHttpClient {
 				},
 				headers,
 				data: request.body,
-				timeout
+				timeout,
+				responseType: request.responseType,
 			};
 
 			const response = await axios.request(config)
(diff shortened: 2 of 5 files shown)


denieler:
@abuaboud can you please check this out? 🙇

abuaboud:
@allcontributors please add @denieler for plugin

abuaboud:
Hi @denieler 

Can you please change it to Property.File, It accepts multiple format (urls / actiivepieces internal format / base64) and gives you `ApFile` where you can choose the format you want to convert it.

https://www.activepieces.com/docs/developers/piece-reference/properties#file

abuaboud:
Example: https://github.com/activepieces/activepieces/blob/768b573d3390f9088df28a44005e8cd7fd7c909d/packages/pieces/openai/src/lib/actions/translation.ts#L13

denieler:
@abuaboud yeah 👍 sounds great, didn't know it existed

denieler:
Sorry, @abuaboud, maybe I'm using `Property.File` in the wrong way, but it seems that it doesn't accept URL as a value. It can only be used as a file upload 😞 Do you think, that to be able to upload from a URL an attachment to Mastodon status update, I need to use 2 pieces: one for uploading the image from the URL and another to post it as uploaded file?

abuaboud:
It should accept the url, If that not the case please update the pr and you can leave it to me hehe

denieler:
@abuaboud yeah, I confirm, it seems if I enter the string URL into this field, the `context.propsValue.media` value is `null` 😞 

Sorry, I didn't fully understand what should I do in this case? I assume that if the value of this prop is URL it's a string and it should keep working in the same way it does now, but in case it's a file I can skip this first `fetch` and just use incoming file Buffer as a form `file` field value directly.

abuaboud:
@denieler Sorry, I mean please upload what you have done so far, I will check it out in my device and check.

I just notice the documentation need to be updated, I don't think you need to do the fetch as activepieces already download it if it link.

denieler:
@abuaboud thank you 🙇 I've pushed my latest changes, but due to the `null` value of the `media` prop I can't test it, sorry 🙇‍♂️

abuaboud:
I've identified the issue. The direct link `https://fastly.picsum.photos/id/237/1024/1024.jpg?hmac=b2xPrkhbcN43JTdFEF0mhAJPrMp6w_hJCTOIDAby14A` works fine.

However, the URL `https://picsum.photos/id/237/1024` doesn't have the HEAD protocol allowed, causing the processor to fail when attempting validation, wether the url is image or not.

You can check the code related to this issue at: [https://github.com/activepieces/activepieces/blob/0f50f4b79902cbc796950541cb2bcdae34e1e17a/packages/pieces/framework/src/lib/processors/processors.ts#L65](https://github.com/activepieces/activepieces/blob/0f50f4b79902cbc796950541cb2bcdae34e1e17a/packages/pieces/framework/src/lib/processors/processors.ts#L65)

Is this currently a blocker for your usecase?

denieler:
@abuaboud thanks for finding this 👍 yes, you are right, I've checked with direct url (without redirect) and it works great 👍 no, it's not a blocker for my case and probably an edge case

PR #3404, opened 2023-12-11
feat: delay to next weekday/day of year

joeworkman:
## What does this PR do?

This adds 2 new actions to the Delay piece:

* Delay the workflow until the next day or the week at a specific time. 
* Delay the workflow until the next day or the year at a specific time. 

I did centralize a `delayUntil` function that was used inside of the `delay-until-used` action.


abuaboud:
Hi @joeworkman 

Thank you for the PR, I believe this should be achieved through data helper and use delay until.

joeworkman:
My original thought was to use date helper and delay until. But then I thought... why require the use of 2 tasks when we can do it all in one? I think many people will not think about using the date helper action with delay until. This makes sense to me.

abuaboud:
@joeworkman I think in-line functions feature will solve these issues will keep the pr open until we finish scooping inline functions and check if it solves the problem or not

joeworkman:
@abuaboud Curious on the status on this. Have inline functions been implemented? Where can I learn more about what that is?

PR #3423, opened 2023-12-13
feat: add app webhooks for self hosted

abuaboud:
## What does this PR do?

Fixes # (issue)


abuaboud:
I would love if there is a contributor who can help me document the setup for slack in self host environment, then we can merge that :D

PR #3456, opened 2023-12-18
feat(discourse): add users to group / change trust level / create post  / create topic / send private message actions

pfernandez98:
## What does this PR do?

This adds a new piece: Discourse.
For now, you can use the following actions:
* Add user(s) to groups
* Change trust level of user
* Send a PM to a user/list of users
* Send new Post
* Send new Topic

I did NOT add any triggers, because Discourse have a webhook feature, so you can use the AP webhooks to trigger ANY action.

ANNOUNCEMENT=true


packages/pieces/discourse/src/lib/actions/new-pm-to-user.action.ts
```diff
@@ -0,0 +1,48 @@
...
+import { discourseAuth } from '../../index';
+import { Property, createAction } from "@activepieces/pieces-framework";
+
+
+
+export const newPrivateMessage = createAction({
+    auth: discourseAuth,
+    name: 'new_private_message',
+    description: 'Create a new private message in Discourse',
+    displayName: 'New Private Message',
```
abuaboud: Can you please change to "Send Private Message"

packages/pieces/discourse/src/lib/actions/new-post.action.ts
```diff
@@ -0,0 +1,66 @@
...
+import { discourseAuth } from '../../index';
+import { Property, createAction } from "@activepieces/pieces-framework";
+
+
+
+export const newPost = createAction({
+    auth: discourseAuth,
+    name: 'new_post',
+    description: 'Create a new post in discourse',
+    displayName: 'New Post',
```
abuaboud: Can you please change to "Create Post", It's better to use verb with the action

packages/pieces/discourse/src/lib/actions/new-topic.action.ts
```diff
@@ -0,0 +1,72 @@
...
+import { discourseAuth } from '../../index';
+import { Property, createAction } from "@activepieces/pieces-framework";
+
+
+
+export const newTopic = createAction({
+    auth: discourseAuth,
+    name: 'new_topic',
+    description: 'Create a new topic in Discourse',
+    displayName: 'New Topic',
```
abuaboud: Can you please change to "Create Topic", It's better to use verb with the action

packages/pieces/discourse/src/lib/actions/new-post.action.ts
abuaboud: I would also do it for the files names and other actions as well :D

abuaboud (commented):
Great Job! :D 

Left some minor comments let's do them and merge!

packages/pieces/discourse/src/lib/actions/new-pm-to-user.action.ts
pfernandez98: done

packages/pieces/discourse/src/lib/actions/new-post.action.ts
pfernandez98: done

packages/pieces/discourse/src/lib/actions/new-topic.action.ts
pfernandez98: done

abuaboud (approved):

PR #3485, opened 2023-12-24
feat(telegram-bot): send media action and `reply_markup` field in messages

alerdenisov:
## Updates telegram-bot integration

* Implements new action **`Send Media`** allows send images, videos, animations and stickers.
* Adds new property `reply_markup` to send inlinekeyboard or bot keyboard. 

image


alerdenisov:
@abuaboud I tried to implement the same action as Discord Send Approval, but for telegram and tried to use run pausing and webhook. It requires reading of post body, what do you think about update of `flow-run-controller` to add body, headers to resume payload?

PR #3520, opened 2023-12-30
test: simple flow test for worker

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/backend/project.json
```diff
@@ -81,6 +81,7 @@
       "executor": "nx:run-commands",
       "options": {
         "commands": [
+          "nx build backend",
```
khaledmashaly, line 84: Can you please explain why this command is needed?

abuaboud: To build the engine and bundle it

PR #3530, opened 2024-01-01
fix(ntfy): support unicode api header

### LINKED ISSUE #3529: [BUG]: ntfy failed to send notification due to invalid HTTP header in `X-Message` / `X-Title`
**Describe the bug**
Process failed when the `message` or `title` of ntfy contains non-ASCII characters.

**To Reproduce**
1. Create a new flow
2. Find ntfy and add it to the flow (and set up a connection, refer to https://docs.ntfy.sh/. You can try with url 'https://ntfy.sh'  and use this topic with name `mytopic`. AccessToken is not required ).
3. Input some characters like an emoji or any other CJK character in the `message` box
4. Test the flow
5. Output: `Invalid character in header content ["X-Message"]`

**Expected behavior**
Successfully sending.

**Screenshots**
Screenshot 2024-01-01 at 10 29 23

**Additional context**
[ntfy docs](https://docs.ntfy.sh/publish/#message-title)
> ntfy supports UTF-8 in HTTP headers, but [not every library or programming language does](https://www.jmix.io/blog/utf-8-in-http-headers/). If non-ASCII characters are causing issues for you in the title (i.e. you're seeing ? symbols), you may also encode any header (including the title) as [RFC 2047](https://datatracker.ietf.org/doc/html/rfc2047#section-2), e.g. =?UTF-8?B?8J+HqfCfh6o=?= ([base64](https://en.wikipedia.org/wiki/Base64)), or =?UTF-8?Q?=C3=84pfel?= ([quoted-printable](https://en.wikipedia.org/wiki/Quoted-printable)).

la3rence:
## What does this PR do?

Fixes #3529

ANNOUNCEMENT=TRUE

### DIFF
--- a/packages/pieces/ntfy/package.json
+++ b/packages/pieces/ntfy/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-ntfy",
-  "version": "0.1.2"
+  "version": "0.1.3"
 }
--- a/packages/pieces/ntfy/src/lib/actions/send-notification.ts
+++ b/packages/pieces/ntfy/src/lib/actions/send-notification.ts
@@ -2,6 +2,10 @@ import { createAction, Property } from "@activepieces/pieces-framework";
 import { AuthenticationType, httpClient, HttpMethod } from "@activepieces/pieces-common";
 import { ntfyAuth } from "../..";
 
+const encodeToRFC2047 = (text: string) => {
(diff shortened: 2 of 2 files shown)


abuaboud:
@allcontributors please add @la3rence for plugin

abuaboud:
please bump patch version in package.json

abuaboud (approved):

PR #3569, opened 2024-01-08
feat(google-drive): export as pdf action

pfernandez98:
## What does this PR do?

This adds a new action for the Google Drive piece for converting a file to PDF, and getting the new ID

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/google-drive/src/lib/action/export-document-as-pdf.action.ts
+++ b/packages/pieces/google-drive/src/lib/action/export-document-as-pdf.action.ts
@@ -1,0 +1,71 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+
+
+import { googleDriveAuth } from '../../index';
+import { Property, createAction } from "@activepieces/pieces-framework";
+import { google } from 'googleapis';
+import { OAuth2Client } from 'googleapis-common';
+import { Stream } from 'stream';
+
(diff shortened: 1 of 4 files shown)


packages/pieces/google-drive/src/lib/action/export-document-as-pdf.action.ts
```diff
@@ -0,0 +1,71 @@
...
+        const templateBuffer = Buffer.from(result.data as any, 'base64');
+
+        const streamm = new Stream.PassThrough().end(templateBuffer);
+
+        const media = {
+            mimeType: 'application/pdf',
+            body: streamm,
+        };
+
+        const file = await drive.files.create({
```
abuaboud: The display name of the action is misleading, I think this one export as pdf and it creates new file inside the google drive right?

Also do you think this action should be in Google Docs rather drive?

pfernandez98: yeah, i fixed the name of the action... But actually, this is a Google drive thing. It uses the google drive API. (I tried to include it on the Docs Piece, but I need the Google drive auth)

https://developers.google.com/drive/api/reference/rest/v3/files/export

abuaboud (approved):

abuaboud (approved):

PR #3570, opened 2024-01-08
fix(schedule): add static dropdown for minutes selection to ensure valid values

### LINKED ISSUE #2983: Issue with "Every X Minutes" Schedule Trigger
The current implementation of the "Every X Minutes" trigger in the schedule feature has a limitation. It can only handle values between 1 and 59. If you input a number outside this range, the schedule won't function correctly, and there's currently no validation to catch this.

Here's the part of the code causing the problem:
```javascript
*/${ctx.propsValue.minutes} * * * *
```

If the value for minutes is not between 1 and 59, it leads to failure.

**Suggested Solution:**
One way to address this issue is to eliminate the short text option and replace it with a static dropdown menu. This dropdown should include only valid values from 1 to 59, preventing users from entering unsupported values.

**Reference:**
[Link to the specific code file on GitHub](https://github.com/activepieces/activepieces/blob/ec67d81bdbdc20ee93822460b59fbe79cd6f73e5/packages/pieces/schedule/src/lib/triggers/every-x-minutes.trigger.ts)

kishanprmr:
## What does this PR do?

Fixes #2983

### DIFF
--- a/packages/pieces/schedule/src/lib/triggers/every-x-minutes.trigger.ts
+++ b/packages/pieces/schedule/src/lib/triggers/every-x-minutes.trigger.ts
@@ -1,35 +1,47 @@
-import { Property, TriggerStrategy, Validators, createTrigger } from "@activepieces/pieces-framework";
+import {
+  Property,
+  TriggerStrategy,
+  createTrigger,
+} from '@activepieces/pieces-framework';
 
 export const everyXMinutesTrigger = createTrigger({
-    name: 'every_x_minutes',
-    displayName: 'Every X Minutes',
-    description: 'Triggers the current flow every X minutes',
-    type: TriggerStrategy.POLLING,
-    sampleData: {},
-    props: {
-        minutes: Property.Number({
-            displayName: "Minutes",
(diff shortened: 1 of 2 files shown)


packages/pieces/schedule/src/lib/triggers/every-x-minutes.trigger.ts
```diff
@@ -1,35 +1,47 @@
...
+  props: {
+    minutes: Property.StaticDropdown({
+      displayName: 'Minutes',
+      description: 'Valid value between 1 to 59.',
+      required: true,
+      defaultValue: 1,
+      options: {
+        disabled: false,
+        options: Array.from({ length: 59 }, (_, index) => ({
+          label: `${index + 1} minutes`,
```
abuaboud: Nit: this will produce '1 minutes' which is not correct

kishanprmr: fixed it.

abuaboud (approved):

PR #3579, opened 2024-01-09
chore: parse string as JSON object

dennisrongo:
## What does this PR do?

The field "meta" is not being sent properly and needs to be converted to a JSON object. Currently, it's being parsed as string.

Fixes # (issue)


abuaboud (dismissed):

abuaboud:
Hi @dennisrongo, I replaced it with processor on json property this should fix it everywhere else too.

abuaboud:
@allcontributors please add @dennisrongo for bug

abuaboud (approved):

PR #3633, opened 2024-01-15
chore: fixed wrong date in changelog

zaaakher:
The latest update in the changelog has 4/1/2023 so I changed it to 4/1/2024


abuaboud:
@allcontributors Please add @zaaakher for doc

PR #3641, opened 2024-01-15
feat: add .nvmrc file

zaaakher:
## What does this PR do?

This PR adds a `.nvmrc` file which makes it easier for nvm to switch to the compatible node version (see nvm [docs](https://github.com/nvm-sh/nvm?tab=readme-ov-file#nvmrc))

---

Use case:

If someone runs `node setup-dev.js` and the scripts exits due to incompatible node version, then the developer can run `nvm use` and it'll automatically detect `.nvmrc` and switch to the version indicated in it.


abuaboud:
please add @zaaakher for code

abuaboud:
@allcontributors please add @zaaakher for code

PR #3644, opened 2024-01-15
feat: prioritize webhook sync calls

abuaboud:
## What does this PR do?

Fixes # (issue)


khaledmashaly (commented):
Nice Job! I would suggest using numbers like 1000 and 5000 for priorities so that we can add higher or lower priorities without having to migrate old jobs

PR #3664, opened 2024-01-16
feat: Adds a File Writer to filesHelper

kanarelo:
## What does this PR do?

Adds a files writer to filesHelper

Fixes # (issue)


abuaboud:
Hi @kanarelo 

What is the use case for this action I am curious?

abuaboud:
@kanarelo as discussed offline, files can be used directly inside other pieces property the same way it's written so writing won't be required.

Closing this issue, please feel free to reopen if there is a use case where this is required.

PR #3667, opened 2024-01-16
feat(engine): use V8's Isolate to sandbox code piece to work with unprivileged docker

### LINKED ISSUE #3652: refactor: sandbox code piece using V8 isolate

khaledmashaly:
## What does this PR do?

Fixes #3652


packages/backend/src/app/helper/system/system.ts
```diff
@@ -104,28 +103,8 @@ const getEnvVar = (prop: SystemProp): string | undefined => {
...
 
 export const validateEnvPropsOnStartup = async (): Promise<void> => {
-    const executionMode = system.get(SystemProp.EXECUTION_MODE)
-    const signedUpEnabled =
-        system.getBoolean(SystemProp.SIGN_UP_ENABLED) ?? false
     const queueMode = system.getOrThrow<QueueMode>(SystemProp.QUEUE_MODE)
-    const environment = system.get(SystemProp.ENVIRONMENT)
     await loadEncryptionKey(queueMode)
-
-    if (
```
abuaboud, line 114: why not check for code execution type + sign up enabled?

packages/backend/src/app/workers/sandbox/isolate-sandbox.ts
```diff
@@ -61,6 +61,7 @@ export class IsolateSandbox extends AbstractSandbox {
                 '--env=HOME=/tmp/',
                 '--env=NODE_OPTIONS=\'--enable-source-maps\'',
                 `--env=AP_PIECES_SOURCE=${pieceSource}`,
+                '--env=AP_CODE_EXECUTOR_SANDBOX_TYPE',
```
abuaboud: Nit: pass the raw value in case we have default for that

abuaboud (changes requested):

packages/backend/src/app/workers/sandbox/isolate-sandbox.ts
khaledmashaly: fixed in 7845ba1

packages/backend/src/app/helper/system/system.ts
khaledmashaly: I think it's best to leave it up to the user's discern. This will be configurable using languages anyway so it doesn't make sense to disable signups if javascript is disabled, right?

abuaboud (approved):

PR #3689, opened 2024-01-18
feat(cli): create piece command 

kishanprmr:
## What does this PR do?

Fixes # (issue)


packages/cli/src/lib/utils/get-available-piece-names.ts
```diff
@@ -0,0 +1,9 @@
+import { readdir } from 'node:fs/promises';
+
+export const getAvailablePieceNames = async (
```
abuaboud: Thiis is outdated, since all pieces now lives in community or custom

Please check the `findAllPiecesDirectory` in tools/script at HEAD

packages/cli/src/index.ts
```diff
@@ -0,0 +1,20 @@
+import { Command } from 'commander';
+import { createPieceCommand } from './lib/commands/create-piece';
+
+const program = new Command();
+
+program.version('0.0.1').description('Activepieces CLI');
```
abuaboud, line 41: Please create alias for the cli in package.json

packages/cli/src/index.ts
```diff
@@ -0,0 +1,20 @@
+import { Command } from 'commander';
```
abuaboud, line 1: commander is not installed in dev dependencies

abuaboud: Please run npm install commander and commit the new package.json

packages/cli/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const program = new Command();
+
+program.version('0.0.1').description('Activepieces CLI');
+
+program
+  .command('create-piece')
+  .argument('<pieceName>', 'piece name')
+  .argument('<packageName>', 'package name')
+  .requiredOption(
```
abuaboud: the type can be interactive cli that allow the user to choose between custom and community

abuaboud: I believe the packageName isn't necessary since it can be deduced from the name. For instance, if the user inputs a piece name like "hello," the default package name would be @activepieces/piece-hello unless the user specifies otherwise.

packages/cli/src/index.ts
```diff
@@ -0,0 +1,20 @@
+import { Command } from 'commander';
+import { createPieceCommand } from './lib/commands/create-piece';
+
+const program = new Command();
+
+program.version('0.0.1').description('Activepieces CLI');
+
+program
+  .command('create-piece')
```
abuaboud: Please check https://www.npmjs.com/package/inquirer

I think it can helps with commanders to make the cli  interactive

packages/cli/src/index.ts
abuaboud: and please check what's needed to be updated in the documentation as well delete the current code for create piece under tools

abuaboud (changes requested):

abuaboud (dismissed):

abuaboud (approved):

PR #3691, opened 2024-01-18
feat: add project switcher /add navigation from and to Admin console

AbdulTheActivePiecer:


abuaboud:
Let's replace "IsDemo" with "Flag." I pushed that change in this pull request: [link to the pull request](https://github.com/activepieces/activepieces/pull/3691/commits/17afa1cd67639f6cd9a7ded5e2367dc29fd0ee71).

The reason why using `isDemo` won't work is that cloud users are inside our platform, so the cloud platform entity is shared between all users, and they will have a single value.

Regarding the community edition, it would be ideal to work just based on that flag without the need for a platform entity. This change would unblock the process from the huge migration.

My suggestion is, as the first step, to replace all instances of "IsDemo" with the flag mentioned above. As a second action item, which can be addressed in a follow-up pull request, we can make it work without the need to call any platform entity so that it functions in the community edition.

abuaboud (approved):

PR #3693, opened 2024-01-18
feat: arbitrary http request action

### LINKED ISSUE #1977: Send custom requests in pieces
To give users a big relief from being limited in the actions just defined in a piece, and according to a user I just chatted with, it will be great to be able to send custom requests in any piece.

This will be the same as the HTTP piece but with predefined fields like Base URL and the ability to choose a connection.

This default action can also have a direct link to the 3rd party's docs so users can get the endpoints they'd like to use.

MoShizzle:
## What does this PR do?

Adds an arbitrary http request action that can be used in all pieces. Added to OpenAI piece as example.

Fixes #1977


packages/pieces/common/src/lib/helpers/index.ts
```diff
@@ -10,3 +12,92 @@ export const getAccessTokenOrThrow = (auth: OAuth2PropertyValue | undefined): st
   return accessToken;
 };
 
+export function createArbitraryHttpRequestAction({ auth, baseUrl, authMapping }: {
```
abuaboud: createCustomApiCallAction

Would be more friendly

PR #3706, opened 2024-01-20
fix(text-helper): validate string input

kishanprmr:
## What does this PR do?

Fixes # (issue)


packages/pieces/community/text-helper/src/lib/actions/replace.ts
```diff
@@ -9,6 +9,7 @@ export const replace = createAction({
     text: Property.ShortText({
       displayName: 'Text',
       required: true,
+      validators: [Validators.string]
```
abuaboud: It's better to add to defaultValidator directly inside the piece framework, another way also to add defaultProcessor where it convert it to a string.

abuaboud:
@kishanprmr Great change! Do you think this will affect the relatable piece? Is it possible to edit?

abuaboud (approved):

abuaboud (approved):

PR #3716, opened 2024-01-22
fix(gmail): body type dropdown/ newline in place of <br>

### LINKED ISSUE #3711: [BUG]: the gmail action "Send email" send raw text HTML
**Describe the bug**
When using the Gmail Send email action with a body that has multilines, it convert all the line breaks into \<br>.
But the br are not intepreted as HTML, so the email is really ugly, see screenshot

**To Reproduce**
1. create a new flow with whatever trigger
2. use the action Gmail : Send email
3. in the body, write text on multiple lines
4. test the action and check your emails
5. you should see \<br> for every line break

**Expected behavior**
HTML is interpreted, we should not see html and just text with new lines

**Screenshots**
Body of the written email in ActivePieces : 
Capture d’écran 2024-01-21 à 12 43 33
Body received by email : 
Capture d’écran 2024-01-21 à 12 43 39

kishanprmr:
## What does this PR do?

Fixes #3711

ANNOUNCEMENT=true


yann120:
Thanks @kishanprmr for fixing this! 🙏🏻

PR #3725, opened 2024-01-23
feat: realtime flow run feedback via websockets

### LINKED ISSUE #3107: Real-Time Flow-Run Feedback via WebSockets
As part of enhancing the developer experience, there is a growing need for real-time feedback on flow-runs. The current asynchronous feedback mechanisms do not provide the immediate, interactive response that a real-time system could offer. To address this, I propose the implementation of an opt-in feature that utilizes WebSockets to deliver live flow-run updates.

Expected Benefits:

Immediate feedback on flow execution status.
Enhanced debugging with real-time logs.
Improved user experience for workflows requiring rapid iteration.
Possible Approach:
A WebSocket-based communication channel could be established for a flow-run, which clients can opt into when initiating a flow-run. This would enable a persistent, low-latency connection allowing clients to receive updates as they happen.

Looking forward to any information regarding existing plans for such a feature or discussions on the feasibility and potential integration into our platform.

MoShizzle:
## What does this PR do?

Fixes #3107


abuaboud:
Thank you for the great PR, @MoShizzle. For some reason, it wasn't stable in my localhost, sometime the test flow would emit back the even and sometimes not.

I strongly suggest we should switch to https://www.npmjs.com/package/@fastify/websocket. It's a maintained and official package, It also depends on `ws`, which has 60 million downloads and is the most popular library (https://www.npmjs.com/package/ws).

Now, for the frontend, ws exist natively in RxJS, for example, https://javascript-conference.com/blog/real-time-in-angular-a-journey-into-websocket-and-rxjs/.

So, we can get rid of Socket.IO and replace it with the above libraries. What do you think?

MoShizzle:
Hi @abuaboud,

Socket.IO is better implementation for us, for a few reasons:
1- Built-in fallback mechanisms in case of connection fail, on WS you'll have to manually handle reconnecting if connection fails for some reason.
2- An option to use polling instead of WS protocol if network policy doesn't allow it
3- Middleware
4- Rooms, which allow you to broadcast to a group of users, this will be mandatory if we implement websockets somewhere that has multiple users active on it (e.g real-time flow changes in builder and collaboration features)

Can you share more information about what happens when the socket fails? Are you sure the frontend is connected? Are you running on docker or locally?

PR #3760, opened 2024-01-28
fix(storage-piece): "Key <key> is not an array" error when using add-to-list action

### LINKED ISSUE #3759: [BUG]: Storage piece fails when trying to use the "Add to list" action.
**Describe the bug**
Inserting an array into the storage and then trying to add to it using the storage piece doesn't work. Get "Key array-data is not an array" error.

**To Reproduce**

storage-bug

wise-introvert:
## What does this PR do?
Array data, in the storage piece, is stored as a string ( despite the JSONB column data type ). This makes it impossible for the user to use the add-to-list action, which assumes that the data retrieved from the storage is an array ( if not empty ) and tries to push items to it.

Fixes #3759

### DIFF
--- a/packages/pieces/community/store/src/lib/actions/store-add-to-list.ts
+++ b/packages/pieces/community/store/src/lib/actions/store-add-to-list.ts
@@ -42,12 +42,19 @@ export const storageAddtoList = createAction({
     }),
   },
   async run(context) {
-    const items =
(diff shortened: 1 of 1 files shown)


abuaboud:
Please check my comment in the issue.

abuaboud (approved):

PR #3763, opened 2024-01-28
feat: authn rate limit

### LINKED ISSUE #3761: Rate-limit authn requests

khaledmashaly:
## What does this PR do?

Fixes #3761


packages/backend/src/app/core/security/rate-limit.ts
```diff
@@ -0,0 +1,31 @@
+import RateLimitPlugin from '@fastify/rate-limit'
+import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox'
+import FastifyPlugin from 'fastify-plugin'
+import { Redis } from 'ioredis'
+import { createRedisClient } from '../../database/redis-connection'
+import { QueueMode, system } from '../../helper/system/system'
+import { SystemProp } from '../../helper/system/system-prop'
+
+export const rateLimitModule: FastifyPluginAsyncTypebox = FastifyPlugin(async (app) => {
+    await app.register(RateLimitPlugin, {
```
abuaboud: I am a bit scared, especially about webhook endpoints. For example, Google servers sending us webhooks, or the engine querying for pieces information / store / connections.

On the other hand, we make a lot of requests from the frontend buit this is least worrying.

packages/backend/src/app/authentication/authentication.controller.ts
```diff
@@ -39,6 +46,7 @@ const SignUpRequestOptions = {
 const SignInRequestOptions = {
     config: {
         allowedPrincipals: ALL_PRINICPAL_TYPES,
+        rateLimit: rateLimitOptions,
```
abuaboud, line 51: Is it a bit extreme to have 2 attempts within 1 minute?

Also, during sign-up, if I input a used, it prompts for another one, or make a wrong attempt login.

I would suggest sign up something low, but sign in I would be more forgiving like 15 per 5 minutes or something like that

packages/backend/src/app/app.ts
```diff
@@ -153,6 +154,8 @@ export const setupApp = async (): Promise<FastifyInstance> => {
 
     await app.register(formBody, { parser: str => qs.parse(str) })
 
+    await app.register(rateLimitModule)
```
abuaboud, line 161: wouldn't this work in community mode?

packages/backend/src/app/core/security/rate-limit.ts
khaledmashaly: disabled globally, enabled for authn only for the time being

packages/backend/src/app/authentication/authentication.controller.ts
khaledmashaly: updated to 50 requests per minute

packages/backend/src/app/app.ts
khaledmashaly: updated to be enforced on cloud only

packages/backend/src/app/authentication/authentication.controller.ts
```diff
@@ -27,9 +28,15 @@ export const authenticationController: FastifyPluginAsyncTypebox = async (app) =
...
+    timeWindow: '1 minute',
+}
+
 const SignUpRequestOptions = {
     config: {
         allowedPrincipals: ALL_PRINICPAL_TYPES,
+        rateLimit: rateLimitOptions,
     },
     schema: {
         body: SignUpRequest,
```
abuaboud, line 44: Can you make this more strict? the rate limit options

packages/backend/src/app/authentication/authentication.controller.ts
```diff
@@ -27,9 +28,15 @@ export const authenticationController: FastifyPluginAsyncTypebox = async (app) =
     })
 }
 
+const rateLimitOptions: RateLimitOptions = {
+    max: 50,
```
abuaboud: I think around 10 should be more than enough for sign in

packages/backend/src/app/core/security/rate-limit.ts
```diff
@@ -0,0 +1,45 @@
...
+        redis: getRedisClient(),
+    })
+})
+
+/**
+ * Extracts the client IP provided by Cloudflare.
+ * @see https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip
+ */
+const getClientIpProvidedByCloudflare = (request: FastifyRequest): string | number | Promise<string | number> => {
+    return request.headers['cf-connecting-ip'] as string
```
abuaboud: Why not make this available to all editions? 

We always prioritize cf ip then x-real-ip then a normal one. and remove this editionIsNotCloud

abuaboud (changes requested):

packages/backend/src/app/authentication/authentication.controller.ts
khaledmashaly: Are you sure about this? multiple devices normally share public IPs, this could make a user unable to sign in because another user sharing the same public IP made a lot of requests.

abuaboud (approved):

abuaboud (approved):

PR #3779, opened 2024-01-30
feat(puppeteer): new piece + replay action

MoShizzle:
## What does this PR do?

Fixes # (issue)


abuaboud:
Unfortunately this piece will require privileged docker to make it work, which will be troublesome in some setup I am closing this pr and we should revisit later.

Chromium do sandboxing and it requires privileged container (SYS_ADMIN), and If the docker is not privileged then we will run into problems in UNSANDBOXED mode, Let's dive into this later.

PR #3803, opened 2024-02-02
fix: multiple delays should work

abuaboud:
## What does this PR do?

Remove the dependencies on execution type. It's poor to depend on that state as the flow can alternate between them. For example, if you have two delay steps after the first delay, the execution type will be resume and it will skip the other delay. 

Solution: Check if the flow is paused from the execution state and decide the state. 

This discovered a hidden bug that if there is a delay inside a loop, it will redo all the steps inside the loop as it overwrites the iteration object inside the loop with an empty one each time the loop is executed, so instead now we try to reuse old loop state if it exists and skip emptying the iteration object.

### DIFF
--- a/packages/shared/src/lib/flow-run/execution/step-output.ts
+++ b/packages/shared/src/lib/flow-run/execution/step-output.ts
@@ -1,5 +1,6 @@
 import { TriggerType } from '../../flows/triggers/trigger'
 import { ActionType } from '../../flows/actions/action'
+import { isNil } from '../../common'
 
 export enum StepOutputStatus {
     FAILED = 'FAILED',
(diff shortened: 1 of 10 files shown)


packages/shared/src/lib/flow-run/execution/step-output.ts
```diff
@@ -120,12 +120,27 @@ export class LoopStepOutput extends GenericStepOutput<ActionType.LOOP_ON_ITEMS,
         })
     }
 
-    addIteration({ item, index }: { item: unknown, index: number }): LoopStepOutput {
+    hasIteration(iteration: number): boolean {
+        return !!(this.output?.iterations[iteration])
```
khaledmashaly: I think we need a more readable predicate like `!== undefined` here

khaledmashaly (changes requested):

PR #3816, opened 2024-02-04
feat(interfaces): add simple form support with file and text input

MoShizzle:
## What does this PR do?

Fixes # (issue)


packages/pieces/community/interfaces/src/lib/triggers/interface-trigger.ts
```diff
@@ -0,0 +1,77 @@
...
+  TriggerStrategy,
+  createTrigger,
+} from '@activepieces/pieces-framework';
+
+export const interfaceTrigger = createTrigger({
+  name: 'interface_trigger',
+  displayName: 'Interface',
+  description: 'Trigger the flow through an interface.',
+  props: {
+    textInputs: Property.Array({
```
abuaboud: This is not an extensible structure. For example, when we add a new file type, we end up adding multiple arrays, and we also impose an order on the interface.

I would suggest renaming this to 'inputs,' removing 'placeholder' from the text field, and adding a static dropdown field called 'Type' with "FILE" and "TEXT" as options.

This will pose a challenge in customizing each object inside the array later, depending on the type. However, I believe this is where we should focus on solving in the future, rather than creating multiple arrays.

packages/pieces/community/interfaces/src/index.ts
```diff
@@ -0,0 +1,13 @@
+import { createPiece, PieceAuth } from '@activepieces/pieces-framework';
+import { interfaceTrigger } from './lib/triggers/interface-trigger';
+import { returnResponse } from './lib/actions/return-response';
+
+export const interfaces = createPiece({
+  displayName: 'Interfaces',
```
abuaboud, line 7: I remember we discussed "Forms" vs "Interfaces".

"Interfaces" is a too generic word for a form 
"form" would give the feeling that it stores responses and normal form software.

I don't really have other suggestions but please try to think about it and let see if there is better word

packages/ui/core/src/app/modules/interfaces/interfaces.component.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
abuaboud: We can remove this file, We don't have tests on the frontend

packages/ui/core/src/app/modules/interfaces/interfaces.component.scss
abuaboud, line 1: This can be removed as well, we kind of prefer tailwind css classes rather than custom css.

packages/ui/core/src/app/modules/interfaces/interfaces.service.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Observable } from 'rxjs';
+import { HttpClient } from '@angular/common/http';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class InterfacesService {
+  constructor(private http: HttpClient) {}
+
+  submitInterface(webhookUrl: string, request: FormData): Observable<any> {
```
abuaboud: I believe we can make this typed instead of `any` following the same structure for interfaces -> return response.

packages/pieces/community/interfaces/src/lib/triggers/interface-trigger.ts
```diff
@@ -0,0 +1,77 @@
...
+          displayName: 'Required',
+          description: 'If checked, the input field will be required.',
+          required: true,
+        }),
+      },
+    }),
+    waitForResponse: Property.Checkbox({
+      displayName: 'Wait for Response',
+      description:
+        'If enabled, the interface will return the flow output to the frontend. Make sure to use HTTP Piece -> Return Response action to return the response.',
```
abuaboud: Make sure to  use `Interfaces Piece -> Return Response` instead of http

packages/pieces/community/interfaces/src/lib/triggers/interface-trigger.ts
```diff
@@ -0,0 +1,77 @@
...
+  Property,
+  TriggerStrategy,
+  createTrigger,
+} from '@activepieces/pieces-framework';
+
+export const interfaceTrigger = createTrigger({
+  name: 'interface_trigger',
+  displayName: 'Interface',
+  description: 'Trigger the flow through an interface.',
+  props: {
```
abuaboud, line 16: I think the user have no idea how to reach the form link :D

Check this piece -> https://github.com/activepieces/activepieces/blob/3d4a76f00a75462a87c5be5e4040a84501bbfde2/packages/pieces/community/gravityforms/src/lib/triggers/new-submission.ts#L14

packages/ui/core/src/app/modules/interfaces/interfaces.component.ts
```diff
@@ -0,0 +1,250 @@
...
+        observer.complete();
+      };
+
+      reader.onerror = (error) => {
+        observer.error(error);
+      };
+    });
+  }
+
+  mimeToExtension(mimeType) {
```
abuaboud: Use that  https://www.npmjs.com/package/mime-types, we already have it installed

abuaboud (changes requested):

PR #3828, opened 2024-02-05
feat(branch): add new condition "(number) is equal to"

pfernandez98:
## What does this PR do?

I added the "(number) is equal to" to the conditions of a branch step.
I get you can do the same with the "(Text) exactly matches", but I think is better to have a dedicated condition for numbers... I mean, you could use the "(Text) exactly matches" for booleans too but instead, we have two separate conditions.

image

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -5731,6 +5731,12 @@
                                                         "enum": [
                                                           "NUMBER_IS_LESS_THAN"
                                                         ]
+                                                      },
+                                                      {
+                                                        "type": "string",
+                                                        "enum": [
+                                                          "NUMBER_IS_EQUAL_TO"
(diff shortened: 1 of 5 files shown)


abuaboud:
Great @pfernandez98 

Can you please add test there ->  https://github.com/activepieces/activepieces/blob/main/packages/engine/test/handler/flow-branching.test.ts

pfernandez98:
> Great @pfernandez98
> 
> Can you please add test there -> https://github.com/activepieces/activepieces/blob/main/packages/engine/test/handler/flow-branching.test.ts

Sure, I'll add them now.

pfernandez98:
@abuaboud done.

abuaboud (approved):

PR #3864, opened 2024-02-08
fix: snackbar after interface submission received

MoShizzle:
## What does this PR do?

Adds a snackbar that appears after the user submits the interface form successfully, the snackbar appears only when not waiting for response

### DIFF
--- a/packages/ui/core/src/app/modules/interfaces/interfaces.component.ts
+++ b/packages/ui/core/src/app/modules/interfaces/interfaces.component.ts
@@ -156,10 +156,15 @@ export class InterfacesComponent implements OnInit {
             link.click();
             // Clean up by revoking the object URL
             URL.revokeObjectURL(fileBase.base64Url);
+          } else {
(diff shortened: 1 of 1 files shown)


packages/ui/core/src/app/modules/interfaces/interfaces.component.ts
```diff
@@ -146,14 +149,17 @@ export class InterfacesComponent implements OnInit {
...
+            const fileBase = result.value as FileResponseInterface;
             link.download = fileBase.fileName;
-            link.href = fileBase.base64Url
+            link.href = fileBase.base64Url;
             link.target = '_blank';
             link.click();
             // Clean up by revoking the object URL
             URL.revokeObjectURL(fileBase.base64Url);
           }
+          this.snackBar.open(`Your submission was successfully received.`, '', {
```
abuaboud: Only do it for non sync operations

abuaboud (changes requested):

PR #3870, opened 2024-02-08
feat: store service operations are now atomic 

MarceloRGonc:
## What does this PR do?

This PR prevents race conditions in changes made to storage.


abuaboud:
Hi @MarceloRGonc 

Great PR, I think the delete lock is not required as postgres already handle these at row level, for insert I think we can replace it with transaction or upsert command.


I will review this soon, what are the problem are you trying to solve I am curious :D

MarceloRGonc:
> Hi @MarceloRGonc
> 
> Great PR, I think the delete lock is not required as postgres already handle these at row level, for insert I think we can replace it with transaction or upsert command.
> 
> I will review this soon, what are the problem are you trying to solve I am curious :D

Hey @abuaboud 

The delete needs the lock because we can do a deletion in the middle of an update. Only if we change the add/update to let postgres take care of everything, but I'm not sure if it's possible.

I'm working on running some parts of the flow in parallel, basically I want the loop to run the iterations in parallel. I will share the result when it is ready.

packages/backend/src/app/store-entry/store-entry.controller.ts
```diff
@@ -36,11 +36,18 @@ export const storeEntryController: FastifyPluginAsyncTypebox = async (fastify) =
...
         ) => {
-            return storeEntryService.getOne({
-                projectId: request.principal.projectId, key: request.query.key,
+            const value = await storeEntryService.getOne({
+                projectId: request.principal.projectId,
+                key: request.query.key,
             })
+
+            if (!value) {
+                return reply.code(404).send('Value not found!')
```
abuaboud: use StatusCode enum instead of 404

packages/backend/src/app/store-entry/store-entry.service.ts
```diff
@@ -1,36 +1,62 @@
...
+
             const entryRequest: Omit<StoreEntry, 'created' | 'updated'> = {
                 id: apId(),
                 key: request.key,
                 value: request.value,
                 projectId,
             }
-            return storeEntryRepo.save(entryRequest)
+
+            return await storeEntryRepo.save(entryRequest)
```
abuaboud: Use upsert, this suppose to be atomic

https://github.com/activepieces/activepieces/blob/09422ca992665ea9d2be4ad7c180a134e6dd8a56/packages/backend/src/app/ee/referrals/referral.service.ts#L28

packages/backend/src/app/store-entry/store-entry.service.ts
```diff
@@ -1,36 +1,62 @@
...
             key,
         })
     },
+
     async delete({ projectId, key }: { projectId: ProjectId, key: string }): Promise<void> {
-        await storeEntryRepo.delete({
-            projectId,
-            key,
-        })
+        const keyLock = await getLock(key)
```
abuaboud: If we made upsert atomic, then the lock would be no longer needed, so please remove this after changing the above to use upsert

abuaboud (changes requested):

abuaboud (dismissed):

khaledmashaly (approved):

packages/backend/src/app/store-entry/store-entry.service.ts
```diff
@@ -6,27 +6,21 @@ const storeEntryRepo = databaseConnection.getRepository<StoreEntry>(StoreEntryEn
...
-                projectId,
-            }
-            return storeEntryRepo.save(entryRequest)
-        }
+        return storeEntryRepo.upsert({
+            id: apId(),
+            key: request.key,
+            value: request.value,
+            projectId,
+        }, ['projectId', 'key'])
```
khaledmashaly: The `Repo#upsert` method doesn't return the upserted entity, i.e. the `StoreEntry` object, please make sure to return the upserted `StoreEntry` object, otherwise, this change breaks the API contract.

I think a possible solution is to use the `Repo#get` method after upserting.

khaledmashaly (changes requested):

packages/backend/src/app/store-entry/store-entry.service.ts
MarceloRGonc: @abuaboud If the upsert doesn't return the same object, we need to use the lock. Because as I said before we can have a delete or another update in between.

abuaboud: Hi @MarceloRGonc I think it's possible to construct the value from parameters (even better construct it from InsertResut that returned from upsert )

So I suggest you try construct it from InsertResult object in upsert

MarceloRGonc: Hi @abuaboud. The upsert doesn't work on this table. I'm getting this error: "there is no unique or exclusion constraint matching the ON CONFLICT specification". To use the upsert I think we need to change the table definition to make 'projectId' and 'Key' the primary key.

abuaboud: I think we should add unique constraint on compound key projectId and key, Let's do that what do you think?

MarceloRGonc: @abuaboud and @khaledmashaly I already made the changes

abuaboud (approved):

PR #3871, opened 2024-02-09
chore: block parallel resume of the same flow run

rita-gorokhod:
## What does this PR do?

This PR acquires a lock on flowRunId, to avoid a situation where multiple resumes are called in parallel for the same run.
For example, if I have a flow that sends a slack approval, I can hit the approve button multiple times.  In the current state of the code, this will resume multiple times and execute the steps after the approval again and again.
By adding a lock, we ensure each flow run can resume only once at a time.

image

Tested with  `await new Promise(resolve => setTimeout(resolve, 10000))` to simulate a delay in the resume step.

### DIFF
--- a/packages/ee/project-members/src/lib/project-members-table/project-members-table.component.html
+++ b/packages/ee/project-members/src/lib/project-members-table/project-members-table.component.html
@@ -2,7 +2,7 @@
   <div class="ap-h-full">
     <ap-page-title [title]="title"></ap-page-title>
     <div class="ap-mb-4" *ngIf="!isFeatureEnabled">
-      <ap-upgrade-note docsLink="https://www.activepieces.com/docs/team/overview"></ap-upgrade-note>
+      <ap-upgrade-note [featureNote]="upgradeNote" ></ap-upgrade-note>
     </div>
     <div class="ap-min-h-[70%]">
       <table mat-table [dataSource]="dataSource" class="ap-w-[100%]" aria-label="Project Members" i18n-aria-label>
--- a/packages/ee/project-members/src/lib/project-members-table/project-members-table.component.ts
+++ b/packages/ee/project-members/src/lib/project-members-table/project-members-table.component.ts
@@ -55,6 +55,7 @@ export class ProjectMembersTableComponent
   displayedColumns = ['email', 'role', 'status', 'created', 'action'];
   title = $localize`Project Members`;
   RolesDisplayNames = RolesDisplayNames;
+  upgradeNote = $localize`Effortlessly invite your teammates to your project and assign them roles, allowing them to view/edit flows or check runs.`;
   StatusDisplayNames: { [k: string]: string } = {
     [ProjectMemberStatus.ACTIVE]: $localize`Active`,
     [ProjectMemberStatus.PENDING]: $localize`Pending`,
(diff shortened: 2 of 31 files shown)


abuaboud:
@rita-gorokhod Good catch!

I think there is another take here, If you have two approval step in same flow, we should generate a unique url for each instead of the same url.

Do you have any idea on how to implement that?

rita-gorokhod:
> I think there is another take here, If you have two approval step in same flow, we should generate a unique url for each instead of the same url.

@abuaboud yep I saw that too. When you get to the second approval, you're already in "resume" state and the webhook action is identical, so it thinks the second approval was already approved too. I'm hacking around it for now by appending a guid to the approval webhook, via a helper `generateUnpauseWebhook` method that also hides the structure of the resume url. I'm not sure what would be a better solution, since you need to assign this value to the slack button, so the piece must know it.

abuaboud:
Hi @rita-gorokhod 

Please take look here -> https://github.com/activepieces/activepieces/pull/3887

I think It can be solved with the other issue, I didn't finish yet

But each pause will have a unique id that can be executed once, that what I am planning for.

rita-gorokhod:
> Hi @rita-gorokhod,
> 
> It's now merged. Do you mind putting the lock on the resume part (the webhook controller) rather than inside the queue? In that case, it will only enter the queue once.
> 
> I think the status of the run is changed to "RUNNING" instead of "PAUSED," so the lock should ignore adding it to the queue again if the status is "RUNNING."
> 
> What do you think?

Hi @MoShizzle , thanks for merging, my apologies for the delay in responding. I took a look at that PR and I think there is still a problem there with resuming multiple approval steps. I left a comment there.

Regarding the suggestion: I think you have to insert it into the queue since the resume can be called for different steps. For example, if you have two approval steps, you can get a resume first for one and then for the other, and you should run the flow for both of them, just not in parallel. 
I think it will be needed for the work @MarceloRGonc is doing to support parallel loop execution.

abuaboud:
Hi @MarceloRGonc and @rita-gorokhod 

I see the problem with asynchronous loop, I never encountered a use case that requires asynchronous loop, I think I am missing a lot of context here.  

What is examples of the real-world use case behind that?

rita-gorokhod:
> Hi @MarceloRGonc and @rita-gorokhod
> 
> I see the problem with asynchronous loop, I never encountered a use case that requires asynchronous loop, I think I am missing a lot of context here.
> 
> What is examples of the real-world use case behind that?

@abuaboud For example sending slack approvals to multiple users. If you don't do it in parallel, you have to wait for each person to approve before sending the message to the next user.

abuaboud:
Hi @rita-gorokhod,

That's a valid use case I have heard before, but I do believe a loop is not an appropriate place for such changes. For a no-code user, it's better to see all the branching on the screen. Additionally, I believe in such case each approval person their own sub-flow, which means they need to perform all handling inside the loop.

I think this guy describes it well:
[Link to the description](https://community.activepieces.com/t/asynchronous-workflows/1210)

My suggestion is let us scope this together and implement it in a more declarative and explicit manner from the end user's point of view. My thoughts are that a normal no-code user would never guess an async loop to fulfill this use case.

What do you think?

rita-gorokhod:
> My suggestion is let us scope this together and implement it in a more declarative and explicit manner from the end user's point of view

Sure, what did you have in mind? I agree that the parallel implementation (and actually, the concept of 'loop' in the first place) is a foreign concept to many users. In our case, our users are not technical, so any UX improvement will be much appreciated and contribute to their ease of use.

PR #3880, opened 2024-02-11
feat: add audit logs

abuaboud:
## What does this PR do?

[X] Add Basic Table to View Events With Pagination
[X] Created Database Migration
[X] Logged Flow / Folder / Connections / Users Events, rest will be added later.

localhost_4200_platform_settings


packages/backend/src/app/app-connection/app-connection.controller.ts
```diff
@@ -10,14 +10,20 @@ import {
...
         const appConnection = await appConnectionService.upsert({
             projectId: request.principal.projectId,
             request: request.body,
         })
-
+        void eventsHooks.get().send(request, {
+            action: ApplicationEventName.UPSERTED_CONNECTION,
+            connection: appConnection,
+            userId: request.principal.id,
+        })
```
khaledmashaly, line 26: Please note that the server process will stop if this promise rejects. You can test this by adding `void Promise.reject('error')` to whatever endpoint you like, and observe how the server explodes when the endpoint receives any request.

I think we should create a shared `rejectedPromiseHandler` for the backend. Logging the error is enough for the time being. The handler can be attached using `Promise#catch` method.

packages/backend/src/app/ee/audit-logs/audit-event-service.ts
```diff
@@ -0,0 +1,119 @@
+import { databaseConnection } from '../../database/database-connection'
+import { AuditEventEntity } from './audit-event-entity'
+import { Cursor, PrincipalType, SeekPage, apId, assertNotNullOrUndefined, isNil } from '@activepieces/shared'
+import { PlatformId } from '@activepieces/ee-shared'
+import { buildPaginator } from '../../helper/pagination/build-paginator'
+import { paginationHelper } from '../../helper/pagination/pagination-utils'
+import { ApplicationEventHooks } from '../../helper/audit-events'
+import { extractClientRealIp } from '../../core/security/rate-limit'
```
khaledmashaly: I don't think it's a good idea that the `audit-event-service` module depends on the `rate-limit` module, as the relation seems unclear. A better approach is to refactor the `extractClientRealIp` method to a shared utility module, for example.

packages/backend/src/app/ee/authentication/enterprise-local-authn/enterprise-local-authn-controller.ts
```diff
@@ -2,14 +2,26 @@ import { FastifyPluginAsyncTypebox } from '@fastify/type-provider-typebox'
...
+import { ApplicationEventName } from '@activepieces/ee-shared'
 
 export const enterpriseLocalAuthnController: FastifyPluginAsyncTypebox = async (app) => {
     app.post('/verify-email', VerifyEmailRequest, async (req) => {
         await enterpriseLocalAuthnService.verifyEmail(req.body)
+        await eventsHooks.get().send(req, {
+            action: ApplicationEventName.VERIFIED_EMAIL,
+            userId: req.body.userId,
+            projectId: req.principal.projectId,
+        })
```
khaledmashaly, line 15: I think adding `await` here was by mistake, right? I think no need to wait for this promise as its return value is ignored anyway. Just make sure to append `.catch`.

khaledmashaly (changes requested):

packages/backend/src/app/ee/authentication/enterprise-local-authn/enterprise-local-authn-controller.ts
abuaboud: send is not longer async, and implemented rejectpromise handler and have it inside the audit service.

abuaboud:
I added flag in the platform entity to enable / disable audit logs as well

PR #3887, opened 2024-02-11
fix: allow multiple approvals in same flow

abuaboud:
## What does this PR do?

Fixes # (issue)

# MAJOR CHANGES
- Moved payload directly inside resumePayload (breaking change for approval piece) 
- Create piece now changes the minimum version is 0.20.0 (not released)
- changed minimum version for approval / slack / discord to 0.20.0 (not released) 

# TESTS

- Tested the pause / resume without sync
- Tested the pause with sync and it returns reply
- Tested running it on mismatch request Id It ignores and with correct one it works.


rita-gorokhod:
@abuaboud I see that pauseRequestId is set only 1 time, at the start of the flow run. If I'm not mistaken, this essentially makes it equal to flowRunId, so if you have 2 pauses in the same flow, the second one will be unpaused as soon as you unpause the first one.

PR #3892, opened 2024-02-11
feat: add AuthValidationContext

AdamSelene:
## What does this PR do?

Pass a new AuthValidationContext to `validate` method of `PieceAuth` to make it possible to store / retrieve values from storage (`PROJECT` scope only obviously - even though it is not enforced)

This will allow us to handle some authentication schemes where a token needs to be periodically refreshed (e.g. [Metabase](https://www.metabase.com/learn/administration/metabase-api#authenticate-your-requests-with-a-session-token))


AdamSelene:
Closing after discussing it with @abuaboud

PR #3893, opened 2024-02-12
feat(metabase): Get Question Action

AdamSelene:
## What does this PR do?

Add Metabase piece with action to get results from a "question"
Metabase auth scheme requires a session token (with a fixed duration) that is delivered by a specific endpoint (using email + password)
The Piece will automatically get the token and refresh it when a HTTP 401 error is received, as per Metabase documentation

The token is encrypted (using code which is very similar to `encryption.ts`) and stored at the Project scope

https://www.metabase.com/learn/administration/metabase-api#authenticate-your-requests-with-a-session-token

Capture d’écran 2024-02-12 à 00 43 33

Capture d’écran 2024-02-12 à 00 44 36

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/community/metabase/.eslintrc.json
+++ b/packages/pieces/community/metabase/.eslintrc.json
@@ -1,0 +1,37 @@
+{
+  "extends": [
+    "../../../../.eslintrc.json"
+  ],
+  "ignorePatterns": [
+    "!**/*"
+  ],
+  "overrides": [
+    {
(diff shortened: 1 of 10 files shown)


AdamSelene:
Logo to be hosted at `https://cdn.activepieces.com/pieces/metabase.png`
Metabase Logo Blue

abuaboud (approved):

PR #3910, opened 2024-02-13
fix: notify on failed runs toggle wasn't working

AbdulTheActivePiecer:
## What does this PR do?

I moved loading projects from app.component.ts to user-logged-in.guard.ts, that way we never have to worry about a page being opened before projects are loaded, that was causing the notify status not being saved correctly.

I also stopped the projects table from reinitializing the projects store every time a project is added/updated and instead opted to adding/updating the project to/in the store instead.

Lastly, I made the frontend use the User model without the password property.


abuaboud (approved):

packages/ui/common/src/lib/guards/user-logged-in.guard.ts
```diff
@@ -31,6 +45,65 @@ export class UserLoggedIn {
...
+          return of(true);
+        }
+        return this.projectService.list().pipe(
+          tap((projects) => {
+            if (!projects || projects.length === 0) {
+              console.error('No projects are assigned to the current user');
+              this.auth.logout();
+            }
+          }),
+          switchMap((projects) => {
```
abuaboud, line 61: Nit you should split this into a couple of declarative functions

PR #3913, opened 2024-02-13
fix: show full step name in tooltip when truncated in the builder

AbdulTheActivePiecer:


packages/ui/feature-builder-canvas/src/lib/components/canvas-utils/drawing/draw-common.ts
```diff
@@ -14,6 +14,7 @@ export type PositionLabel = {
 };
 
 export const FLOW_ITEM_WIDTH = 250;
```
abuaboud, line 16: We have space in terms of horizontal, why not increase it back to 290? what do you think?

abuaboud (changes requested):

packages/ui/feature-builder-canvas/src/lib/components/canvas-utils/drawing/draw-common.ts
AbdulTheActivePiecer: I will discuss the design we have with Ash later and see what we edits we come up with, leave it for now.

abuaboud (approved):

PR #3916, opened 2024-02-13
fix: http body for non-json requests

### LINKED ISSUE #3915: [BUG]: http body with anything other than valid json sends empty body
**Describe the bug**
When sending a http request using the http piece and setting the body to everything other than valid json it will instead send an empty body

**To Reproduce**
1. Create a http request step
2. Set the url to https://echo.free.beeceptor.com
3. Set the body to '1234'
4. Test the step
5. Check the response for any body

**Expected behavior**
It should send the body like it did in prior versions

**Additional context**
Already checked it with @abuaboud and we found the issue (Property.Json now has validation), I'll file a PR to fix this issue by changing the property type to LongText

JanHolger:
## What does this PR do?

Fixes #3915

### DIFF
--- a/packages/pieces/community/http/src/lib/actions/send-http-request-action.ts
+++ b/packages/pieces/community/http/src/lib/actions/send-http-request-action.ts
@@ -27,7 +27,7 @@ export const httpSendRequestAction = createAction({
       displayName: 'Query params',
       required: true,
     }),
-    body: Property.Json({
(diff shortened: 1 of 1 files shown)


AbdulTheActivePiecer:
could you please remove this as well from the piece ? 
image

abuaboud (approved):

PR #3930, opened 2024-02-14
chore: add parallel loop

MarceloRGonc:
## What does this PR do?

This PR adds support for running iterations of a loop in parallel.
- We start by triggering all iterations.
- Then we wait for the iterations to finish.
-  If any of the iterations are in a paused state, the results are saved to be used after the resume
- We only exit the loop after all iterations have finished, that is, if any are paused we will wait for the resume.

Depends one:
- https://github.com/activepieces/activepieces/pull/3870
- https://github.com/activepieces/activepieces/pull/3871

### DIFF
--- a/packages/engine/src/lib/handler/context/step-execution-path.ts
+++ b/packages/engine/src/lib/handler/context/step-execution-path.ts
@@ -17,4 +17,8 @@ export class StepExecutionPath {
         const newPath = this.path.slice(0, -1)
         return new StepExecutionPath(newPath)
     }
+
+    toString(): string {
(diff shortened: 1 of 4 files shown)


abuaboud:
Hi @MarceloRGonc,

This is a bit surprising for Activepieces Engine. I have never encountered a use case where it requires loops that run asynchronously.

What is the real-world use case behind that?

abuaboud:
Hi @MarceloRGonc,

I am temporarily closing this PR until we find a solution together for the mentioned comment. Please reopen this PR as soon as we make progress on the main PR.

Thank you,
Mo.

PR #3955, opened 2024-02-18
feat: show suggested triggers/actions while searching

AbdulTheActivePiecer:


packages/ui/feature-pieces/src/lib/services/piece-meta.service.ts
```diff
@@ -274,25 +275,38 @@ export class PieceMetadataService {
...
       this.clearCache$.asObservable().pipe(startWith(void 0)),
     ]).pipe(
       switchMap(([edition, release]) => {
+        let params:Record<string,boolean|string>= {
+          release,
+          edition
+        };
+        if(includeHidden)
+        {
+          params={...params,includeHidden}
```
abuaboud: use SpreadIfDefined that works well

abuaboud:
```ts
        const params:Record<string,boolean|string>= {
          release,
          edition,
          ...spreadIfDefined('includeHidden', includeHidden),
          ...spreadIfDefined('searchQuery', searchQuery),
          ...spreadIfDefined('suggestActionsAndTrigger', suggestActionsAndTrigger)
        };
```

packages/server/api/src/app/pieces/piece-metadata-service/index.ts
```diff
@@ -48,3 +49,48 @@ export const getPiecePackage = async (
...
+                triggers: Object.keys(pieceMetadataEntity.triggers).length,
+            
+            }
+        
+        })
+    }
+
+    return pieceMetadataEntityList.map((pieceMetadataEntity) => {
+     
+        const suggestedActions = Object.keys(pieceMetadataEntity.actions).map((key) => {
```
abuaboud: Change that to Object.values(piecemetadataEntity.actions), same for below

packages/server/api/src/app/pieces/piece-metadata-service/index.ts
```diff
@@ -48,3 +49,48 @@ export const getPiecePackage = async (
         }
     }
 }
+
+export function toPieceMetadataModelSummary<T extends PieceMetadataSchema | PieceMetadataModel>(
+    pieceMetadataEntityList: T [],
+    showSuggestedActionsAndTriggers?: boolean,
+): PieceMetadataModelSummary[] {
+    if (!showSuggestedActionsAndTriggers) {
```
abuaboud: This function can be replaced with

```ts
export function toPieceMetadataModelSummary<T extends PieceMetadataSchema | PieceMetadataModel>(
    pieceMetadataEntityList: T[],
    showSuggestedActionsAndTriggers?: boolean,
): PieceMetadataModelSummary[] {
    return pieceMetadataEntityList.map((pieceMetadataEntity) => {
        return {
            ...pieceMetadataEntity,
            actions: Object.keys(pieceMetadataEntity.actions).length,
            triggers: Object.keys(pieceMetadataEntity.triggers).length,
            suggestedActions: showSuggestedActionsAndTriggers ? Object.values(pieceMetadataEntity.actions) : undefined,
            suggestedTriggers: showSuggestedActionsAndTriggers ? Object.values(pieceMetadataEntity.triggers) : undefined,
        }
    })
}```

File: packages/server/api/src/app/pieces/piece-metadata-service/hooks/piece-filtering.ts [hunk 4]
```diff
@@ -63,3 +73,103 @@ const filterBasedOnCategories = ({
...
+            ...piece,
+            actions,
+            triggers,
+        }
+    })
+
+    const keysInAdditionToActionsAndTriggers = [ 
+        ...defaultFilterKeys,
+        {
+            name: 'actions.displayName',
```
abuaboud: If all weights are same why not just?
```ts
    const fuse = new Fuse(putActionsAndTriggersInAnArray, {
        isCaseSensitive: false,
        shouldSort: true,
        keys: ['display', 'description',
            'actiions.displayName', 'actions.description',
            'triggers.description', 'triggers.displayName'],
        threshold: 0.2,
    })
 ```


    
    Let's also remove defaultFilterKeys variable and make it inline as it's more readable

File: packages/server/api/src/app/pieces/piece-metadata-service/hooks/piece-filtering.ts [hunk 5]
```diff
@@ -1,51 +1,61 @@
 import { PieceCategory } from '@activepieces/shared'
 import { PieceMetadataSchema } from '../../piece-metadata-entity'
 import Fuse from 'fuse.js'
+import { ActionBase, TriggerBase } from '@activepieces/pieces-framework'
+
+
+const defaultFilterKeys = [  {
```
abuaboud: REmove it and replace it in line ['name', 'description'] also no need for wieght if they are same

packages/server/api/src/app/pieces/piece-metadata-service/hooks/piece-filtering.ts
```diff
@@ -63,3 +73,103 @@ const filterBasedOnCategories = ({
...
+            
+        })
+
+}
+function searchForRelatedActionsAndTriggers(originalPiece: PieceMetadataSchema, searchQuery: string):
+{
+    suggestedActionsAndTriggers: { description: string, displayName: string, key: string }[]
+} {
+  
+    const actionsAndTriggers = [
```
abuaboud: This thing can be replaced with

TLDR if return type is one variable why return an object?
Onject.keys is same as Object.values.
Remove variable and make it inline in the end
```ts
function searchForRelatedActionsAndTriggers(originalPiece: PieceMetadataSchema, searchQuery: string): (ActionBase | TriggerBase)[] {
    const nestedFuse = new Fuse([
        ...Object.values(originalPiece.actions),
        ...Object.values(originalPiece.triggers),
    ], {
        isCaseSensitive: false,
        shouldSort: true,
        keys: ['displayName', 'description'],
        threshold: 0.2,
    })
    return nestedFuse.search(searchQuery, { limit: 3 }).map(({ item }) => item)
}

```

abuaboud (changes requested):

packages/ui/feature-pieces/src/lib/services/piece-meta.service.ts
AbdulTheActivePiecer: <3

packages/server/api/src/app/pieces/piece-metadata-service/index.ts
AbdulTheActivePiecer: I don't want to return full actions and triggers that's why I do the mapping between actions/triggers to {displayName, name}, it's an optimization.

packages/server/api/src/app/pieces/piece-metadata-service/hooks/piece-filtering.ts
AbdulTheActivePiecer: I removed the weights and renamed the arrays to be more descriptive, (pieceFilterKeys and pieceWithTriggersAndActionsFilterKeys) the reason being I don't want to have to edit two arrays when I edit pieceFilterKeys or even update one and forget the other.

packages/server/api/src/app/pieces/piece-metadata-service/hooks/piece-filtering.ts
AbdulTheActivePiecer: I don't like inline values when there is a chance we edit them in the future, it forces me to re-read the code and find out where to edit the value, that's why I use consts, I moved it to the top, in regards for the return type, it is made on purpose, so we can map the array back to a Record<string, ActionBase | TriggerBase>

abuaboud (approved):

PR #3993, opened 2024-02-22
fix: stop flows filter in runs table from flickering

AbdulTheActivePiecer:


packages/ui/feature-dashboard/src/lib/pages/runs-table/runs-table.component.ts
```diff
@@ -124,6 +125,7 @@ export class RunsTableComponent implements OnInit {
         limit: 1000,
       })
       .pipe(
+        delay(55000),
```
abuaboud: Is this by mistake?

abuaboud (changes requested):

packages/ui/feature-dashboard/src/lib/pages/runs-table/runs-table.component.ts
AbdulTheActivePiecer: yes

PR #4014, opened 2024-02-25
feat(api): update activity endpoint

khaledmashaly:


packages/server/api/src/app/ee/activity/activity-worker-controller.ts
```diff
@@ -12,6 +12,15 @@ export const activityWorkerController: FastifyPluginAsyncTypebox = async (app) =
...
             .send(newActivity)
     })
+
+    app.post('/:id', UpdateActivityRequest, async (req, res) => {
+        await activityService.update({
+            id: req.params.id,
+            ...req.body,
+        })
+
+        return res.status(StatusCodes.NO_CONTENT).send()
```
abuaboud, line 22: Nit: make it return the updated value like every other endpoint

abuaboud (approved):

PR #4044, opened 2024-02-27
refactor: create default platform for each project

abuaboud:
### Changes Made:
- PlatformId is now mandatory in JWT. It adds a platform with an anonymous id for anonymous users.
- The browser logs out when the platform id is not present in the JWT token.
- Generated Migration for Enterprise Postgres.
- Generated Migration for Community Postgres / Community Sqlite3.
  - Both create a platform, then create a default platform for each project, and finally update the user and project with the platform id.
  
Please review that:

Sqlite3 migration 
Postgres migration 

### Testing:
- [x] Tested on Enterprise Edition and Postgres / Community Postgres / Community Sqlite3.
- [ ] Created an account before this change.
- [ ] Ran this change and verified that the account now has a platform, and the browser logged out due to the lack of a platform id.

Please review the changes and let me know if there are any further adjustments needed.


packages/server/api/src/app/authentication/authentication-service/hooks/community-service-hooks.ts
```diff
@@ -11,45 +16,52 @@ export const communityAuthenticationServiceHooks: AuthenticationServiceHooks = {
...
     async postSignUp({ user }) {
-        const project = await projectService.create({
+        const platformCreated = await flagService.getOne(ApFlagId.PLATFORM_CREATED)
+        if (platformCreated?.value) {
+            return getProjectAndToken(user)
+        }
+        const platform = await platformService.create({
+            ownerId: user.id,
+            name: DEFAULT_PLATFORM_NAME,
+        })
```
khaledmashaly, line 26: Should this bit be atomic?

packages/shared/src/lib/authentication/model/principal.ts
```diff
@@ -1,13 +1,12 @@
...
-import { ProjectId, ProjectType } from '../../project/project'
+import { ProjectId } from '../../project/project'
 
 export type Principal = {
     id: ApId
     type: PrincipalType
     projectId: ProjectId
-    projectType?: ProjectType
-    platform?: {
+    platform: {
```
khaledmashaly, line 9: Please note that older tokens might be missing the `platform` prop.

abuaboud: Good catch, I will force logout there

PR #4058, opened 2024-02-29
feat(stable-diffusion-web-ui): text to image action

AdamSelene:
## What does this PR do?

Add Piece to use the text-to-image API from [Stable Diffusion web UI](https://github.com/AUTOMATIC1111/stable-diffusion-webui) - returns base64 images (similar to image-helper)

Capture d’écran 2024-02-28 à 22 59 54

### DIFF
--- a/packages/pieces/community/stable-diffusion-webui/src/lib/actions/text-to-image.ts
+++ b/packages/pieces/community/stable-diffusion-webui/src/lib/actions/text-to-image.ts
@@ -1,0 +1,100 @@
+import { createAction, Property } from '@activepieces/pieces-framework';
+import { randomBytes } from 'node:crypto';
+import { kebabCase } from 'lodash';
+
+import {
+  httpClient,
+  HttpMethod,
+  HttpRequest,
+} from '@activepieces/pieces-common';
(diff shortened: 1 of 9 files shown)


packages/pieces/community/stable-diffusion-webui/src/lib/actions/text-to-image.ts
```diff
@@ -0,0 +1,86 @@
...
+        },
+        override_settings_restore_afterwards: true,
+      }),
+    };
+    const response = await httpClient.sendRequest(request);
+    return {
+      images: response.body['images']?.map((image: string) => {
+        return {
+          data: `data:image/png;base64,${image}`,
+        };
```
abuaboud, line 93: We can try to use:

You can give each image a random name or something similar. This will return a local path URL in production mode. When using the test step, it will return a URL that can be viewed by the user instead of base64.
```
        await context.files
          .write({
            fileName: `image.png`,
            data: Buffer.from(imageBase64, 'base64'),
          })
 ```

AdamSelene: Done!

abuaboud (approved):

PR #4079, opened 2024-03-04
refactor: remove ExecutionOutput

abuaboud:
## What does this PR do?

This PR does the following:
- Removes everything from `ExecutionOutput`, as it's redundant in `Flow Run` except for steps.
- Makes steps return in FlowRun on get and skip in list.
- Documents Get/List API.
- Renames ExecutionOutputStatus to FlowExecutionOutput.
- Removes terminationStatus from the API and the enum but still needs migration in follow-up.

There should be no breaking change in this PR.

Fixes # (issue)


packages/shared/src/lib/flow-run/execution/flow-execution.ts
```diff
@@ -0,0 +1,76 @@
+
+import { Static, Type } from '@sinclair/typebox'
+import { ExecutionState } from './execution-output'
+
+export enum FlowExecutionStatus {
```
khaledmashaly: Since we are renaming, why not rename to `FlowRunStatus`? I think this will make it more clear instead of the mix between `run` and `execution` which is confusing.

packages/shared/src/lib/flow-run/execution/flow-execution.ts
```diff
@@ -0,0 +1,76 @@
...
+    tasks: Type.Number(),
+    tags: Type.Optional(Type.Array(Type.String())),
+    error: Type.Optional(Type.Object({
+        stepName: Type.String(),
+        message: Type.String(),
+    })),
+    stopResponse: Type.Optional(StopResponse),
+}
+
+export const FlowExecutionResponse = Type.Union([
```
khaledmashaly: Also `FlowRunResponse` here will be much clearer

PR #4092, opened 2024-03-05
feat: delete users

khaledmashaly:
## What does this PR do?

Fixes # (issue)


packages/ee/shared/src/lib/signing-key/signing-key-model.ts
```diff
@@ -16,6 +16,9 @@ export const SigningKey = Type.Object({
     algorithm: Type.Enum(KeyAlgorithm),
     /* ID of user who generated this key pair */
     generatedBy: ApId,
+    generatedByEmail: Type.String({
```
abuaboud: This can be removed :D with the generated by

abuaboud (approved):

abuaboud (dismissed):

PR #4099, opened 2024-03-06
feat: support run scope for storage piece / framework

wise-introvert:
This is a rough implementation for a run-specific scope for the storage piece ( or any `store` available in pieces through context ). I'd like to get your opinion on my approach and ensure I'm doing it right. I'm sure I'm butchering the implementation somewhere :sweat_smile: Please let me know if you feel I need to do things differently or if they could be done better.


packages/engine/src/lib/helper/trigger-helper.ts
```diff
@@ -50,6 +52,7 @@ export const triggerHelper = {
                 prefix,
                 flowId: params.flowVersion.flowId,
                 workerToken: params.workerToken,
+                runId: get(params, 'flowRunId', ''),
```
abuaboud: remove `lodash` and the console log above.

I think we should forbidden the scope `RUN` inside the trigger because it's not attached to run any run.

packages/shared/src/lib/engine/engine-operation.ts
```diff
@@ -91,6 +91,7 @@ export type ExecuteTriggerOperation<HT extends TriggerHookType> = BaseEngineOper
     edition?: string
     appWebhookUrl?: string
     webhookSecret?: string
+    flowRunId?: FlowRunId
```
abuaboud: Please remove this

packages/server/api/src/app/webhooks/webhook-service.ts
```diff
@@ -128,6 +134,8 @@ export const webhookService = {
                 hookType: HookType.BEFORE_LOG,
                 projectId,
                 executionType: ExecutionType.BEGIN,
+                flowRunId,
```
abuaboud: Please remove this as triggers are not related to runs

packages/server/api/src/app/flows/trigger/hooks/execute-trigger-hooks.ts
```diff
@@ -32,6 +33,7 @@ export async function executeTrigger(
                     simulate,
                 }),
                 projectId,
+                flowRunId,
```
abuaboud: Same for changes in this file.

packages/engine/src/lib/services/storage.service.ts
```diff
@@ -51,24 +51,24 @@ export const createStorageService = ({ workerToken }: { workerToken: string }) =
     }
 }
 
-export function createContextStore({ prefix, flowId, workerToken }: { prefix: string, flowId: FlowId, workerToken: string }): Store {
+export function createContextStore({ prefix, flowId, workerToken, runId }: { prefix: string, flowId: FlowId, workerToken: string, runId: FlowRunId }): Store {
```
abuaboud: I would make this file generic `<T>` that returns `Store<T>`, and then I can pass one enum with `FLOW` and `PROJECT` into triggers and another with `runs` as well to the action.

abuaboud (changes requested):
Good job, you are close!

packages/engine/src/lib/helper/trigger-helper.ts
wise-introvert: Ah, okay. So a trigger won't have the run-scope at all?

packages/shared/src/lib/engine/engine-operation.ts
wise-introvert: Gotcha. So, basically remove all references to run-id in all trigger-related helpers, right?

packages/engine/src/lib/services/storage.service.ts
wise-introvert: Could you please give me an example :sweat_smile:. I didn't get the part about making the 'file' generic.

abuaboud:
```ts
export interface Store<SCOPE> {
  put<T>(key: string, value: T, scope?: SCOPE): Promise<T>;
  get<T>(key: string, scope?: SCOPE): Promise<T | null>;
  delete(key: string, scope?: SCOPE): Promise<void>;
}

export enum StoreScope {
  // Collection were deprecated in favor of project
  PROJECT = 'COLLECTION',
  FLOW = 'FLOW',
  RUN = 'RUN',
}

```

And then you need to change the create context store

```ts
export function createContextStore<SCOPE extends StoreScope>({ prefix, flowId, workerToken, runId, defaultScope }: { prefix: string, flowId: FlowId, workerToken: string, runId: FlowRunId,
defaultScope: SCOPE  }): Store<SCOPE> {
    return {
        async put<T>(key: string, value: T, scope = defaultScope): Promise<T> {
            const modifiedKey = createKey(prefix, scope, flowId, key, runId)
            await createStorageService({ workerToken }).put({
                key: modifiedKey,
                value,
            })
            return value
        },
        async delete(key: string, scope = defaultScope): Promise<void> {
            const modifiedKey = createKey(prefix, scope, flowId, key, runId)
            await createStorageService({ workerToken }).delete({
                key: modifiedKey,
            })
        },
        async get<T>(key: string, scope = defaultScope): Promise<T | null> {
            const modifiedKey = createKey(prefix, scope, flowId, key, runId)
            const storeEntry = await createStorageService({ workerToken }).get(modifiedKey)
            if (storeEntry === null) {
                return null
            }
            return storeEntry.value as T
        },
    }
}

function createKey<SCOPE extends StoreScope>(prefix: string, scope: SCOPE, flowId: FlowId, key: string, runId: FlowRunId): string {
    switch (scope) {
        case StoreScope.RUN:
            return prefix + 'run_' + runId + '/' + key
        case StoreScope.FLOW:
            return prefix + 'flow_' + flowId + '/' + key
        default:
            return prefix + key
    }
}

```

Now you need to pass <Omit<StoreScope, 'RUN'> when creating it for trigger and without for action, so you make it return a type that doesn't accept run in compilation time.

Sorry for bad explanation, If it's not clear Please let me know

wise-introvert: Quick question: since the generic in `createContextStore` function extends `StoreScope` and "RUN" in `StoreScope` is required,  wouldn't passing `Omit<StoreScope, 'RUN'>` throw an error?

abuaboud: You are correct. I would look for something like ActionStoreScope / TriggerStoreScope, or try to differentiate them. I wrote the code quickly above and didn't test it. 

So please take it with a grain of salt. Sorry.

wise-introvert: @abuaboud, can you please check my latest [commit](https://github.com/activepieces/activepieces/pull/4099/commits/0494a1bb67fb74af4c21e4fbd16eec133108862d) on this?

abuaboud: @wise-introvert I took look quick seems correct, I will try to give it a detailed review and merge it tmw :D

wise-introvert: Thanks a ton @abuaboud!

abuaboud (approved):

PR #4120, opened 2024-03-07
feat(dust): new action to reply to existing conversation

AdamSelene:
## What does this PR do?
Add action to reply to an existing conversation (possibly using a different agent)
Capture d’écran 2024-03-07 à 11 03 53

### DIFF
--- a/packages/pieces/community/dust/package.json
+++ b/packages/pieces/community/dust/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-dust",
-  "version": "0.0.1"
+  "version": "0.0.2"
 }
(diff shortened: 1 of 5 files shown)


abuaboud (approved):

abuaboud (changes requested):
Good job! Could you please bump the version in package.json?

I assume it's not a breaking change, so it should be changed to 0.0.1. If it is, then it should be 0.1.0.

For more information, you can visit: https://www.activepieces.com/docs/developers/piece-reference/piece-versioning

AdamSelene:
@abuaboud done! Thanks :)

PR #4121, opened 2024-03-07
feat(dust): add action for upserting a document into a database

AdamSelene:
## What does this PR do?
Allow upserting a document into a Dust data source (so it can be indexed and used by assistants)

⚠️ **depends on #4120 (stacked PR)**

Capture d’écran 2024-03-07 à 13 33 03

### DIFF
--- a/packages/pieces/community/dust/package.json
+++ b/packages/pieces/community/dust/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-dust",
-  "version": "0.0.1"
+  "version": "0.0.2"
 }
--- a/packages/pieces/community/dust/src/lib/actions/upsert-document.ts
+++ b/packages/pieces/community/dust/src/lib/actions/upsert-document.ts
@@ -1,0 +1,74 @@
+import { createAction, Property } from '@activepieces/pieces-framework';
+import {
+  httpClient,
+  HttpMethod,
+  HttpRequest,
+} from '@activepieces/pieces-common';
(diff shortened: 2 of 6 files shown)


packages/pieces/community/dust/src/lib/actions/upsert-document.ts
```diff
@@ -0,0 +1,76 @@
...
+          source_url: propsValue.sourceUrl,
+          tags: tags,
+        },
+        (key, value) => (typeof value === 'undefined' ? null : value)
+      ),
+    };
+
+    console.log('TAGS', propsValue.tags);
+    const response = await httpClient.sendRequest(request);
+    console.log('RESPONSE', response);
```
abuaboud: nit: please remove console.log in general, for example response it might leaks headers which has the token

packages/pieces/community/dust/package.json
```diff
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-dust",
-  "version": "0.0.1"
+  "version": "0.0.3"
```
abuaboud: Nit: you can make this 0.0.2 and I will merge this pr directly instead of sequentially and github auto close the previous pr

abuaboud (changes requested):

packages/pieces/community/dust/src/lib/actions/upsert-document.ts
AdamSelene: 🤦 sorry forgot to cleanup

PR #4177, opened 2024-03-14
fix(piece-framework): axios http client to support array query parameter multiple times

MarceloRGonc:
## What does this PR do?

This PR add support for the same query parameter to be used multiple times in the same request. When an endpoint we are using has a query parameter that expects an array, generally the request looks like:

- '...?key=value1&key=value2'

We cannot support this using a Record object. So I changed it to a URLSearchParams object


abuaboud:
Hii @MarceloRGonc 

Good catch, approved

abuaboud (approved):

PR #4178, opened 2024-03-14
fix: trimming large payload of jsons

AbdullahBitar:
## What does this PR do?
Treats json as a tree and keeps trimming the leaves until the size of json is acceptable.

Fixes # (issue)


packages/engine/src/lib/helper/logging-utils.ts
```diff
@@ -121,3 +124,81 @@ const isObject = (obj: unknown): obj is Record<string, unknown> => {
 const bySizeDesc = (a: [string, unknown], b: [string, unknown]): number => {
     return sizeof(b[1]) - sizeof(a[1])
 }
+
+function trimJson(json: any){
```
abuaboud: You are adding function inside function that's bad practice.

packages/engine/src/lib/helper/logging-utils.ts
```diff
@@ -121,3 +124,81 @@ const isObject = (obj: unknown): obj is Record<string, unknown> => {
...
+    type Node = {
```
abuaboud: Move the type outside

packages/engine/src/lib/helper/logging-utils.ts
```diff
@@ -1,123 +1,95 @@
...
+            }
         }
+        
+        leaves.sort((a, b) => a.size - b.size)
     }
-
-    if (!objectExceedMaxSize(obj)) {
-        return obj
+    
+    let cnt = 0
```
abuaboud: This is should be local variable inisde the functiion

packages/engine/src/lib/helper/logging-utils.ts
```diff
@@ -1,123 +1,95 @@
...
-            case 'empty':
-                for (const key in iteration) {
-                    trimmedIteration[key] = await trimStepOutput(iteration[key], mode)
-                }
-                break
-            case 'trim':
-                trimmedIteration = await loggingUtils.trimExecution(iteration)
-                break
+    
+    const nodes: Node[] = []
```
abuaboud: When you making nodes global, you are making jsonToNode and other functions not pure as they depend on out nodes which life cycle is not controlled by them

packages/engine/src/lib/helper/logging-utils.ts
```diff
@@ -1,123 +1,95 @@
...
+    const leaves: Node[] = []
+    let totalJsonSize = sizeof(json)
+    
+    function jsonToNodes(curNode: unknown, curNodeId: number) {
+        if(curNode !== null && typeof curNode == "object" ) {
```
abuaboud: Try rename this expression to const isObject or create function for example :P

https://refactoring.guru/consolidate-conditional-expression

abuaboud (changes requested):

packages/engine/src/lib/helper/logging-utils.ts
abuaboud: This should have better name than `cnt`

abuaboud (approved):

PR #4227, opened 2024-03-20
feat(pdf): parse pdf

nyamkamunhjin:
## What does this PR do?

This adds a new piece to ActivePieces. It downloads a PDF file from the given URL and extracts its content to text.


packages/pieces/community/pdf-to-text/src/lib/actions/pdf-to-text.ts
```diff
@@ -0,0 +1,24 @@
...
+  description: 'Parses texts from PDF file',
+  props: {
+    file: Property.LongText({
+      displayName: 'PDF File URL',
+      required: true,
+    }),
+  },
+  async run(context) {
+    // Action logic here
+    console.dir(context, { depth: null });
```
abuaboud: Nit: please remove this, we shouldn't log props value they might contains sensitive values

packages/pieces/community/pdf-to-text/project.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "pieces-pdf-to-text",
```
abuaboud: Can you regenerate the piece and call it `PDF` we can extend it to many other actions.

packages/pieces/community/pdf-to-text/src/lib/actions/pdf-to-text.ts
```diff
@@ -0,0 +1,24 @@
+import { createAction, Property } from '@activepieces/pieces-framework';
+import { readPdfText } from 'pdf-text-reader';
+
+export const parseTextFromPdfFile = createAction({
+  // auth: check https://www.activepieces.com/docs/developers/piece-reference/authentication,
+  name: 'parseTextFromPdfFile',
+  displayName: 'Parse text from PDF file',
+  description: 'Parses texts from PDF file',
+  props: {
+    file: Property.LongText({
```
abuaboud: Let use Property.File

It can parse multiple format (url / base64 / files passed by other steps) into buffer 

https://www.activepieces.com/docs/developers/piece-reference/properties#file

abuaboud (changes requested):

packages/pieces/community/pdf-to-text/project.json
nyamkamunhjin: Deleted current piece regenerated a new one named `pdf`

packages/pieces/community/pdf-to-text/src/lib/actions/pdf-to-text.ts
nyamkamunhjin: Separated the current action into two pieces that parses from PDF file and PDF url

packages/pieces/community/PDF/src/lib/actions/parse-pdf-from-url.ts
```diff
@@ -0,0 +1,23 @@
+import { createAction, Property } from '@activepieces/pieces-framework';
+import { readPdfText } from 'pdf-text-reader';
+
+export const parsePdfFromUrl = createAction({
+  // auth: check https://www.activepieces.com/docs/developers/piece-reference/authentication,
+  name: 'parsePdfFromUrl',
+  displayName: 'Parse text from PDF URL',
+  description: 'Downloads PDF file from internet and parses the text',
+  props: {
+    'file-url': Property.ShortText({
```
abuaboud: @nyamkamunhjin Sorry for my bad explanation, Property.File does accept urls too, out of the box too. 

So no need for this action.

abuaboud (approved):
I renamed PDF -> pdf, I removed the second action as Property.File accept both file and url, and created icon and updated the url.

abuaboud:
also removed the auto retry on error handling option, as it's not api call

abuaboud:
@allcontributors please add @nyamkamunhjin for plugin

Good job <3

PR #4232, opened 2024-03-20
fix(gmail): add missing trigger

AdamSelene:
## What does this PR do?

The "new email" trigger already exists but is not referenced


abuaboud:
Read is sensitive scope for google which require hard approval for us on cloud, although it's easy if you specify your own google credentials, that's why we are separating into another piece.

https://github.com/activepieces/activepieces/pull/4544

PR #4279, opened 2024-03-26
feat: increase sidenav caption width

takaram:
## What does this PR do?
When you switch the language to Japanese, you will notice that the word "コネクション" (connection) in the sidenav is wrapped into four lines.
While in English a single word cannot be wrapped by default, in some languages like Japanese this is not the case.

This PR increases the width of the captions in the sidenav to minimize the word wrapping.


abuaboud:
Thank you @takaram , @AbdulTheActivePiecer can you please take look

AbdulTheActivePiecer:
Well done, I removed a little bit of extra padding as well to it your word fully on one line, cheers 😄  @takaram

PR #4285, opened 2024-03-27
feat: sync pieces from cloud to db

abuaboud:
## What does this PR do?

- Add Sync Button in Pieces Table
- Add Sync endpoint
- Add Versions endpoint to sync
- Deprecated cloud_and_db
- Add service to sync pieces, first we call list and check if latest version is inserted if not we start inserting from oldest to newest if they don't exist

Fixes # (issue)


packages/ee/ui-platform/src/lib/pages/pieces-table/pieces-table.component.html
```diff
@@ -1,26 +1,37 @@
-<ap-upgrade-note [featureNoteTitle]="upgradeNoteTitle" [featureNote]="upgradeNote" videoUrl="https://cdn.activepieces.com/videos/showcase/pieces.mp4"
-    *ngIf="isDemo"></ap-upgrade-note>
+<ap-upgrade-note [featureNoteTitle]="upgradeNoteTitle" [featureNote]="upgradeNote"
+    videoUrl="https://cdn.activepieces.com/videos/showcase/pieces.mp4" *ngIf="isDemo"></ap-upgrade-note>
```
khaledmashaly, line 2: This link is better moved to the component class.

packages/ee/ui-platform/src/lib/pages/pieces-table/pieces-table.component.html
```diff
@@ -138,4 +151,5 @@
     <ng-container *ngIf="toggelCloudOAuth2$ | async"> </ng-container>
     <ng-container *ngIf="dialogClosed$ | async"></ng-container>
     <ng-container *ngIf="addPackageDialogClosed$ | async"></ng-container>
+    <ng-container *ngIf="syncPieces$ | async"></ng-container>
```
khaledmashaly, line 154: Should this file be refactored to use the new flow control syntax?

packages/server/api/src/app/pieces/base-piece-module.ts
```diff
@@ -25,12 +26,60 @@ import {
...
+        config: {
+            allowedPrincipals: ALL_PRINCIPAL_TYPES,
+        },
+        schema: {
+            params: Type.Object({
+                name: Type.String(),
+                scope: Type.String(),
+            }),
+            querystring: ListVersionRequestQuery,
+        },
```
khaledmashaly: Can you please extract the request in a variable and give it a declarative name? These endpoints are not standard so we need some way of describing the endpoint.

packages/server/api/src/app/pieces/base-piece-module.ts
```diff
@@ -25,12 +26,60 @@ import {
...
+        const { name } = req.params
+        const decodedName = decodeURIComponent(name)
+        return pieceMetadataService.getVersions({
+            name: decodedName,
+            projectId: req.principal.projectId,
+            release: req.query.release,
+            edition: req.query.edition ?? ApEdition.COMMUNITY,
+            platformId: req.principal.type === PrincipalType.UNKNOWN ? undefined : req.principal.platform.id,
+        })
+    })
```
khaledmashaly, line 46: Can we merge these two endpoints into one? From what I understand they are the same endpoint but one for scoped packages and one for normal packages. If the problem is with path params we can just use a query param or body param instead.

packages/server/api/src/app/pieces/piece-metadata-service/index.ts
```diff
@@ -17,11 +16,10 @@ const initPieceMetadataService = (): PieceMetadataService => {
     const source = system.getOrThrow<PiecesSource>(SystemProp.PIECES_SOURCE)
     switch (source) {
         case PiecesSource.DB:
+        case PiecesSource.CLOUD_AND_DB:
```
khaledmashaly, line 19: Nice! This will allow the feature not to introduce breaking changes.

packages/server/api/src/app/pieces/piece-sync-service.ts
```diff
@@ -0,0 +1,108 @@
...
+                }
+                catch (error) {
+                    logger.error({ error, name: summary.name }, 'Error syncing piece')
+                }
+            }
+        }
+        catch (error) {
+            logger.error({ error }, 'Error syncing pieces')
+        }
+    },
```
khaledmashaly, line 58: Please note that this bit here could have all sorts of race conditions, I suggest using a lock here, Also to allow customers to bring multiple instances of the API server if they require so.

khaledmashaly (commented):
Good! Here are some comments for you.

packages/ee/ui-platform/src/lib/pages/pieces-table/pieces-table.component.html
abuaboud: It's lint and not inside the scope of the task, let's me do it in related pr

packages/ee/ui-platform/src/lib/pages/pieces-table/pieces-table.component.html
abuaboud: I think we should migrate package by package but i didn't want to introduce it in this pr

packages/server/api/src/app/pieces/base-piece-module.ts
abuaboud: Done and all other endpoints

packages/server/api/src/app/pieces/base-piece-module.ts
abuaboud: Done, made it one endpoint with query param since the name can contain /

PR #4318, opened 2024-03-31
chore(cli): generate starter code for new piece actions

### LINKED ISSUE #4252: Scale up connectors MVP
We should start experimenting with how to utilize the CLI to generate the code required for the pieces, and then use OpenAI to fill most of the fields using outside knowledge such as OpenAPI specs and other open-source connectors.

[Proof of Concept to scale up pieces from (https://github.com/microsoft/PowerPlatformConnectors)](https://github.com/microsoft/PowerPlatformConnectors)

My suggestion for the first step is to create a CLI that asks users for the properties when they create an action. Then, generate a boilerplate with specific comments in the body. After that, we can instruct OpenAI to fill in these code-specific comment areas.

khaledmashaly:
## What does this PR do?

Closes #4252

### DIFF
--- a/packages/cli/src/lib/commands/create-action.ts
+++ b/packages/cli/src/lib/commands/create-action.ts
@@ -1,12 +1,14 @@
 import { writeFile } from 'node:fs/promises';
+import { execSync } from 'node:child_process';
 import chalk from 'chalk';
 import { Command } from 'commander';
 import inquirer from 'inquirer';
 import { displayNameToCamelCase, findPieceSourceDirectory, displayNameToKebabCase } from '../utils/piece-utils';
 import { checkIfFileExists, makeFolderRecursive } from '../utils/files';
 import { join } from 'node:path';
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/commands/create-action.ts
```diff
@@ -67,16 +76,34 @@ export const createActionCommand = new Command('create')
...
-        name: 'actionName',
+        name: 'actionDisplayName',
         message: 'Enter the action display name',
       },
       {
         type: 'input',
         name: 'actionDescription',
         message: 'Enter the action description',
+      },
+      {
```
abuaboud, line 87: I think we should have flag to turn on the `OpenAI` otherwise we will break everyone experience and the tutorial.

Since this is in beta, and we need proper documentation for that but I will leave it for another pr, when doing the cli.

abuaboud (changes requested):
Please check my comments

PR #4324, opened 2024-04-01
fix: make json editor theme same as all the others

AbdulTheActivePiecer:


abuaboud (changes requested):
Then delete defineTheme cobalt and see if it can be removed from the assets and bundle

PR #4325, opened 2024-04-01
feat: display error invalid template

### LINKED ISSUE #4311: Show error on Importing Template
When implementing invalid templates/json or when the server replies with a 400 error, it gets stuck on the loading screen. It should display an error.

https://github.com/activepieces/activepieces/blob/84b4aadde7c0d64cf2e590919a7624a13398a86b/packages/ui/common/src/lib/components/dialogs/import-flow-dialog/import-flow-dialog.component.ts#L70

There should be a catchError here with a snackbar.

Example:
https://github.com/activepieces/activepieces/blob/84b4aadde7c0d64cf2e590919a7624a13398a86b/packages/ui/common/src/lib/guards/user-logged-in.guard.ts#L124

shravankshenoy:
Fixes #4311 

## Things to Consider 
Initially had written the code below

```
catchError((err: HttpErrorResponse) => {
   const status = err?.status;
   if (status === StatusCodes.BAD_REQUEST) {
     this.snackBar.open($localize`The uploaded template is invalid`);
     this.loading = false;
     }
   return of(void 0);
})
```

However when I tried different json files, sometimes instead of an HTTPErrorResponse, it was giving a TypeError, as shown below:

image

Hence instead of looking for any particular type of error, I decided to just show the message as long as we get any error. Hope that is ok.

### DIFF
--- a/packages/ui/common/src/lib/components/dialogs/import-flow-dialog/import-flow-dialog.component.ts
+++ b/packages/ui/common/src/lib/components/dialogs/import-flow-dialog/import-flow-dialog.component.ts
@@ -71,12 +71,13 @@ export class ImportFlowDialogComponent {
           projectId: this.data.projectId,
         }).pipe(
           switchMap((flow) => this.importFlow(flow.id, template, false)),
-          catchError((err) => {            
+          catchError((err) => {
+            console.error(err);
             this.snackBar.open($localize`The uploaded template is invalid`);
-            this.loading = false;            
+            this.loading = false;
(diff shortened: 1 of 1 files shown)


abuaboud:
@allcontributors please add @shravankshenoy for code

You are awesome!

AbdulTheActivePiecer:
Hi @shravankshenoy  @shravankshenoy, thanks for your contribution, I would usually add a console.error to show people what the error is.

Issue #4333, opened 2024-04-02
[BUG]: Connection Description Rendered in Incorrect Location for Pieces with CustomAuth


kishanprmr:
**Describe the bug**

**To Reproduce**
1. Create a flow with any of the following pieces or any piece with Custom Authentication type, such as ActiveCampaign, NocoDB, or Shopify.
2. Select any action and observe the text rendering below the connection component.

**Expected behavior**
Steps to create connection should only be shown in the create connection popup.

**Screenshots**

![image](https://github.com/activepieces/activepieces/assets/135701940/d1fbef13-c15f-4f76-9ce2-7da5d78354bd)
![image](https://github.com/activepieces/activepieces/assets/135701940/72af54a3-99bb-434e-8c4d-6e00cf47bb62)

**Additional context**

PR #4335, opened 2024-04-02
fix: only show connection description inside new/edit connection dialogs

AbdulTheActivePiecer:


AbdulTheActivePiecer:
closes #4333

PR #4382, opened 2024-04-07
feat: add community rewards information in dashboard 

AbdulTheActivePiecer:


packages/server/api/.env
```diff
@@ -15,4 +15,3 @@ AP_TELEMETRY_ENABLED=false
 AP_TRIGGER_DEFAULT_POLL_INTERVAL=1
 AP_WEBHOOK_TIMEOUT_SECONDS=30
 AP_WEBHOOK_URL="http://localhost:3000"
-AP_PIECES_SYNC_MODE=NONE
```
abuaboud, line 18: Revert

packages/ui/feature-dashboard/src/lib/components/sidenav-routes-list/sidenav-routes-list.component.ts
```diff
@@ -6,36 +6,62 @@ import {
...
 import { ApFlagId, ProjectMemberRole, supportUrl } from '@activepieces/shared';
 import { DashboardService, FlagService } from '@activepieces/ui/common';
+import { SidenavRouteItemComponent } from '../sidenav-route-item/sidenav-route-item.component';
+import { CommonModule } from '@angular/common';
 
 type SideNavRoute = {
   icon: string;
   caption: string;
-  route: string;
+  route?: string;
```
abuaboud: can you change this to `route: string | undefined`
instead of optional, so the at least we will be aware when it's not set in the route.

abuaboud (changes requested):

abuaboud (dismissed):

PR #4402, opened 2024-04-09
fix: wasn't able to attach a debugger to the running process

cr0fters:
## What does this PR do?

It looks like the folder structure was changed in February to rename backend to server/api. This meant that when you try and attach the debugger in vscode it was giving an error (directory not found)

Fixes # (issue)


abuaboud:
Thank you @cr0fters

Issue #4460, opened 2024-04-13
Send Announcement / Notifications 


abuaboud:
In this ticket in the admin console, we should create upgrade tab in admin console -> settings the first version should have current version and latest version.

Both of these information is available at flag service in the frontend.

abuaboud:
![Image](https://github.com/activepieces/activepieces/assets/1812998/c339f275-0fb5-4069-a58f-fbf1cb7b9ea8)

PR #4467, opened 2024-04-14
feat: show updates in admin console

islamaf:
## What does this PR do?

Implementation of feature #4460

## Things done in this PR
* Create upgrade component 
* Add upgrade component to platform settings


packages/ee/ui-platform/src/lib/components/upgrades/upgrades.component.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Observable, map, combineLatest } from 'rxjs';
+import { PlatformSettingsBaseComponent } from '../platform-settings-base.component';
+import { ApFlagId } from '@activepieces/shared';
+import { AsyncPipe } from '@angular/common';
+
+const compareVersions = (latestVersion: string, currentVersion: string) => {
+  const latest = latestVersion.split('.').map(Number);
+  const current = currentVersion.split('.').map(Number);
+
+  if (latest[1] > current[1]) {
```
abuaboud: This is great but currently we are under < 1.0.0 but this will break in future.

The best thing to do is use `semver` package

```typscript
const compareVersions = (
  latestVersion: string,
  currentVersion: string
): VersionComparisonResult => {
  if (semver.gt(latestVersion, currentVersion)) {
    const diff = semver.diff(latestVersion, currentVersion);
    let message = '';
    if (diff === 'minor'  || diff == 'major') {
      message = 'Major update is available. Please update.';
    } else if (diff === 'patch') {
      message = 'Patch update is available. Please update.';
    }
    return {
      emoji: '😞',
      message: message,
    };
  }

  return {
    emoji: '🙂',
    message: 'Up to date!',
  };
};
```

something along these line

islamaf: Yes! Makes sense. Will work on the change now.

Issue #4459, opened 2024-04-13
Self Serve Adding Custom Domain on Cloud


abuaboud:

abuaboud:
Done by @islamaf

PR #4474, opened 2024-04-15
feat: self serve custom domains

islamaf:
## What does this PR do?

Feature [#4459](https://github.com/activepieces/activepieces/issues/4459)

Allowing self serving custom domains by creating, reading, updating and deleting them.

## What is in this PR?
* Using Cloudflare API with the CRUD operations for custom domains.
* Verifying domain with Cloudflare API.
* Adding SSL to create request body.
* Custom domain error.


packages/server/shared/src/lib/system/system-prop.ts
```diff
@@ -1,4 +1,7 @@
 export enum SystemProp {
+    AUTH_EMAIL = 'AUTH_EMAIL',
```
abuaboud: CLOUDFLARE_AUTH_EMAIL
CLOUDFLARE_ZONE_ID
CLOUDFLARE_API_KEY

packages/server/api/src/app/ee/custom-domains/custom-domain.service.ts
```diff
@@ -33,17 +70,61 @@ export const customDomainService = {
...
+    async verifyDomain(request: {
         platformId: string
-    }): Promise<CustomDomain> {
-        const customDomain = customDomainRepo.create({
-            id: apId(),
-            domain: request.domain,
+        id: string
+    }): Promise<boolean> {
+        const customDomain = await customDomainRepo
+            .createQueryBuilder('custom_domain')
```
abuaboud: Use findOneBy

packages/server/api/src/app/ee/custom-domains/custom-domain.service.ts
```diff
@@ -33,17 +70,61 @@ export const customDomainService = {
...
+            }
+
+            return await customDomainRepo.save(customDomain)
+        }
+        catch (e) {
+            logger.error(e)
+            throw new ActivepiecesError({
+                code: ErrorCode.CUSTOM_DOMAIN_FAILED,
+                params: {
+                    message: 'Failed to create a custom domain. Make sure you are not using domains like *.example.com and *.example.net.',
```
abuaboud: Let's remove the try -> catch, If the error is unexpected, it will return 500 and we can monitor 5xx 

Now If you are afraid of the user enters invalid custom domain, you can use the request model to add validation using sinclair that way we will fail fast even before sending to cloudflare.

packages/server/api/src/app/ee/custom-domains/custom-domain.service.ts
```diff
@@ -1,23 +1,60 @@
...
+                    await cloudflareHostnameServices.delete(hostnameDetails.data.result.id)
+                }
+            }
+            
+            await customDomainRepo.delete({
+                id: request.id,
+                platformId: request.platformId,
+            })
+        }
+        catch (e) {
```
abuaboud: Remove this for same reason below

packages/server/api/src/app/ee/custom-domains/custom-domain.service.ts
```diff
@@ -1,23 +1,60 @@
...
 
 export const customDomainService = {
     async delete(request: { id: string, platformId: string }): Promise<void> {
-        await customDomainRepo.delete({
-            id: request.id,
-            platformId: request.platformId,
-        })
+        try {
+            const edition = getEdition()
+            if (edition === ApEdition.CLOUD) {
```
abuaboud: We need to allow people other than cloud to delete custom domain but we just want to avoid calling cloudflare

packages/ee/shared/src/lib/custom-domains/index.ts
```diff
@@ -18,6 +18,13 @@ export type CustomDomain = Static<typeof CustomDomain>;
 
 export const AddDomainRequest = Type.Object({
     domain: Type.String(),
+    ssl: Type.Object({
```
abuaboud: Let's remove this as we will provide static one in the backend same as cloudflare user interface

packages/ee/ui/platform/src/lib/components/dialogs/create-custom-domain-dialog/create-custom-domain-dialog.component.ts
```diff
@@ -50,6 +50,14 @@ export class CreateCustomDomainDialogComponent {
       this.addCustomDomain$ = this.customDomainService
         .create({
           domain: rawData.domain,
+          ssl: {
+            // test params
```
abuaboud: If above removed this will also get removed

packages/server/api/src/app/ee/custom-domains/custom-domain.service.ts
```diff
@@ -1,23 +1,60 @@
...
 import {
     CustomDomain,
     CustomDomainStatus,
     ListCustomDomainsRequest,
 } from '@activepieces/ee-shared'
-import { apId, SeekPage } from '@activepieces/shared'
+import { logger } from '@activepieces/server-shared'
+import { ActivepiecesError, ApEdition, apId, ErrorCode, SeekPage } from '@activepieces/shared'
+
+export type SSLParams = {
```
abuaboud: Same for this one, as it will be inline in the axios, and we need to remove it from all methods

abuaboud (changes requested):

packages/server/api/src/app/ee/custom-domains/custom-domain.service.ts
islamaf: Yup, it checks if edition is cloud, if yes then it will start deleting from cloudflare then from DB, otherwise it will skip cloudflare and go delete from DB directly

abuaboud (approved):

PR #4479, opened 2024-04-16
feat(google-drive): Move file and set file as public action 

pfernandez98:
## What does this PR do?

this adds the capability for the user to set a file or folder as public in their google drive and get the download URL.

https://community.activepieces.com/t/google-drive-update-the-edit-permissions-action/3798


packages/pieces/community/google-drive/src/lib/action/set-public-access.ts
```diff
@@ -0,0 +1,40 @@
...
+      type: 'anyone',
+    };
+    const res = await drive.permissions.create({
+      fileId: fileId,
+      requestBody: permission,
+    });
+
+    const file = await drive.files.get({
+      fileId: fileId,
+      fields: 'webContentLink',
```
kishanprmr: @pfernandez98 The files.get response doesn't contain any webContentLink. Is it necessary in action output?

pfernandez98: Actually, It's needed for the "downloadUrl" field I return with the response @kishanprmr

kishanprmr (approved):

PR #4536, opened 2024-04-23
feat(linear): enrich actions output

AdamSelene:
## What does this PR do?

Enrich various Linear actions output to include the actual object (e.g. created issue, project, comment, etc.) instead of a useless internal ID

The output looks like this:
```
{
  lastSyncId: <timestamp>,
  success: true,
  _issue: {
    id: <UUID>
  },
  issue: {
    <actual object>
  }
}
```
I've kept the `_issue` proxy object to avoid breaking existing flows but it should probably be removed at some point

### DIFF
--- a/packages/pieces/community/linear/package.json
+++ b/packages/pieces/community/linear/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-linear",
-  "version": "0.0.5"
+  "version": "0.0.6"
 }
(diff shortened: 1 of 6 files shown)


abuaboud (changes requested):
Hi @AdamSelene 

Why not remove the proxy object and increase the minor version, the new flows will have to manual upgrade the pieces :D

AdamSelene:
@abuaboud done!

abuaboud (approved):

abuaboud (approved):

PR #4596, opened 2024-04-30
chore: add different colors for step output values depending on their types.

AbdulTheActivePiecer:


packages/ui/feature-builder-test-steps/src/lib/test-polling-trigger/test-polling-trigger.component.html
```diff
@@ -49,7 +49,7 @@
                 <div class="ap-typography-caption ap-text-description ap-my-2" i18n>
                     This sample data can be used in the next steps
                 </div>
-                <ap-json-viewer class="json-viewer" title="Output" 
+                <ap-json-viewer class="json-viewer" viewTitle="Output" i18n-viewTitle
```
abuaboud, line 52: i18n is broken here @AbdulTheActivePiecer

PR #4599, opened 2024-04-30
feat: piece branching

### LINKED ISSUE #2065: Approvals & Human Tasks
The current problems with Approvals:

- There is no Testing Experience
- The user has know they have to create two steps, first to create links then send it somewhere then add wait for approval piece, this is confusing.
- What happens when the user click on the link (What should it show?) what if the user click twice?

islamaf:
## What does this PR do?

Fixes #2065


packages/engine/src/lib/handler/piece-executor.ts
```diff
@@ -108,6 +108,10 @@ const executeAction: ActionHandler<PieceAction> = async ({ action, executionStat
                 url.search = new URLSearchParams(params.queryParams).toString()
                 return url.toString()
             },
+            generateApprovalUrl: (params) => {
```
abuaboud: Adding a method to the engine is irreversible, the resume URL is a generic method

packages/pieces/community/approval/src/lib/actions/wait-for-approval.ts
```diff
@@ -1,35 +1,55 @@
...
+        denied: boolean;
+      }>(request);
+
+      return {
+        approved: res.body.approved,
+        denied: res.body.denied,
+      };
+    } else {
+      return {
+        approved: ctx.resumePayload.queryParams['action'] === 'approve',
```
abuaboud: I don't get this part

abuaboud (changes requested):

packages/pieces/community/approval/src/lib/actions/wait-for-approval.ts
```diff
@@ -1,35 +1,55 @@
...
+  props: {},
+  errorHandlingOptions: {
+    continueOnFailure: {
+      hide: true,
+    },
+    retryOnFailure: {
+      hide: true,
+    },
+  },
+  outputs: ['approved', 'denied'],
```
abuaboud: outputs: [
{
   name: "approved"
},
{
  name: "approved
}
]

packages/pieces/community/framework/src/lib/context.ts
```diff
@@ -127,6 +127,9 @@ export type BaseActionContext<
   generateResumeUrl: (params: {
     queryParams: Record<string, string>
   }) => string;
+  generateApprovalUrl: (params: {
```
abuaboud: remove

packages/server/api/src/app/flows/flow/approval/approval.controller.ts
```diff
@@ -0,0 +1,25 @@
+import { FastifyPluginAsyncTypebox, Type } from '@fastify/type-provider-typebox'
+import { approvalService } from './approval.service'
+import { ALL_PRINCIPAL_TYPES, ApId } from '@activepieces/shared'
+
+
+export const approvalController: FastifyPluginAsyncTypebox = async (app) => {
```
abuaboud: remove

packages/server/api/src/app/flows/flow/approval/approval.service.ts
```diff
@@ -0,0 +1,41 @@
+import { flowVersionService } from '../../flow-version/flow-version.service'
+import { flowRepo } from '../flow.repo'
+import { ActivepiecesError, ApprovalResponse, ErrorCode, FlowId, isNil, PopulatedFlow } from '@activepieces/shared'
+
+// type ApprovalRequestParam = 'approved' | 'denied'
+
+const APPROVAL_PIECE_NAME = '@activepieces/piece-approval'
+
+export const approvalService = {
+    getApprovalByFlowIdOrThrow: async (flowId: string, action: string): Promise<ApprovalResponse> => {
```
abuaboud: remove

docs/developers/piece-reference/piece-branching.mdx
```diff
@@ -0,0 +1,57 @@
...
+          response: {},
+        },
+      });
+
+      // Return an empty Map while execution is paused
+      return branchedPieceResponse()
+    } else {
+      // Return a Map populated with the expected returned data
+      return branchedPieceResponse({
+        version: 'v1', // Optional: adding a version for the expected response. Defaults to 'v1'.
```
abuaboud, line 50: Move this to inside function, don't give it as option to the contributor, It's just for us to maintain the engine.

docs/developers/piece-reference/piece-branching.mdx
```diff
@@ -0,0 +1,57 @@
...
+  errorHandlingOptions: {
+    continueOnFailure: {
+      hide: true,
+    },
+    retryOnFailure: {
+      hide: true,
+    },
+  },
+  // Here we define the names of the outputs expected to be branched from the piece
+  outputs: [
```
abuaboud, line 32: Instead of using a name since you are using it as an id

We can change this to an object where the key is the name of the branch

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -190,6 +190,10 @@ export const Action = Type.Recursive(action => Type.Union([
     })]),
     Type.Intersect([PieceActionSchema, Type.Object({
         nextAction: Type.Optional(action),
+        children: Type.Optional(Type.Array(Type.Object({
```
abuaboud: Why not object where the key is the name and value is the actualon?

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -700,7 +706,13 @@ function removeAnySubsequentAction(action: Action): Action {
             delete clonedAction.firstLoopAction
             break
         }
-        case ActionType.PIECE:
+        case ActionType.PIECE: {
+            if (clonedAction.settings.outputs && clonedAction.settings.outputs.length > 1) {
+                delete clonedAction.onSuccessAction
```
abuaboud: Why? Piece action shouldn't have this property to start with

abuaboud (changes requested):

docs/developers/piece-reference/piece-branching.mdx
islamaf: I think the issue would be mainly with adding other properties to the branch in case we would need to add this later, if we would go with an object, I suggest this way:
```ts
{
    branchName: {
       expectedOutput: true,
       ...otherProperties,
    },
} 
```
by this way, we would be able to extend the functionality further in the future. What do you think?

abuaboud: map of objects work

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -58,6 +58,9 @@ export const PieceActionSettings = Type.Object({
     actionName: Type.Optional(Type.String({})),
     input: Type.Record(Type.String({}), Type.Any()),
     inputUiInfo: SampleDataSettingsObject,
+    outputs: Type.Optional(Type.Array(Type.Object({
```
abuaboud: Editing flow definition is bad idea and this information is already duplicated.

You can use action definition in the piece to obtain the branch names.  

image

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -190,6 +193,10 @@ export const Action = Type.Recursive(action => Type.Union([
     })]),
     Type.Intersect([PieceActionSchema, Type.Object({
         nextAction: Type.Optional(action),
+        children: Type.Record(
+            Type.String({}),
+            Type.Union([action, Type.Undefined()]),
```
abuaboud: `action` as value should be  enough

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -65,6 +66,23 @@ function deleteAction(
             parentStep.nextAction = stepToUpdate.nextAction
         }
         switch (parentStep.type) {
+            case ActionType.PIECE: {
+                if (
```
abuaboud: The frontend should tell us the branch name (which is the same as what is defined in the outputs in the piece), and we should update the action in children map. We should never hardcode 'onFailureAction' and 'onSuccessAction'.

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -349,6 +403,22 @@ function updateAction(
                 parentStep.firstLoopAction = createAction(request, actions)
             }
         }
+        if (isPieceBranched(parentStep)) {
```
abuaboud, line 418: Same comment as above this is hardcoded.

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -245,10 +294,15 @@ function getAllStepsAtFirstLevel(step: Trigger): (Action | Trigger)[] {
     }
     return steps
 }
-function getAllChildSteps(action: LoopOnItemsAction | BranchAction): Action[] {
+function getAllChildSteps(action: LoopOnItemsAction | BranchAction | PieceAction): Action[] {
     switch (action.type) {
         case ActionType.LOOP_ON_ITEMS:
             return traverseInternal(action.firstLoopAction) as Action[]
+        case ActionType.PIECE: 
+            return [
```
abuaboud: This is hardcoded, It can be rewritten to 

```ts
        case ActionType.PIECE: {
            const children: Action[] = []
            Object.values(action.children).forEach((child) => {
                const childSteps: Action[] = traverseInternal(child) as Action[]
                children.push(...childSteps)
            })
            return children
        }
```

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -499,6 +570,35 @@ function addAction(
                 )
             }
         }
+        else if (isPieceBranched(parentStep)) {
+            if (request.stepLocationRelativeToParent === StepLocationRelativeToParent.INSIDE_TRUE_BRANCH) {
```
abuaboud: This is hardcoded again, the frontend should just send the branch name and we should update the corresponding key in chidlren object

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -918,9 +1053,18 @@ function isStepLastChildOfParent(child: Step, trigger: Trigger): boolean {
...
 }
 
-function doesStepHaveChildren(step: Step): step is LoopOnItemsAction | BranchAction {
-    return step.type === ActionType.BRANCH || step.type === ActionType.LOOP_ON_ITEMS
+function doesStepHaveChildren(step: Step): step is LoopOnItemsAction | BranchAction | PieceAction {
+    return step.type === ActionType.BRANCH || step.type === ActionType.LOOP_ON_ITEMS || isPieceBranched(step)
+}
+
+function isPieceBranched(step: Step): step is PieceAction {
+    if (step.settings.outputs) {
```
abuaboud: You should depend on the piece definition  rather the piece settings.

packages/ui/canvas-utils/src/lib/drawing/branch-drawer.ts
```diff
@@ -26,9 +29,17 @@ export class BranchDrawer {
...
-  static handleBranchAction(branchStep: BranchAction): FlowDrawer {
+  static handleBranchAction(
+    branchStep: BranchAction | PieceAction
+  ): FlowDrawer {
     let resultDrawer = FlowDrawer.construct(undefined);
-    const actions = [branchStep.onSuccessAction, branchStep.onFailureAction];
+    const actions =
+      branchStep.type === ActionType.BRANCH
+        ? [branchStep.onSuccessAction, branchStep.onFailureAction]
+        : [
```
abuaboud: This is hardcoded this should be just Object.values(children)

packages/ui/canvas-utils/src/lib/drawing/branch-drawer.ts
```diff
@@ -117,22 +127,38 @@ export class BranchDrawer {
     };
   }
 
+  private static doesBranchHaveChildren(
```
abuaboud, line 130: This is unreadable and hardcoded.

It can be written like that, but again If we start depend on the branch name as key instead of `stepLocationRelativeToParent` It will be much better
```ts
  private static doesBranchHaveChildren(
    branchStep: BranchAction | PieceAction,
    stepLocationRelativeToParent: StepLocationRelativeToParent
  ) {
    switch (branchStep.type) {
      case ActionType.BRANCH:
        switch (stepLocationRelativeToParent) {
          case StepLocationRelativeToParent.INSIDE_TRUE_BRANCH:
            return !!branchStep.onSuccessAction;
          case StepLocationRelativeToParent.INSIDE_FALSE_BRANCH:
            return !!branchStep.onFailureAction;
        }
        break;
      case ActionType.PIECE:
        return !isNil(branchStep?.children[BRANCH_NAME]);
      default:
        throw new Error('Invalid branch type');
    }
  }
  ```

File: packages/engine/src/lib/handler/piece-executor.ts [hunk 20]
```diff
@@ -144,6 +166,52 @@ const executeAction: ActionHandler<PieceAction> = async ({ action, executionStat
...
+async function runBranchablePieceWithVersion({ 
+    pieceOutput,
+    executionState, 
+    action, 
+    constants,
+    stepOutput, 
+    version = 'v1', 
+}: RunBranchablePieceWithVersion): Promise<FlowExecutorContext> {
+    const versions = {
+        v1: async (): Promise<FlowExecutorContext> => {
```
abuaboud, line 182: This is disaster, If approval piece is changed that mean we need to change the engine and make engine doesn't work at all.

We have documented how branching should work in the framework but this will not work at all as it require approved to be hardcoded and onSuccessAction 

It can be rewritten like that
```ts
async function runBranchablePieceWithVersion({
    pieceOutput,
    executionState,
    action,
    constants,
    stepOutput,
    version = 'v1',
}: RunBranchablePieceWithVersion): Promise<FlowExecutorContext> {
    const versions = {
        v1: async (): Promise<FlowExecutorContext> => {
            let newExecutionContext = executionState
            for (const [branchKey, _branchValue] of pieceOutput.output.entries()) {
                const childAction = action.children[branchKey]
                if (isNil(childAction)) {
                    continue
                }
                newExecutionContext = await flowExecutor.execute({
                    action: childAction,
                    executionState: newExecutionContext,
                    constants,
                })
                break;
            }
            return newExecutionContext.upsertStep(action.name, stepOutput
                .setOutput(pieceOutput.output))
                .increaseTask()
                .setVerdict(ExecutionVerdict.RUNNING, undefined)
        },
    }

    return versions[version]()
}
```

packages/pieces/community/approval/src/lib/actions/wait-for-approval.ts
```diff
@@ -1,35 +1,38 @@
...
+      ctx.run.pause({
+        pauseMetadata: {
+          type: PauseType.WEBHOOK,
+          response: {},
+        },
+      });
+
+      return branchedPieceResponse()
+    } else {
+      return branchedPieceResponse({
```
abuaboud: You are depending on the value to decide where to branch this is wrong, what if both were true.

What if i want to include information like it were approved by who? I think you should use key as to execute branch or not this can be rewritten like that

```ts
async function runBranchablePieceWithVersion({
    pieceOutput,
    executionState,
    action,
    constants,
    stepOutput,
    version = 'v1',
}: RunBranchablePieceWithVersion): Promise<FlowExecutorContext> {
    const versions = {
        v1: async (): Promise<FlowExecutorContext> => {
            let newExecutionContext = executionState
            for (const [branchKey, _branchValue] of pieceOutput.output.entries()) {
                const childAction = action.children[branchKey]
                if (isNil(childAction)) {
                    continue
                }
                newExecutionContext = await flowExecutor.execute({
                    action: childAction,
                    executionState: newExecutionContext,
                    constants,
                })
                break;
            }
            return newExecutionContext.upsertStep(action.name, stepOutput
                .setOutput(pieceOutput.output))
                .increaseTask()
                .setVerdict(ExecutionVerdict.RUNNING, undefined)
        },
    }

    return versions[version]()
}
```


Also you can change generate approval link to

```ts
 async run(ctx) {
    return {
      approveLink: ctx.generateResumeUrl({
        queryParams: { action: 'approve' },
      }),
      denyLink: ctx.generateResumeUrl({
        queryParams: { action: 'deny' },
      }),
    };
    ```

File: packages/pieces/community/approval/src/lib/actions/wait-for-approval.ts [hunk 22]
```diff
@@ -1,35 +1,38 @@
...
-      return {
-        approved: ctx.resumePayload.queryParams['action'] === 'approve',
-      };
-    }
-  },
-});
+import { createAction } from '@activepieces/pieces-framework';
+import { ExecutionType, PauseType, branchedPieceResponse } from '@activepieces/shared';
+
+export const waitForApprovalLink = createAction({
```
abuaboud, line 4: You will also have to bump `package.json`

packages/ui/feature-builder-right-sidebar/src/lib/input-forms/piece-input-form/piece-input-form.component.ts
```diff
@@ -415,6 +415,7 @@ export class PieceInputFormComponent extends InputFormCore {
                 ...selectedTriggerOrAction.props,
                 ...spreadIfDefined(AUTHENTICATION_PROPERTY_NAME, authProperty),
               },
+              outputs: selectedTriggerOrAction.outputs,
```
abuaboud: This has to be removed, It's duplicated information, you already have it in definition `selectedTriggerOrAction`

packages/ui/feature-builder-right-sidebar/src/lib/step-type-sidebar/step-type-sidebar.component.ts
```diff
@@ -383,6 +383,7 @@ export class StepTypeSidebarComponent implements AfterViewInit {
               pieceVersion:
                 flowItemDetails.extra?.pieceVersion ?? 'NO_APP_VERSION',
               actionName: suggestion?.name,
+              outputs: suggestion?.outputs,
```
abuaboud: This has to be removed, It's duplicated information, you already have it in definition `suggestion.outputs`

packages/ui/canvas-utils/src/lib/drawing/branch-drawer.ts
```diff
@@ -198,4 +224,17 @@ export class BranchDrawer {
...
   }
+
+  private static setLabel(
+    index: number,
+    step: BranchAction | PieceAction
+  ): string {
+    if (step.type === ActionType.BRANCH) {
+      return index === 0 ? $localize`True` : $localize`False`;
+    } else {
+      return index === 0
```
abuaboud: You already allow array in the framework, this will not work for the documentation you have wrote.

The children array / outputs array in the piece can be longer than 2 elements

abuaboud (changes requested):

PR #4619, opened 2024-05-03
fix: delay step were failing after loop

abuaboud:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/engine/src/lib/handler/context/flow-execution-context.ts
+++ b/packages/engine/src/lib/handler/context/flow-execution-context.ts
@@ -67,7 +67,8 @@ export class FlowExecutorContext {
             return undefined
         }
         assertEqual(stepOutput.type, ActionType.LOOP_ON_ITEMS, 'stepOutput.type', 'LOOP_ON_ITEMS')
-        return stepOutput as LoopStepOutput
+        // The new LoopStepOutput is needed as casting to LoopStepOutput will just cast the data but the class methods will not be available
+        return new LoopStepOutput(stepOutput as LoopStepOutput)
     }
 
     public isCompleted({ stepName }: { stepName: string }): boolean {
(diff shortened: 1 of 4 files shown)


packages/pieces/community/google-sheets/src/lib/triggers/new-or-updated-row.trigger.ts
```diff
@@ -107,7 +110,7 @@ export const newOrUpdatedRowTrigger = createTrigger({
     );
 
     const range =
-      triggerColumn === 'all_columns'
+      triggerColumn === ALL_COLUMNS
```
HoldYourWaffle, line 113: What happens when a column is called "all_columns"?
An array-based mechanism might be more appropriate here, then we'd also get multiple trigger columns for free.

packages/engine/src/lib/handler/context/flow-execution-context.ts
```diff
@@ -67,7 +67,8 @@ export class FlowExecutorContext {
             return undefined
         }
         assertEqual(stepOutput.type, ActionType.LOOP_ON_ITEMS, 'stepOutput.type', 'LOOP_ON_ITEMS')
-        return stepOutput as LoopStepOutput
+        // The new LoopStepOutput is needed as casting to LoopStepOutput will just cast the data but the class methods will not be available
```
HoldYourWaffle, line 70: I'm looking at this on a phone so pardon me if this is a dumb question, but why is it safe to cast to `LoopStepOutput` at all?

packages/pieces/community/google-sheets/src/lib/triggers/new-or-updated-row.trigger.ts
```diff
@@ -44,14 +47,14 @@ export const newOrUpdatedRowTrigger = createTrigger({
     trigger_column: Property.Dropdown({
       displayName: 'Trigger Column',
       description: `Trigger on changes to cells in this column only.Select **All Columns** if you want the flow to trigger on changes to any cell within the row.`,
```
HoldYourWaffle, line 49: Shouldn't these be translation keys?

PR #4663, opened 2024-05-10
feat(clickup): Assignee Dropdown for Create Comment Action

founderblocks-sils:
I've kept the original (weird) behaviour of setting the API user as assignee by default to not break past scenarios.

## What does this PR do?

Allow people to set the assignee when posting comments on clickup.


abuaboud:
@allcontributors please add @founderblocks-sils for plugin

abuaboud (approved):

abuaboud:
@founderblocks-sils Thank you a lot!

PR #4676, opened 2024-05-13
feat(formbricks): support self hosted instance

abuaboud:
## What does this PR do?

Fixes # (issue)


mantis64:
Still does not work under the latest build.  No survey's will load

PR #4700, opened 2024-05-15
feat(metabase): switch to new API key auth mechanism

AdamSelene:
## What does this PR do?

Metabase 0.49 introduced a new API key auth mechanism which is simpler and safer to use
https://www.metabase.com/releases/metabase-49#authenticate-api-requests-with-api-keys


abuaboud (approved):
That's awesome!

PR #4747, opened 2024-05-22
feat: allow retrieving platform info using API key

AdamSelene:
## What does this PR do?

We want to be able to get the list of allowed / denied pieces via an API call authenticated with an API key

**This has not been tested** (API keys are an `ee` feature which is harder to test on a dev setup 😬 )


abuaboud:
Merged to the main on another branch

PR #4763, opened 2024-05-24
feat: add new piece html-helper

pfernandez98:
## What does this PR do?

This adds a new piece for manipulating HTML. For now, it has 2 actions: 

- Get Element By Id
- Get Element by Query Selector

This needs a new dependency: 
```
"node-html-parser": "6.1.13"
```


abuaboud:
Hi @pfernandez98 


I think this is great piece, but I was thinking of scoping it to "Web scrapper" / "Web Extractor", I think most of the user will not find easy to fetch the html instead they should provide links, I think this idea to be refined, what do you think?

PR #4827, opened 2024-05-30
feat(cli): generate piece & auth from open api spec using open ai

saifmohamedsv:
### What does this CLI do currently?
1. Create a normal piece using the original method
2. Using an open AI function calling for information extraction
3. You can pass the URL for json open API spec or local path 
`eg: folder-name/openapi.json OR http://localhost:3000/openapi.json`
4. Generate piece authentication template OAuth2 / API Key
5. Generate a basic piece action from open API spec using open AI  function calling


abuaboud:
Hi @saifmohamed-dev 

Can you make it take the OPENAI key from the environment variable, currently it's empty in the code, can you also open docs to document it usage and I think we are good to go

saifmohamedsv:
Hi, @abuaboud
You can't push any secret keys to commits
Also normal `.env` way does not work for me I think you have a custom way to handle secrets which I don't know.

abuaboud:
@saifmohamedsv 

You can have them inline like that

```
ENV_VARIABLE='asd' npm run cli pieces scale http://localhost/slackOpenAPI.json
```

saifmohamedsv:
> @saifmohamedsv
> 
> You can have them inline like that, It's not necessarily to use .env
> 
> ```
> ENV_VARIABLE='asd' npm run cli pieces scale http://localhost/slackOpenAPI.json
> ```

@abuaboud OK Got it 
You can already use it now in the last commit I sat this up just run this command for example:
`AP_OPENAI_KEY='openai-api-key' npm run cli pieces scale openAPI/Slack.json`

abuaboud (approved):

PR #4851, opened 2024-06-03
chore: parameterise createCustomApiCallAction parameters to be able t…

hamedsh:
…o customize custom actions

## What does this PR do?

parameterize createCustomApiCallAction parameters to be able to customize custom actions.
adding some parameters to createCustomApiCallAction makes it possible to add custom actions without adding files.

### DIFF
--- a/packages/pieces/community/common/src/lib/helpers/index.ts
+++ b/packages/pieces/community/common/src/lib/helpers/index.ts
@@ -12,15 +12,19 @@ export const getAccessTokenOrThrow = (auth: OAuth2PropertyValue | undefined): st
   return accessToken;
 };
 
-export function createCustomApiCallAction({ auth, baseUrl, authMapping }: {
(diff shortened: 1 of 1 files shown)


abuaboud:
Hi @hamedsh 

Custom API Call is like standard in all pieces, I don't get what the reason behind this change do you mind sharing an example or what you trying to do?

hamedsh:
> Hi @hamedsh
> 
> Custom API Call is like standard in all pieces, I don't get what the reason behind this change do you mind sharing an example or what you trying to do?

for some services, like customer.io, there are 2 base URLs (track.customer.io/api/v1/ and api.customer.io/v1/). By using the provided change, it is possible to add these 2 Custom API Calls:
1. Custom Track API Call
2. Custom API Call

abuaboud:
@allcontributors please add @hamedsh for code

abuaboud:
@allcontributors please add @hamedsh for code

Issue #4854, opened 2024-06-03
Streamlining Premium Pieces


islamaf:
## Problem Statement
Handling the premium pieces in regards of deployment, showing them to end user and licensing.

## What needs to be done
* Make sure CI/CD works for premium pieces in deployment.
* Send the premium pieces in the pieces array on the Website **for the premium users**.
* Ability to specify them in the License.

PR #4855, opened 2024-06-03
feat: streamline premium pieces

islamaf:
## What does this PR do?

Fixes [#4854](https://github.com/activepieces/activepieces/issues/4854)

### DIFF
--- a/packages/server/api/src/app/database/migration/sqlite/1717443603235-AddPremiumPiecesColumnSqlite.ts
+++ b/packages/server/api/src/app/database/migration/sqlite/1717443603235-AddPremiumPiecesColumnSqlite.ts
@@ -1,0 +1,336 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+
+export class AddPremiumPiecesColumnSqlite1717443603235 implements MigrationInterface {
+    name = 'AddPremiumPiecesColumnSqlite1717443603235'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            ALTER TABLE "platform"
+            ADD "premiumPieces" character varying array
+        `)
+        await queryRunner.query(`
+            UPDATE "platform"
+            SET "premiumPieces" = '{}'
+        `)
+        await queryRunner.query(`
+            ALTER TABLE "platform"
+            ALTER COLUMN "premiumPieces" SET NOT NULL
+        `)
+        await queryRunner.query(`
+            CREATE TABLE "temporary_platform" (
+                "id" varchar(21) PRIMARY KEY NOT NULL,
+                "created" datetime NOT NULL DEFAULT (datetime('now')),
+                "updated" datetime NOT NULL DEFAULT (datetime('now')),
+                "ownerId" varchar(21) NOT NULL,
+                "name" varchar NOT NULL,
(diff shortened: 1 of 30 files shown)


packages/server/api/src/app/database/migration/sqlite/1717443603235-AddPremiumPiecesColumnSqlite.ts
```diff
@@ -0,0 +1,336 @@
...
+            ALTER TABLE "platform"
+            ADD "premiumPieces" character varying array
+        `)
+        await queryRunner.query(`
+            UPDATE "platform"
+            SET "premiumPieces" = '{}'
+        `)
+        await queryRunner.query(`
+            ALTER TABLE "platform"
+            ALTER COLUMN "premiumPieces" SET NOT NULL
```
abuaboud: I don't think this will work, the same issue existed in alerts hehe

tools/scripts/utils/piece-script-utils.ts
```diff
@@ -61,19 +62,42 @@ export async function findAllPiecesDirectoryInSource(): Promise<string[]> {
...
+    const piecesFolder = await findAllPremiumPiecesDirectory()
+    if (piecesFolder.length === 0) {
+        return []
+    }
+    for (const pieceFolder of piecesFolder) {
+        const projectJson = await readProjectJson(pieceFolder);
+        await exec('npx nx build ' + projectJson.name);
+    }
+
+    const enterprisePiecesPath = resolve(cwd(), 'dist', 'packages', 'ee', 'pieces')
```
abuaboud: Can you restrict them by edition, you don't want people to see them in the dev mode and think it okay to use them

packages/server/api/src/app/pieces/piece-metadata-service/piece-metadata-service.ts
```diff
@@ -27,6 +27,7 @@ type ListParams = {
     orderBy?: PieceOrderBy
     searchQuery?: string
     suggestionType?: SuggestionType
+    premiumPieces?: string
```
abuaboud: A better name would be includePremiumPiece and its type is boolean, i am thinking of prefer an enum over the boolean rule but i don't think it's applicable here

abuaboud (changes requested):

packages/server/api/src/app/pieces/piece-sync-service.ts
```diff
@@ -106,10 +107,17 @@ async function getOrThrow({ name, version }: { name: string, version: string }):
...
+    let premiumPieces: string[] | undefined
+    if (platformId) {
+        const platform = await platformService.getOne(platformId)
+        premiumPieces = platform?.premiumPieces
+    }
+
     const queryParams = new URLSearchParams()
     queryParams.append('edition', getEdition())
     queryParams.append('release', await flagService.getCurrentRelease())
+    premiumPieces && queryParams.append('premiumPieces', premiumPieces.join(','))
```
abuaboud: Don't do things like that, make it array, queryParams is smart enough to serialize array, actually that's the wrong way to serialize an array.

abuaboud (changes requested):

abuaboud (approved):

PR #4892, opened 2024-06-10
fix: show PT in created issues

islamaf:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/server/api/src/app/ee/helper/email/email-service.ts
+++ b/packages/server/api/src/app/ee/helper/email/email-service.ts
@@ -1,3 +1,4 @@
+import dayjs from 'dayjs'
 import { jwtUtils } from '../../../helper/jwt-utils'
 import { getEdition } from '../../../helper/secret-helper'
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/ee/helper/email/email-service.ts
```diff
@@ -56,11 +57,12 @@ export const emailService = {
         if (EDITION_IS_NOT_PAID) {
             return
         }
+        const createdAtFormatted = dayjs(createdAt.replace(' UTC', '')).tz('PT').format('DD MMM YYYY, HH:mm [PT]')
```
abuaboud: const createdAtFormatted = dayjs(createdAt).tz('PT').format('DD MMM YYYY, HH:mm [PT]')

Why not this?

islamaf: found a better way to do it, moved it to issues service instead!

abuaboud (approved):

PR #4901, opened 2024-06-11
feat: selectable status filter in runs and correct redirection on failed issues

islamaf:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
+++ b/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
@@ -124,11 +124,7 @@ export const flowRunService = {
             environment: RunEnvironment.PRODUCTION,
         })
         if (status) {
-            status.forEach((s, index) => {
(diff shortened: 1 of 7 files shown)


packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -121,9 +121,15 @@ export const flowRunService = {
         let query = flowRunRepo.createQueryBuilder('flow_run').where({
             projectId,
             ...spreadIfDefined('flowId', flowId),
-            ...spreadIfDefined('status', status),
             environment: RunEnvironment.PRODUCTION,
         })
+        if (status) {
+            status.forEach((s, index) => {
```
abuaboud: Use In operator it takes an array and native typeorm operation and Postgres

abuaboud (changes requested):

PR #4921, opened 2024-06-13
feat: suport false value when creating google-contact

ikus060:
## What does this PR do?

Using dynamic value for phoneNumber, companyName and email when creating Google Contact, if the dynamic value has a boolean value of false, it get converted into string "false" and the value created in Google Contact is "false".

I'm not sure this would be the expected result for users. I'm not sure of the policies regarding data conversion between types in ActivePieces either. So you are free to recommend me alternative implementation. I will do the same for the update-contact action I'm working on.

e.g.: in Google Contacts
image

e.g.: In Activepieces
image

### DIFF
--- a/packages/pieces/community/google-contacts/package.json
+++ b/packages/pieces/community/google-contacts/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-google-contacts",
-  "version": "0.3.6"
+  "version": "0.3.7"
 }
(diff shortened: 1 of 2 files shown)


ikus060:
Working on new implementation.

abuaboud:
I feel this is intended @ikus060 why not pass for example an empty / null string then It would skip phone number / email.

Not sure if i understand you correctly.

ikus060:
@abuaboud I come from Python world where False, None and Empty string are all equals. That might just be it.

abuaboud:
Hi @ikus060 

I think this will has side effect, I think it's better to try empty string or send undefined / null 

I am closing this PR, please let me know if the other way doesn't work

PR #4930, opened 2024-06-16
feat(reoon-verifier): single email and bulk email verification actions

AnneMariel95:
## What does this PR do?

This adds a new piece [Reoon Email Verifier](https://www.reoon.com/email-verifier/).

Actions:
- Verify a single email
- Verify bulk emails

> Note: The piece name is `reoon-verifier` because Reoon has several products. `Email` is omitted because `reoon-email-verifier` would be too long, and Reeon only has one product that is used for verification.


kishanprmr:
@allcontributors please add @AnneMariel95  for plugin

kishanprmr (approved):

PR #4979, opened 2024-06-24
feat: filter runs by issue start date

### LINKED ISSUE #4977: Filter Runs after Certain date (Issue First Seen)

islamaf:
## What does this PR do?

Fixes #4977


packages/ui/feature-dashboard/src/lib/pages/runs/runs.component.ts
```diff
@@ -173,7 +173,8 @@ export class RunsComponent
         FlowRunStatus.INTERNAL_ERROR,
         FlowRunStatus.QUOTA_EXCEEDED,
       ],
-      issue.flowId
+      issue.flowId,
+      new Date(issue.created.split('T')[0])
```
abuaboud: You are removing the timezone, I assume you are doing this to avoid the place where it set hour to 0 and end date hour to 23:59:59.

I suggest to rewrite this area.

islamaf: yes i do this to remove he time, however, it doesnt mess up the PT timezone becuase we add it in the email service separately, check attached photo
image

abuaboud (approved):

PR #4980, opened 2024-06-24
feat(captain-data): schedule workflow and get job results actions

AdamSelene:
## What does this PR do?

Add new integration with [Captain Data](https://www.captaindata.co) ([API doc](https://docs.captaindata.co/#intro))
- schedule a workflow
- retrieve results
- custom API call

### DIFF
--- a/packages/pieces/community/captain-data/src/lib/actions/launch-workflow.ts
+++ b/packages/pieces/community/captain-data/src/lib/actions/launch-workflow.ts
@@ -1,0 +1,68 @@
+import {
+  createAction,
+  DynamicPropsValue,
+  Property,
+} from '@activepieces/pieces-framework';
+import { httpClient, HttpMethod } from '@activepieces/pieces-common';
+import {
+  CAPTAIN_DATA_BASE_URL,
+  captainDataAuth,
+  CaptainDataAuthType,
+} from '../..';
+import { workflowProp } from '../common';
+
+export const launchWorkflow = createAction({
+  // auth: check https://www.activepieces.com/docs/developers/piece-reference/authentication,
+  name: 'launchWorkflow',
+  displayName: 'Launch a workflow',
+  description: '',
+  auth: captainDataAuth,
+  props: {
+    workflow: workflowProp,
+    jobName: Property.ShortText({
+      displayName: 'Job name',
+      required: false,
+    }),
+    inputs: Property.Json({
+      displayName: 'Inputs',
+      description: 'Inputs are the starting point of your workflow.',
+      defaultValue: [],
+      required: false,
+    }),
+    steps: Property.Json({
+      displayName: 'Steps',
+      description:
+        'You need to configure each steps with:\n' +
+        '- accounts : The identifiers of the accounts you want to use for this step. The accounts indicated must match the integration used in the step (e.g A Linkedin account for a Search Linkedin People Profile step). You can find the UIDs in the "Integrations\' tab on the platform by clicking on a specific account "Action" button (copy account UID).\n' +
+        '- accounts_rotation_enabled: (Optional) Whether or not you want to enable the Accounts Rotation feature for this step (only certain Linkedin & Outlook automations are applicable for this feature).\n' +
+        '- parameters: The specific parameters for this given step - (Can be empty but is required)\n' +
+        '- step_uid: The UID of the step to configure. You can find it in the API Playground, on Captain Data\'s platform by clicking on "view body to schedule".\n.' +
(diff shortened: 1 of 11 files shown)


AdamSelene:
Logo: 

03ab0ac731008522d253a02cb939c9fb7645f807cbfba4cd064866356b6e9a8e

packages/pieces/community/captain-data/src/lib/actions/launch-workflow.ts
```diff
@@ -0,0 +1,36 @@
...
+  description: '',
+  auth: captainDataAuth,
+  props: {
+    workflow: workflowProp,
+    payload: Property.Json({
+      displayName: 'Payload',
+      description:
+        'See https://docs.captaindata.co/#36e905b6-3a31-4bcd-8c6f-0eb6093b5a8a for details',
+      required: true,
+    }),
```
kishanprmr, line 47: Is it possible to create separate field for each workflow step configuration instead of JSON?

packages/pieces/community/captain-data/src/lib/actions/launch-workflow.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  async run({ auth, propsValue }) {
+    const response = await httpClient.sendRequest({
+      url: `${CAPTAIN_DATA_BASE_URL}/workflows/${propsValue.workflow}/schedule`,
+      method: HttpMethod.POST,
+      headers: {
+        Authorization: `x-api-key ${(auth as CaptainDataAuthType).apiKey}`,
+        'x-project-id': (auth as CaptainDataAuthType).projectId,
+      },
```
kishanprmr, line 63: please add payload in HTTP POST request.

AdamSelene: 🤦 thanks

packages/pieces/community/captain-data/src/lib/actions/launch-workflow.ts
AdamSelene: In theory yes but it would require more work to generate the dynamic props - I'd rather do it in a second stage (particularly since I'm myself not very familiar with Captain Data)
My specific use case here is to prevent my users from using a generic HTTP pieces with unsafe API keys in headers 😬

kishanprmr: We can add this action in the second stage, as it will make it easier for the user to set up the workflow configuration from dynamic properties. What do you think?

AdamSelene: You mean remove the action for now and add it back in a second PR when it's more "baked"? I can do that yes!
As long as we have the custom api call users won't be blocked

kishanprmr: Yes, please add this action in the next PR. Let's merge it after you remove it.

AdamSelene: @kishanprmr I've spent a few hours trying to have something nice with dynamic props, but it turns out it's not possible today - I would need to nest dynamic drop-downs and objects and arrays within an array and ActivePieces does not allow it
I ranted a bit on our shared Slack channel 😬 

Bottom line: either we ship the action with a several JSON fields (`inputs` and `steps` at least), or we drop it and leave users with custom api call.
I think it's still better to have the explicit action, even if not perfect - WDYT?

```
{
    "job_name": "My Linkedin Company Search job ", -> ShortText, easy
    "unstructured_meta": true, -> Checkbox but effect of "true" is not documented
    "inputs": [ -> Array but why? Should be a single object :face_palm: Can't have nested object in array yet
        { 
            "linkedin_company_search_url": "https://www.linkedin.com/search/results/companies/mySearch"        
        }
    ],
    "steps": [{ -> this gets wild
        "accounts": ["40a0b02-1d2e-406e-a2b7-ad3d83948264"], // A Linkedin account UID - but there is no way to tell from the API that for this specific STEP of this specific WORKFLOW you need a LINKEDIN account and not KASPR or LUSHA or GITHUB or any of the 29 available integrations
        "parameters": {"max_results": 42}, // Can't nest an Object in an Array
        "step_uid": "f34de6f7-da13-4dc2-bfdb-27601e651be4", // By wrapping the Array within a DynamicProperties I can use a StaticDropdown here
    }],
}
```

kishanprmr: Okay. Let's merge it.

kishanprmr (approved):

PR #4991, opened 2024-06-26
feat(poper): Update new lead polling trigger to a webhook trigger

thirstycode:
## What does this PR do?

Provides a new way to integrate with Poper. New method relies on Webhook thus the polling overhead.


kishanprmr:
@thirstycode Thank you for the webhook setup guide.

kishanprmr (approved):

PR #4993, opened 2024-06-26
feat: send alerts on flow run failures

islamaf:
## What does this PR do?

Fixes # (issue)


packages/ui/feature-dashboard/src/lib/components/alerts-table/alerts-table.component.html
```diff
@@ -5,11 +5,46 @@
...
+            </div>
+        </div>
+
+        <div class="ap-mt-4 ap-w-full ap-flex ap-flex-col">
+            <span class="ap-typography-subtitle-1">Settings</span>
+            <mat-form-field class="ap-mt-2 ap-w-[20%]">
+                <mat-label>Alerts on Runs Frequency</mat-label>
+                <mat-select i18n-placeholder [formControl]="notificationControl">
+                    <mat-select-trigger>
+                        {{selectTriggerValue}}
```
AbdulTheActivePiecer: Instead of this why not just use a map, where the key is the possible notification value, and the value $localize`DISPLAY_NAME`

packages/server/api/src/app/flows/flow-run/flow-run-side-effects.ts
```diff
@@ -51,11 +52,13 @@ export const flowRunSideEffects = {
...
         if (flowRun.environment === RunEnvironment.PRODUCTION) {
             if (isFailedState(flowRun.status)) {
-                await issuesService.add({
+                const issue = await issuesService.add({
                     flowId: flowRun.flowId,
                     projectId: flowRun.projectId,
                     flowRunCreatedAt: flowRun.created,
                 })
+
+                await alertsService.sendAlertOnFinish({ issue, flowRunId: flowRun.id })
```
AbdulTheActivePiecer: Can we rename this to sendAlertsOnRunFinish ? and rename the method this code is in to calculateDelayForPausedRun?

packages/server/api/src/app/ee/alerts/alerts-handler.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export const alertsHandler = {
+    [NotificationStatus.NEVER]: async (_: IssueParams): Promise<void> => Promise.resolve(),
+    [NotificationStatus.ALWAYS]: async (params: IssueParams): Promise<void> => sendAlertOnFlowRun(params),
+    [NotificationStatus.NEW_ISSUE]: async (params: IssueParams): Promise<void> => sendAlertOnNewIssue(params),
+}
+
+async function sendAlertOnNewIssue(params: IssueParams): Promise<void> {
+    const { platformId, issueCount } = params
+    if (issueCount > 1) {
```
AbdulTheActivePiecer: extract this into a const and call it isOldIssue

AbdulTheActivePiecer (changes requested):

AbdulTheActivePiecer (approved):

PR #4996, opened 2024-06-27
feat: send email reminders on flow run issues 

islamaf:
## What does this PR do?

Fixes # (issue)


packages/server/api/src/app/ee/helper/email/email-service.ts
```diff
@@ -81,6 +85,67 @@ export const emailService = {
...
     },
 
+    async sendIssuesReminder({ projectId }: { projectId: string }): Promise<void> {
+        const project = await projectService.getOneOrThrow(projectId)
+        const platform = await platformService.getOneOrThrow(project.platformId)
+        if (!platform.flowIssuesEnabled || platform.embeddingEnabled) {
+            return
+        }
+        
+        const issues = await issuesService.list({ projectId, cursor: undefined, limit: 50 })
```
AbdulTheActivePiecer: extract the limit into a const and call it something like maxNumberOfIssuesToRemindUserOf

AbdulTheActivePiecer (dismissed):

PR #4998, opened 2024-06-27
feat(wedof): add new actions

obenazouz-wedof:
## What does this PR do?

Fixes # (issue)


packages/ui/common/src/lib/environments/environment.ts
```diff
@@ -1,7 +1,9 @@
 export const environment = {
   production: false,
-  apiUrl: 'http://localhost:3000/v1',
+  apiUrl:
```
kishanprmr: @obenazouz-wedof can you please revert these changes in this file?

obenazouz-wedof: done :) ! thank you that was a little mistake

packages/server/api/.env
```diff
@@ -15,4 +15,4 @@ AP_TELEMETRY_ENABLED=false
 AP_TRIGGER_DEFAULT_POLL_INTERVAL=1
 AP_WEBHOOK_TIMEOUT_SECONDS=30
 AP_PIECES_SYNC_MODE=NONE
-AP_WEBHOOK_URL="http://localhost:3000"
```
abuaboud: Can you please reverse the whitespace there, We have hooks to forbid commit .env and this will cause an issue for other when they are merging.

obenazouz-wedof: done 👍 , thnx

abuaboud (approved):

PR #5006, opened 2024-06-28
feat(slack): revamp the piece to use offical sdk / listing channels is now fixed.

AdamSelene:
## What does this PR do?

- Use official SDK to retrieve users and channels in props
- Remove hard-coded 600 limit
- Migrate all actions and triggers


packages/pieces/community/slack/src/lib/triggers/new-channel.ts
```diff
@@ -1,81 +1,66 @@
...
   },
-  async test(context) {
-    const { store, auth, propsValue } = context;
-    return await pollingHelper.test(polling, { store, auth, propsValue });
+  test: async (context) => {
+    const client = new WebClient(context.auth.access_token);
+    const response = await client.conversations.list({
+      exclude_archived: true,
+      limit: 10,
+      types: 'public_channel',
```
abuaboud, line 40: @AdamSelene should we include other types?

AdamSelene: no strong opinion, it's only for testing anyway

PR #5069, opened 2024-07-08
feat(grist): New/ Updated Record Trigger, Create/Update/Search Record and Upload Attachment Actions

kishanprmr:
## What does this PR do?

Fixes # (issue)


abuaboud (approved):

revolunet:
Hey @kishanprmr, thanks for the GRIST piece !

Trying it on my grist instance, looks like the connection fails because there is no default `current` organisation in my instance.

For me, using `0` instead of `current` works correctly. could you try on your instance ?

https://github.com/activepieces/activepieces/blob/main/packages/pieces/community/grist/src/index.ts#L42

thanks

PR #5078, opened 2024-07-09
feat(notion): get children blocks, append markdown content

AdamSelene:
## What does this PR do?

- Add action to retrieve a page's content (or a given block's children): https://developers.notion.com/reference/get-block-children
- Update "append to page" action to support markdown


kishanprmr:
Hello @AdamSelene ,

"Get Page Content" action uses [Retrieve Children Block API](https://developers.notion.com/reference/get-block-children) which only returns children block of page without page content like page properties and columns. You can use [Retrieve Page API](https://developers.notion.com/reference/retrieve-a-page) to fetch page content. 

If you are creating this action to fetch blocks then please rename it to "Get Block Content".

AdamSelene:
@kishanprmr Done! I also added a `depth` param to recursively fetch children

AdamSelene:
@kishanprmr I've also added markdown support to "append to page"

kishanprmr:
@AdamSelene Thanks for the markdown option. Can you please fix the lint issue for notion piece in PR? You can check it by running ```npx nx run pieces-notion:lint``` locally.

AdamSelene:
@kishanprmr done thanks!

kishanprmr (approved):

PR #5104, opened 2024-07-12
feat(gmail): add options to reply to thread and draft messages

AdamSelene:
## What does this PR do?

- add option to specify `Message-ID` of previous message to reply to it. It will also be added to the correct Gmail thread.
- add option to draft message instead of sending it (works with the "in reply to" option too)

⚠️ I've added a new scope (for drafting messages) and I was not able to test the feature with the default ActivePieces OAuth app - only with my own. I suspect the new scope needs to be added on your side - otherwise I got an "google has blocked this app" error when creating a new connection


kishanprmr:
@AdamSelene, thank you for adding the 'create draft' action. Could you please implement a separate action for 'create draft'? It should function similarly to 'send email', but with a different API call. Also, ensure that both actions return response.data as the response, since you are using the Google SDK.

I will check the new scope permissions for the Cloud OAuth2 app.

AdamSelene:
@kishanprmr I hesitated at first between 2 actions, or a single action with a draft checkbox
Your suggestion makes total sense, but I decided to go for the single action because I **think** is some cases, a builder would like to go back and forth between "draft" (I need to review) and "send" (no need to review). Or it could be determined dynamically (e.g. with an LLM). A single action makes it simpler (otherwise you need to basically duplicate the step which is more cumbersome)
And since the semantics between the 2 are the same (same input)…
WDYT?

abuaboud:
Hi @kishanprmr 

It's now added

kishanprmr:
@AdamSelene Could you please increment the piece version to 0.7.0? The current response data format for the action is different.

PR #5129, opened 2024-07-17
fix(wedof): file/url parameter for send a file action

obenazouz-wedof:
## What does this PR do?

Fixes # (issue)


packages/pieces/community/wedof/src/lib/actions/send-file.ts
```diff
@@ -82,18 +82,56 @@ export const sendFile = createAction({
...
+    },
+  }),
   },
   async run(context) {
-    const message = {
+     const message = {
         title: context.propsValue.title ?? null,
-        typeId: context.propsValue.typeId,
-        file: context.propsValue.file,
+        typeId: String(context.propsValue.typeId['undefined'])?? null,
```
kishanprmr, line 132: @obenazouz-wedof, could you please recheck the typeId prop? It appears to be accessing an 'undefined' field.

packages/pieces/community/wedof/src/lib/actions/send-file.ts
```diff
@@ -38,23 +38,23 @@ export const sendFile = createAction({
         required: false,
     }),
     typeId: Property.DynamicProperties({
```
kishanprmr, line 40: @obenazouz-wedof You can directly use Dropdown prop instead of DynamicProperties. As per [endpoint docs](https://www.wedof.fr/api/doc/operations/get_app_api_certificationfolder_listallcertificationfoldersfile) , use typeId as value and name as label.

kishanprmr (approved):

PR #5134, opened 2024-07-18
feat: make the number of polls from scheduled job configurable

hazemadelkhalel:
## What does this PR do?

I have made the worker take an argument _pollingCount in the function run & poll, whose default value is 1.

In case it is a scheduled job: I have made it take min(5, number of scheduled jobs) and execute them concurrently.


packages/server/worker/src/lib/job-polling.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const POLLING_CONCURRENCY = system.getNumberOrThrow(WorkerSystemProps.POLLING_CONCURRENCY)
+
+const pollLocks = {
+    [QueueName.ONE_TIME]: new Semaphore(POLLING_CONCURRENCY),
+    [QueueName.SCHEDULED]: new Semaphore(POLLING_CONCURRENCY),
+    [QueueName.WEBHOOK]: new Semaphore(POLLING_CONCURRENCY),
+}
+
+export async function pollJob(workerToken: string, queueName: QueueName): Promise<ApQueueJob | null> {
```
abuaboud: It's better not to pollute the global namespace so the IDE doesn't suggest too many functions.

Put it inside a constant.

An example:
```javascript
const jobPolling = {
    // function definition here
    pollJobWithConnectionPool: function() {
        // function logic here
    }
};
export { jobPolling };
```

abuaboud (approved):

PR #5177, opened 2024-07-25
Feat/data to insert

AbdulTheActivePiecer:
## What does this PR do?

Fixes # (issue)


packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,169 @@
...
+export const TextInputWithMentions = ({
+  className,
+  originalValue,
+  onChange,
+  placeholder,
+  extraClasses,
+}: TextInputWithMentionsProps) => {
+  const flowVersion = useBuilderStateContext((state) => state.flowVersion);
+  const steps = flowHelper.getAllSteps(flowVersion.trigger);
+  const piecesMetadata = piecesHooks
```
abuaboud: Just change the functions to use step of actions or triggers directly
```ts
  useStepsMetadata: (props: (Action | Trigger)[]) => {
    const queries = props.map(((step) => stepMetadataQueryBuilder({ step })));
    return useQueries({
      queries,
    });
  },
  ```

File: packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx [hunk 2]
```diff
@@ -0,0 +1,169 @@
...
+      renderHTML({ node }) {
+        const mentionAttrs: MentionNodeAttrs =
+          node.attrs as unknown as MentionNodeAttrs;
+        return textMentionUtils.generateMentionHtmlElement(mentionAttrs);
+      },
+    }),
+  ];
+};
+
+const linkMetadataWithStepsThatUseThem = (
```
abuaboud: This function is not required

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,169 @@
...
+}: TextInputWithMentionsProps) => {
+  const flowVersion = useBuilderStateContext((state) => state.flowVersion);
+  const steps = flowHelper.getAllSteps(flowVersion.trigger);
+  const piecesMetadata = piecesHooks
+    .useStepsMetadata(steps.map((step) => ({ step })))
+    .filter((res) => res.data !== undefined)
+    .map((res) => res.data)
+    .map((res) => {
+      return linkMetadataWithStepsThatUseThem(res, steps);
+    });
```
abuaboud: You can replace the two lines with this
```ts
  const steps = flowHelper.getAllSteps(flowVersion.trigger);
  const piecesMetadata = piecesHooks.useStepsMetadata(steps);
```

package.json
```diff
@@ -223,6 +223,7 @@
     "pickleparser": "0.1.0",
     "pino-loki": "2.1.3",
     "posthog-js": "1.140.1",
+    "primereact": "10.7.0",
```
abuaboud: Please uninstall primereact as we agreed

packages/react-ui/src/styles.css
```diff
@@ -81,4 +80,18 @@
     }
 
 
-}
\ No newline at end of file
+}
+
+*[data-pd-ripple="true"]{
```
abuaboud, line 85: Remove these as we are removing primereact.

packages/react-ui/src/app/builder/data-selector/data-selector-node-content.tsx
```diff
@@ -0,0 +1,116 @@
...
+const ToggleIcon = ({ expanded }: { expanded: boolean }) => {
+  const toggleIconSize = 15;
+  return expanded ? (
+    <ChevronUp height={toggleIconSize} width={toggleIconSize}></ChevronUp>
+  ) : (
+    <ChevronDown height={toggleIconSize} width={toggleIconSize}></ChevronDown>
+  );
+};
+type NodeTemplateProps = {
+  node: MentionTreeNode;
```
abuaboud:
```suggestion
type DataSelectorNodeContentProps = {
```
Rename to Component Name + props.

packages/react-ui/src/app/builder/data-selector/data-selector-size-togglers.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const DataSelectorSizeTogglers = ({
+  state,
+  setListSizeState: setDataSelectorSizeState,
+}: DataSelectorSizeTogglersPorps) => {
+  const handleClick = (newState: DataSelectorSizeState) => {
+    setDataSelectorSizeState(newState);
+  };
+  const buttonClassName = (btnState: DataSelectorSizeState) =>
+    state === btnState ? 'text-outline' : 'text-outline opacity-50';
+  return (
```
abuaboud, line 31:
```suggestion
  const buttonClassName = (btnState: DataSelectorSizeState) => cn(
    '',
    {
      'text-outline': state === btnState,
      'text-outline opacity-50': state !== btnState,
    },
  );
```
This more extensible version

packages/react-ui/src/app/builder/data-selector/data-selector.css
```diff
@@ -0,0 +1,24 @@
+
+*[data-pd-ripple="true"]{
```
abuaboud: I assume they will be removed as they are tied to primereact.

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+        return (
+          step.settings.pieceName === metadata.pieceName &&
+          step.settings.pieceVersion === metadata.pieceVersion
+        );
+      }
+      return (
+        (step.type === ActionType.CODE ||
+          step.type === ActionType.BRANCH ||
+          step.type === TriggerType.EMPTY ||
+          step.type === ActionType.LOOP_ON_ITEMS) &&
```
abuaboud: Why check all of the remaining  types ? These are all the remaining types.

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+        step.type === metadata.type
+      );
+    })
+    .map((step) => step.name);
+  return { ...metadata, stepNamesThatUseThisMetadata };
+};
+const defaultClassName =
```
abuaboud: Please move this to where it's used, It's hard to move up and down in the file to find it.

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+  className,
+  originalValue,
+  onChange,
+  placeholder,
+  extraClasses,
+}: TextInputWithMentionsProps) => {
+  const flowVersion = useBuilderStateContext((state) => state.flowVersion);
+  const steps = flowHelper.getAllSteps(flowVersion.trigger);
+  const piecesMetadata = piecesHooks
+    .useStepsMetadata(steps.map((step) => ({ step })))
```
abuaboud: Is there any reason this doesn't take array of steps instead of object of { step }, please make useStepsMetadata takes an array of steps and not an object of steps.

packages/react-ui/src/features/pieces/lib/pieces-hook.ts
```diff
@@ -47,54 +53,10 @@ export const piecesHooks = {
...
-          }
-        }
-      },
-      staleTime: Infinity,
+  useStepsMetadata: (props: UseStepMetadata[]) => {
+    const queries = props.map(({ step }) => stepMetadataQueryBuilder({ step }));
+    return useQueries({
+      queries,
     });
   },
```
abuaboud:
```suggestion
  },
  useStepsMetadata: (steps: (Action | Trigger)[]) => {
    return useQueries({
      queries: steps.map((step) => stepMetadataQueryBuilder({ step })),
    });
  },
```
If you need array of steps, then you can make it array of steps

packages/react-ui/src/lib/text-input-utils.ts
```diff
@@ -87,68 +80,84 @@ const parseTextAndHardBreakNodes = (item: string) => {
...
+function convertTextToTipTapJsonContent({
+  propertyPath: userTextInput,
+  stepMetadataFinder,
+}: {
+  propertyPath: string;
+  stepMetadataFinder: (
+    path: string,
+  ) => (StepMetadata & { dfsIndex: number }) | undefined;
+}): {
   type: TipTapNodeTypes.paragraph;
```
abuaboud:
```suggestion
function convertTextToTipTapJsonContent(
  propertyPath: string,
  userTextInput: string,
  stepMetadata: (StepMetadata & { dfsIndex: number }),
): {
```

I think this function is unnecssary,  you can just pass the metadata directly.

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+    .filter((res) => res.data !== undefined)
```
abuaboud: It doesn't make sense to filter data coming from back in hooks. we are not handling the loading.

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+    .map((res) => res.data)
+    .map((res) => {
+      return linkMetadataWithStepsThatUseThem(res, steps);
```
abuaboud: This function is not really needed, the function can be rewritten

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+import { textMentionUtils } from '@/lib/text-input-utils';
+
+import {
+  piecesHooks,
+  StepMetadata,
+} from '../../../features/pieces/lib/pieces-hook';
+import { useBuilderStateContext } from '../builder-hooks';
+
+const extensions = (placeholder?: string) => {
```
abuaboud, line 24: Can you move this down, as It make sense to be around the editor, so I have some sense of what extensions are there

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+export const TextInputWithMentions = ({
+  className,
+  originalValue,
+  onChange,
+  placeholder,
+  extraClasses,
+}: TextInputWithMentionsProps) => {
+  const flowVersion = useBuilderStateContext((state) => state.flowVersion);
+  const steps = flowHelper.getAllSteps(flowVersion.trigger);
+  const piecesMetadata = piecesHooks
```
abuaboud: All of the code below and please remove the function `linkMetadataWithStepsThatLinkthem`

The whole code be replaced with the following:

```ts
    
  const steps = flowHelper.getAllSteps(flowVersion.trigger);
  const {data: pieceMetadata} = piecesHooks.useStepsMetadata(steps);

  const removeBarackets = (path: string) => { 
    return path.slice(2, path.length - 2);
  }
  const getStepMetadataFromPath = (path: string) => {
    const stepName = textMentionUtils.keysWithinPath(removeBarackets(path))[0];
    const idx = steps.findIndex((step) => step.name === stepName);
    const metadata = data[idx];
    return metadata;
  };

```

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
```diff
@@ -0,0 +1,171 @@
...
+  return { ...metadata, stepNamesThatUseThisMetadata };
+};
+const defaultClassName =
+  ' w-full  rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring focus-visible:ring-offset-1 disabled:cursor-not-allowed disabled:opacity-50';
+type TextInputWithMentionsProps = {
+  className?: string;
+  originalValue?: string;
+  onChange: (value: string) => void;
+  placeholder?: string;
+  extraClasses?: string;
```
abuaboud: I think having extraClasses and className is redundant. I'd just use className.

packages/react-ui/src/features/pieces/components/piece-icon-list.tsx
```diff
@@ -1,45 +1,64 @@
...
+    .useStepsMetadata(steps)
+    .map((data) => data.data)
+    .filter((data) => !!data);
 
-  const visibleSteps = steps.slice(0, 2);
-  const extraStepsCount = steps.length - visibleSteps.length;
-  const extraSteps = steps.slice(2);
+  const uniqueMetadata = [
+    ...new Map(stepsMetadata.map((item) => [item['pieceName'], item])).values(),
+  ];
```
abuaboud, line 24: This is can be written in the following way:

```ts
const steps = flowHelper.getAllSteps(trigger);
const { data: stepsMetadata } = piecesHooks.useStepsMetadata(steps);

const uniqueMetadata: StepMetadata[] = Array.from(
  new Map<string, StepMetadata>(
    stepsMetadata.map(item => [item.pieceName, item])
  ).values()
);
```

packages/react-ui/src/features/pieces/lib/pieces-hook.ts
```diff
@@ -105,3 +67,73 @@ export const piecesHooks = {
...
+    staleTime: Infinity,
+  };
+}
+
+async function getStepMetadata(
+  type: ActionType | TriggerType,
+  pieceName: string,
+  pieceVersion: string,
+): Promise<StepMetadata> {
+  const metadataExtractor = async () => {
```
abuaboud: I don't think this should be function inside function, I don't understand how pieceName and pieceVersion is always defined for types that are not PIECE.

packages/react-ui/src/features/pieces/lib/pieces-hook.ts
```diff
@@ -105,3 +67,73 @@ export const piecesHooks = {
...
+        type,
+        pieceName,
+        pieceVersion,
+      };
+    },
+    staleTime: Infinity,
+  };
+}
+
+async function getStepMetadata(
```
abuaboud: Why not make this take step instead?

packages/react-ui/src/features/pieces/lib/pieces-hook.ts
```diff
@@ -105,3 +67,73 @@ export const piecesHooks = {
     });
   },
 };
+
+function stepMetadataQueryBuilder({ step }: { step: Action | Trigger }) {
+  const isPieceStep =
+    step.type === ActionType.PIECE || step.type === TriggerType.PIECE;
+  const pieceName = isPieceStep ? step.settings.pieceName : `${step.type}`;
+  const pieceVersion = isPieceStep ? step.settings.pieceVersion : ``;
```
abuaboud: This is super hacky, just make the getStepMetadata takes step and It would use pieceName and pieceVerison when it's really required.

packages/react-ui/src/features/pieces/lib/pieces-hook.ts
```diff
@@ -29,6 +29,9 @@ export type StepMetadata = {
   displayName: string;
   logoUrl: string;
   description: string;
+  pieceName: string;
```
abuaboud, line 34: Why these returns pieceName / pieceVersion to begin with?

I think if you are looking for pieces, then it's better to useMultiplePieces and these return the full model from the backend.

packages/react-ui/src/features/templates/components/share-template.tsx
```diff
@@ -43,61 +38,66 @@ const ShareTemplate: React.FC<{ templateId: string }> = ({ templateId }) => {
...
+    staleTime: 0,
+  });
+  if (!data) {
+    return (
+      <div className="w-full h-full flex items-center justify-center">
+        <LoadingSpinner size={50}></LoadingSpinner>
+      </div>
+    );
+  }
+  return <TemplateViewer template={data}></TemplateViewer>;
```
abuaboud, line 100: Why this can't be single component where FlowTemplate handle the loading as well? I don't see where TemplateViewer can be reused, If it has to be resuable later then It's better to seperate in different component.

packages/react-ui/src/components/theme-provider.tsx
```diff
@@ -71,3 +71,28 @@ export const useTheme = () => {
 
   return context;
 };
+
+export const useRipple = () => {
```
abuaboud: Remove this, It's related to primereact.

abuaboud (changes requested):

packages/react-ui/src/app/builder/text-input-with-mentions/text-input-with-mentions.tsx
AbdulTheActivePiecer: we discussed this.

packages/react-ui/src/components/theme-provider.tsx
AbdulTheActivePiecer: I need primereact for the ripple, how else am I going to make it ?

packages/react-ui/src/features/templates/components/share-template.tsx
AbdulTheActivePiecer: Template viewer uses hooks, and those hooks depend on the template, making sure the template is defined before the component is rendered makes it easier to handle those hooks instead of checking for the template being defined or not.

packages/react-ui/src/lib/text-input-utils.ts
AbdulTheActivePiecer: You can't call hooks inside functions, so I needed to pass the function that has access to the result of the hook.

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+
+const getStep = (stepName: string) => {
+  return (state: BuilderState) => {
+    const { flowVersion } = state;
+    if (!flowVersion) {
+      return undefined;
+    }
+    return flowHelper.getStep(flowVersion, stepName);
+  };
+};
```
abuaboud, line 284: Remove it's unused

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+              },
+              key: s.name,
+            },
+          ]
+        : stepMentionNode.children,
+    };
+  });
+};
+
+const getStep = (stepName: string) => {
```
abuaboud: This is currently unused, either call it in places where it used or remove it.

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
   });
   return targetMap;
 }
+
+const getAllStepsMentions: (state: BuilderState) => MentionTreeNode[] = (
```
abuaboud: This is only used in one place data-selector, Let's move it to the file that used it instead of gobally shared file as there is no other place using it.

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  return stepOutput;
+}
+const MAX_ARRAY_LENGTH_BEFORE_SLICING = 100;
+const isStepName = (name: string) => {
+  const regex = /^(trigger|step_\d+)$/;
+  return regex.test(name);
+};
+
+export const dataSelector = {
+  isStepName,
```
abuaboud: This is unused, please remove it

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    return {
+      key: propertyPath,
+      data: {
+        propertyPath: propertyPath,
+        displayName: displayName,
+        isSlice: false,
+        value: Object.keys(stepOutput).length === 0 ? 'Empty List' : undefined,
+      },
+      children: Object.keys(stepOutput).map((k) => {
```
abuaboud: k is one letter varriable, It's better to be renamed to key

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+        const escapedKey = k
+          .replaceAll(/\\/g, '\\')
+          .replaceAll(/"/g, '\\"')
+          .replaceAll(/'/g, "\\'")
+          .replaceAll(/\n/g, '\\n')
+          .replaceAll(/\r/g, '\\r')
+          .replaceAll(/\t/g, '\\t')
+          .replaceAll(/’/g, '\\’');
+        const newPath = `${propertyPath}['${escapedKey}']`;
+        const newDisplayName = k;
```
abuaboud: both of these can be in-lined, they are only used once.

Ie.


```suggestion
     return traverseStepOutputAndReturnMentionTree({
          stepOutput: value,
          propertyPath: `${propertyPath}['${escapedKey}']`,
          displayName: key,
        });
```

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  displayName,
+}: {
+  stepOutput: unknown;
+  propertyPath: string;
+  displayName: string;
+}): MentionTreeNode {
+  if (stepOutput && typeof stepOutput === 'object') {
+    if (Array.isArray(stepOutput)) {
+      return handlingArrayStepOutput(stepOutput, propertyPath, displayName);
+    }
```
abuaboud: I think javascript is considering array is Object but it's doesn't have to be written like this.

a better way to write this like that, is to move array outside the check of object.

and please extract the stepoutput check to something like this
const isObject = stepOutput && typeof stepOutput === 'object';

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+    displayName: string;
+    /**value for json key */
+    value?: string | unknown;
+    isSlice: boolean;
+    isTestStepNode?: boolean;
+  };
+  children?: MentionTreeNode[];
+};
+/**Traverses an object to find its child properties and their paths, stepOutput has to be an object on first invocation */
+function traverseStepOutputAndReturnMentionTree({
```
abuaboud: I would split this into at least 3 functions for internal handing.

if (isObject) {
    // handle object
}
if (isArray) {
    // handle array function
}
return {

}

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+    return 'null';
+  }
+
+  if (typeof stepOutput === 'string') {
+    return `"${stepOutput}"`;
+  }
+
+  return stepOutput;
+}
+const MAX_ARRAY_LENGTH_BEFORE_SLICING = 100;
```
abuaboud: this is used in one function, please move it to inside function.

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+      data: {
+        propertyPath: path,
+        displayName: lastDisplayName,
+        value: stepOutput.length === 0 ? 'Empty List' : undefined,
+        isSlice: false,
+      },
+    };
+  }
+
+  const numberOfSlices = new Array(Math.ceil(stepOutput.length / 100)).fill(0);
```
abuaboud: You have defined variable to the "100" why not use it?

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  const children: MentionTreeNode[] = [];
+  numberOfSlices.forEach((_, i) => {
+    const startingIndex = i * 100;
+    const endingIndex = Math.min((i + 1) * 100, stepOutput.length) - 1;
+    const newPath = `${path}`;
+    const displayName = `${lastDisplayName} ${startingIndex}-${endingIndex}`;
```
abuaboud: What's point of defining newPath if it's same as as path?

It's only used in one place, please in-line it.

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+  const step = flowHelper.getStep(flowVersion, selectedStep.stepName);
+  if (!step) {
+    return [];
+  }
+  const path = flowHelper.findPathToStep({
+    stepToFind: step,
+    trigger: flowVersion?.trigger,
+  });
+
+  return path.map((s) => {
```
abuaboud: s one letter named variable, a better name would be step

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+) => {
+  const { selectedStep, flowVersion } = state;
+  if (!selectedStep || !flowVersion || !flowVersion.trigger) {
+    return [];
+  }
+  const step = flowHelper.getStep(flowVersion, selectedStep.stepName);
+  if (!step) {
+    return [];
+  }
+  const path = flowHelper.findPathToStep({
```
abuaboud: steps is better thanme than path, because it describe the output of the function, and it is more readable

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+    stepToFind: step,
+    trigger: flowVersion?.trigger,
+  });
+
+  return path.map((s) => {
+    const stepMentionNode: MentionTreeNode =
+      dataSelector.traverseStepOutputAndReturnMentionTree({
```
abuaboud: So this is confusing, because why would you traverse the tree then check if It's undefined? and do special case handling for  that?

a better thinking thougt If it's need test return the testStepNode otherwise traverse tree and return the testStepNoode

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+                propertyPath: s.name,
+                isTestStepNode: true,
+                isSlice: false,
+              },
+              key: s.name,
+            },
+          ]
+        : stepMentionNode.children,
+    };
+  });
```
abuaboud:
```suggestion
   const stepNeedsTesting =
      s.settings.inputUiInfo?.currentSelectedData === undefined;
    if (stepNeedsTesting) {
      return {
        key: s.name,
        data: {
          displayName: `${s.dfsIndex}. ${s.displayName}`,
        },
        children: [
          {
            key: s.name,
            data: {
              displayName: 'Testing Step',
              propertyPath: s.name,
              isTestStepNode: true,
              isSlice: false,
            },
            children: [],
          },
        ]
      };
    }
    const stepMentionNode: MentionTreeNode =
      dataSelector.traverseStepOutputAndReturnMentionTree({
        stepOutput: s.settings.inputUiInfo?.currentSelectedData,
        propertyPath: s.name,
        displayName: s.displayName,
      });
    return stepMentionNode;
```
Something like this, (This is not exact code, but it should give you an idea of what I am trying to do)

packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -223,3 +224,67 @@ function getStateAtPath({
...
+  if (!step) {
+    return [];
+  }
+  const path = flowHelper.findPathToStep({
+    stepToFind: step,
+    trigger: flowVersion?.trigger,
+  });
+
+  return path.map((s) => {
+    const stepMentionNode: MentionTreeNode =
```
abuaboud: Please use isNil in these cases isNils.settings.inputUiInfo?.currentSelectedData )

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+      data: {
+        propertyPath: propertyPath,
+        displayName: displayName,
+        isSlice: false,
+        value: Object.keys(stepOutput).length === 0 ? 'Empty List' : undefined,
+      },
+      children: Object.keys(stepOutput).map((k) => {
+        const escapedKey = k
+          .replaceAll(/\\/g, '\\')
+          .replaceAll(/"/g, '\\"')
```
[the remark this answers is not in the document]
abuaboud: you can replaceAll with this
```
 const escapedKey = key.replaceAll(/[\\"'\n\r\t’]/g, (char) => `\\${char}`);
 ```

File: packages/react-ui/src/lib/data-selector-utils.ts [hunk 43]
```diff
@@ -0,0 +1,173 @@
...
+      path: propertyPath,
+      startingIndex: 0,
+    });
+  }
+  const isObject = stepOutput && typeof stepOutput === 'object';
+
+  if (isObject) {
+    return handleObjectStepOutput({ propertyPath, displayName, stepOutput });
+  } else {
+    const value = formatStepOutput(stepOutput);
```
abuaboud: Again this is single variable that can be inlined

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,173 @@
...
+      lastDisplayName: displayName,
+      path: propertyPath,
+      startingIndex: 0,
+    });
+  }
+  const isObject = stepOutput && typeof stepOutput === 'object';
+
+  if (isObject) {
+    return handleObjectStepOutput({ propertyPath, displayName, stepOutput });
+  } else {
```
abuaboud: No need for else as you are returning the value of function.

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,173 @@
...
+  lastDisplayName: string;
+  startingIndex: number;
+}): MentionTreeNode => {
+  const MAX_ARRAY_LENGTH_BEFORE_SLICING = 100;
+  if (stepOutput.length <= MAX_ARRAY_LENGTH_BEFORE_SLICING) {
+    return {
+      key: lastDisplayName,
+      children: stepOutput.map((v, idx) => {
+        const newPath = `${path}[${idx + startingIndex}]`;
+        const newDisplayName = `${lastDisplayName} ${idx + startingIndex}`;
```
abuaboud: This can be in-lined newPath and newDisplayName

packages/react-ui/src/lib/data-selector-utils.ts
```diff
@@ -0,0 +1,173 @@
...
+  stepOutput: unknown[];
+  path: string;
+  lastDisplayName: string;
+  startingIndex: number;
+}): MentionTreeNode => {
+  const MAX_ARRAY_LENGTH_BEFORE_SLICING = 100;
+  if (stepOutput.length <= MAX_ARRAY_LENGTH_BEFORE_SLICING) {
+    return {
+      key: lastDisplayName,
+      children: stepOutput.map((v, idx) => {
```
abuaboud: this is single character name, maybe name it a child

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -943,6 +940,91 @@ function isStepLastChildOfParent(child: Step, trigger: Trigger): boolean {
...
+        throw new Error('Step not found while traversing to find it ')
+    }
+    const pathWithIndex = path.map((f) => {
+        return {
+            ...f,
+            dfsIndex: findStepDfsIndex(trigger, f.name),
+        }
+    })
+    return [{ ...trigger, dfsIndex: 1 }, ...pathWithIndex]
+}
```
abuaboud:
```suggestion

type StepWithIndex = Step & { dfsIndex: number }

function findPathToStep({ targetStep: stepToFind, trigger }: {
    targetStep: Step
    trigger: Trigger
}): StepWithIndex[] {
    const steps = getAllSteps(trigger).map((step, dfsIndex) => ({
        ...step,
        dfsIndex,
    }));
    return steps.filter((step) => isPartOfInnerFlow({
        parentStep: step,
        childName: stepToFind.name,
    })).filter((step) => step.name !== stepToFind.name)
}
```

You can replace the functionn with this

PR #5196, opened 2024-07-28
feat: date-filter-runs

islamaf:
## What does this PR do?

Fixes # (issue)


packages/react-ui/src/features/flows/lib/flows-hooks.ts
```diff
@@ -14,7 +14,7 @@ export const flowsHooks = {
           cursor: undefined,
         });
       },
-      staleTime: 0,
+      staleTime: Infinity,
```
abuaboud: Don't cache infinity what if the user adds flow, It will be buggy

PR #5224, opened 2024-07-31
chore: animate closing/opening sidebars

abuaboud:
## What does this PR do?

Fixes # (issue)


packages/react-ui/src/app/builder/builder-page.tsx
```diff
@@ -30,55 +53,77 @@ const BuilderPage = () => {
...
+            withHandle={rightSidebar !== RightSideBarType.NONE}
+            onDragging={setIsDraggingHandle}
+          />
+
+          <ResizablePanel
+            ref={rightHandleRef}
+            id="right-sidebar"
+            defaultSize={0}
+            minSize={0}
+            maxSize={45}
```
abuaboud: Please revert it to 60% this intentional changes from previous design.

Reason: I compared the design with previous one, It's much better than this way and all other similar designs / platforms allow you for bigger right sidebar, specially for code settings sidebar, which is now easier to use in the smaller screen.

packages/react-ui/src/app/builder/builder-page.tsx
```diff
@@ -30,55 +53,77 @@ const BuilderPage = () => {
...
-
+  const [isDraggingHandle, setIsDraggingHandle] = useState(false);
+  const rightHandleRef = useAnimateSidebar(
+    useCallback(() => rightSidebar === RightSideBarType.NONE, [rightSidebar]),
+    rightSidebar,
+  );
+  const leftHandleRef = useAnimateSidebar(
+    useCallback(() => leftSidebar === LeftSideBarType.NONE, [leftSidebar]),
+    leftSidebar,
+  );
```
abuaboud:
```suggestion
  const [isDraggingHandle, setIsDraggingHandle] = useState(false);
  const rightHandleRef = useAnimateSidebar(
    rightSidebar,
  );
  const leftHandleRef = useAnimateSidebar(
    leftSidebar,
  );
```

I don't think useCallback is needed, you already know the values of isSidebarClosed() and in case it's changed, the component will re-render.

packages/react-ui/src/app/builder/builder-page.tsx
```diff
@@ -12,14 +14,35 @@ import {
...
+        handleRef.current?.getSize() === undefined)
+    ) {
+      handleRef.current?.resize(25);
+    } else if (isSidebarClosed()) {
+      console.log(handleRef.current);
+      handleRef.current?.resize(0);
+    }
+  }, [handleRef, isSidebarClosed, sidebarValue]);
+  return handleRef;
+};
```
abuaboud:
```suggestion
const useAnimateSidebar = (
  sidebarValue: LeftSideBarType | RightSideBarType,
) => {
  const handleRef = useRef<ImperativePanelHandle>(null);
  const sidebarbarClosed = [LeftSideBarType.NONE, RightSideBarType.NONE].includes(sidebarValue);
  useEffect(() => {
    const sidebarSize = handleRef.current?.getSize() ?? 0;
    if (sidebarbarClosed) {
      handleRef.current?.resize(0);
    } else if (sidebarSize === 0){
      handleRef.current?.resize(25);
    }
  }, [handleRef, sidebarValue]);
  return handleRef;
};
```
- I removed isSidebarClosed() and replaced it with boolean since you already have value.
- I moved the sidebarClosed check to the first condition so we don't have to check it in else.
- I removed isSidebarClosed() from the depdendencies array since it's not used.
- I extracted the sideBar and varriable and used it.
- Removed console log

AbdulTheActivePiecer (approved):

PR #5229, opened 2024-07-31
feat: added rate limiter for jobs per project

hazemadelkhalel:


packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+import { DefaultJobOptions, Job, Queue, Worker } from 'bullmq'
+import dayjs from 'dayjs'
+
+import { createRedisClient } from '../../database/redis-connection'
+import { AddParams, JOB_PRIORITY } from '../queue/queue-manager'
+import { redisHandler } from '../redis/redis-handler'
+
+
+const EIGHT_MINUTES_IN_MILLISECONDS = 8 * 60 * 1000
+const defaultJobOptions: DefaultJobOptions = {
```
abuaboud: It's only used once, move it down to where it's used

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+import { isNil } from '@activepieces/shared'
+import { DefaultJobOptions, Job, Queue, Worker } from 'bullmq'
+import dayjs from 'dayjs'
+
+import { createRedisClient } from '../../database/redis-connection'
+import { AddParams, JOB_PRIORITY } from '../queue/queue-manager'
+import { redisHandler } from '../redis/redis-handler'
+
+
+const EIGHT_MINUTES_IN_MILLISECONDS = 8 * 60 * 1000
```
abuaboud: Inline this inside defaultJobOptions

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+            return
+        }
+
+        await queue?.remove(pulledJobId)
+
+        await activeQueue.add(pulledJobId, pulledJob.data, {
+            jobId: pulledJobId,
+            priority: JOB_PRIORITY[pulledJob.data.priority as keyof typeof JOB_PRIORITY],
+        })
+
```
abuaboud: Remove this and read my comment above

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    connection: createRedisClient(),
+                    defaultJobOptions,
+                },
+            )
+            await queue.waitUntilReady()
+        }
+        if (isNil(worker)) {
+            const lockDuration = dayjs.duration(triggerTimeoutSandbox, 'seconds').add(3, 'minutes').asMilliseconds()
+            worker = new Worker('rateLimitJobs', null, {
```
abuaboud: the null can be a function directly and add the code that handle jobs here.

check the example of how the function is inside, you can figure out which queue it belongs to by looking at job.data.type

https://github.com/taskforcesh/bullmq#:~:text=Process%20the%20jobs%20in%20your%20workers%3A

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        await redisHandler.push(data.payload.userId, data.payload.jobId)
+        const delayedJob = await Job.fromId(queue, id)
+        if (isNil(delayedJob)) {
+            return
+        }
+        await delayedJob.changeDelay(1000000000)
+    },
+
+    async activeJob(job: Job, activeQueue: Queue): Promise<void> {
```
abuaboud: rename it to `onCompleteOrFailedJob`

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
abuaboud: Another reason why this is bad, Is you might remove the job then fail to add it to ActiveQueue just move the adding job inside the worker funtion like above you don't need the remove, as the queue already have option to removeOnComplete

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+                'rateLimitJobs',
+                {
+                    connection: createRedisClient(),
+                    defaultJobOptions,
+                },
+            )
+            await queue.waitUntilReady()
+        }
+        if (isNil(worker)) {
+            const lockDuration = dayjs.duration(triggerTimeoutSandbox, 'seconds').add(3, 'minutes').asMilliseconds()
```
abuaboud: Why you need that long queue? The queue handler should only add job to right queue make it static or short like 30 seconds

packages/server/api/src/app/workers/redis/redis-handler.ts
```diff
@@ -0,0 +1,33 @@
...
+            return null
+        }
+        const jobId = await redis.spop(userKey)
+        await redis.decr(userKey)
+        return jobId
+    },
+
+    async shouldBeLimited(userId: string): Promise<boolean> {
+        const userKey = `active_runs:${userId}`
+        const limit = 100 //TODO: The limit can be configurable
```
abuaboud: True, add environment variable to it

packages/server/api/src/app/workers/redis/redis-handler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const redisHandler = {
+    async init(): Promise<void> {
+        redis = getRedisConnection()
+    },
+
+    async push(userId: string, jobId: string): Promise<void> {
+        const userKey = `active_runs:${userId}`
+        await redis.sadd(userId, jobId)
+        await redis.incr(userKey)
```
abuaboud: For example what if we had a bug that this counter is wrong and we had to wrong, It will not self correct it's self, I would add TTL for the record but not sure if this is optimal answer
```ts
        await redis.expire(userKey, TTL_SECONDS) // Reset TTL on access
```

packages/server/api/src/app/workers/helper/redis-rate-limiter.ts
```diff
@@ -0,0 +1,96 @@
...
+    removeOnComplete: true,
+}
+
+
+let worker: Worker | null = null
+
+let queue: Queue | null = null
+
+
+export const redisRateLimiter = {
```
abuaboud: This is should be merged with redis-Handler, most of the functions should be internal not inside export const

I think what you need to expose three functions as they are only one used inside redis-queue
- shouldRateLimit
- rateLimitJob is (which is now called delayJob)
- onJobCompleteOrFail

packages/server/api/src/app/workers/redis/redis-consumer.ts
```diff
@@ -55,6 +58,7 @@ export const redisConsumer: ConsumerManager = {
         }
     },
     async init(): Promise<void> {
+        await redisRateLimiter.init()
```
abuaboud: Keep it in one place, remove this from here, It's already called redis-queue

packages/server/api/src/app/workers/redis/redis-rate-limiter.ts
```diff
@@ -0,0 +1,110 @@
...
+                await redisQueue.add(job.data)
+            }, {
+                connection: createRedisClient(),
+                lockDuration: dayjs.duration(30, 'seconds').asMilliseconds(),
+                maxStalledCount: 5,
+                stalledInterval: 30000,
+            })
+            await worker.waitUntilReady()
+        }
+    },
```
abuaboud: I would remove the null checks and add the following.

Add this to assertions.ts
```suggestion

export function assertNull<T>(
    value: T | null,
    fieldName: string,
): asserts value is T {
    if (value !== null) {
        throw new Error(`${fieldName} should be null`)
    }
} 
```

and then use it there
```suggestion
    async init(): Promise<void> {
        assertNull(redis, 'redis is not null')
        assertNull(queue, 'queue is not null')
        assertNull(worker, 'worker is not null')
        redis = getRedisConnection()
        queue = new Queue(
            'rateLimitJobs',
            {
                connection: createRedisClient(),
                defaultJobOptions: {
                    attempts: 5,
                    backoff: {
                        type: 'exponential',
                        delay: 8 * 60 * 1000,
                    },
                    removeOnComplete: true,
                },
            },
        )
        await queue.waitUntilReady()
        worker = new Worker<AddParams<JobType.ONE_TIME | JobType.WEBHOOK>>('rateLimitJobs', async (job) => {
            await redisQueue.add(job.data)
        }, {
            connection: createRedisClient(),
            lockDuration: dayjs.duration(30, 'seconds').asMilliseconds(),
            maxStalledCount: 5,
            stalledInterval: 30000,
        })
        await worker.waitUntilReady()
    },
```

packages/server/shared/src/lib/system/system.ts
```diff
@@ -78,6 +78,7 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
     [AppSystemProp.TEMPLATES_SOURCE_URL]:
         'https://cloud.activepieces.com/api/v1/flow-templates',
     [AppSystemProp.TRIGGER_DEFAULT_POLL_INTERVAL]: '5',
+    [AppSystemProp.RATE_LIMIT]: '10',
```
abuaboud: This is a genric name, how about `PROJECT_RATE_LIMIT`

packages/server/api/src/app/workers/redis/redis-rate-limiter.ts
```diff
@@ -0,0 +1,110 @@
...
+                {
+                    connection: createRedisClient(),
+                    defaultJobOptions: {
+                        attempts: 5,
+                        backoff: {
+                            type: 'exponential',
+                            delay: 8 * 60 * 1000,
+                        },
+                        removeOnComplete: true,
+                    },
```
abuaboud, line 35: move this to 
const rateLimitQueueName = 'rateLimitJobs'

packages/server/api/src/app/workers/redis/redis-rate-limiter.ts
```diff
@@ -0,0 +1,110 @@
...
+
+let redis = getRedisConnection() 
+
+const RATE_LIMIT = system.getNumberOrThrow(AppSystemProp.RATE_LIMIT)
+
+
+let worker: Worker | null = null
+
+let queue: Queue | null = null
+
```
abuaboud: Nit: I would move const to top and keep let variables together.

packages/server/api/src/app/workers/redis/redis-rate-limiter.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        assertNotNullOrUndefined(worker, 'Worker is not initialized')
+        const pulledJob = await Job.fromId(worker, pulledJobId)
+        assertNotNullOrUndefined(pulledJob, 'Pulled job is not found')
+        await pulledJob.changeDelay(0)
+    },
+
+    async getQueue(): Promise<Queue> {
+        if (isNil(queue)) {
```
abuaboud: Nit: replace with assertNotNullOrDefined

abuaboud (approved):

abuaboud (approved):

PR #5237, opened 2024-08-01
fix: remove email format in sign up schema

islamaf:
## What does this PR do?

Fixes # (issue)


abuaboud:
Closing we need validators on emails

PR #5268, opened 2024-08-05
fix: improve file handling when downloading from URL

no10xcoder:
## What does this PR do?

This PR improves file handling when downloading files from URL:

- Remove `Content-Type` check: it should not matter to AP what the Content-Type is of the resulting file
- Add support for `Content-Disposition`: when this header exists it now tries to read the filename from that
- Only use the last part of the filename when there is a file extension

I also made a minor improvement for the base64 file handling, it tries to base the extension on the content-type, but this would give very odd results in some cases, for example:

- .docx has the mimetype `application/vnd.openxmlformats-officedocument.wordprocessingml.document`, which would result in the filename `unknown.vnd.openxmlformats-officedocument.wordprocessingml.document`
- An unknown mimetype `application/octet-stream` would become `unknown.octet-stream`.

I think using a hardcoded filename of `unknown.bin` is better in this case.

Another improvement I want to add is adding the possibility of adding the Content-Type to the file object, but I want to discuss that first.

### DIFF
--- a/packages/pieces/community/framework/src/lib/processors/processors.ts
+++ b/packages/pieces/community/framework/src/lib/processors/processors.ts
@@ -82,57 +82,68 @@ export class Processors {
       if (isBase64(urlOrBase64, { allowMime: true })) {
         const matches = urlOrBase64.match(/^data:([A-Za-z-+/]+);base64,(.+)$/);
         let base64 = urlOrBase64;
-        let contentType: string | null = null;
 
         if (matches && matches?.length === 3) {
-          contentType = matches[1];
           base64 = matches[2];
 
-          // You need to provide how you decide filename and extension in case of base64 string
-          const filename = 'unknown';
-          const extension = contentType.split('/')[1];
-
           return new ApFile(
-            filename + '.' + extension,
+            'unknown.bin',
             Buffer.from(base64, 'base64'),
-            extension
           );
         }
       }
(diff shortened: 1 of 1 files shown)


anasbarg (approved):

abuaboud:
Very good PR! It triggered a series of discussions.

We refactored the entire area in [this PR](https://github.com/activepieces/activepieces/pull/5410).

**Base64:**
Instead of using `unknown.bin`, we will try to determine the file type using the `mime-types` package when dealing with Base64.

**Checking for Content Type:**

The reason we were checking for the content type was to prevent the downloading of websites like google.com. However, this doesn't make much sense, as browsers generally allow you to save files regardless.

So, instead of editing this test case, we decided to approach it in the following order:

- If the "Content-Disposition" header exists, it means the file is downloadable. We extract the file name and extension from the provided file name.
- If not, it could be something that can be downloaded, like `https://cdn.activepieces.com/brand/logo.svg`, but it doesn't trigger the auto-download when you open it in the URL.
- We try to extract the file name from the URL (first we clean the query strings and protocol, then rely on the file name). If it has a dot, it means there is an extension, so we try to use it.

If all else fails, we try to guess the file type from the "Content-Type" header, and the file name will default to `unknown.bin`.

abuaboud:
I am closing this pr in favor of the refactoring!

PR #5281, opened 2024-08-05
feat(react-templates): support creating flows from templates

anasbarg:
## What does this PR do?

This PR adds support for creating flows from predefined templates.

### Optimizations

I noticed that we always return the entire list of templates from the backend (the seek cursor is always `null`). So I decided to make the same assumption on the frontend as it is on the backend and use it to:

1. Optimistically preload the templates as soon as the flows table renders so the user doesn't have to wait for the templates to load when they open the templates dialog
2. Make the search to be a front-end search for snappier search results avoiding the frequent loading spinner

### Not implemented yet

1. Filtering using tags & pieces
3. "learn more" button to navigate to a description screen .. for now I made it open the blog URL in a new tab.


packages/react-ui/src/features/flows/components/select-flow-template-dialog.tsx
```diff
@@ -0,0 +1,122 @@
+import { useMutation } from '@tanstack/react-query';
```
abuaboud: Let's discuss the structure of the project (I am kind of hesitant now )

I suggest moveing this the app level, since it doesn't use features template and pieces.

packages/react-ui/src/features/flows/components/select-flow-template-dialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+  children,
+  templates: { templates, isLoading },
+}: {
+  children: React.ReactNode;
+  templates: { templates?: FlowTemplate[]; isLoading: Boolean };
+}) => {
+  const navigate = useNavigate();
+  const [search, setSearch] = useState<string>('');
+
+  // Filter templates if the template name or the template description contains the search query
```
abuaboud: Nit: I would move templates / isLoading to inside the selectFlowTemplateDialog.

```ts
type SelectFlowTemplateDialogProps = {
  children: React.ReactNode
}
const SelectFlowTemplateDialog = ({
  children,
}: SelectFlowTemplateDialogProps) => {
  const navigate = useNavigate();
  const [search, setSearch] = useState<string>('');


  const { data: templates, isLoading } = useQuery<FlowTemplate[], Error>({
    queryKey: ['templates'],
    queryFn: async () => {
      const templats = await templatesApi.list()

    },
    staleTime: 0,
  });
  ```

File: packages/react-ui/src/features/flows/components/select-flow-template-dialog.tsx [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+                  <PieceIconList
+                    trigger={template.template.trigger}
+                    maxNumberOfIconsToShow={2}
+                  />
+                </div>
+                <div className="text-sm font-medium leading-none px-4 leading-5">
+                  {template.name}
+                </div>
+                <div className="py-2 flex">
+                  <Button variant="link" onClick={() => createFlow(template)}>
```
abuaboud: Add loading={isPending}

PR #5294, opened 2024-08-07
feat(redis-rate-limiter): made an early return for failed web hooks

hazemadelkhalel:


packages/server/api/src/app/workers/redis/redis-queue.ts
```diff
@@ -37,7 +38,8 @@ export const redisQueue: QueueManager = {
     async add(params): Promise<void> {
         const { type, data } = params
         const { shouldRateLimit } = await redisRateLimiter.shouldBeLimited(jobTypeToQueueName[type], data.projectId, 1)
-        if (shouldRateLimit) {
+
+        if (PROJECT_RATE_LIMITER_ENABLED && shouldRateLimit) {
```
abuaboud, line 42: Let's move it to inside redis handler and make early return in both on complete and for shouldRateLimit make it always return false

PR #5319, opened 2024-08-11
fix: show invitation link when inviting someone already invited and smtp isn't configured

AbdulTheActivePiecer:


packages/server/api/src/app/user-invitations/user-invitation.service.ts
```diff
@@ -217,6 +208,20 @@ async function generateInvitationLink(userInvitation: UserInvitation): Promise<s
         path: `invitation?token=${token}&email=${encodeURIComponent(userInvitation.email)}`,
     })
 }
+const handleInvitationLink = async ({ platform, userInvitation }: { platform: Platform, userInvitation: UserInvitation }) => {
```
abuaboud, line 211: nit: enrichWithInvitationLink

and you can use direct parameters, they are only two and have different type

PR #5335, opened 2024-08-12
feat: multi select dropdown with select all & search

alexandrudanpop:
## What does this PR do?

- make PropertyType.MULTI_SELECT_DROPDOWN render a multi-select dropdown
- fix some small bugs in `multi-select-piece-property.tsx` & `multi-select.tsx` (when having multiple options scrollbar is not shown)
- add option to select all / clear all 
- add search

image

image


packages/react-ui/src/app/builder/piece-properties/multi-select-piece-property.tsx
```diff
@@ -23,79 +23,114 @@ type MultiSelectPiecePropertyProps = {
...
   return (
     <MultiSelect
-      modal={true}
-      value={selectedIndicies}
-      onValueChange={sendChanges}
+      value={initialValues as string[]}
+      onValueChange={onChange}
       disabled={disabled}
-      onSearch={(searchTerm) => setSearchTerm(searchTerm ?? '')}
+      filter={true}
```
alexandrudanpop, line 55: I know the implementation has diverged from the original, this is because it's easier to deal with values than indexes, and filtering is already enabled by the MultiSelect component.

PR #5347, opened 2024-08-13
fix: exiting runs didn't take you back to draft + when should the flow be readonly

AbdulTheActivePiecer:


packages/react-ui/src/app/builder/builder-hooks.ts
```diff
@@ -270,7 +273,17 @@ export const createBuilderStore = (initialState: BuilderInitialState) =>
         flowUpdatesQueue.add(updateRequest);
         return { flowVersion: newFlowVersion };
       }),
-    setVersion: (flowVersion: FlowVersion) => set({ flowVersion, run: null }),
+    setVersion: (flowVersion: FlowVersion) => {
+      set((state) => ({
+        flowVersion,
+        run: null,
+        readonly:
+          state.flow.publishedVersionId !== flowVersion.id &&
```
abuaboud, line 281: Shouldn't this be only flowVersion.state === FlowersionState.LOCKED 

Any lock version should be read only.

AbdulTheActivePiecer: If you publish it instantly makes the builder readonly therefore you won't be able to edit anything anymore.

PR #5360, opened 2024-08-14
fix: multi-selects weren't working + block interaction outside the dropdown until user clicks outside like on cloud + selects width

AbdulTheActivePiecer:


alexandrudanpop:
Hi @AbdulTheActivePiecer, I've been working on similar thing, perhaps you could also incorporate it, basically making multi-selects work + making search and enable all / clear all work:
https://github.com/activepieces/activepieces/pull/5335

AbdulTheActivePiecer:
hello @alexandrudanpop we are adding the clear all and add all option later, it is in the punted column in the react board we have atm: https://github.com/orgs/activepieces/projects/95/views/1?pane=issue&itemId=74921265

PR #5362, opened 2024-08-14
fix: make engine reinstall across restarts and make file copying atomic

hazemadelkhalel:

### DIFF
--- a/packages/server/worker/src/lib/utils/cache-handler.ts
+++ b/packages/server/worker/src/lib/utils/cache-handler.ts
@@ -20,10 +20,7 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
         if (!cacheExists) {
             await saveToCache({}, folderPath)
         }
-        const newCache = await readCache(folderPath)
-        if (typeof newCache === 'object') {
-            cached[folderPath] = newCache
-        }
+        cached[folderPath]= await readCache(folderPath)
     }
     const cache = (cached[folderPath] as CacheMap) || {}
     return cache
--- a/packages/server/worker/src/lib/utils/cache-handler.ts
+++ b/packages/server/worker/src/lib/utils/cache-handler.ts
@@ -1,4 +1,3 @@
-import { UUID } from 'node:crypto'
 import { readFile } from 'node:fs/promises'
 import { join } from 'path'
 import { fileExists, memoryLock, threadSafeMkdir } from '@activepieces/server-shared'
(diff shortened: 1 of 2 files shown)


packages/server/worker/src/lib/utils/cache-handler.ts
```diff
@@ -8,7 +9,7 @@ export enum CacheState {
     READY = 'READY',
     PENDING = 'PENDING',
 }
-type CacheMap = Record<string, CacheState>
+type CacheMap = Record<string, CacheState | UUID>
```
abuaboud: replace it with | string

packages/server/worker/src/lib/utils/engine-installer.ts
```diff
@@ -1,12 +1,17 @@
...
 import { logger, memoryLock, SharedSystemProp, system } from '@activepieces/server-shared'
 import { ApEnvironment } from '@activepieces/shared'
-import { cacheHandler, CacheState } from './cache-handler'
+import { cacheHandler } from './cache-handler'
 
 const engineExecutablePath = system.getOrThrow(
     SharedSystemProp.ENGINE_EXECUTABLE_PATH,
 )
 const isDev = system.getOrThrow(SharedSystemProp.ENVIRONMENT) === ApEnvironment.DEVELOPMENT
+const ENGINE_CACHE_ID = randomUUID()
```
abuaboud: you can use nanoId(), instead of UUID we use it everywhere else.

packages/server/worker/src/lib/utils/engine-installer.ts
```diff
@@ -15,27 +20,40 @@ export const engineInstaller = {
...
-
-            const cache = cacheHandler(path) 
-            
-            const engineFileExists =  await cache.cacheCheckState('main.js') === CacheState.READY
             logger.debug({ path }, '[engineInstaller#install]')
-            if (!engineFileExists || isDev) {
-                await copyFile(engineExecutablePath, `${path}/main.js`) 
-                await cache.setCache('main.js', CacheState.READY)
+            const cache = cacheHandler(path) 
+            const isEngineInstalled = await cache.cacheCheckState('ENGINE_INSTALLED') === ENGINE_CACHE_ID
```
abuaboud: move `ENGINE_INSTALLED` to const variable in the top

abuaboud (changes requested):

PR #5422, opened 2024-08-25
feat: reconnect button

AbdulTheActivePiecer:
## What does this PR do?

Fixes # (issue)


packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
```diff
@@ -187,22 +230,25 @@ const CreateOrEditConnectionDialog = React.memo(
                   <OAuth2ConnectionSettings
                     authProperty={piece.auth as OAuth2Property<OAuth2Props>}
                     piece={piece}
+                    reconnectConnection={reconnectConnection}
```
abuaboud, line 230: You only need reconnectConnection name not the whole object, can you please change it to the name only

packages/react-ui/src/app/builder/step-settings/piece-settings/connection-select.tsx
```diff
@@ -40,41 +47,87 @@ const ConnectionSelect = React.memo((params: ConnectionSelectProps) => {
     pieceName: params.piece.name,
     cursor: undefined,
     limit: 100,
-    projectId: authenticationSession.getProjectId(),
+    projectId: authenticationSession.getProjectId() ?? '',
   });
-
-  const addBrackets = (str: string) => `{{connections['${str}']}}`;
+  const currentlySelectedConnection = useMemo(() => {
```
abuaboud: You no longer need to traverse the whole connections page, you just need the name :") and it's already in form value.

packages/react-ui/src/app/builder/step-settings/piece-settings/connection-select.tsx
```diff
@@ -40,41 +47,87 @@ const ConnectionSelect = React.memo((params: ConnectionSelectProps) => {
...
+            <Select disabled={params.disabled}>
+              <SelectContent>
+                <SelectLoader />
+              </SelectContent>
+            </Select>
+          )}
+          {!isLoading && (
+            <FormItem>
+              <CreateOrEditConnectionDialog
+                reconnectConnection={currentlySelectedConnection}
```
abuaboud: This breaks +New Connection functionality

packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
```diff
@@ -74,15 +88,34 @@ const CreateOrEditConnectionDialog = React.memo(
...
+                ' Name can only contain letters, numbers and underscores',
+              ),
+            }),
+          }),
+        ]),
       }),
       overrideSchema,
     ]);
 
+    useEffect(() => {
```
abuaboud: I don't think you need useEffect, reconnectionConnection is property, so it will trigger render

Check this: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

abuaboud (changes requested):

packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
AbdulTheActivePiecer: I need it to know the connection type for the oauth dialog.

packages/react-ui/src/app/builder/step-settings/piece-settings/connection-select.tsx
AbdulTheActivePiecer: I need it to know the connection type for the oauth dialog.

packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
AbdulTheActivePiecer: I now set the name in the createDefaultValues function and I added a key to the component so it rerenders every time the user changes the connection.

packages/react-ui/src/app/builder/step-settings/piece-settings/connection-select.tsx
AbdulTheActivePiecer: fixed.

PR #5444, opened 2024-08-28
feat(slack): set user's status, find user by slack handle  actions and make channel optional for trigger

AdamSelene:
## What does this PR do?

- add action to set the connected user's status
- add action to find a user by Slack handle
- make channel optional for "new message" trigger (to receive messages on **all** public channels)

ANNOUNCEMENT=true

### DIFF
--- a/packages/pieces/community/slack/package.json
+++ b/packages/pieces/community/slack/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-slack",
-  "version": "0.6.3"
+  "version": "0.7.0"
 }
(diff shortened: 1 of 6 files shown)


kishanprmr:
@AdamSelene new message trigger gives sample data if channel is not selected. Can you test it again and let me know if you face any issue.

AdamSelene:
@kishanprmr what do you want me to test exactly?
I had to return the static test data because I have no channel to retrieve dynamic data from (I could technically list all channels, get the channel ID of the mandatory `#general` one, and hope the connection has access to that channel, but I did not think it was worth it) - happy to add a better solution if you have one in mind!

kishanprmr (approved):

Issue #4350, opened 2024-04-03
[BUG]: hardcoded schema name for Postgres migrations


tgranqvist:
**Describe the bug**
I am running Activepieces in Docker, my composefile spins up a Postgres server as well. I tried using a custom schema name `activepieces` in the database, but it fails on the migrations. E.g. `1677286751592-billing.ts` has hardcoded the schema `public` when dropping all the indexes. Same goes for all the migration files I checked, they all hardcode the schema to public.

`js
await queryRunner.query(
    'DROP INDEX "public"."idx_plan_stripe_customer_id"',
)
`

**To Reproduce**
1. Set up a Postgres database with a custom schema
2. Add a user for Activepieces
3. Alter the DB user for Activepieces to use that schema as default: `ALTER ROLE activepieces SET search_path TO activepieces;`
4. Configure Docker container/composefile to use above DB user
5. Start activepieces
6. See error

**Expected behavior**
I expect the container to start and do all the migrations. I does create tables, as they are not prefixed with hardcoded public scheme.

**Additional context**
SQL to create user, schema, grant access, and set default schema for said user
`sql
CREATE USER activepieces WITH ENCRYPTED PASSWORD '1234';
CREATE SCHEMA activepieces;
GRANT ALL ON SCHEMA activepieces TO activepieces;
ALTER ROLE activepieces SET search_path TO activepieces;
`
These tables are created **in the correct schema** before the migration failure. So the database setup works.
![image](https://github.com/activepieces/activepieces/assets/24589919/037a6d01-b12f-4038-835f-58d9839ced30)

abuaboud:
Hi @tgranqvist 

Interesting bug, Let me check `typeorm` docs and if there is way to support multiple schemas.

tgranqvist:
The (to me) obvious solution is that it shouldn't specify the schema at all. As I could see, tt doesn't for anything else, neither creating nor dropping tables or even creating indices. Only when dropping an index it seems to prepend the schema, hardcoded to `public`. Bug, oversight or intentional, I can't say. Thanks for having a look!

And maybe stating the obvious here: I'm not trying to use multiple schemas with one Activepiece installation. Just **another** one than the default `public`.

abuaboud:
Hi @tgranqvist 

I understand, I think the main problem is that `typeorm` seems designed to generate schema for your own environment and not for configurable environments, one possible way is to pass the schema as environment variable and use it in their all entity manager and remove public. from all migration.

[...]

PR #5450, opened 2024-08-29
fix: remove all references to the public schema in migration files

jasonfill:
## What does this PR do?

Several migration files reference the `public` schema while others did not. This is inconsistent with the rest of the application and creates challenges with Activepieces is not being run in the `public` schema. 

Fixes #[4350](https://github.com/activepieces/activepieces/issues/4350)


abuaboud:
I reverted the one command, that migrate existing data of user table, It will conflict with "user" table in postgres, so I kept it as it is.

abuaboud (approved):

PR #5490, opened 2024-09-03
feat: add customizable expiry date (in seconds) to user invitations and include in request payload

hazemadelkhalel:
This PR introduces a new feature for customizing user invitation expiry dates. The expiry date is now set in seconds and included in the request payload when sending an invitation (it's not required & by default, it stays for a day).

in the body of the request
```
{
   ....,
   ....,
   ....,
   "invitationExpirySeconds" : 24 * 60 * 60        //One Day
}


abuaboud:
https://github.com/activepieces/activepieces/pull/5493

PR #5669, opened 2024-09-20
feat: improve piece selector experience 

cezudas:
## What does this PR do?

The Angular version of the pieces search result included the suggestions (displayed as chips).
It was useful as the results included information such as:
- the action|trigger relevant to the search query
- the name of the piece

The Reactjs re-write no longer includes the suggestions and forces the user to perform an additional click on the piece name to reveal all available actions. To make matter worse, image that the list of available actions is very long and now the user needs to scroll down and find the search query relevant piece.

With this PR the piece search result view will list all piece suggestions together for any given piece (under with piece name), rendering the suggestions as clickable list items. This should simplify selection.

### Current UI: 

Screenshot 2024-09-20 at 16 14 45

### Suggested UI:

Screenshot 2024-09-20 at 16 03 17

### DIFF
--- a/packages/react-ui/src/app/builder/flow-canvas/nodes/step-node.tsx
+++ b/packages/react-ui/src/app/builder/flow-canvas/nodes/step-node.tsx
@@ -185,9 +185,9 @@ const ApStepNode = React.memo(({ data }: { data: ApNode['data'] }) => {
         setAllowCanvasPanning(true);
       }}
       key={data.step?.name}
-      ref={setNodeRef}
-      {...attributes}
-      {...listeners}
+      ref={openPieceSelector ? null : setNodeRef}
+      {...(!openPieceSelector ? attributes : {})}
+      {...(!openPieceSelector ? listeners : {})}
     >
       <div
         className="absolute text-accent-foreground text-sm opacity-0 transition-all duration-300 group-hover:opacity-100 "
--- a/packages/react-ui/src/app/builder/piece-properties/text-input-with-mentions/text-input-utils.ts
+++ b/packages/react-ui/src/app/builder/piece-properties/text-input-with-mentions/text-input-utils.ts
@@ -1,5 +1,6 @@
 import { MentionNodeAttrs } from '@tiptap/extension-mention';
 import { JSONContent } from '@tiptap/react';
+
 import { StepMetadata } from '@/features/pieces/lib/types';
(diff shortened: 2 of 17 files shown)


cezudas:
I created this screenshots in my localhost environment. Unfortunately I was not able to install more pieces :( so that I can include multiple search result entries in my screenshots.

Screenshot 2024-09-20 at 15 35 26

```
err: {
[API]       "type": "Error",
[API]       "message": "directoryPath for @activepieces/piece-amazon-s3 is null or undefined is null or undefined",
[API]       "stack":
[API]           Error: directoryPath for @activepieces/piece-amazon-s3 is null or undefined is null or undefined
[API]               at assertNotNullOrUndefined (/Users/alexandru-danpop/openops/workflow-editor/dist/packages/server/api/main.js:1311:15)
[API]               at LocalPieceManager.<anonymous> (/Users/alexandru-danpop/openops/workflow-editor/dist/packages/server/api/main.js:34669:55)
[API]               at Generator.next (<anonymous>)
[API]               at fulfilled (/Users/alexandru-danpop/openops/workflow-editor/node_modules/tslib/tslib.js:166:62)
[API]       "context": {
[API]         "[PieceManager#install]": {
[API]           "projectPath": "/Users/alexandru-danpop/openops/workflow-editor/cache",
[API]           "pieces": [
[API]             {
[API]               "packageType": "REGISTRY",
[API]               "pieceName": "@activepieces/piece-amazon-s3",
[API]               "pieceVersion": "0.3.5",
[API]               "scope": "PROJECT",
[API]               "pieceType": "CUSTOM"
[API]             }
[API]           ]
[API]         }
[API]       }
[API]     }

```

abuaboud:
Hi @cezudas 

Thank you for opening the PR, You came at the right time! 

Here is my thoughts, I am thinking If we should hide the right sidebar when there is search query, that way It shows the piece and it suggestion under one column view.

the one column is less confusing, but I am worried about a case where the search went wrong for example you looked for "add row" but google sheets suggestion have "Insert row" so let say add row didn't show up then the user will not be able to explore the action of google sheets alone, except going to scroll without a search on google sheet and click on it to check whether add row is there or not. 

I will leave it to @AbdulTheActivePiecer to take it from here.

AbdulTheActivePiecer (approved):

PR #5681, opened 2024-09-24
feat(github): use octokit sdk for actions and triggers

AdamSelene:
## What does this PR do?

Switch existing actions and triggers to official octokit SDK
The output changes (hence the major version bump)
- `success: true` -> `status: 200`
- `<entity_name>: {…}` -> `data: {…}`
- response headers are available


packages/pieces/community/github/src/lib/common/index.ts
```diff
@@ -1,180 +1,120 @@
...
-		const response = await httpClient.sendRequest<GithubRepository[]>(request);
-		repos.push(...response.body);
-
-		hasNext = response.headers?.link?.includes('rel="next"');
-		page += 1;
-	} while (hasNext);
-
-	return repos;
+async function getUserRepo(authProp: OAuth2PropertyValue) {
+  const client = new Octokit({ auth: authProp.access_token });
```
kishanprmr, line 94: @AdamSelene can you use pagination for these dropdown? You can check it [here](https://github.com/octokit/octokit.js?tab=readme-ov-file#pagination)

packages/pieces/community/github/src/lib/common/index.ts
```diff
@@ -1,180 +1,120 @@
...
-}
-export interface GithubIssueLabel {
-	id: string;
-	name: string;
+  authProp: OAuth2PropertyValue,
+  owner: string,
+  repo: string
+) {
+  const client = new Octokit({ auth: authProp.access_token });
+  return (await client.rest.issues.listLabelsForRepo({ owner, repo })).data;
```
kishanprmr: pagination here also

packages/pieces/community/github/src/lib/common/index.ts
AdamSelene: Good catch, sorry I missed it!

kishanprmr (approved):

PR #5696, opened 2024-09-26
fix: unneeded frontend remove loops sample data iterations code

AbdulTheActivePiecer:


packages/react-ui/src/app/builder/test-step/test-sample-data-viewer.tsx
```diff
@@ -34,12 +29,11 @@ const TestSampleDataViewer = React.memo(
     currentSelectedData,
     errorMessage,
     lastTestDate,
-    type,
     children,
   }: TestSampleDataViewerProps) => {
     const formattedData = useMemo(
```
abuaboud: This variable can be removed, just use currentSelectedData

PR #5700, opened 2024-09-27
fix: store all action / trigger files into file storage (s3 / db)

abuaboud:
## What does this PR do?

- Removes Local File / Memory File Implementation in Engine
- Store Actions File Into S3 / Database, to support re-run from failed step, or to be able to read the file in paused run, previously the file on local disk would not persist between runs 
- Sending multipart files to webhook now converts them to links instead of base64, which would cause 
- When user request a file, and It's s3, we generate presigned link and redirect them there to avoid download file into API server and serve it again to the request.

Fixes # (issue)


anasbarg (approved):

anasbarg:
lgtm

PR #5717, opened 2024-09-29
fix: move invalid code template to local variable

abuaboud:
## What does this PR do?

Fixes # (issue)


abuaboud:
Thank you @kamal for catching this!

PR #5724, opened 2024-10-01
fix: import flows inside builders now overwrites current flow

AbdulTheActivePiecer:


packages/react-ui/src/components/ui/form.tsx
```diff
@@ -158,7 +158,7 @@ const FormMessage = React.forwardRef<
     <p
       ref={ref}
       id={formMessageId}
-      className={cn('text-sm font-medium text-destructive', className)}
+      className={cn(formMessageClass, className)}
```
abuaboud: nit: It's better to keep it inline to close where your need it, the variable is also used once, It's better inline.

Because imagine someone reading it, they will have to scroll up and check the variable deceleration

packages/react-ui/src/components/ui/input.tsx
```diff
@@ -1,22 +1,61 @@
...
         type={type}
-        className={cn(
-          'flex-grow h-10 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:cursor-pointer file:float-right file:mx-2 file:border-1 file:border-solid file:border-muted file:rounded-sm file:bg-accent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring focus-visible:ring-offset-1 disabled:cursor-not-allowed disabled:opacity-50 box-border',
-          className,
-          {
-            'cursor-pointer': type === 'file',
-          },
-        )}
-        ref={ref}
+        className={cn(inputClass, className)}
```
abuaboud, line 57: why not inline inputClassName?

packages/react-ui/src/features/flows/components/import-flow-dialog.tsx
```diff
@@ -106,7 +164,10 @@ const ImportFlowDialog = ({
           <Button onClick={handleSubmit} loading={isPending}>
             {t('Import')}
           </Button>
-        </DialogFooter>
+        </div>
+        {errorMessage && (
+          <div className={cn(formMessageClass, 'mt-4')}>{errorMessage}</div>
```
abuaboud: Why not use ErrorMessage? or inline the formMessageClass if it not create a component or use `ErrorMessage`?

abuaboud (changes requested):

packages/react-ui/src/components/ui/form.tsx
AbdulTheActivePiecer: the variable is exported, it is not intended to be used once, we have at least two places where we don't use the FormMessage class but still need validation, that's where the variable is used, so error messages don't look different in multiple places.

packages/react-ui/src/components/ui/input.tsx
AbdulTheActivePiecer: Because the class is used twice, once for the file input and once for the normal input, both must always be the same.

packages/react-ui/src/features/flows/components/import-flow-dialog.tsx
AbdulTheActivePiecer: Answered in a previous comment

PR #5763, opened 2024-10-05
fix: dev setup on windows

HoldYourWaffle:
## What does this PR do?

As discussed on Discord, this fixes weird shenanigans when setting up a dev environment on Windows. See [this StackOverflow answer](https://stackoverflow.com/a/59830014/10434371).

Keeping this as a draft for now to investigate the best way to avoid an RCE oopsie.


abuaboud:
This file is not used outside the development environment, good job on the fix <3

abuaboud (approved):

PR #5775, opened 2024-10-07
fix: uncontrolled to controlled input warning in SignInForm component

cezudas:
## What does this PR do?

Fixes a warning on local-dev indicating that an input element is switching from uncontrolled to controlled state by providing a `defaultValues` object to useForm hook which ensures all form input values are initialized.

Screenshot 2024-10-07 at 09 56 30

Reference material [here](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable).


abuaboud:
@allcontributors please add @cezudas for bug

PR #5779, opened 2024-10-07
fix: long step name label overlapping the step node

kamal:
## What does this PR do?

Long step names grow into the step node. By using `left: 100%` instead of negative `right`, we ensure the label grows to the right.

Markup 2024-10-07 at 12 16 23

Fixes # (issue)


AbdulTheActivePiecer:
Thanks @kamal, I always forget writing that bug down when I see it :joy and forget it.

PR #5810, opened 2024-10-11
feat: improve human input piece user experience 

abuaboud:
## What does this PR do?

- Renamed "User Input" to "Human Input"
- Removed "Return Chat Response"
- Added markdown above "Waiting for response" as it's more obvious for the user, so they don't forget to add action to last step
- You can use Respond with Markdown to reply for Chatbot
- No longer need to save the chat in  the store, later we can extend that the frontend send the whole context again in webhook, or use separate solution so we don't have store constraints.
- Changed the piece icon
- Fixed Form description for Toggle

**Todo in the next PR**
- Support Respond with File for Chat
- Add Meaningful error in the user interface If the flow fails to reply or has an error.
- Revisit testing instructions for testing chat interface.


abuaboud:
@anasbarg Can you please handle the TODO in the next pr

anasbarg (approved):

PR #5830, opened 2024-10-14
feat(file-helper): Add Encoding property to createFile action

janwytze:
## What does this PR do?

This PR adds an encoding property to the `createFile` action. We need this to convert base64 to a file. The default is still utf8, so no flows should break because of this PR.

Fixes # (issue)

ANNOUNCEMENT=true


abuaboud:
Hi @janwytze 

I bumped the version in package.json and also added ?? operator in case the ui didn't update it.

Thank you :D

PR #5834, opened 2024-10-14
feat(core): add pieceConfig mechanism

AdamSelene:
## What does this PR do?

- add `AP_PIECES_CONFIG` env var to store piece-specific configuration - e.g. `{"@activepieces/piece-http": {"someKey": "someValue"}}`
- inject `pieceConfig` in actions' `run()` method


abuaboud:
Let's take it offline, I believe this behavior might be disruptive if  pieces secretly started to depend on these values, and they are also will not be easy to setup for other customers.

I understand the concern! Let's scope it as part of review process ticket

PR #5835, opened 2024-10-15
Olivier sambourg/feat http add configurable deny list of headers

AdamSelene:
## What does this PR do?

This PR builds on top of #5834 (and will need to be rebased) to add a configurable deny list of headers
e.g. `AP_PIECES_CONFIG='{"@activepieces/piece-http": {"deniedHeaders": ["Authorization", "X-Api-Key"]}}'`


abuaboud:
Let's take it offline, I believe this behavior might be disruptive if  pieces secretly started to depend on these values, and they are also will not be easy to setup for other customers.

I understand the concern! Let's scope it as part of review process ticket

PR #5856, opened 2024-10-16
fix: project.externalId uniqueness apply to non-deleted projects only

kamal:
## What does this PR do?

Currently, you can't create a new project with an externalId of a deleted project

Fixes # (issue)


abuaboud:
Thank you, I added the condition the index as well, and updated misalignment on platform entity on community edition.

abuaboud (approved):

PR #5909, opened 2024-10-22
fix(send-file)

bilal-devv:
## What does this PR do?

Fixes # (issue)

This fix my upload file method, using content type formdata instead of json


abuaboud:
@obenazouz-wedof can you please take look

PR #5964, opened 2024-10-31
feat: google-calendar add `Expand Recurring Event` option

ikus060:
## What does this PR do?

Add options to Goolgle Calendar Get Events to expand recurring events or not similar to Google Calendar Trigger

### DIFF
--- a/packages/pieces/community/google-calendar/src/lib/actions/get-events.ts
+++ b/packages/pieces/community/google-calendar/src/lib/actions/get-events.ts
@@ -54,6 +54,12 @@ export const getEvents = createAction({
       displayName: 'Date to',
       required: false,
     }),
+    singleEvents: Property.Checkbox({
+			displayName: 'Expand Recurring Event?',
+      description: "Whether to expand recurring events into instances and only return single one-off events and instances of recurring events, but not the underlying recurring events themselves.",
+			required: true,
+			defaultValue: false,
+		}),
   },
   async run(configValue) {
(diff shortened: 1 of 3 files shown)


ikus060:
I would need some help to understand why the CICD pipeline is failing.

packages/pieces/community/google-calendar/src/lib/actions/get-events.ts
```diff
@@ -63,11 +69,14 @@ export const getEvents = createAction({
       end_date,
       search,
       event_types,
+      singleEvents,
     } = configValue.propsValue;
     const { access_token: token } = configValue.auth;
     const queryParams: Record<string, string> = { showDeleted: 'false' };
     let url = `${googleCalendarCommon.baseUrl}/calendars/${calendarId}/events`;
 
+    queryParams['singleEvents'] = singleEvents ? 'true' : 'false';
```
abuaboud: Can you please check if singleEvents is not null then set this parameter, in that way we can achieve backward compatibility and then please change the version to 0.5.6

ikus060: Sure, I will plan some time to work on this today.

Thanks for your review.

ikus060: I'm not sure this null verification was required because the default value is false. Any how. I've made it explicit.

Change done.

abuaboud (approved):

PR #5967, opened 2024-11-03
fix(bulk-action): resolve issue with selecting multiple rows across different pages in tables for connections, flows, and runs

hazemadelkhalel:


hazemadelkhalel:
NOTE: Check whether the run table pagination works fine or not.
@abuaboud

packages/react-ui/src/app/connections/connection-table.tsx
```diff
@@ -186,14 +75,180 @@ const filters = [
...
+    RowDataWithActions<AppConnectionWithoutSensitiveData>,
+    unknown
+  >[] = [
+    {
+      id: 'select',
+      header: ({ table }) => (
+        <Checkbox
+          checked={
+            table.getIsAllPageRowsSelected() ||
+            (table.getIsSomePageRowsSelected() && 'indeterminate')
```
AbdulTheActivePiecer: what is this condition ? I see it everywhere, like why is this string 'indeterminate' in the condition ?

packages/react-ui/src/app/routes/flows/index.tsx
```diff
@@ -302,8 +356,10 @@ const FlowsPage = () => {
   const bulkActions: BulkAction<PopulatedFlow>[] = useMemo(
     () => [
       {
-        render: (selectedRows, resetSelection) => {
-          const isDisabled = selectedRows.length === 0;
+        render: (_, resetSelection) => {
+          const isDisabled = selectedRowIds.length === 0;
+          console.log('selectedRowIds', selectedRowIds);
+          console.log('isDisabled', isDisabled);
```
AbdulTheActivePiecer: remove these, please.

packages/react-ui/src/app/routes/flows/index.tsx
```diff
@@ -324,7 +380,7 @@ const FlowsPage = () => {
                     hasPermission={userHasPermissionToPushToGit}
                   >
                     <PushToGitDialog
-                      flowIds={selectedRows.map((flow) => flow.id)}
+                      flowIds={selectedRowIds.map((flow) => flow.version.id)}
```
AbdulTheActivePiecer: rename selectedRowIds to selectedRows because that's what they are, if you want them to be selectedRowIds, change the type to string array and set the value to the ids only in the setter, and do this in all the components that have that variable name.

packages/react-ui/src/app/routes/flows/index.tsx
```diff
@@ -186,15 +190,65 @@ const FlowsPage = () => {
...
+                return !table
+                  .getRowModel()
+                  .rows.some((r) => r.original.version.id === row.version.id);
+              });
+              setSelectedRowIds(filteredRowIds);
+            }
+          }}
         />
       ),
+      cell: ({ row }) => {
```
AbdulTheActivePiecer, line 222: I will leave this comment here because I can't comment on the place that's broken.
Basically, all places that should refresh the table after an action don't work, setRefresh is not making the table refresh nor the folders list.

AbdulTheActivePiecer (changes requested):

packages/react-ui/src/app/routes/flows/index.tsx
AbdulTheActivePiecer: This bug is also in the connections table, ~~probably in the runs table as well~~

AbdulTheActivePiecer (approved):

PR #5982, opened 2024-11-05
fix: add error message extractor from trigger failed test

AbdulTheActivePiecer:
Currently when a trigger test fails we don't show a comprehensive error message we need to fix that, I added a way for us to show it in the UI, the only issue is that it is being returned from the backend as {}.


kishanprmr:
Testing this for Slack and Google Sheets triggers for false values only displays the same message, “Request failed with status code 400,” rather than the actual underlying API error.

PR #6005, opened 2024-11-07
refactor: delete branch and switch to router

abuaboud:
## What does this PR do?

Fixes # (issue)


abuaboud:
It's now tested on both sqlite and postgres.

AngelPedroza:
Hi everyone,

I would like to ask why this "Router" piece was removed. I was following forums and it seemed like a great addition and a useful tool to reduce the time needed to create flows.

I understand if removing the Branch piece wasn't right, but why not keep both pieces? @abuaboud

PR #6063, opened 2024-11-17
feat(voice-ai): speech to text and text to speech actions

kishanprmr:
## What does this PR do?

Fixes # (issue)


kishanprmr:
@abuaboud the Speech-to-Text endpoint requires Form Data as input, but it seems the proxy endpoint doesn't support it. I tried using the toFile upload functionality from OpenAI, but it doesn't seem to work with JSON input.

PR #6097, opened 2024-11-21
fix: prevent false matches in piece directory detection

### LINKED ISSUE #6096: [BUG]: Piece directory detection fails when working path contains piece name
**Describe the bug**
The current piece directory detection logic using `includes()` causes false matches when the working directory path contains a piece name as a substring. This makes it impossible to develop pieces when your working path contains the piece name.

**To Reproduce**
1. Clone activepieces into a path that contains a piece name
   Example: `/home/user/mycompany/activepieces`
2. Try to develop a piece named 'mycompany'
3. The piece directory detection fails because it matches 'mycompany' in the working path

**Expected behavior**
- Piece directory detection should only match exact directory names
- Working path should not interfere with piece detection
- `mycompany` piece should work regardless of the development path

**Impact**
This affects:
- Enterprise customers who might have company names in their development paths
- Any developer whose working directory path contains strings that match piece names
- Development environment reliability

**Proposed Solution**
Replace `includes(pieceName)` with `endsWith('/'+pieceName)` to:
1. Only match exact directory names
2. Require a directory separator before the name

dev-instasent:
## What does this PR do?

This PR fixes the piece directory detection logic to prevent false matches when working paths contain piece names or when piece names are substrings of other directories.

The current implementation using `includes()` causes false matches in scenarios like:
- Working in `/home/user/mycompany/activepieces` with a piece named 'mycompany'
- A piece named 'mail' would incorrectly match a 'gmail' directory

The solution:
- Changes directory matching from `includes(pieceName)` to `endsWith('/'+pieceName)`
- Ensures exact directory name matching by requiring:
  1. The full piece name
  2. A directory separator before the name

Fixes #6096


abuaboud (dismissed):

dev-instasent:
@abuaboud Done!

PR #6176, opened 2024-11-28
fix(dropbox): fix support of accented characters in path property

simonc:
## What does this PR do?

It is currently impossible to upload a file in Dropbox using AP if its path contains accented characters.

Dropbox's API does not support UTF-8 characters so they need be encoded a specific way. The encoding code can be found here https://www.dropbox.com/developers/reference/json-encoding

This changes applies the proposed JS code to the `path` property.


kishanprmr:
@allcontributors please add @simonc for plugin

kishanprmr (approved):

PR #6210, opened 2024-12-02
fix(common): use async validation in zod

abuaboud:
## What does this PR do?

Fixes # (issue)


abuaboud:
Thank you @kamal for pointing that all pieces are waiting while we don't use parseAsync for async refinement

PR #6219, opened 2024-12-03
feat(crypto): generate hmac signature

matthieu-lombard:
## What does this PR do?

Adding a new **Hmac Signature** action in **Crypto** piece

### DIFF
--- a/packages/pieces/community/crypto/src/index.ts
+++ b/packages/pieces/community/crypto/src/index.ts
@@ -2,15 +2,21 @@ import { createPiece, PieceAuth } from '@activepieces/pieces-framework';
 import { PieceCategory } from '@activepieces/shared';
 import { generatePassword } from './lib/actions/generate-password';
 import { hashText } from './lib/actions/hash-text';
+import { hmacSignature } from './lib/actions/hmac-signature';
+
+export const hmacSecretKey = PieceAuth.SecretText({
+  displayName: 'Secret key',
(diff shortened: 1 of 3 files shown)


packages/pieces/community/crypto/src/index.ts
```diff
@@ -2,15 +2,21 @@ import { createPiece, PieceAuth } from '@activepieces/pieces-framework';
...
+export const hmacSecretKey = PieceAuth.SecretText({
+  displayName: 'Secret key',
+  required: true,
+});
 
 export const Crypto = createPiece({
   displayName: 'Crypto',
   description: 'Generate random passwords and hash existing text',
-  auth: PieceAuth.None(),
+  auth: hmacSecretKey,
```
matthieu-lombard: It seems to have added the connection to all actions ... I just want on my `hmacSignature` action ? It is not possible ?

abuaboud: Hi @matthieu-lombard It's not currently possible, I think we will handle it within the dealing with secrets inside the flow in general.

Can you please in time being move it to the action?

matthieu-lombard: Yes I see that, I push this right now with the secret key in a simple text input

abuaboud (approved):

abuaboud (approved):

PR #6290, opened 2024-12-10
fix(google-sheets): "cannot read properties of null" error when using…

wise-introvert:
Fixes the following error when using the `add-or-update-row` trigger in the google sheets piece:
```
TypeError: Cannot read properties of null (reading '0')
  at ITrigger.<anonymous> (/usr/src/app/"/tmp"/sandbox/PIECE-pieceName-@activepieces/piece-google-sheets-pieceVersion-0.11.7/node_modules/packages/pieces/community/google-sheets/src/lib/triggers/new-or-updated-row.trigger.ts:202:38)
  at Generator.next (<anonymous>)
  at fulfilled (/usr/src/app/"/tmp"/sandbox/PIECE-pieceName-@activepieces/piece-google-sheets-pieceVersion-0.11.7/node_modules/.pnpm/tslib@1.14.1/node_modules/tslib/tslib.js:112:62)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  ```

### CODE REVIEW

**Comment by wise-introvert** [comment]
I've just added sanity checks to variables that could be null or undefined. This helps avoid the error from bubbling up to the client side. Please let me know if you want me to change anything or use a different approach. Thanks!

**Comment by abuaboud** [comment]
I don't think current code solves any issue, as both arrays will not be null.

You need to change the line 202 so it handle array out of bound, because oldValuesHashes can be smaller than oldValuesHashes[row].

I believe this is the issue.
```
			const oldRowHash = oldValuesHashes[row];
```

**Comment by wise-introvert** [comment]
@abuaboud gotcha. Thanks for the quick reply 😄. I'll make changes and push it here. Thanks, again!

**Comment by wise-introvert** [comment]
@abuaboud how does this look:
```typescript
			const oldRowHash = row < oldValuesHashes.length ? oldValuesHashes[row] : undefined;
```

abuaboud (approved):

abuaboud (approved):

PR #6314, opened 2024-12-12
feat(tables): add tables piece

### LINKED ISSUE #6269: Restful API & Data Types
The goal to create the restful structure and the data types.

amrdb:
## What does this PR do?

Types and Rest API Controllers for the tables feature.

Fixes #6269


packages/server/api/src/app/tables/field/field.controller.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    fastify.get('/:id/fields', GetFieldsRequest, async (request, reply) => {
+        await reply.status(StatusCodes.OK).send({})
+    },
+    )
+}
+
+const CreateRequest = {
+    config: {
+        allowedPrincipals: ALL_PRINCIPAL_TYPES,
```
abuaboud: Let's only Allow [ENGINE, USER]

The unknown means unauthenticated and api key is not applicable.

packages/shared/src/lib/tables/cell.ts
```diff
@@ -0,0 +1,11 @@
+import { Static, Type } from '@sinclair/typebox'
+import { BaseModelSchema } from '../common'
+
+export const Cell = Type.Object({
+    ...BaseModelSchema,
+    recordId: Type.String(),
+    fieldId: Type.String(),
+    value: Type.Any(),
```
abuaboud: Nit: Type.Uknowen()  would be better, so we don't use it freely in the code without casting.

abuaboud (changes requested):

packages/server/api/src/app/tables/field/field.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+import { fieldService } from './field.service'
+
+export const fieldController: FastifyPluginAsyncTypebox = async (fastify) => {
+
+    fastify.post('/:id/fields', CreateRequest, async (request, reply) => {
+        const response = await fieldService.create({
+            tableId: request.params.id,
+            request: request.body,
+        })
+        await reply.status(StatusCodes.OK).send(response)
```
abuaboud: Nit: change it to created 201

packages/server/api/src/app/tables/field/field.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+    )
+
+    fastify.get('/:id/fields/:fieldId', GetFieldByIdRequest, async (request, reply) => {
```
abuaboud: Let remove the nesting, usually in all other endpoints we decided that all endpoints shouldn't have nesting in general as we might run in situations in future where there is multiple parent for the entity

packages/server/api/src/app/tables/field/field.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+            return
+        }
+
+        await reply.status(StatusCodes.OK).send(response)
+    },
+    )
+
+    fastify.delete('/:id/fields/:fieldId', DeleteFieldRequest, async (request) => {
+        return fieldService.delete({
+            tableId: request.params.id,
```
abuaboud: We need to handle the security by appending project id from the principal to all queries

packages/server/api/src/app/tables/field/field.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    fastify.get('/:id/fields/:fieldId', GetFieldByIdRequest, async (request, reply) => {
+        const response = await fieldService.getById({
+            tableId: request.params.id,
+            id: request.params.fieldId,
+        })
+
+        if (!response) {
+            await reply.status(StatusCodes.NOT_FOUND).send('Field not found')
```
abuaboud: We throw entity not found exception and it automatically converted to 404 by error handlee in certain style

packages/server/api/src/app/tables/table/table.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (!response) {
+            await reply.status(StatusCodes.NOT_FOUND).send('Table not found')
+            return
+        }
+        
+        await reply.status(StatusCodes.OK).send(response)
+    },
+    )
+
+    fastify.post('/:id/import', ImportRequest, async (request, reply) => {
```
abuaboud: Let remove it and we can always add it later

abuaboud: Let remove it and we can always add it later

packages/server/api/src/app/tables/table/table.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+
+    async delete({ projectId, id }: { projectId: string, id: string }): Promise<void> {
+        await tableRepo().delete({
+            projectId,
+            id,
+        })
+    },
+
+    // async import({ projectId, id, data }: { projectId: string, id: string, data: any }): Promise<void> {
```
abuaboud: Deadcode are usually stay there for while, let remove it and add it back when it's implemented

PR #6417, opened 2024-12-29
feat(tables-api): fields ordering using position

amrdb:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/server/api/src/app/database/database-connection.ts
+++ b/packages/server/api/src/app/database/database-connection.ts
@@ -169,4 +169,4 @@ export function APArrayContains<T>(
 }
 
 // Uncomment the below line when running `nx db-migration server-api name=<MIGRATION_NAME>` and recomment it after the migration is generated
-export const exportedConnection = databaseConnection()
+// export const exportedConnection = databaseConnection()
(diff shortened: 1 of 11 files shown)


packages/server/api/src/app/database/database-connection.ts
```diff
@@ -169,4 +169,4 @@ export function APArrayContains<T>(
 }
 
 // Uncomment the below line when running `nx db-migration server-api name=<MIGRATION_NAME>` and recomment it after the migration is generated
-// export const exportedConnection = databaseConnection()
\ No newline at end of file
+export const exportedConnection = databaseConnection()
```
abuaboud: comment this again

packages/server/api/src/app/tables/field/field.entity.ts
```diff
@@ -19,6 +19,11 @@ export const FieldEntity = new EntitySchema<FieldSchema>({
             type: String,
             enum: FieldType,
         },
+        position: {
+            type: Number,
+            nullable: false,
+            unique: true,
```
abuaboud: It should not be unique, as there be multiple tables.

Either you make it unique based on other attributes like table Id

PR #6450, opened 2025-01-05
feat: add default value action

### LINKED ISSUE #5418: New Text Action: Default Value
One useful feature other platforms have is a simple "Default Value" function. This can be emulated with branch and storage, but that's quite a few steps!

I would like a new Action in the Text Helper called "Default Value". It would have two inputs: An input string (likely from a previous output) and a default string. The output would simply be the input string or the default if the input is null. Something very similar already exists in the Storage Get function.

This can be emulated by using a branch and Storage but that takes a few steps. It can also be emulated with a put and immediate get but that is inefficient too.

**Additional context**
Just a simple feature request!

Anmol-Gup:
## What does this PR do?

Default value action in text helper piece.

Fixes #5418

### DIFF
--- a/packages/pieces/community/text-helper/package.json
+++ b/packages/pieces/community/text-helper/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-text-helper",
-  "version": "0.2.7"
+  "version": "0.3.0"
 }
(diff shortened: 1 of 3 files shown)


AbdulTheActivePiecer:
Cheers @Anmol-Gup, thank you for your contribution.
@allcontributor please add @Anmol-Gup  for plugin

PR #6463, opened 2025-01-07
feat(spotify): get saved tracks action

### LINKED ISSUE #6462: Spotify - Add Saved Tracks action
**Describe your usecase.**
The existing 'Get Playlist Items' action returns a list of tracks in a playlist, but this can't be used to fetch a user's saved tracks because Spotify has [a separate endpoint](https://developer.spotify.com/documentation/web-api/reference/get-users-saved-tracks) for this.

Add a new action so that user's can fetch their saved tracks using this endpoint.

I tried using the existing 'Custom API Call' for this. However, the scopes in the Spotify OAuth2 object are hardcoded, and this endpoint requires an additional scope `user-library-read`, meaning request fails due to insufficient scope. 

I'm sure there is some trickery one could do to add additional scopes to the auth request before it reaches Spotify, but this would be a hack. Ideally, I feel like the scope should be user-configurable when adding a new connection, but this is a separate issue. Regardless, this seems like a common operation so I think it deserves its own action.

jerboa88:
## What does this PR do?

Add a new Action to the Spotify Piece for fetching a list of the user's saved tracks, as this can't be done with the existing 'Get Playlist Items' endpoint.

Closes #6462

### DIFF
--- a/packages/pieces/community/spotify/package.json
+++ b/packages/pieces/community/spotify/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-spotify",
-  "version": "0.3.4"
+  "version": "0.3.5"
 }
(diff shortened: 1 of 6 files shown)


kishanprmr:
@allcontributors please add @jerboa88 for plugin

kishanprmr (approved):

PR #6473, opened 2025-01-07
feat(matrix): enable markdown formatting

revolunet:
## What does this PR do?

💅 Add markdown support to the matrix piece

### DIFF
--- a/packages/pieces/community/matrix/package.json
+++ b/packages/pieces/community/matrix/package.json
@@ -1,7 +1,5 @@
 {
   "name": "@activepieces/piece-matrix",
   "version": "0.3.4",
-  "dependencies": {
-    "marked": "^15.0.6"
-  }
+  "dependencies": {}
 }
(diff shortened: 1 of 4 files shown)


packages/pieces/community/matrix/package.json
```diff
@@ -1,4 +1,7 @@
 {
   "name": "@activepieces/piece-matrix",
-  "version": "0.3.4"
-}
\ No newline at end of file
+  "version": "0.3.4",
+  "dependencies": {
+    "marked": "^15.0.6"
```
abuaboud: Can you please run `npm install marked` in the global `package.json` and remove it from the dependencies section?  

We use Nx (nx.dev), which automatically detects dependencies during build time and includes them in the generated `package.json`. This approach helps us maintain a single version of each dependency in our `package.json`.

revolunet: ok should be fixed

abuaboud (approved):

PR #6540, opened 2025-01-15
fix(nocoDB): added compatibility with self-hosted versions

loudotdigital:
## What does this PR do?

This PR fixes compatibility issues with self-hosted versions of NocoDB by addressing the following:

Removes dependency on workspaces: Self-hosted versions of NocoDB do not support the workspace function, so this update eliminates the necessity of having a workspace to retrieve a base.

Adds API version selection: Users can now select which version of the NocoDB API they are using, ensuring compatibility with various versions.

Fixes # ( added compatibility with self-hosted versions)

### DIFF
--- a/packages/pieces/community/crypto/package.json
+++ b/packages/pieces/community/crypto/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-crypto",
-  "version": "0.0.5"
+  "version": "0.0.6"
 }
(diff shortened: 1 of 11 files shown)


loudotdigital:
@kishanprmr seems that changes have not been merged. Should I resubmit?

kishanprmr:
@loudotdigital I merged only the Groq piece. Can you reopen the PR for only the NocoDB changes? Apologies for the inconvenience.

PR #6578, opened 2025-01-21
chore: add worker version to workers table

AbdulTheActivePiecer:


packages/server/worker/src/lib/api/server-api.service.ts
```diff
@@ -16,8 +16,9 @@ export const workerApiService = (workerToken: string) => {
     const client = new ApAxiosClient(apiUrl, workerToken)
 
     return {
-        async heartbeat(): Promise<WorkerMachineHealthcheckResponse | null> {
+        async heartbeat(version: string): Promise<WorkerMachineHealthcheckResponse | null> {
             const request: WorkerMachineHealthcheckRequest = await workerMachine.getSystemInfo()
```
abuaboud: Please revert everything else, and just make workerMachine.getSystemInfo return the version.

If you need the current version and the function is only available  in app server, then move it to shared folder.

abuaboud (changes requested):

PR #6598, opened 2025-01-21
feat: add admin controller for 14 day trial key

amrabuaza:
Create new apis that's help us to generate trial key and extend trial

**Add new env variables**
AP_CLOUDFLARE_API_TOKEN = 'CLOUDFLARE_API_TOKEN',
AP_CLOUDFLARE_API_BASE = 'CLOUDFLARE_API_BASE',
AP_SECRET_MANAGER_API_KEY='AP_SECRET_MANAGER_API_KEY'


packages/server/api/src/app/user/user-service.ts
```diff
@@ -120,6 +120,13 @@ export const userService = {
             platformId,
         })
     },
+    async getByEmail(email: string): Promise<User | null> {
```
abuaboud: use identityService then use get by identity

abuaboud (approved):

PR #6617, opened 2025-01-22
fix: update error status on Dust piece to match changes in the API

Blightwidow:
## What does this PR do?

The API for Dust changed, and the error status changed from `errored` to `failed`. Because of this, the pieces are timing out and never resolve. This ensures it resolve with a proper failure instead of a time-out.


kishanprmr:
@allcontributors please add @Blightwidow for plugin

kishanprmr (approved):

PR #6618, opened 2025-01-22
feat(nocodb): add compatibility with self-hosted NocoDB instances

loudotdigital:
## What does this PR do?
This PR fixes compatibility issues with self-hosted versions of NocoDB by addressing the following:

Removes dependency on workspaces: Self-hosted versions of NocoDB do not support the workspace function, so this update eliminates the necessity of having a workspace to retrieve a base.

Adds API version selection: Users can now select which version of the NocoDB API they are using, ensuring compatibility with various versions.


packages/pieces/community/nocodb/src/lib/common/index.ts
```diff
@@ -7,10 +7,24 @@ export function makeClient(auth: PiecePropValueSchema<typeof nocodbAuth>) {
 }
 
 export const nocodbCommon = {
+	version: Property.StaticDropdown({
```
abuaboud: Please move this to the authentication property as optional parameter

BuiltByLou: API Version has been moved to Authentication and it's an optional parameter

kishanprmr (approved):

PR #6636, opened 2025-01-23
fix(platform-billing): prevent user to edit ai credit and fix design …

amrabuaza:
Prevent user to edit ai credit and fix design issues


packages/server/api/src/app/ee/platform-billing/platform-billing.service.ts
```diff
@@ -24,7 +24,7 @@ export const platformBillingService = (log: FastifyBaseLogger) => ({
         return platformBilling
     },
 
-    async update(platformId: string, tasksLimit: number | undefined, aiCreditsLimit: number | undefined): Promise<PlatformBilling> {
```
abuaboud, line 27: nit: make this named constructor It would be easier to read when it's being called

{platformId, taskLimit} : {platformId: string, taskLimits: number | undefined}

packages/react-ui/src/app/routes/platform/billing/dialogs/tasks.tsx
```diff
@@ -41,7 +41,7 @@ export const TasksLimitDialog = ({
   });
   useEffect(() => {
```
abuaboud: Do you really need useEffect? The initalLimit is property and it would trigger rerender

https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

packages/react-ui/src/app/routes/platform/billing/dialogs/tasks.tsx
```diff
@@ -90,7 +90,13 @@ export const TasksLimitDialog = ({
                         type="button"
                         variant="transparent"
                         className="absolute right-1 top-1/2 -translate-y-1/2 text-xs"
-                        onClick={() => field.onChange('')}
+                        onClick={() => {
+                          field.onChange('');
+                          form.setError('tasks', {
+                            type: 'manual',
```
abuaboud: Why do you need to set it manual? If the typebox allow for null or undefined / empty string then you might not need it at all.

from the user perspective they should be able to remove the limit

abuaboud (changes requested):

abuaboud (approved):

PR #6642, opened 2025-01-23
feat: add trigger and resume payloads into flow run logs file

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)


abuaboud:
We don't need new column for it, we can store it inside the state, please note make sure you do this in async matter so it doesn't block the starting of the flow as it will make webhook with /sync calls slower

image

packages/shared/src/lib/flow-run/execution/flow-execution.ts
```diff
@@ -62,6 +62,8 @@ const BaseExecutionResponse = {
     tags: Type.Optional(Type.Array(Type.String())),
     error: Type.Optional(FlowError),
     stopResponse: Type.Optional(StopResponse),
+    triggerPayload: Type.Optional(Type.String()),
```
abuaboud, line 65: We already do have triggerPayload in the steps, I don't think we need it for here.

For `resumePayload` it make sense, but did we try to test If the flow is paused and failed (due to quota exceeded) can we retry it?

PR #6652, opened 2025-01-23
fix: task limit dialog

amrabuaza:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/react-ui/src/app/routes/platform/billing/dialogs/tasks.tsx
+++ b/packages/react-ui/src/app/routes/platform/billing/dialogs/tasks.tsx
@@ -80,7 +80,7 @@ export const TasksLimitDialog = ({
                         field.onChange(value);
                       }}
                     />
-                    {form.watch('tasks') !== undefined && form.watch('tasks') !== '' && (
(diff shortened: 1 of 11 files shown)


packages/react-ui/src/app/routes/platform/billing/dialogs/tasks.tsx
```diff
@@ -79,13 +80,13 @@ export const TasksLimitDialog = ({
                         field.onChange(value);
                       }}
                     />
-                    {form.watch('tasks') !== undefined && (
+                    {form.watch('tasks') !== undefined && form.watch('tasks') !== '' && (
```
abuaboud: you can use field.value 

form.watch is a hook, and this can be problematic in certain case for example if this is not rendered on cycle and render in another cycle, it would show error 

https://medium.com/@navneetskahlon/understanding-react-hooks-rendered-fewer-hooks-error-and-order-of-hooks-called-cd526ce1e183#:~:text=Error%201%3A%20Rendered%20Fewer%20Hooks,associated%20state%20and%20effect%20dependencies.

abuaboud (approved):

PR #6701, opened 2025-01-29
feat: fetch current user data from api not localstorage

amrabuaza:
fetch user from api instead of localstorage


packages/react-ui/src/hooks/platform-user-hooks.ts
```diff
@@ -13,4 +17,13 @@ export const platformUserHooks = {
       },
     });
   },
+  useCurrentUser: () => {
+    return useQuery<UserWithMetaInformationAndProject, Error>({
+      queryKey: ['currentUser'],
```
abuaboud: You should include user id from the token because it will cache it

abuaboud: You should include user id from the token because it will cache it

abuaboud (changes requested):

packages/react-ui/src/app/components/initial-data-guard.tsx
```diff
@@ -1,14 +1,11 @@
...
 
-import { flagsHooks } from '@/hooks/flags-hooks';
-
 import { LoadingScreen } from './loading-screen';
 
 type InitialDataGuardProps = {
   children: React.ReactNode;
 };
 export const InitialDataGuard = ({ children }: InitialDataGuardProps) => {
-  flagsHooks.prefetchFlags();
```
AbdulTheActivePiecer, line 11: no need to prefetch, useFlags use suspense query which will show the loading screen, which is what we need.

packages/react-ui/src/hooks/platform-user-hooks.ts
```diff
@@ -13,4 +18,22 @@ export const platformUserHooks = {
...
+      queryKey: ['currentUser', userId],
+      queryFn: async () => {
+        if (!userId) {
+          return null;
+        }
+        const result = await platformUserApi.getCurrentUser();
+        return result;
+      },
+      staleTime: Infinity,
+    });
```
AbdulTheActivePiecer: This addresses Mo's comment + fixes a bug where if this hook is called before the user logs in we simply just return null, instead of letting the endpoint through 403

AbdulTheActivePiecer: That would make the global handler do an endless loop where it redirects to sign-in.

packages/server/api/src/app/user/platform/platform-user-controller.ts
```diff
@@ -6,16 +6,42 @@ import {
...
+            platformId: user.platformId,
+            firstName: identity.firstName,
+            lastName: identity.lastName,
+            email: identity.email,
+            trackEvents: identity.trackEvents,
+            newsLetter: identity.newsLetter,
+            verified: identity.verified,
+            projectId: req.principal.projectId,
+        }
+    })
```
AbdulTheActivePiecer: This endpoint is used in a controller that only allows platform owners (admin platform role && ownerId === userId) to access it, so anyone who is not a platform owner will always get 403 resulting in breaking frontend for anyone who signs in as non-platform owner (i.e embedding)

packages/shared/src/lib/user/user.ts
```diff
@@ -50,3 +50,22 @@ export const UserWithMetaInformation = Type.Object({
...
+    lastName: Type.String(),
+    created: Type.String(),
+    updated: Type.String(),
+    projectId: Type.String(),
+    trackEvents: Type.Boolean(),
+    newsLetter: Type.Boolean(),
+    verified: Type.Boolean(),
+})
+
+export type UserWithMetaInformationAndProject = Static<typeof UserWithMetaInformationAndProject>
```
AbdulTheActivePiecer, line 71: Bump shared package version after adding anything related to it.

AbdulTheActivePiecer (commented):
Please fix the /me endpoint access issue and merge this at the start of next week so we mitigate any issues, after thoroughly testing it.

AbdulTheActivePiecer (dismissed):

AbdulTheActivePiecer (approved):

AbdulTheActivePiecer (approved):

PR #6705, opened 2025-01-29
fix: wrong error for missing projectId in api key authn

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "activepieces",
-  "version": "0.39.8",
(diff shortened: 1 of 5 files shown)


abuaboud (approved):

AbdulTheActivePiecer:
don't remove the optional operator here
 `const principalProjectId = request.principal?.projectId`
even though the type of principal might not include that it can be undefined or null, during execution it can be and we should figure out a way to fix that.

Issue #6297, opened 2024-12-11
Change Created At to Last Modified in Flows Table


abuaboud:
In the flows table, we should show the last modified instead of created, as it's more accurate ordering to most relevant flows to the user. 

<!-- Pylon-Issue-ID:82c5b955-b939-4e1a-8169-6da1bafaa8be --!>

Issue #6228, opened 2024-12-04
Replace Seek Pagination with something user friendly


abuaboud:
Good morning,  

We’ve identified a bug in the pagination of the **"Runs"** tab. Specifically:  
- When pressing **"Next"**, it does not navigate to the correct page.  
- After pressing **"Next"**, the **"Previous"** button can be clicked multiple times, which should not be possible.  

It's related to our pagination system, that means there is newer runs that happened so previous page will show the newer one. 💀

I understand the confusion we are replacing this seek pagination to different one based on page number / offset

This version is clear, concise, and focused. Link: https://activepieces.slack.com/archives/C07QQ62GNUE/p1732010181406329

abuaboud:
Read more https://github.com/activepieces/activepieces/issues/6228

PR #6719, opened 2025-01-31
fix: show last modified instead of created for flows

geekyme-fsmk:
### Before modifying

Screenshot 2025-01-31 at 2 22 02 PM

### Modifying
https://github.com/user-attachments/assets/9a196436-ef2e-4be5-931d-45d1ab835bf3

Fixes # (issue)
#6297


AbdulTheActivePiecer:
Hi there @geekyme-fsmk, currently the table is sorted by createdBy by default, removing it from the UI and not editing the sorting  field is going to look odd in the UI because the user will not see the most recently modified flow first.

Will take a look to see if that can be fixed with our current pagination library, we were looking at rewriting it in another issue.
https://github.com/activepieces/activepieces/issues/6228

PR #6720, opened 2025-01-31
feat: Adding support for syncing specific pieces

ArvindEnvoy:
## What does this PR do?

I am adding an optional flag for the `sync-pieces` CLI command called `--pieces` that lets a user specify which pieces they actually want to sync. I am adding this flag because our team has quite a few custom pieces in our AP project, and syncing them requires all of them to build, even though we only care about the one we are developing at the moment. The existing sync setup makes development unnecessarily slow, so I am adding a flag that we can use to sync only the piece we are working on. I was thinking that this feature could be useful more broadly for other teams facing the same issue.


abuaboud:
@allcontributors please add @ArvindEnvoy for code

hazemadelkhalel:
@ArvindEnvoy 
I opened another pr that was required to merge it: https://github.com/activepieces/activepieces/pull/6797

Thank you!

Issue #6722, opened 2025-01-31
Discard SMTP Settings in Platform Admin


abuaboud:
When people connects faulty SMTP, they need a way to discard it temporary so it does show up invitations 

<!-- Pylon-Issue-ID:0d4786a4-5591-4161-bccc-881e62c9b7d5 --!>

PR #6737, opened 2025-02-03
feat: added button to disable mail server with confirmation beforehand

AbdulTheActivePiecer:
closes [#6722](https://github.com/activepieces/activepieces/issues/6722)


abuaboud (dismissed):

packages/react-ui/src/hooks/platform-hooks.ts
```diff
@@ -26,6 +38,7 @@ export const platformHooks = {
       refetch: async () => {
         await query.refetch();
       },
+      setCurrentPlatform,
```
abuaboud: nit: inline the definition there and extract "platform" as variable and reuse in the above 2 methods

PR #6788, opened 2025-02-05
feat: add project role users table

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)


packages/server/api/src/app/authentication/user-identity/user-identity-service.ts
```diff
@@ -86,6 +86,15 @@ export const userIdentityService = (log: FastifyBaseLogger) => ({
             newsLetter: user.newsLetter,
         }
     },
+    async getBasicInformationByEmail(email: string): Promise<Pick<UserIdentity, 'email' | 'firstName' | 'lastName'>> {
```
abuaboud: I think you have getMetaINformation in userService you will need this instead of this one.

PR #6797, opened 2025-02-05
feat: Adding support for syncing specific pieces

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/cli/src/lib/utils/piece-utils.ts
+++ b/packages/cli/src/lib/utils/piece-utils.ts
@@ -8,18 +8,29 @@ import chalk from 'chalk'
 import FormData from 'form-data';
 import fs from 'fs';
 
-export async function findAllPieces(inputPath?: string): Promise<string[]> {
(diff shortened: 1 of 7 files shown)


AbdulTheActivePiecer (approved):

packages/cli/src/lib/utils/piece-utils.ts
```diff
@@ -27,7 +27,6 @@ export async function findPieces(inputPath?: string, pieces?: string[]): Promise
               return normalizedPath.endsWith(path.sep + piece);
           });
           if (!folder) {
-              console.error(chalk.red(`🚨 Piece ${piece} not found`));
```
ArvindR12: @hazemadelkhalel I feel like this log line might have been helpful? For example, if someone tries to sync pieces that don't exist (e.g. because of a typo), they would know right away if this log line was here. I see that you moved it to `assertPieceExists`, but that function doesn't get called from all callsites of `findPieces`. Just my 2 cents.

PR #6821, opened 2025-02-09
feat(bulk-rerun): add bulk rerurn to run flow page

amrabuaza:
Add bulk rerurn to run flow page (table) to allow user select all runs inside all pages


packages/react-ui/src/app/routes/runs/index.tsx
```diff
@@ -309,7 +181,13 @@ const FlowRunsPage = () => {
...
                       <PlayIcon className="mr-2 h-3 w-4" />
                       {selectedRows.length > 0
-                        ? `${t('Retry')} (${selectedRows.length})`
+                        ? `${t('Retry')} (${
+                            selectedAll
+                              ? data?.totalCount
+                                ? data?.totalCount - excludedRows.size
+                                : selectedRows.length
+                              : selectedRows.length
+                          })`
```
AbdulTheActivePiecer: Please remove total count and instead just make it say all or (all except X), where X is the number of excluded runs.
We don't return total count in any of the paginated requests and we shouldn't have inconsistent api responses.

Once the pagination issue is fixed and we show this in all tables, we can add it back.

PR #6826, opened 2025-02-09
fix: fix bug related to handling null/undefined formValues

s31w4n:
When disabling and re-enabling the form via dynamic values, the "add item"  button was not functioning due to null formValues. Added null check to prevent  "properties of undefined" error.

## What does this PR do?

Fixes # (issue)


AbdulTheActivePiecer:
Thanks @s31w4n, this will be addressed in the  https://github.com/activepieces/activepieces/pull/6094 pr

s31w4n:
> Thanks @s31w4n, this will be addressed in the #6094 pr

Hey @AbdulTheActivePiecer,
Could you please update my contribution icon to "Code" for this PR? Thanks! 😊

PR #6847, opened 2025-02-12
fix(grist): fix default workspace

revolunet:
This PR allows to use self-hosted GRIST instances

Looks like `current` doest not always exist, while `0` works everywhere i tested

Following comment here : https://github.com/activepieces/activepieces/pull/5069#issuecomment-2654926619


abuaboud:
Can you please bump the version in package.json

kishanprmr:
@revolunet  I just tested the workspace dropdown with '0' as the parameter instead of 'current'. For a team site, it only shows a few workspaces instead of all. Can you share the steps you used for team sites in the self-hosted version?

revolunet:
Hey, i just tested with the API. looks like `docs` also works in both cases. can you try if it works fo you ?

vviers:
My two cents : wouldn't something like a simple `GET /orgs` be an easier way to check that the API can be called ?

revolunet:
> My two cents : wouldn't something like a simple `GET /orgs` be an easier way to check that the API can be called ?

Indeed! added a `client.listOrgs()` to avoid this issue

revolunet:
@kishanprmr should be better now

abuaboud (approved):

PR #6871, opened 2025-02-17
feat: support multiple functions in v8 code SANDBOX_CODE_ONLY

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/engine/src/lib/core/code/v8-isolate-code-sandbox.ts
+++ b/packages/engine/src/lib/core/code/v8-isolate-code-sandbox.ts
@@ -87,16 +87,13 @@ const executeIsolate = async ({ isolate, isolateContext, code }: ExecuteIsolateP
 }
 
 const serializeCodeModule = (codeModule: CodeModule): string => {
-    let serializedCodeFunction = ''
-    for (const key in codeModule) {
-        serializedCodeFunction += `const ${key} = ${(codeModule as any)[key].toString()};`
-    }
+    const serializedCodeFunction = Object.keys(codeModule)
+        .reduce((acc, key) => 
+            acc + `const ${key} = ${(codeModule as any)[key].toString()};`, 
+            '');
+
     // replace the exports.function_name with function_name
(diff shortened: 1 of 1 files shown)


packages/engine/src/lib/core/code/v8-isolate-code-sandbox.ts
```diff
@@ -87,8 +87,16 @@ const executeIsolate = async ({ isolate, isolateContext, code }: ExecuteIsolateP
 }
 
 const serializeCodeModule = (codeModule: CodeModule): string => {
-    const serializedCodeFunction = codeModule.code.toString()
-    return `const code = ${serializedCodeFunction}; code(inputs);`
+    let serializedCodeFunction = ''
+    for (const key in codeModule) {
+        serializedCodeFunction += `const ${key} = ${(codeModule as any)[key].toString()};`
+    }
```
AbdulTheActivePiecer: Not all keys in JSON objects can be valid JS variable names, please make a processor function that would make any string a valid js variable, and make sure you don't make any duplicates as well.

AbdulTheActivePiecer: and you don't need as any, please remove it.

packages/engine/src/lib/core/code/v8-isolate-code-sandbox.ts
```diff
@@ -87,8 +87,16 @@ const executeIsolate = async ({ isolate, isolateContext, code }: ExecuteIsolateP
 }
 
 const serializeCodeModule = (codeModule: CodeModule): string => {
-    const serializedCodeFunction = codeModule.code.toString()
-    return `const code = ${serializedCodeFunction}; code(inputs);`
+    let serializedCodeFunction = ''
+    for (const key in codeModule) {
```
AbdulTheActivePiecer: you should actually make this a reduce on Object.keys(codeModule) and return serializedCodeFunction  inside the method passed to the reduce function.

PR #6922, opened 2025-02-23
fix(embed): activepieces.connect resolving without the connection

### LINKED ISSUE #6907: Return External Id from The Connect Method in Embed SDK
Currently the connect method isn't returning the correct schema, it should include the external Id of the connection

kamal:
## What does this PR do?

Fixes #6907

### DIFF
--- a/packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
+++ b/packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
@@ -67,7 +67,7 @@ type ConnectionDialogProps = {
   piece: PieceMetadataModelSummary | PieceMetadataModel;
   open: boolean;
   onConnectionCreated: (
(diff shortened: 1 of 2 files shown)


packages/react-ui/src/app/connections/create-edit-connection-dialog.tsx
```diff
@@ -145,7 +145,6 @@ const CreateOrEditConnectionDialog = React.memo(
         return appConnectionsApi.upsert(formValues);
       },
       onSuccess: (connection) => {
-        setOpen(false);
```
kamal, line 148: this is the cause for the `window.activepieces.connect` to resolve the promise prematurely

AbdulTheActivePiecer:
Thanks for discovering the issue @kamal, I opened another PR to solve this because the solution is quite more nuanced.
https://github.com/activepieces/activepieces/pull/6924

PR #6930, opened 2025-02-24
feat: show input data for testing a step

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)


packages/react-ui/src/app/builder/pieces-selector/index.tsx
```diff
@@ -201,6 +203,8 @@ const PieceSelector = ({
     switch (operation.type) {
       case FlowOperationType.UPDATE_TRIGGER: {
         setSampleData(stepData.name, undefined);
+        // TODO THINK OF THAT WHAT TO DO WITH THE INPUT
+        setSampleDataInput(stepData.name, undefined);
```
abuaboud: Keep it {}

packages/react-ui/src/features/flows/lib/sample-data-hooks.ts
```diff
@@ -53,4 +56,36 @@ export const sampleDataHooks = {
...
+      staleTime: 0,
+      retry: 4,
+      refetchOnWindowFocus: false,
+      queryFn: async () => {
+        const steps = flowStructureUtil.getAllSteps(flowVersion!.trigger);
+        const singleStepSampleDataInput = await Promise.all(
+          steps.map(async (step) => {
+            return {
+              [step.name]: await getSampleData(
+                flowVersion!,
```
abuaboud: Make sure test when there is no sample data what will happen

abuaboud (changes requested):

AbdulTheActivePiecer:
1. The input for testing a step is not being resolved, so the user will see something like ` {{ trigger['method'] }}` and that shouldn't be the case.
2. The input for the trigger in runs is not showing the real inputs of the trigger and is always empty, this will cause confusion for users comparing trigger input in runs vs in the builder in the test step section.

abuaboud (approved):

abuaboud (approved):

PR #6946, opened 2025-02-26
feat: add action categories collapsible

reemayoush:
A collapsible action category feature enhances user experience by organizing related functions into expandable and collapsible sections, making navigation more intuitive and efficient. It allows users to focus on relevant options while keeping others hidden, reducing clutter and improving readability. This structure streamlines workflows by providing quick access to frequently used actions while maintaining a clean and organized interface. By enabling users to collapse sections they are not using, it optimizes screen space and ensures a more structured, efficient interaction with the tool or application.

Screenshot 2025-02-26 at 5 48 49 PM

### DIFF
--- a/packages/pieces/community/framework/src/lib/action/action.ts
+++ b/packages/pieces/community/framework/src/lib/action/action.ts
@@ -27,6 +27,7 @@ type CreateActionParams<PieceAuth extends PieceAuthProperty, ActionProps extends
   auth?: PieceAuth
   displayName: string
   description: string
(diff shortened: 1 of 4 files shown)


AbdulTheActivePiecer:
Hi there @reemayoush, for the time being we would like to keep things as they are and not introduce an extra "category" for actions or triggers and keep things simple.

If you would like you can still apply these changes to your repo but you would have conflicts every time you sync with us.

PR #7009, opened 2025-03-05
fix(xml-webhook-response): fix response return when flow execution is…

amrabuaza:
Fix return response to webhook when flow execution is respond and continue


packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -399,4 +408,31 @@ const RemoveFlowRequest = {
...
+                status: Type.Number(),
+                body: Type.Any(),
+                headers: Type.Record(Type.String(), Type.String()),
+            }),
+        }),
+    },
+}
+
+const stoppedResponse = (result: FlowRunResponse): EngineHttpResponse => {
+    const response = result.response as StopResponse
```
abuaboud, line 365: We can use `StopResponse` in result.response type and other places, casting would cause unnecessary issue if they the types is not same

PR #7070, opened 2025-03-12
feat(slack): add slack command handling

geekyme:
## What does this PR do?

A different variation of `New mention` that only triggers on specific command, and can handle the arguments after the command. This prevents the need for users to write custom code to parse the slack message. 

Screenshot 2025-03-12 at 9 39 53 PM


geekyme:
Raised here - https://github.com/activepieces/activepieces/pull/7080

PR #7072, opened 2025-03-12
fix: support duplicate query params for arrays

danielpoonwj:
## What does this PR do?

The `HTTP (Send HTTP request)` action does not support sending query params in an array style eg. `https://httpbin.org/get?ids[]=1&ids[]=2`. It ends up only sending the last element `ids[]=2` even when providing the query params explicitly on the URL.

Screenshot 2025-03-12 at 10 41 45 PM

## Approach

Currently the httpClient extracts url params from the url (type `URLSearchParams`), then merges that into the provided query parameters (type `Record`/object). `URLSearchParams` supports duplicate keys but `Record` doesn't. This is passed into [`AxiosRequestConfig.params`](https://axios-http.com/docs/req_config) which supports either a plain object or a URLSearchParams object.

This PR merges the provided query params into the `URLSearchParams` object instead, supporting duplicate keys like `ids[]`.

Screenshot 2025-03-12 at 10 41 06 PM

## Caveats

- This should be backwards compatible for most cases, unless there are cases where people are expecting implicit query param deduplication.
- This doesn't fix the UI form for inputting query params. It still automatically picks the last value. Workaround is to provide it explicitly on the URL itself like my example.

### DIFF
--- a/packages/pieces/community/common/package.json
+++ b/packages/pieces/community/common/package.json
@@ -1,5 +1,5 @@
 {
   "name": "@activepieces/pieces-common",
-  "version": "0.4.2",
+  "version": "0.4.3",
   "type": "commonjs"
 }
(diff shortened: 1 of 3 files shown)


danielpoonwj:
btw should i bump version specifically for the http piece? or would this require upgrading AP version?

kishanprmr:
@allcontributors please add @danielpoonwj for code

kishanprmr (approved):

PR #7138, opened 2025-03-18
fix(hubspot): convert checkbox property to static dropdown

kishanprmr:
## What does this PR do?

Fixes # (issue)


packages/pieces/community/hubspot/src/lib/common/props.ts
```diff
@@ -178,9 +178,22 @@ async function createReferencedPropertyDefinition(
...
+						{
+							label:'Yes',
+							value:'true'
+						},
+						{
+							label:'No',
+							value:'false'
+						}
+					]
+				}
```
AbdulTheActivePiecer: We need to add a value "Unanswered" which is null.

Are you sure that hubspot's api accepts 'true'/'false' instead of actual boolean values ?

kishanprmr: added the unanswered option

AbdulTheActivePiecer (approved):

PR #7170, opened 2025-03-20
feat(slack): simulate triggers instead of using sample data

AdamSelene:
## What does this PR do?

Since https://github.com/activepieces/activepieces/pull/6187 it's possible to properly simulate triggers (loading real data) which is often much more useful than static sample data

### DIFF
--- a/packages/pieces/community/slack/src/lib/triggers/new-mention.ts
+++ b/packages/pieces/community/slack/src/lib/triggers/new-mention.ts
@@ -10,38 +10,6 @@ import { WebClient } from '@slack/web-api';
 import { isNil } from '@activepieces/shared';
 import { getFirstFiveOrAll } from '../common/utils';
 
-const sampleData = {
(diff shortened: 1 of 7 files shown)


packages/pieces/community/slack/src/lib/triggers/new-mention.ts
```diff
@@ -96,39 +64,6 @@ export const newMention = createTrigger({
...
-          channel_type: 'channel',
-        };
-      }).sort((a, b) => parseFloat(b.ts!) - parseFloat(a.ts!));
-
-      return getFirstFiveOrAll(messages);
-  },
-
   run: async (context) => {
     const payloadBody = context.payload.body as PayloadBody;
     const channels = context.propsValue.channels as string[];
```
kishanprmr, line 69: @AdamSelene can you make channels empty list if its not available like `    const channels = (context.propsValue.channels as string[]) ?? [];
` ?

kishanprmr (changes requested):

kishanprmr (approved):

PR #7206, opened 2025-03-24
fix(oauth2): change code challenge to 43 char

briceflaceliere:
The Oauth2 challenge code (pkce) does not comply with RFC 7636 (https://datatracker.ietf.org/doc/html/rfc7636#section-4.1).

It must be between 43 and 128 characters long, which is not the case with the default use of `nanoid` (21 characters). I therefore changed the length of `nanoid` to 43. 
This should not have any impact on existing integrations unless the connector does not fully comply with the Oauth2 specifications.


abuaboud:
@allcontributors please add @briceflaceliere for security

PR #7350, opened 2025-04-11
fix(ui+server): add Actions column to Pieces table and log the actual…

hakrsh:
… error

## What does this PR do?

Adds an "Actions" column to the Pieces table UI and improves error logging in the piece sync service.

### Relevant User Scenarios

Fixes # (issue)


abuaboud:
Hi @hakrsh 

We intentionally keep the actions column without a header, the reasoning it will feel less crowded as it's actionable thing and not information.

For example take look at this:
https://ui.shadcn.com/docs/components/data-table

PR #7355, opened 2025-04-12
fix: clicking on an issue doesn't switch to the history tab

hazemadelkhalel:
### Relevant User Scenarios

Fixes # (issue)


zjawj:
https://code.mayoubang.cn  —  你的免费 AI 代码加速器！

Issue #6133, opened 2024-11-26
Support Custom Fields for Clickup Task Actions


kishanprmr:
https://community.activepieces.com/t/clickup-custom-task-fields-how-to-add/6960/2

CarefulGuru:
I'm working on tweaking the clickup integration right now so that it supports ClickUp chat, will try fit this in as well

PR #7412, opened 2025-04-17
feat: clickup piece enhancements

CarefulGuru:
## What does this PR do?

- adds a trigger for every relevant section of the webhooks api
- allows triggers to be filtered down to task level
- add a piece for finding tasks based on their name
- add what I would consider to be the most frequently used functions based on the new ClickUp Chat API

Still going through testing but I believe the code to be mostly complete at this point


CarefulGuru:
Latest commit solves #6133 - as far as I can tell everything is working!

kishanprmr:
@allcontributors please add @CarefulGuru  for plugin

kishanprmr (approved):

PR #7420, opened 2025-04-17
feat(metabase): new action get graph

Kevinyu-alan:
## What does this PR do?

Add a new action for metabase that allows to get the graph of a question and download it.
This is only a V1 and we might change the way we do that by parsing the HTML later. I'm waiting confirmations to know if this is enough.
There's already an action for that, but the result and method are different. Let's keep both actions until this one works perfectly.

For that we need an embedding key and that the question is embed (only possible with an admin).
I put the embedding key as a ShortText instead of a SecretText because SecretText are all mandatory (even if the required is false). But by doing this, the field isn't hidden with dots, can be an issue.

Fixes Q-1275

### DIFF
--- a/packages/pieces/community/metabase/package.json
+++ b/packages/pieces/community/metabase/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-metabase",
-  "version": "0.1.5"
+  "version": "0.1.6"
 }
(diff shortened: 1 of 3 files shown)


Kevinyu-alan:
Let me know if this works for you

kishanprmr:
Hi @Kevinyu-alan , could you please sign the CLA certificate when you get a chance?

Kevinyu-alan:
Hi @kishanprmr, can I have more context and how to do it? I didn't heard about this before.

PR #7430, opened 2025-04-20
feat: enhance async and refactor webhook handler

### LINKED ISSUE #7174: Optimize Execution Time for Sync webhooks
We should optimize:

* The overhead for execution webhook sync by just executing it in single engine operation, instead of doing two hops.
* Cache flow in memory so it minimize the flow loading from database.

<!-- Pylon-Issue-ID:24cd148f-ed7c-46bb-9aa5-ebc5933ea725 --!>

hazemadelkhalel:
### Relevant User Scenarios

Fixes #7174


packages/server/worker/src/lib/executors/webhook-job-executor.ts
```diff
@@ -141,14 +142,14 @@ async function handleSampleData(
...
     log: FastifyBaseLogger,
 ): Promise<void> {
-    const { flowId, payload } = data
+    const { flowId, payload, flowVersionIdToRun } = data
     const latestFlowVersion = await engineApiService(
         engineToken,
         log,
     ).getFlowWithExactPieces({
         flowId,
         type: GetFlowVersionForWorkerRequestType.LATEST,
```
abuaboud: I don't think you need the type GetFlowVersionForWorkerRequestType to fetch the flow, I would edit the parent endpoint since you already have flowId, and versionId you can fetch it directly in the endpoint.

packages/server/shared/src/lib/job/job-data.ts
```diff
@@ -89,6 +90,7 @@ export const WebhookJobData = Type.Object({
         Type.Literal(GetFlowVersionForWorkerRequestType.LOCKED),
         Type.Literal(GetFlowVersionForWorkerRequestType.LATEST),
     ])),
+    flowVersionIdToRun: Nullable(Type.String()),
```
abuaboud, line 93: Let's remove `flowVersionToRun` and keep the flowVersionIdToRun

packages/server/api/src/app/webhooks/webhook-handler.ts
```diff
@@ -0,0 +1,137 @@
...
+                schemaVersion: LATEST_JOB_DATA_SCHEMA_VERSION,
+                requestId: webhookRequestId,
+                synchronousHandlerId,
+                payload,
+                flowId: flow.id,
+                saveSampleData,
+                flowVersionToRun,
+                flowVersionIdToRun,
+            },
+            priority: await getJobPriority(synchronousHandlerId),
```
abuaboud, line 58: I don't think we need this anymore, since one time is added into it's own queue

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -502,6 +505,70 @@ export const flowService = (log: FastifyBaseLogger) => ({
...
+                    return null
+                }
+                if (isNil(rawFlow.publishedVersionId)) {
+                    break
+                }
+                flowVersionIdToRun = rawFlow.publishedVersionId
+                break
+            }
+        }   
+        if (isNil(flowVersionIdToRun) && type !== GetFlowVersionForWorkerRequestType.LATEST) {
```
abuaboud: Isn't this is same as latest?

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -502,6 +505,70 @@ export const flowService = (log: FastifyBaseLogger) => ({
...
+                const flow = await flowService(log).getOnePopulated({
+                    id: request.flowId,
+                    projectId,
+                })
+                if (!isNil(flow)) {
+                    flowVersionIdToRun = flow?.version.id
+                }
+                break
+            }
+            case GetFlowVersionForWorkerRequestType.EXACT: {
```
abuaboud: The case is not used anymore

abuaboud (changes requested):

abuaboud (approved):

PR #7437, opened 2025-04-21
feat(store): bulk add items in list

### LINKED ISSUE #6866: Add Item to List Should be Bulk
We need to modify the storage Add Item to List to take a bulk of items after [AP-38](https://linear.app/activepieces/issue/AP-38/improve-handling-of-line-item-data) this is merged.

Note: Adding multiple items in same action is much faster than doing one by one inside a loop

<!-- Pylon-Issue-ID:a86969dc-1ef0-476d-9666-a638546074c6 --!>

kishanprmr:
### Relevant User Scenarios

Fixes #6866


packages/pieces/community/store/src/lib/actions/store-add-to-list.ts
```diff
@@ -1,72 +1,76 @@
...
+			throw new Error(`Provided value is not an array.`);
+		}
+		// Get existing items from store
+		let items = (await context.store.get<unknown[]>(key, scope)) ?? [];
+		try {
+			if (typeof items === 'string') {
+				items = JSON.parse(items);
+			}
+			if (!Array.isArray(items)) {
+				throw new Error(`Key ${context.propsValue['key']} is not an array`);
```
AbdulTheActivePiecer: is not a list*

packages/pieces/community/store/src/lib/actions/store-add-to-list.ts
```diff
@@ -1,72 +1,76 @@
...
+			}
+		} catch (err) {
+			throw new Error(`Key ${context.propsValue['key']} is not an array`);
+		}
```
AbdulTheActivePiecer, line 63: please try to show the error in the message as well, so in case it happens users can understand why

PR #7462, opened 2025-04-23
feat(rss): new trigger, take a list of feed

Kevinyu-alan:
## What does this PR do?

We created a new trigger that allow a list instead of a single link feed for rss

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/rss/src/lib/common/props.ts
+++ b/packages/pieces/community/rss/src/lib/common/props.ts
@@ -2,5 +2,20 @@ import { Property } from '@activepieces/pieces-framework';
 
 export const rssFeedUrl = Property.ShortText({
   displayName: 'RSS Feed URL',
+  description: 'Single RSS feed URL',
   required: true,
+});
+
+export const rssFeedUrls = Property.Array({
+  displayName: 'RSS Feed URLs',
+  description: 'List of RSS feed URLs',
+  required: true,
+  defaultValue: [],
+  properties: {
+    url: Property.ShortText({
+      displayName: 'URL',
+      description: 'RSS feed URL',
+      required: true,
+    }),
+  },
 });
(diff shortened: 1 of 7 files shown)


packages/pieces/community/rss/src/lib/common/props.ts
```diff
@@ -2,5 +2,20 @@ import { Property } from '@activepieces/pieces-framework';
...
+  description: 'Single RSS feed URL',
   required: true,
 });
+
+export const rssFeedUrls = Property.Array({
+  displayName: 'RSS Feed URLs',
+  description: 'List of RSS feed URLs',
+  required: true,
+  defaultValue: [],
+  properties: {
```
kishanprmr: You can remove the properties so that it becomes a direct list of URLs, as shown in the image.

image

packages/pieces/community/rss/src/lib/triggers/new-item-list-triggers.ts
```diff
@@ -0,0 +1,289 @@
...
+import dayjs from 'dayjs';
+import { getId } from '../common/getId';
+
+type PollingProps = {
+  rss_feed_urls: { url: string }[];
+};
+
+export const rssNewItemListTrigger = createTrigger({
+  name: 'new-item-list',
+  displayName: 'New Item In Feed List',
```
kishanprmr: Can you change the display name to `New Items in Multiple Feeds`?

packages/pieces/community/rss/src/lib/triggers/new-item-list-triggers.ts
```diff
@@ -0,0 +1,289 @@
...
+} from '@activepieces/pieces-framework';
+import { rssFeedUrls } from '../common/props';
+import FeedParser from 'feedparser';
+import axios from 'axios';
+import { isNil } from '@activepieces/shared';
+import dayjs from 'dayjs';
+import { getId } from '../common/getId';
+
+type PollingProps = {
+  rss_feed_urls: { url: string }[];
```
kishanprmr: If you remove `properties` from the array field, it will change to `string[]`.

kishanprmr (changes requested):

kishanprmr (approved):

Issue #7289, opened 2025-04-02
Show Flows used by Certain Connections


abuaboud:
From the user's perspective, the connections table should display the list of flows that are actively using this connection.
Why:

* Makes it easier to see where the connection is used.
* Prevents deletion when it is in use.

Technical details:
We should establish a database relationship between connections and flows, on each update we extract connections being used in the flow json and store the relation.

zjawj:
体验未来编程！✨ https://code.mayoubang.cn AI 智能驱动，免费代码生成，让你的创意更快实现！

PR #7484, opened 2025-04-25
feat: introduce replace connections and show flows using connection dialog

### LINKED ISSUE #7290: Transfer / Merge Connections
If an employee who set up a Slack bot leaves the company, you may need to merge their connection with another, like the system Slack connection.
Workaround
Right now, the only option is to rename and reconnect the connection, but this creates two connections instead of one.
Solution
Add a Merge/Replace Connection feature so that all flows using the first connection switch to the second one.

hazemadelkhalel:
### Relevant User Scenarios

Fixes #7290 #7289

### DIFF
--- a/packages/shared/test/flow/flow-helper.test.ts
+++ b/packages/shared/test/flow/flow-helper.test.ts
@@ -108,6 +108,7 @@ const flowVersionWithBranching: FlowVersion = {
         },
         displayName: 'Cron Expression',
     },
+    connectionIds: [],
     valid: true,
     state: FlowVersionState.DRAFT,
 }
@@ -153,6 +154,7 @@ const emptyScheduleFlowVersion: FlowVersion = {
     },
     valid: true,
     state: FlowVersionState.DRAFT,
+    connectionIds: [],
 }
 
 describe('Flow Helper', () => {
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -95,6 +97,28 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (app, _opts
         }
     },
     )
+
+    app.get('/:id/flows', ListFlowsFromAppConnectionRequest, async (request, reply) => {
+        const flows = await appConnectionService(request.log).listFlowsFromAppConnection({
```
abuaboud: It's wrong to write it in that way, I think you should make list flows in flow controller accept connection Ids instead.

packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
```diff
@@ -189,6 +196,81 @@ export const appConnectionService = (log: FastifyBaseLogger) => ({
         }))
     },
 
+    async applyOperation(params: ApplyOperationParams): Promise<PopulatedFlow[]> {
```
abuaboud: Remove the Operation idea, we only need UPDATE_CONNECTION,  will be moved listFlowsFromAppConnection to flows,

I think this file is to large, you can start new file for this.

app-connection-replace-utils.ts, all it does list from flows service and replace it.

PR #7518, opened 2025-04-29
feat(dust): upload file piece

TimPetricola:
## What does this PR do?

Add support for [Dust's file upload](https://docs.dust.tt/reference/post_api-v1-w-wid-files).

A new piece allows uploading a file, and uploads it to Dust, using its official SDK. As a result, the uploaded file can be referenced in downstream pieces.

The file upload concept is replacing the fragment one on Dust's side. For now we are keeping support for both.

The "create conversation" piece is also updated so we can reference a previously uploaded file (instead of fragment's content)

Also took the opportunity to use the official Dust JS SDK (instead of direct API calls). I only migrated a couple of calls though.

### Relevant User Scenarios

Getting a file/image (from a webhook, a form, ...) and using it as part of a conversation with an assistant on Dust

Screenshot 2025-04-29 at 14 32 57

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "@dnd-kit/core": "6.1.0",
         "@dnd-kit/modifiers": "7.0.0",
         "@dnd-kit/sortable": "8.0.0",
+        "@dust-tt/client": "1.0.41",
         "@fastify/basic-auth": "5.0.0",
         "@fastify/cors": "8.3.0",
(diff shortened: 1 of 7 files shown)


kishanprmr:
@TimPetricola can you sync your PR with the latest changes from the main branch?

TimPetricola:
@kishanprmr done :)

kishanprmr (approved):

PR #7554, opened 2025-05-05
feat(docusign): add basic docusign integration

AdamSelene:
## What does this PR do?

Add a basic `docusign` piece with support for authentication (JWT) and custom API call


AdamSelene:
Logo to be hosted on https://cdn.activepieces.com/pieces/docusign.png:

Docusign Vertical_White

kishanprmr:
@AdamSelene ,

Could you please add more  actions/triggers in addition to the custom API call?
Also, could you clarify the purpose of the server parameter in the custom API call configuration?

Thanks!

AdamSelene:
@kishanprmr I will add more actions very shortly yes (in another PR), but I need this bare bone version to unblock some users and better understand what they will use it for. It's required because we can't use the `http` basic piece (as credentials would be leaked to git in our setup)

I added the extra `serverContext` param because I need to be able to be able to generate a "consent URL" that requires the generic /redirect URL of the ActivePieces instance
But maybe I could move this logic to the `validate` method of the connection - it would be much cleaner, I'll git it a try

AdamSelene:
@kishanprmr I've updated the PR and tested the new mechanism, let me know if it's better for you!

kishanprmr:
> I added the extra serverContext param because I need to be able to be able to generate a "consent URL" that requires the generic /redirect URL of the ActivePieces instance
But maybe I could move this logic to the validate method of the connection - it would be much cleaner, I'll git it a try

@abuaboud Could you please help suggest the best way to achieve this?

AdamSelene:
@kishanprmr I added a few actions

packages/engine/src/lib/helper/piece-helper.ts
```diff
@@ -108,6 +108,11 @@ export const pieceHelper = {
         const { piece: piecePackage } = params
 
         const piece = await pieceLoader.loadPieceOrThrow({ pieceName: piecePackage.pieceName, pieceVersion: piecePackage.pieceVersion, piecesSource })
+        const server = {
+            token: params.engineToken,
```
AbdulTheActivePiecer: Hi @AdamSelene, can you please tell us why you need this exposed ? I don't see it used anywhere else other than here, exposing it for validate might be a security concern.

AdamSelene: It was for consistency: I re-used the server context you inject in `run()` 
I don't need the token, just the public URL

kishanprmr: @AdamSelene , could you remove it if it's not required?

AdamSelene: I would need to change the type of `ServerContext` to allow null values - or use a different type
Just to understand: why would it be a security concern in `validate` when it's not in `run`? I well understand the "if it's not used, remove it" argument, but consistency is also a good argument IMO

mercuryyy:
Looks like a great addition! looking fwd for this merged, tried to pull to my fork!

packages/engine/src/lib/helper/piece-helper.ts
AbdulTheActivePiecer: Hi there @AdamSelene sorry for responding 4 days later, I was off 😂.

We always try to remove anything unused just so it doesn't add an extra unnecessary layer of complexity to the code.
You could do this:
replace ` validate?: (params: { auth: AuthValueSchema; server: ServerContext }) => Promise<` with 
` validate?: (params: { auth: AuthValueSchema; server: Omit<ServerContext, 'token'> }) => Promise<` in 
 packages/pieces/community/framework/src/lib/property/authentication/common.ts
and remove the token you pass to auth.validate. 

Can you also please give us permission to edit the PRs you open just so next time you don't have to go through all of this back and forth with us when it comes to this type of comments, we can just do them ourselves :D

AdamSelene: on it thanks!

AdamSelene: @AbdulTheActivePiecer Done, thanks ✅

kishanprmr:
@AdamSelene could you please update the versions of the pieces that were modified?

kishanprmr:
@AdamSelene its merged.

Issue #6156, opened 2024-11-27
Verify the grace shutdown


abuaboud:
Link: https://activepieces.slack.com/archives/C07GT9LCCMT/p1732185748328769

PR #7622, opened 2025-05-08
feat: support resuming execution from the last state after container restart

### LINKED ISSUE #7148: Durable Execution for Long Runs across restarts
Currently If the container is restarted due, It sends shutdown signal and wait 30 seconds before terminating the instance.

The long running flows will be interrupted and run again on another instance, since we have the logs we should continue running from where it stopped instead of start again from beginning

hazemadelkhalel:
### Relevant User Scenarios

Fixes #7148 #6156


packages/server/api/src/app/workers/worker-controller.ts
```diff
@@ -102,6 +102,7 @@ export const flowWorkerController: FastifyPluginAsyncTypebox = async (app) => {
         )
         const createFlowRuns = filterPayloads.map((payload) =>
             flowRunService(request.log).start({
+                flowRunId: apId(),
```
abuaboud: Why you specify this? I think the method has logic if this is preexisting one for some reason

PR #7635, opened 2025-05-08
feat: translate pieces

### LINKED ISSUE #3181: Enable Piece Translation Support
**Goal:**
Make piece details and properties translatable.

**Implementation Steps:**

**Step 1 - Folder Structure and Default Strings:**

Create a 'translations' folder inside the package. Add a script to generate the default string file for better organization.

**Example Structure:**
```
- piece/
  - translations/
    - en.json
  - other_package_files...
```

**Step 2 - Translation Format:**

Find a standard well used format for translations

1. Marking translated strings (state attribute)
3. can be used with platforms like Crowdin (Standard Format)

**Step 3 - GitHub Action for Crowdin:**

Update the GitHub commit and Crowdin with file paths for translation sources. so the non technical users can contribute back using crowd in, as well it's easier as they have approval process too.

**GitHub Action Example:**

Configure the action to read up-to-date sources and push them back to Crowdin.

[GitHub Action Configuration](https://github.com/activepieces/activepieces/blob/main/.github/workflows/crowdin.yml)

AbdulTheActivePiecer:
closes #3181


docs/developers/piece-reference/i18n.mdx
```diff
@@ -0,0 +1,57 @@
...
+            [LocalesEnum.FRENCH]: fr,
+        },
+});
+
+
+```
+ 
+  </Step>
+   <Step title="Testing">
+    In your dashboard go to Project Settings -> Appearance -> Lanugages
```
abuaboud: typo

docs/developers/piece-reference/i18n.mdx
```diff
@@ -0,0 +1,57 @@
...
+        displayName: "Webhook",
+        description: "Receive HTTP requests and trigger flows using unique URLs.",
+        auth: PieceAuth.None(),
+        categories: [PieceCategory.CORE],
+        minimumSupportedRelease: "0.52.0",
+        logoUrl: "https://cdn.activepieces.com/pieces/webhook.svg",
+        authors: ["abuaboud", "pfernandez98", "kishanprmr", "AbdulTheActivePiecer"],
+        actions: [returnResponse, returnResponseAndWaitForNextWebhook],
+        triggers: [catchWebhook],
+        i18n: {
```
abuaboud: Can we not have it? and auto detect it using the code.

This doesn't have to be in every createPiece, It can be inside createPiece function or maybe not at all and just when we are reading pieceMetadata or so

```
    migrations: [path.join(__dirname, '../database/migrations/**/*{.ts,.js}')],
```

packages/server/api/src/app/pieces/piece-metadata-service/file-piece-metadata-service.ts
```diff
@@ -50,14 +51,23 @@ export const FilePieceMetadataService = (_log: FastifyBaseLogger): PieceMetadata
                 pieces: originalPiecesMetadata,
                 suggestionType: params.suggestionType,
             })
+            console.log('--------------------------------')
```
abuaboud: I think this should be removed

docs/developers/piece-reference/i18n.mdx
```diff
@@ -0,0 +1,57 @@
...
+icon: 'globe'
+---
+
+
+<Steps>
+  <Step title="Generate The Translation File">
+    After you are done with editing your piece, generate the translation file by running the follwing command:
+    
+    ```bash
+    npm run cli pieces generate-translation-file
```
abuaboud: I suggest adding a `generate-i18n` target in your `project.json` in each piece and edit the piece generator CLI to have this target there too.

Then, update the `"build"` target to include `"dependsOn": ["generate-i18n"]`.

Here’s an example: instead of running it as a prebuild step

image

this target will generate all i18n files by default for every locale.

For each locale file:

* It removes keys that no longer exist.
* It adds any missing keys with default english translation or empty, You should check how crowdin or standard format and how they know if this is already been translated or no.

abuaboud (changes requested):

PR #7638, opened 2025-05-09
Update local.mdx

gotoxp45:
Update the URL to land on sign-in page

### Relevant User Scenarios

Fixes # (issue)


abuaboud:
Hi @gotoxp45 

I don't think it does matter as both redirect to sign-in page, localhost:4200 is easier to type for someone.

gotoxp45:
Yeah I got it but when I was visiting localhost:4200 it wasn't redirecting
when I raise it, but now it is working properly

PR #7699, opened 2025-05-12
feat: revamp folders

Louai-Zokerburg:
## What does this PR do?
- Move folders component to new row
- add emojis
- refactor delete model so it's open state can be used uncontrolled or controlled

Fixes # (issue)


packages/server/api/src/app/flows/folder/folder.module.ts
```diff
@@ -104,6 +105,19 @@ const folderController: FastifyPluginAsyncTypebox = async (fastify) => {
...
+
+    fastify.patch(
+        '/order',
+        UpdateFolderOrderParams,
+        async (request, reply) => {
+            await folderService(request.log).updateOrder({
+                projectId: request.principal.projectId,
+                folderOrders: request.body.folderOrders,
+            })
+            // TODO: Add event logging
```
Louai-Zokerburg: @abuaboud This is supposed to log folders sorted event.

would really appreciate some help here

abuaboud (approved):

PR #7701, opened 2025-05-12
feat(metabase): add timeout so graph can load

Kevinyu-alan:
## What does this PR do?

When using playwright to go to the iframe of a metabase graph, it doesn't let enough time for the graph to load so it only screens a loading page. Added a property wait time for the timeout.

### Relevant User Scenarios

Fixes # (issue)


kishanprmr:
@Kevinyu-alan Could you sync your PR with the main branch? There are some merge conflicts.

PR #7705, opened 2025-05-12
fix(handshake): handle undefined payload by fetching data when necessary

abuaboud:
### Relevant User Scenarios

Fixes # (issue)


igs-raiderler:
Very clean code with good test coverage. This is a great example of best practices.

PR #7734, opened 2025-05-14
feat: support multiple mcps

### LINKED ISSUE #7869: MCP run history

Gamal72:
### Relevant User Scenarios

Fixes #7869 #7872 #7871 #7870


abuaboud:
We need the following for multiple MCPs:

* Give each one a proper name (using Server ID looks messy). Check the tables in the Editing Experience for reference.
* Only show the "Last Modified" instead of created . yodate column in the table.
* Show tool icons in the table, like we do in the Flows section.

abuaboud:
I still can save without selecting connection
image

abuaboud:
We do not like the design that show icons on the hover, It should be three dots always visible with delete menu inside it.

image

abuaboud:
This has quite different style, I don't mind it but we can't use different one, I would use the same component used in the runs.

MCP Runs:
image

Existing: 
image

packages/react-ui/src/app/mcp/mcp-connect.tsx
```diff
@@ -0,0 +1,559 @@
...
+    | 'secondary'
+    | 'link';
+  icon: React.ReactNode;
+  className?: string;
+  disabled?: boolean;
+  hasPermission?: boolean;
+};
+
+// Reusable ButtonWithTooltip component
+const ButtonWithTooltip = ({
```
abuaboud: You can move it to main components

packages/react-ui/src/app/mcp/mcp-connect.tsx
```diff
@@ -0,0 +1,559 @@
+import { ReloadIcon } from '@radix-ui/react-icons';
```
abuaboud: This place can have it's own folder

mcp-connect with index.ts

and then split the component in each file, we strongly recommend one component per file.

abuaboud (approved):

PR #7740, opened 2025-05-15
feat(dropbox): download file action

Daniel-Klippa:
## What does this PR do?

Created a new action that allows the user to download a file

The action downloads a file from the specified dropbox path. There was a previous action: "Get temporary file link", but this action doesn't return an ActivePieces file so it could not be linked to other pieces and/or actions.

### Relevant User Scenarios

 - Download a file from Dropbox


kishanprmr:
@allcontributors please add @Daniel-Klippa for code

kishanprmr (approved):

PR #7835, opened 2025-05-23
feat: add env var to disable automatic flow enabling

AdamSelene:
## What does this PR do?

Add a way to stop automatically enabling flows when a new version is published.
By default the current behavior is unchanged.

### Relevant User Scenarios

Fixes # (issue)


hazemadelkhalel:
https://github.com/activepieces/activepieces/pull/7953

PR #7853, opened 2025-05-25
fix: change patch to post in billing and folders

Louai-Zokerburg:
## What does this PR do?

Fixes # (issue)


abuaboud:
It seems missing sqlite migration

PR #7908, opened 2025-05-31
refactor: revamp ai provider

amrdb:
### Relevant User Scenarios

Fixes # (issue)


packages/server/api/src/app/flags/flag.service.ts
```diff
@@ -280,6 +287,8 @@ export const flagService = {
         if (!cloudPlatformId || !platformId) {
             return false
         }
+        console.log('platformId', platformId)
+        console.log('cloudPlatformId', cloudPlatformId)
```
abuaboud: Nit: remove?

packages/react-ui/src/lib/api.ts
amrdb, line 1: Will this cause anything to break?
I did this because I wanted the api request to /flags to have the auth token.
Maybe instead of my change I can just remove /flags from disallowedRoutes.

abuaboud (changes requested):

abuaboud (approved):

PR #7918, opened 2025-06-02
fix: display error notification

### LINKED ISSUE #7932: [BUG]: Prevent activation on trigger error
**Describe the bug**
When the onEnable method of a trigger returns an error (e.g., with a non-OK response), the workflow is still activated. Additionally, no error notification is displayed to the user, making it unclear that the trigger failed.

**To Reproduce**
1. Go to a flow with a trigger that can fail on enable (e.g., due to invalid configuration/ api error).
2. Set the trigger and try to enable the flow.
3. onEnable returns an error (e.g., EngineResponseStatus not OK).
4. The flow is activated anyway, and no error is shown in the UI.

**Expected behavior**
The workflow should not be activated when onEnable fails, and an error notification should be shown in the interface to inform the user of the failure reason.

**Screenshots**
Not applicable at the moment.

**Additional context**
Throwing a simple error inside the onEnable function or the preUpdateStatus logic did not prevent the flow from being activated nor trigger UI feedback.

obenazouz-wedof:
## What does this PR do?

display error notification and prevent activation if trigger onEnable fails

Implemented error handling in the flow activation logic to ensure that when the onEnable method of a trigger returns an error, the workflow is not activated. If the trigger fails to enable (i.e., returns a non-OK status), an error is now thrown and properly propagated, allowing the interface to display an error notification to the user. This prevents invalid or incomplete workflow activations and improves visibility into trigger issues.

### Relevant User Scenarios

Fixes #7932


hazemadelkhalel:
Hello @obenazouz-wedof 

Thanks for fixing this! Two small suggestions:

1. **Replace ```Error``` with ```ActivepiecesError```**

    - For consistency, could you please use ActivepiecesError instead of Error?

2. **Centralize the Error Condition**

   - I noticed that the same condition is checked in three places:

       - Lines 42 and 95 in packages/server/api/src/app/flows/flow/flow-service-side-effects.ts

       - Line 30 in packages/server/api/src/app/webhooks/webhook-simulation/webhook-simulation-side-effects.ts

    - To avoid duplication, could you move the error-throwing logic inside the ```enablePieceTrigger``` function?
    
    
image


Let me know if that works!

obenazouz-wedof:
Hello @hazemadelkhalel ,

Thank you for your suggestions. I have made the requested changes,

You can verify the changes at your convenience

hazemadelkhalel (approved):

PR #7970, opened 2025-06-04
feat(google-drive): duplicate file as a specific mime type (gdocs or gsheet)

geekyme:
## What does this PR do?

Allows opening a csv file in google drive in google spreadsheet format by supplying an optional param when using `Duplicate File`.

In activepieces:
Screenshot 2025-06-04 at 4 18 08 PM

Result in gdrive.
Screenshot 2025-06-04 at 4 25 24 PM

### Relevant User Scenarios

- if user need to use google sheets actions on a .csv, its not possible until they open that .csv in google sheets.

### DIFF
--- a/packages/pieces/community/google-drive/src/lib/action/duplicate-file.action.ts
+++ b/packages/pieces/community/google-drive/src/lib/action/duplicate-file.action.ts
@@ -25,6 +25,19 @@ export const duplicateFileAction = createAction({
       description: 'The ID of the folder where the file will be duplicated',
       required: true,
     }),
+    mimeType: Property.StaticDropdown({
+      displayName: 'Convert to Google Workspace Format',
+      description: 'Optionally convert the file to a Google Workspace format',
+      required: false,
+      options: {
+        options: [
+          {
+            label: 'CSV to Google Spreadsheet',
+            value: 'application/vnd.google-apps.spreadsheet',
+          }
+        ],
+      },
+    }),
     include_team_drives: common.properties.include_team_drives,
   },
   async run(context) {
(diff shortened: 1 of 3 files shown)


packages/pieces/community/google-drive/src/lib/action/duplicate-file.action.ts
```diff
@@ -25,6 +25,19 @@ export const duplicateFileAction = createAction({
...
+      required: false,
+      options: {
+        options: [
+          {
+            label: 'CSV to Google Spreadsheet',
+            value: 'application/vnd.google-apps.spreadsheet',
+          }
+        ],
+      },
+    }),
```
AbdulTheActivePiecer: Can you describe to us the workflow @geekyme , like is the file that's being duplicated a google sheet or a csv file ?

If it is a CSV and you would like to modify the mimetype of the file as you are duplicating it, then it would be better to change the dropdown name to "Mime Type" and add options that represent the possible types one can use like application/vnd.google-apps.spreadsheet and so on.

geekyme-fsmk: the file that is being duplicated is a .csv file. 
@AbdulTheActivePiecer I named the dropdown this way as most non-technical users will not understand what is a "mime type".

geekyme: the file that is being duplicated is a .csv file. 
@AbdulTheActivePiecer I named the dropdown this way as most non-technical users will not understand what is a "mime type".

AbdulTheActivePiecer: Hi @geekyme, sorry for the delayed reply, the team is on holiday and I was sick as well 😂

Anyhow, the problem with the drop down is that no one will understand its use without your context, maybe instead of mime type, you can name it "Duplicate as" and put two options one for google sheets and another for google docs and put the values of the mime types there, and in the description say "if left unselected the file will be duplicated as it is", what do you think ?

geekyme: Updated according to your suggestion thanks!

AbdulTheActivePiecer: tested, works fine, when the conversion is not possible, gdrive throws: "The requested conversion is not supported." 

Merging.

PR #7979, opened 2025-06-04
feat(issues): link issues to runs, archive issues and more enhancements

7adidaz:
## What does this PR do?
-


packages/server/api/src/app/app.ts
```diff
@@ -2,11 +2,13 @@ import { ApplicationEventName, AuthenticationEvent, ConnectionEvent, FlowCreated
 import { PieceMetadata } from '@activepieces/pieces-framework'
 import { AppSystemProp, exceptionHandler, rejectedPromiseHandler } from '@activepieces/server-shared'
 import { ApEdition, ApEnvironment, AppConnectionWithoutSensitiveData, Flow, FlowRun, FlowTemplate, Folder, McpWithTools, ProjectRelease, ProjectWithLimits, spreadIfDefined, UserInvitation } from '@activepieces/shared'
+import { fastifySchedule } from '@fastify/schedule'
```
abuaboud: You can remove this and replace it with our system jobs (look for system-jobs.ts), It's already implemented and based on redis.

Which mean one of the app server at most will execute it and it's persistent

packages/shared/src/lib/issues/issue-dto.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export enum IssueStatus {
+    RESOLVED = 'RESOLVED',
+    UNRESOLVED = 'UNRESOLVED',
+    ARCHIVED = 'ARCHIVED',
+}
+
+const Step = Type.Object({
+    stepId: Type.String(),
+    // status: Type.String(),
```
abuaboud: Remove the commented code

packages/server/shared/src/lib/system-props.ts
```diff
@@ -78,6 +78,8 @@ export enum AppSystemProp {
     SHOW_CHANGELOG = 'SHOW_CHANGELOG',
     ENABLE_FLOW_ON_PUBLISH = 'ENABLE_FLOW_ON_PUBLISH',
     
+    ISSUE_ARCHIVE_DAYS = 'ISSUE_ARCHIVE_DAYS',
```
abuaboud, line 81: Add to environment-variables.md, in the same code base under docs folder

https://www.activepieces.com/docs/install/configuration/environment-variables

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -112,6 +111,15 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
         else {
             app.io.to(request.principal.projectId).emit(WebsocketClientEvent.FLOW_RUN_PROGRESS, runId)
         }
+ 
+        await flowRunService(request.log).updateStatus({
```
abuaboud: Why this was moved?

.env.example
```diff
@@ -26,3 +26,4 @@ AP_REDIS_PORT=6379
 AP_FLOW_TIMEOUT_SECONDS=600
 AP_TELEMETRY_ENABLED=true
 AP_TEMPLATES_SOURCE_URL="https://cloud.activepieces.com/api/v1/flow-templates"
+AP_ISSUE_ARCHIVE_DAYS=1
```
abuaboud: Remove this, there is default value already provided in system.ts, use that place to provide a default value and I would suggest something like 7 days as start.

packages/react-ui/src/features/issues/api/issues-api.ts
```diff
@@ -4,16 +4,23 @@ import {
...
-} from '@activepieces/ee-shared';
-import { SeekPage } from '@activepieces/shared';
+  SeekPage,
+} from '@activepieces/shared';
 
 export const issuesApi = {
   list(request: ListIssuesParams): Promise<SeekPage<PopulatedIssue>> {
     return api.get<SeekPage<PopulatedIssue>>('/v1/issues', request);
   },
   resolve(issueId: string) {
```
abuaboud: Why is this needed? the user should be only to archive now from the user interface right?

packages/server/api/.env.tests
```diff
@@ -24,3 +24,4 @@ AP_CLOUD_PLATFORM_ID="cloud-id"
 AP_APPSUMO_TOKEN="app-sumo-token"
 AP_PIECES_SYNC_MODE=NONE
 AP_CONTAINER_TYPE=WORKER_AND_APP
+AP_ISSUE_ARCHIVE_DAYS=1
```
abuaboud: Remove and use move the default value to system.ts

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -1,49 +1,113 @@
...
+        projectId: flowRun.projectId, 
+        fileId: flowRun.logsFileId ?? '', 
+        type: FileType.FLOW_RUN_LOG,
+    })
+    const rawData = Buffer.from(flowSteps.data).toString('utf-8')
+    const executionOutput = JSON.parse(rawData) as ExecutioOutputFile
+    const executionState = executionOutput.executionState
+
+    const failedStep = Object.entries(executionState.steps).find(
+        ([_, step]) => step.status === StepOutputStatus.FAILED,
```
abuaboud: I think this doesn't work on loops  as they are nested and not directly in top level steps.

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -1,49 +1,113 @@
...
+
+    return failedStep[0]
+}
+
+const getStepFromFlow = async (flowId: string, stepId: string, log: FastifyBaseLogger) => {
+    const flowVersion = await flowVersionService(log).getLatestLockedVersionOrThrow(flowId)
+    const allSteps = flowStructureUtil.getAllSteps(flowVersion.trigger)
+    const step = allSteps.find(s => s.name === stepId)
+    
+    if (!step) {
```
abuaboud: I think this can be moved under `issueServices` as basically this is helper function and people usually reads code from above to below.

It's confusing to read helper function than understand what is the main purpose of file.

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -1,49 +1,113 @@
...
+        throw new ActivepiecesError({
```
abuaboud: You can just use assertNotNullOrUndefined, because we don't expect it to not be there?

packages/server/api/src/app/workers/engine-controller.ts
7adidaz: long story short... before moving this... I couldn't access the run flow logs to know which step failed. After doing this, since the logs are already saved, I can access them later in the issue service. This had me debugging for about 6 hours, but Abdul helped by pointing out this file.

.env.example
7adidaz: Thanks for pointing this out...

packages/react-ui/src/features/issues/api/issues-api.ts
7adidaz: The resolve function was already implemented... I could remove it, but I see there is no need to... maybe we need it later.

packages/server/api/src/app/flows/issues/issues-service.ts
7adidaz: No, steps inside the logs are not nested.. I have inspected this multiple times, and it's correct.. example:

```
{
  "executionState": {
    "steps": {
      "trigger": {
        "type": "PIECE_TRIGGER",
        "status": "SUCCEEDED",
        "input": {},
        "output": {
          "method": "GET",
          "headers": {
            "host": "127.0.0.1:4200",
            "user-agent": "curl/8.5.0",
            "accept": "*/*",
            "connection": "close"
          },
          "queryParams": {},
          "rawBody": ""
        }
      },
      "step_1": {
        "type": "PIECE",
        "status": "SUCCEEDED",
        "input": {
          "mapping": {
            "newProperty": "oldProperty"
          }
        },
        "output": {
          "newProperty": "oldProperty"
        },
        "duration": 0.17397800000253483
      },
      "step_3": {
        "type": "PIECE",
        "status": "PAUSED",
        "input": {
          "unit": "seconds",
          "delayFor": "20"
        },
        "output": {},
        "duration": 305.2004029999998
      }
    }
  }
}
```

packages/server/api/src/app/flows/issues/issues-service.ts
7adidaz: you have a point... i will do it

packages/server/shared/src/lib/system-props.ts
7adidaz: ack

packages/shared/src/lib/issues/issue-dto.ts
7adidaz: ack

packages/server/api/.env.tests
7adidaz: ack

packages/server/api/src/app/flows/issues/issues-service.ts
7adidaz: ack

packages/react-ui/src/features/issues/api/issues-api.ts
abuaboud: I think let's remove it we can always add it later if needed

7adidaz: ack

packages/server/api/src/app/flows/issues/issues-service.ts
7adidaz: i missed your point about the loop, and I will update it in the next patch

packages/server/api/src/app/app.ts
7adidaz: Good point.

packages/react-ui/src/app/routes/flows/issues-table/columns.tsx
```diff
@@ -72,4 +81,36 @@ export const issuesTableColumns: ColumnDef<
       );
     },
   },
+  {
+    id: 'status',
+    accessorKey: 'status',
+    header: t('Status'),
```
abuaboud: image

**Major**
It looks different, It's not using data header component

packages/react-ui/src/app/routes/flows/issues-table/columns.tsx
```diff
@@ -72,4 +81,36 @@ export const issuesTableColumns: ColumnDef<
...
+                ? '#92400E'
+                : issue.status === 'RESOLVED'
+                ? '#065F46'
+                : '#374151',
+          }}
+        >
+          {issue.status.toLowerCase()}
+        </div>
+      );
+    },
```
abuaboud, line 103: The status is all lower case that's incorrect in English, first letter should be upper case.

packages/react-ui/src/app/routes/flows/issues-table/columns.tsx
```diff
@@ -72,4 +81,36 @@ export const issuesTableColumns: ColumnDef<
...
   },
+  {
+    id: 'status',
+    accessorKey: 'status',
+    header: t('Status'),
+    cell: ({ row }) => {
+      const issue = row.original;
+      return (
+        <div
+          className="inline-flex items-center px-2.5 py-0.5 rounded-full text-xs font-medium capitalize"
```
abuaboud: **Design**:

Instead of using random colors, we can use  `StatusIconWithText`

```ts
    cell: ({ row }) => {
      const issue = row.original;
      const status = issue.status === IssueStatus.ARCHIVED ? t('Archived') : issue.status === IssueStatus.RESOLVED ? t('Resolved') : t('Unresolved');
      const icon = issue.status === IssueStatus.ARCHIVED ? Archive : issue.status === IssueStatus.RESOLVED ? Check : X;
      return (
        <div className="flex justify-center">
          <StatusIconWithText
            variant={
              issue.status === IssueStatus.ARCHIVED
                ? 'default'
                : issue.status === IssueStatus.RESOLVED
                ? 'success'
                : 'error'
            }
            icon={icon}
            text={status}
          />
        </div>
        ```

File: packages/react-ui/src/app/routes/flows/issues-table/index.tsx [hunk 14]
```diff
@@ -19,83 +19,144 @@ import { toast } from '@/components/ui/use-toast';
...
+  const statusValues = searchParams.getAll('status');
+  const hasStatusFilter = statusValues.length > 0;
+
+  useEffect(() => {
+    if (!hasStatusFilter) {
+      const newSearchParams = new URLSearchParams(searchParams);
+      newSearchParams.append('status', IssueStatus.UNRESOLVED);
+      setSearchParams(newSearchParams);
+    }
+  }, [hasStatusFilter, searchParams, setSearchParams]);
```
abuaboud: I think you just need to `useEffectOnce`

image

If you unselect this, It will be reverted to
image

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -1,51 +1,71 @@
...
+                flowId: flowRun.flowId,
                 id: issueId,
                 lastOccurrence: date,
-                count: 0,
-                status: IssueStatus.ONGOING,
+                stepName: flowRun.failedStepName,
+                status: IssueStatus.UNRESOLVED,
                 created: date,
                 updated: date,
+                count: 0,
```
abuaboud: Nit: Count is not part of schema, you should remove

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -1,51 +1,71 @@
...
-            status: IssueStatus.ONGOING,
+        const issue = await repo().findOneByOrFail({
+            projectId: flowRun.projectId,
+            flowId: flowRun.flowId,
+            stepName: flowRun.failedStepName,
+            status: IssueStatus.UNRESOLVED,
         })
         
-        return updatedIssue
+        issue.step = await getStepFromFlow(flowRun.flowId, flowRun.failedStepName, log)
```
abuaboud: This should be centralized

```
async function enrichIssue(issue: Issue, log: FastifyBaseLogger) {
    const flowVersion = await flowVersionService(log).getLatestLockedVersionOrThrow(issue.flowId)
    const count = await flowRunRepo().countBy({
        flowId: issue.flowId,
        failedStepName: issue.stepName,
        status: In([FlowRunStatus.FAILED, FlowRunStatus.INTERNAL_ERROR, FlowRunStatus.TIMEOUT]),
    })
    return {
        ...issue,
        step: isNil(issue.stepName) ? undefined : await getStepFromFlow(flowVersion, issue.stepName),
        flowDisplayName: flowVersion.displayName ?? '',
        count,
    }
}

```

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+
+        if (status && status.length > 0) {
+            query = query.andWhere('status IN (:...statuses)', { statuses: status })
+        }
 
         const { data, cursor: newCursor } = await paginator.paginate(query)
 
-        const populatedIssues = await Promise.all(data.map(async issue => {
+        const issuesWithCounts = await Promise.all(data.map(async issue => {
+            // Fixed query with proper join between issue and flow_run
```
abuaboud: This LLM comment should be removed :)

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+            const count = await repo()
```
abuaboud: I think this function joins flow_run with issues but that really no need for the joining at all.

It's simple count by flowId, failedStepName and failed status.

```
    const count = await flowRunRepo().countBy({
        flowId: issue.flowId,
        failedStepName: issue.stepName,
        status: In([FlowRunStatus.FAILED, FlowRunStatus.INTERNAL_ERROR, FlowRunStatus.TIMEOUT]),
    })
    ```

File: packages/server/api/src/app/flows/issues/issues-service.ts [hunk 19]
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+                .select('COUNT(*)', 'count')
+                .innerJoin('flow_run', 'fr', 'i."flowId" = fr."flowId" AND i."stepName" = fr."failedStepName"')
+                .where('fr."flowId" = :flowId', { flowId: issue.flowId })
+                .andWhere('fr.status IN (:...statuses)', { statuses: [FlowRunStatus.FAILED, FlowRunStatus.INTERNAL_ERROR, FlowRunStatus.TIMEOUT] })
+                .andWhere('fr."failedStepName" = :stepName', { stepName: issue.stepName })
+                .andWhere('i.id = :issueId', { issueId: issue.id })
+                .getRawOne()
+                .then(result => parseInt(result.count, 10))
+
+            if (count === 0 && issue.status === IssueStatus.UNRESOLVED) {
```
abuaboud: Resolving issues inside the enrichment is wrong. 

Image the user requests 10 per page, we only returns something less than 10 but they have next page.

You should make function and call it at beginning of listing.


````ts
async function resolveIssueWithZeroCount(log: FastifyBaseLogger) {
    await repo().createQueryBuilder()
        .update(IssueEntity)
        .set({
            status: IssueStatus.RESOLVED,
            updated: new Date().toISOString(),
        })
        .where('status = :status', { status: IssueStatus.UNRESOLVED })
        .andWhere(_issue => {
            const subQuery = flowRunRepo()
                .createQueryBuilder('flowRun')
                .select('flowRun.id')
                .where('flowRun."flowId" = IssueEntity."flowId"')
                .andWhere('flowRun."failedStepName" = IssueEntity."stepName"')
                .andWhere('flowRun.status IN (:...failedStatuses)', { FAILED_STATES })
                .getQuery()
            return `NOT EXISTS (${subQuery})`
        })
        .execute()

    log.info('Resolved issues with zero failed runs')
}
````

File: packages/server/api/src/app/flows/issues/issues-service.ts [hunk 20]
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+                await this.updateById({
+                    projectId: issue.projectId,
+                    id: issue.id,
+                    status: IssueStatus.RESOLVED,
+                })
+                issue.status = IssueStatus.RESOLVED
+            }
+
+            const actualStatus = issue.status === IssueStatus.ARCHIVED
```
abuaboud: The first statement doesn't make sense at all, If it's archived then archived?

It can be rewritten like that:
(count === 0? resolved: same)

abuaboud: Okay this is even worse issue.status=RESOLVED in previous if statement block already done this, why we need this?

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+
+            return {
+                issue,
+                count,
+                actualStatus,
+            }
+        }))
+
+        // Filter issues based on the actual calculated status
+        const filteredIssues = issuesWithCounts.filter(({ actualStatus }) => {
```
abuaboud: Again we should not filter after fetching the page.

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+            if (!status || status.length === 0) {
+                return true
+            }
+            
+            return status.includes(actualStatus)
+        })
+
+        const populatedIssues = await Promise.all(filteredIssues.map(async ({ issue, count, actualStatus }) => {
```
abuaboud: The whole function is catastrophic, the three blocks can be done in one line.


```ts
     const populatedIssues = await Promise.all(data.map(issue => enrichIssue(issue, log)))
        return paginationHelper.createPage<PopulatedIssue>(populatedIssues, newCursor)
    },
    ```

File: packages/server/api/src/app/flows/issues/issues-service.ts [hunk 23]
```diff
@@ -98,43 +169,74 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
+    },
 })
 
 type UpdateParams = {
     projectId: string
     id: string
     status: IssueStatus
+}
+
+const getStepFromFlow = async (flowId: string, stepName: string, log: FastifyBaseLogger) => {
```
abuaboud, line 172: You are fetching flowVersion twice, once in list and one here.

But if you have enrichIssue function that just takes issue and use it both in list and get it would be much better.

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
             const flowVersion = await flowVersionService(log).getLatestLockedVersionOrThrow(issue.flowId)
+            if (!isNil(issue.stepName)) {
+                issue.step = await getStepFromFlow(issue.flowId, issue.stepName ?? '', log) 
+            }
+            
             return {
                 ...issue,
                 flowDisplayName: flowVersion.displayName,
+                count,
+                status: actualStatus,
```
abuaboud: If you are using actualStatus there, then issue.status=RESOLVED in the if statement block for count = 0 is not used

packages/server/api/src/app/flows/issues/issues-service.ts
```diff
@@ -57,20 +77,71 @@ export const issuesService = (log: FastifyBaseLogger) => ({
...
             }
         }))
+
         return paginationHelper.createPage<PopulatedIssue>(populatedIssues, newCursor)
     },
```
abuaboud, line 91: Here is the full rewrite for this function

```ts
    async list({ projectId, cursor, limit, status }: ListIssuesParams): Promise<SeekPage<PopulatedIssue>> {
        await resolveIssueWithZeroCount(log)

        const decodedCursor = paginationHelper.decodeCursor(cursor ?? null)
        const paginator = buildPaginator({
            entity: IssueEntity,
            query: {
                limit,
                order: 'ASC',
                afterCursor: decodedCursor.nextCursor,
                beforeCursor: decodedCursor.previousCursor,
            },
        })

        let query = repo().createQueryBuilder(IssueEntity.options.name)
            .where({ projectId })

        if (!isNil(status) && status.length > 0) {
            query = query.andWhere('status = :status', { status: In(status) })
        }

        const { data, cursor: newCursor } = await paginator.paginate(query)

        const populatedIssues = await Promise.all(data.map(issue => enrichIssue(issue, log)))
        return paginationHelper.createPage<PopulatedIssue>(populatedIssues, newCursor)
    },
    ```

abuaboud (changes requested):

PR #7985, opened 2025-06-05
chore: sync connections on flows pushing/deleting to git

hazemadelkhalel:
### Relevant User Scenarios

Fixes # (issue)


packages/server/api/src/app/ee/projects/project-release/git-sync/git-sync.service.ts
```diff
@@ -179,6 +173,37 @@ export const gitRepoService = (_log: FastifyBaseLogger) => ({
                     })
                 }
                 await gitHelper.commitAndPush(git, gitRepo, request.commitMessage ?? `chore: updated flows ${request.flowIds.join(', ')}`)
+
+                const connections = await appConnectionService(log).list({
```
abuaboud: I think this can be rewritten in unified way.

Both of DELETE_FLOW and PUSH_FLOW is essentially pushing new state, instead you can have one function.

- It does read all flows in the directory
- Find all connections related to these flows
- Just override the connections state instead of figure which one to delete and which one to add

PR #8006, opened 2025-06-07
fix: adjust sidebar background for dark mode (#8005)

### LINKED ISSUE #8005: [BUG]: the code piece fails
**Describe the bug**  
In dark mode, the sidebar background color does not adapt properly and remains light, similar to the light mode styling. This creates a visual inconsistency and affects the overall dark theme experience.

**To Reproduce**  
1. Switch the application to dark mode  
2. Observe the sidebar background color  
3. Compare it with the main layout/background  
4. Notice that the sidebar remains a different color that doesn't match the rest of the theme

**Expected behavior**  
The sidebar background should change to a darker tone that aligns with the rest of the UI when dark mode is active.

**Screenshots**  

Image

**Additional context**  
This is a UI inconsistency specific to dark mode. The fix would involve updating the sidebar container's background color in dark mode styles.

jadhavharshh:
## What does this PR do?

This PR resolves a UI inconsistency in dark mode where the sidebar background remained different colored, matching the light theme instead of adapting to the dark mode styling. The fix ensures the sidebar now correctly reflects the active theme, improving overall visual coherence.

The theme setup was already in place — the sidebar was using the wrong variable for its background. This PR updates it to use the correct theme-aware value so it displays appropriately in dark mode.

### Relevant User Scenarios

- Users browsing the app in **dark mode** will now see a sidebar background that matches the rest of the UI theme.
- This improves readability and visual consistency in dark mode.

### Before / After Screenshots (Optional)

**Before (Dark Mode):**  
image->

**After (Dark Mode):**  
image

Fixes #8005

### DIFF
--- a/packages/react-ui/src/styles.css
+++ b/packages/react-ui/src/styles.css
@@ -113,7 +113,7 @@
     --chart-4: 280 65% 60%;
     --chart-5: 340 75% 55%;
 
-    --sidebar-background: 240 5.9% 10%;
+    --sidebar-background: 20 14.3% 4.1%;
     --sidebar-foreground: 240 4.8% 95.9%;
     --sidebar-primary: 224.3 76.3% 48%;
     --sidebar-primary-foreground: 0 0% 100%;


aadarsh-nagrath:
i guess it was looking fine with before bg color, now sidebar seems to blend too much. You could fix that logo color maybe, it should be white in dark mode.

packages/react-ui/src/styles.css
```diff
@@ -113,7 +113,7 @@
     --chart-4: 280 65% 60%;
     --chart-5: 340 75% 55%;
 
-    --sidebar-background: 240 5.9% 10%;
+    --sidebar-background: 20 14.3% 4.1%;
```
AbdulTheActivePiecer: please set this to  --sidebar-background: 20 14.3% 6%; and I will merge it afterwards

Issue #8010, opened 2025-06-08
[BUG]: The Simplified Chinese translation is incorrect.


caizhiqiangzeyuan:
**Describe the bug**
The translation.json file located at activepieces/packages/react-ui/public/locales/zh contains translations that are not in Simplified Chinese (e.g., '斷開連接' instead of '断开连接'). This may cause inconsistency in the UI for Chinese users.

**To Reproduce**
1. mouse hover on 'Project Settings'
2. Click on 'Appearance'
3. Select the language  '简体中文'
4. Some Simplified Chinese translations are incorrect.

**Expected behavior**

<img width="1300" alt="Image" src="https://github.com/user-attachments/assets/5ea3e2f0-c96f-4d07-a91a-ad656a1eda4f" />

**Screenshots**

<img width="1301" alt="Image" src="https://github.com/user-attachments/assets/d156d6b3-9a5d-46c1-8ea2-4f87f7ccc4c6" />

**Additional context**

abuaboud:
Please use crowdin we have updated it and fixed it

PR #8014, opened 2025-06-09
fix: the simplified chinese translation is incorrect #8011

caizhiqiangzeyuan:
## What does this PR do?

The translation.json file located at activepieces/packages/react-ui/public/locales/zh contains translations that are not in Simplified Chinese (e.g., '斷開連接' instead of '断开连接'). This may cause inconsistency in the UI for Chinese users.

I added separate JSON files for Simplified Chinese and Traditional Chinese.

### Relevant User Scenarios

image

Fixes # (issue)
[bug address](https://github.com/activepieces/activepieces/issues/8010)


abuaboud:
Hi @caizhiqiangzeyuan 

Thank you for contributing this, but if we merge it it will be overridden by crowdin.

Is it okay if you contribute it directly to https://crowdin.com/project/activepieces?

https://www.activepieces.com/docs/about/i18n

abuaboud:
We have used machine translation on simplified chineese
https://crowdin.com/project/activepieces

PR #8035, opened 2025-06-11
feat(slack): add customizable connection scopes

AdamSelene:
## What does this PR do?

Scopes can now be customized to avoid `missing_scope` errors when doing a Custom API call

Capture d’écran 2025-06-11 à 11 18 18

### Relevant User Scenarios

Fixes # (issue)


kishanprmr:
Thanks for the PR! I’ll discuss with the team how we can make the scopes more customizable and will update you on the PR accordingly.

PR #8062, opened 2025-06-13
feat: add run.id to trigger context

kamal:
## What does this PR do?

Currently, only actions have access to run within the run() context. For triggers, run is accessible in the onStart callback, but not within run itself. This capability is useful for enabling logging inside triggers for log correlation purposes.

### Relevant User Scenarios

Fixes # (issue)


abuaboud:
Hi @kamal 

I think the correct way is to use onStart, the reason is that run is not really associated to any run as it's outside run lifecycle for example.

If there is polling job like google sheets, the run method may return multiple rows (array of payload) in that case, each payload will have it's own run.

So we have created onStart where if you want to do an side effect before start executing the flow.

PR #8103, opened 2025-06-18
chore: added deps to custom property

AbdulTheActivePiecer:


AbdulTheActivePiecer:
example: 

```js
import { isNil } from '@activepieces/shared';
function randomColor(): string {
  return '#' + Math.floor(Math.random() * 16777215).toString(16).padStart(6, '0');
}

export const test = createAction({
  name: 'test',
  displayName: 'Test Custom Property',
  description: 'Test Custom Property',
  errorHandlingOptions: {
    continueOnFailure: {
      hide: true,
    },
    retryOnFailure: {
      hide: true,
    },
  },
  props: {
    test:Property.Custom({
      code:(({value,onChange,containerId, deps})=>{
      console.log(deps.randomColor())
       console.log(deps.isNil(null))
      }),
      displayName: 'Custom Property',
      required: true,
      deps: {
        randomColor,
        isNil,
      }
    }), 
  },
  async run(context) {
    return {
      success: true,
    };
  },
});
```

packages/pieces/community/framework/src/lib/property/input/index.ts
```diff
@@ -235,24 +235,32 @@ export const Property = {
...
+      deps?: undefined;
+    } : {
+      /**
+       * This is a record of dependencies that will be serialized and passed to the custom property code function, if a dependency is a function, it must be a pure function that does not depend on any external state.
+       */
+      deps:D;
+    })
   ): R extends true ? CustomProperty<true> : CustomProperty<false> {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const seralize = require('serialize-javascript');
```
abuaboud, line 255: Why not use import at beginning of file?

AbdulTheActivePiecer: because this package is imported in the frontend and serialize javascript is a node package, loading it in the react would break react.

abuaboud: That's tricky I don't think that will work because nx will most likly skip packaging this, maybe we can go one level up or so?

PR #8117, opened 2025-06-20
feat(jira-cloud): Updated status trigger

prasanna2000-max:
### What does this PR do?
This PR introduces a new Jira trigger which uses **statuscategorychangedate** that enhances update detection by allowing users to poll based on the status change timestamp.

The new trigger supports polling based on statuscategorychangedate, enabling more precise control — so it only fires when the ticket’s status changes, avoiding false triggers from unrelated field such as assignee, description are updated.

UI Preview:
image

### Relevant User Scenarios
Previously:

> Suppose a ticket’s status changed to Approved a week ago.
> If the issue is later updated (e.g. someone edits the description), the trigger fires again — even though the status didn’t change.
> This is because the trigger was based solely on issue.fields.updated.

Now:
> Users can use updatedIssueStatus trigger , so it triggers only when the status changes.

### DIFF
--- a/packages/pieces/community/jira-cloud/package.json
+++ b/packages/pieces/community/jira-cloud/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-jira-cloud",
-  "version": "0.1.3"
+  "version": "0.1.4"
 }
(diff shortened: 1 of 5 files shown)


kishanprmr:
@allcontributors please add @prasanna2000-max for code

PR #8146, opened 2025-06-23
Chore/refactor piece selector utils

AbdulTheActivePiecer:
### Relevant User Scenarios

Fixes # (issue)


packages/react-ui/src/features/pieces/lib/steps-hooks.ts
```diff
@@ -188,15 +225,36 @@ function passSearch(
...
 };
 
 type UseMetadataProps = {
-  searchQuery?: string;
+  searchQuery: string;
   enabled?: boolean;
   type: 'action' | 'trigger';
 };
+
+export const filterOutPiecesWithNoSuggestions = (
```
abuaboud: I don't like random exported function, can they become part of const? if step hooks is wrong place, maybe it should be in 

piece-selector-utilts?

packages/react-ui/src/features/pieces/lib/steps-hooks.ts
```diff
@@ -2,63 +2,94 @@ import { useQueries, useQuery } from '@tanstack/react-query';
...
+} from '../../../lib/types';
+
+import { piecesApi } from './pieces-api';
+import {
+  CORE_ACTIONS_METADATA,
+  CORE_STEP_METADATA,
+  stepUtils,
+} from './step-utils';
+
+const isPopularPieces = (
```
abuaboud: Now I am more convinced for these to exists in piece-selector-utils

packages/react-ui/src/features/pieces/lib/steps-hooks.ts
```diff
@@ -2,63 +2,94 @@ import { useQueries, useQuery } from '@tanstack/react-query';
...
+};
+
+const isUniversalAiPiece = (stepMetadata: StepMetadata) => {
+  if (stepMetadata.type === ActionType.PIECE) {
+    return stepMetadata.categories.includes(PieceCategory.UNIVERSAL_AI);
+  }
+  return false;
+};
+
+const getQueryKeyForStepMetadata = (
```
abuaboud: I would move this below const.

People read from top to bottom, you can't tell what is this file if you start reading helper functions, I usually move them down.

packages/react-ui/src/features/pieces/lib/piece-selector-utils.ts
```diff
@@ -0,0 +1,275 @@
...
+    return acc;
+  }, true);
+};
+
+const maxListHeight = 300;
+const minListHeight = 100;
+const searchInputDivHeight = 48;
+
+// Adjusts piece list height to prevent overflow on short screens
+const useAdjustPieceListHeightToAvailableSpace = () => {
```
abuaboud, line 238: Do we really need this? can it be done with css like max-height and leave it for radix to figure it out?

abuaboud (changes requested):

packages/react-ui/src/features/pieces/lib/piece-selector-utils.ts
AbdulTheActivePiecer: You can't I tried and the solutions output isn't same as this, people on small screens will have their selector clipping the screen and unable to either see the list or the search bar.

packages/react-ui/src/features/pieces/lib/steps-hooks.ts
AbdulTheActivePiecer: this function shouldn't be exported in the first place, will remove the export.

AbdulTheActivePiecer: Moved this logic wholly to pieces-hooks.

PR #8147, opened 2025-06-23
Chore/refactor mcp and todos

AbdulTheActivePiecer:


packages/react-ui/src/app/builder/pieces-selector/create-todo-dialog.tsx
```diff
@@ -0,0 +1,441 @@
...
+  operation: PieceSelectorOperation;
+  hidePieceIcon: boolean;
+};
+
+const getTodoActionName = (todoType: TodoType) => {
+  switch (todoType) {
+    case TodoType.INTERNAL:
+      return TODO_ACTIONS.createTodoAndWait;
+    case TodoType.EXTERNAL:
+      return TODO_ACTIONS.wait_for_approval;
```
abuaboud: rename to `waitForApproval` atleast the key

packages/react-ui/src/app/builder/pieces-selector/create-todo-dialog.tsx
```diff
@@ -0,0 +1,441 @@
...
+  );
+  if (isNil(result)) {
+    toast(INTERNAL_ERROR_TOAST);
+    console.error(`Action ${actionName} not found in piece metadata`);
+    return null;
+  }
+  return result;
+};
+
+const createRouterStep = ({
```
abuaboud: The file is too long, I would suggest to break it down, first It seems those helper functions are good to be moved to create-todo-dialog-utils.ts or so

packages/react-ui/src/app/builder/pieces-selector/create-todo-dialog.tsx
```diff
@@ -0,0 +1,441 @@
...
+                                  className="w-full h-auto rounded"
+                                />
+                              </div>
+                            </div>
+                          </TooltipContent>
+                        </Tooltip>
+                      </h4>
+                      <div className="flex-shrink-0 w-5 h-5">
+                        <div
+                          className={`w-5 h-5 rounded-full grid place-items-center border ${
```
abuaboud: Sounds like we should split this area into components,a nd it should use card.tsx

packages/react-ui/src/app/builder/pieces-selector/create-todo-dialog.tsx
```diff
@@ -0,0 +1,441 @@
...
+                    <h3 className="text-md font-medium mb-3 text-center">
+                      {displayImageType === TodoType.INTERNAL
+                        ? t('Preview (Activepieces Todos)')
+                        : t('Preview (External channel)')}
+                    </h3>
+
+                    <div className="w-full h-[350px] overflow-hidden rounded mb-2 flex items-center justify-center bg-muted/50 relative">
+                      <img
+                        src={
+                          displayImageType === TodoType.INTERNAL
```
abuaboud: Instead of having if statements everywhere, shouldn't we  have an object or so and just refer it there, so all information related the variance are in one place.


One way you can do it is to make these as paramters, and have another components and call this component based on the conidtion and pass the correct customization there.

PR #8148, opened 2025-06-23
feat: introduce retry on failure mode\

abuaboud:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/http/src/lib/actions/send-http-request-action.ts
+++ b/packages/pieces/community/http/src/lib/actions/send-http-request-action.ts
@@ -201,14 +201,22 @@ export const httpSendRequestAction = createAction({
       displayName: 'Timeout(in seconds)',
       required: false,
     }),
-    failsafe: Property.Checkbox({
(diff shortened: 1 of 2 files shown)


packages/pieces/community/http/src/lib/actions/send-http-request-action.ts
```diff
@@ -201,14 +201,22 @@ export const httpSendRequestAction = createAction({
...
+          { label: 'Continue flow', value: 'continue' },
+          { label: 'Retry on internal errors (5xx)', value: '5xx' },
+        ],
+      },
     }),
   },
   errorHandlingOptions: {
     continueOnFailure: { hide: true },
-    retryOnFailure: { defaultValue: true },
+    retryOnFailure: { hide: true },
```
kishanprmr: Shouldn't we provide the value true for retrying on 5xx errors, as the default is false?

abuaboud: Yes can you please do it and merge

PR #8155, opened 2025-06-24
fix: enhance error messages for testing steps

hazemadelkhalel:
If testing step fails it wasn't showing the error and instead saving it so next time you visit the flow you would see the error as sample data.


packages/react-ui/src/app/builder/test-step/test-step-utils.ts
```diff
@@ -4,10 +4,15 @@ function formatErrorMessage(errorMessage: string): string {
     return errorMessage;
   }
 
+  const filteredParts = errorMessagesSplit.filter((part) => part.trim() !== '');
+
```
AbdulTheActivePiecer: call this trimmedErrorMessages. 

The idea is that filtered can mean many things, so it is better to always describe what the filter is doing.

packages/react-ui/src/app/builder/test-step/test-step-utils.ts
```diff
@@ -4,10 +4,15 @@ function formatErrorMessage(errorMessage: string): string {
...
   const indentationStep = '  ';
-  return errorMessagesSplit.reduce((acc, current, index) => {
+  return filteredParts.reduce((acc, current, index) => {
     const indentation = indentationStep.repeat(index);
-    return `${acc}${indentation}Error ${index + 1}: ${current.trim()}\n`;
+    const cleanedPart = current.trim().replace(/^\n+|\n+$/g, '');
```
AbdulTheActivePiecer: I am guessing you are removing endlines here, so just name it errorMessageWithoutEndlines

hazemadelkhalel: Yes removing the trailling whitespaces endlines

PR #8173, opened 2025-06-26
feat/todos - allow for message to be included when approving a todo

joshuaheslin:
# PR is for illustration purposes only!

## What does this PR do?

- Add query params to the Todos webhook so the 'approve' status can include a 'message' field.
- It allows for webhooks to include `message` as a query param so it's available in the flow outputs (for usage downstream such as WHY the user approved)
- UI can be improved to the user can "Approve with Message".
- Notice that this is different to Todo Comments. Todo Comments are a great feature but the system does not allow for the comments to be included back into the flow context.

WARNING: PR is for illustration purposes only!

If AP team is happy with this, we will submit a complete PR.


packages/shared/src/lib/todos/index.ts
```diff
@@ -85,6 +86,7 @@ export const StatusOption = Type.Object({
     description: Nullable(Type.String()),
     variant: Type.Union([Type.Literal(STATUS_VARIANT.POSITIVE), Type.Literal(STATUS_VARIANT.NEGATIVE), Type.Literal(STATUS_VARIANT.NEUTRAL)]),
     continueFlow: Type.Boolean(),
+    message: Type.Optional(Type.String()),
```
joshuaheslin, line 89: UI can "Approve with Message" here

packages/pieces/community/todos/src/lib/actions/wait-for-approval.ts
```diff
@@ -52,6 +58,7 @@ export const waitForApproval = createAction({
     } else {
       return {
         status: ctx.resumePayload.queryParams['status'],
+        message: ctx.resumePayload.queryParams['message'],
```
joshuaheslin, line 61: This allows an optional approve message to be added into the flow context.

PR #8195, opened 2025-06-29
feat: resolve dynamic properties in MCP tools

### LINKED ISSUE #7370: [BUG]: Dynamic Properties missing schema (MCP)
The MCP clients fail to know the schema for any dynamic properties.

It is not accepting (any or known) values.

We need to provide a schema for each one of them.

hazemadelkhalel:
## What does this PR do?

Fixes #7370 

### Token Costs

| Type             | Price per 1M tokens |
|------------------|---------------------|
| **Input**        | $4.00               |
| **Cached Input** | $1.00               |
| **Output**       | $16.00              |
| **Training**     | $100.00 / hour      |

---

### Token Usage Estimates per Request

| Component              | Token Range | Average Tokens |
|------------------------|-------------|----------------|
| Base Prompt Template   | ~200        | 200            |
| User Instructions      | 50 – 200    | 125            |
| Property Schema Values | 100 – 500   | 300            |
| Output Schema          | 50 – 200    | 125            |
| **Total Input Tokens** | –           | **~600**       |
| **Total Output Tokens**| –           | **~200**       |

---

### Cost Per Request

- **Input**: `(600 / 1,000,000) × $1.00` = **$0.0006**
- **Output**: `(200 / 1,000,000) × $16.00` = **$0.0032**

**Total Cost per Request** = **$0.0038**

---

### Cost at Maximum Execution Depth

- **Average Depth**: 3
- **Total Cost per Tool Execution**:  
  `$0.0038 × 3 = $0.0114`

---

### Estimated Costs

| Executions      | Estimated Cost     |
|-----------------|--------------------|
| 100 executions  | $0.76 – $1.14     |
| 1000 executions | $7.6 – $11.40     |


packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
```diff
@@ -71,32 +78,41 @@ async function addPieceToServer(
...
+                        pieceType: pieceMetadata.pieceType,
+                    },
                 )
-                const pieceConnectionExternalId = !isNil(toolPieceMetadata.connectionExternalId) ? `{{connections['${toolPieceMetadata.connectionExternalId}']}}` : undefined
-                const parsedInputs = {
-                    ...defaultValues,
-                    ...params,
-                    auth: pieceConnectionExternalId,
-                }
+                const pieceConnectionExternalId = `{{connections['${toolPieceMetadata.connectionExternalId}']}}`
```
AbdulTheActivePiecer: move this logic inside parseActionParametersFromInstructions

packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
```diff
@@ -253,6 +269,126 @@ async function addFlowToServer(
...
+        platformId,
+        projectId,
+    })
+
+    const actionProperties = actionMetadata.props
+    const parsedParameters: Record<string, unknown> = {
+        'auth': pieceConnectionExternalId,
+    }
+
+    for (const [propertyName, propertyDefinition] of Object.entries(actionProperties)) {
```
AbdulTheActivePiecer: instead of this for, filter the properties by their type, then use .reduce to return parsedParameters

packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
```diff
@@ -71,32 +79,40 @@ async function addPieceToServer(
...
         const actionMetadata = pieceMetadata.actions[action]
         const actionName = mcpToolNaming.fixTool(actionMetadata.displayName, mcpTool.id, McpToolType.PIECE)
 
-        const actionSchema = Object.fromEntries(
-            Object.entries(actionMetadata.props)
-                .filter(([_key, prop]) => prop.type !== PropertyType.MARKDOWN)
-                .map(([key, prop]) => [key, piecePropertyToZod(prop)]),
-        )
+        const toolSchema = {
+            instructions: z.string().describe('The instructions to follow when executing the tool'),
```
abuaboud: I think the description is useless there, I think It should be something very specific like what to include and how detailed is it.

packages/server/worker/src/lib/runner/process/engine-process-manager.ts
```diff
@@ -237,7 +237,7 @@ function getFlowVersionId(operation: EngineOperation, type: EngineOperationType)
         case EngineOperationType.EXECUTE_FLOW:
             return (operation as ExecuteFlowOperation).flowVersion.id
         case EngineOperationType.EXECUTE_PROPERTY:
-            return (operation as ExecutePropsOptions).flowVersion.id
+            return (operation as ExecutePropsOptions).flowVersion?.id ?? DEFAULT_MCP_DATA.flowVersionId
```
abuaboud: I see the function already accept undefined as answer? why not just returns undefined instead of random  string that doesn't represent anything

packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
```diff
@@ -253,6 +269,135 @@ async function addFlowToServer(
...
+    assertNotNullOrUndefined(pieceConnectionExternalId, 'Tool has no connection with the piece, please try to add a connection to the tool')
+
+    const aiModel = await initializeOpenAIModel({
+        platformId,
+        projectId,
+    })
+
+    const actionProperties = actionMetadata.props
+    const actionPropertiesEntries = Object.entries(actionProperties)
+
```
abuaboud: I don't think that's correct, can you explain how this is the right order to process action properties?

packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
```diff
@@ -253,6 +269,135 @@ async function addFlowToServer(
...
+CONTEXT:
+- Previously filled properties: ${JSON.stringify(acc, null, 2)}
+
+INSTRUCTIONS:
+1. Extract the property value from user instructions
+2. For required properties without clear instructions, use defaults or make reasonable assumptions
+3. Return raw values for simple types, valid JSON for complex types
+
+RESPONSE FORMAT:
+- Return only the value (raw for simple types, JSON for complex types)
```
abuaboud: I don't think you should specific the format and then try to parse it, most of AI models allow you to specify JSON (Zod Schema) and it will answer it based on your schema.

packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
```diff
@@ -253,6 +269,135 @@ async function addFlowToServer(
...
+- No explanations or additional text
+
+Respond with ONLY the value, no explanations or additional text.`
+
+        const aiResponse = await generateText({
```
abuaboud: I don't think this is the right function to use in vercel ai, check the other one for output data.

Note: We already have extract structured data action also in ap.

packages/server/api/src/app/mcp/mcp-server/mcp-server.ts
hazemadelkhalel: If our actions follow good UX, they would be sorted by their dependencies as you select them — but that’s not always the case.

So instead, we can build a graph from the refreshers and perform a topological sort.

abuaboud: I can't see anywhere we mention this, and probably many people will not follow as the pieces are created by people.

It's bad to design based on assumption

hazemadelkhalel: Yeah, I have built a graph and solved it using Topological Sort.

packages/server/api/src/app/mcp/mcp-utils.ts
```diff
@@ -1,11 +1,59 @@
...
+
+function buildFinalExtractionPrompt({
+    parameterExtractionPrompt,
+    propertySchemaValues,
+}: BuildFinalExtractionPromptParams): string {
+    return `
+    ${parameterExtractionPrompt}
+
+    YOU MUST USE THESE VALUES TO FILL OUT THE PROPERTIES ACCORDING TO USER INSTRUCTIONS:
+    ${JSON.stringify(propertySchemaValues, null, 2)}
```
abuaboud, line 50: Why we are providing this? It's already in the tool?

hazemadelkhalel: These are the values returned by the engine after executing the property. We should provide them so the AI can select from these options.

PR #8238, opened 2025-07-03
fix(elevenlabs): update library and support API regions

### LINKED ISSUE #8218: Support for ElevenLabs European Data Residency API
Hello ActivePieces team!

ElevenLabs introduced European Data Residency on May 2nd, providing a separate API endpoint for EU customers. Currently, the ElevenLabs Custom API Call node does not support this new setup. While it’s possible to use an arbitrary URL, it’s not possible to create a proper ElevenLabs connection with our credentials.

We need to allow dropdown in the piece to select the region.

<!-- Pylon-Issue-ID:b170592c-eaee-4c3b-935e-1d1fd7e730f8 --!>

marapper:
## What does this PR do?

ElevenLabs introduced [European Data Residency](https://elevenlabs.io/blog/introducing-european-data-residency) on 2nd May with a [separate API for EU customers](https://elevenlabs.io/docs/conversational-ai/legal/gdpr#faq). Although it's possible to change the domain in a Custom API call, creating a connection for EU users is blocked due to validation.

[elevenlabs](https://www.npmjs.com/package/elevenlabs) package is deprecated in favour of [Elevenlabs JS SDK](https://www.npmjs.com/package/@elevenlabs/elevenlabs-js), which is actively maintained and supports new API and configuration.

I keep the logic untouched, only adding a model to Text to Speech piece (models vary in quality and speed of generation, which is useful in flows).

[Demonstration on a video](https://youtu.be/ZEMJoxluQn8)

1. Elevenlabs connection: converted to CustomAuth with two fields
- region: dropdown with code that relates to a specific ElevenLabsEnvironment constant from the library
image

- apiKey: secret text field, remains the same

2. Auth validation uses the new library by passing the region as a parameter

3. Custom API call: API base URL depends on the region of the selected Elevenlabs connection. *Note*: when the connection is changed, it overrides URL field (the behaviour exists in the actual version of the piece; it could be addressed as a bug of CustomAPICallAction).

4. Text to Speech piece: refactored to use method from SDK library, introduces new `model` parameter that is supported by Elevenlabs API.

image

5. Extracted common methods to manage shared logic and provide backward compatibility.

### Relevant User Scenarios

No changes in the business logic of Custom API Call and Text to Speech.

**Backward compatibility**. We change PieceAuth.Secret to PieceAuth.CustomAuth with two options: region and API key. Fallback in Custom API Call and Text to Speech pieces is introduced to support already created connections.

Because a connection can't be edited but only reconnected, there is no breaking impact on existing users.

--------

Fixes #8218


marapper:
Removed claim from description, excuse me for this slip. What do you think should be done to move forward with this PR?

kishanprmr:
@allcontributors please add @marapper for plugin

Issue #8259, opened 2025-07-04
Step input isn't showing the most recent piece patch version inputs.


AbdulTheActivePiecer:
Currently if you are editing the draft version of your flow, and you have a step of version 0.1.X and there was an optional property added to the action/trigger, it doesn't show it.

PR #8268, opened 2025-07-06
fix: make draft flow version show correct piece action/trigger props for latest patched version

AbdulTheActivePiecer:
On cloud atm, if a piece is on version ~0.2.0 and you are using it in a draft flow version, it will not show the most recent patch version props in the step settings form, instead it will show 0.2.0 .


AbdulTheActivePiecer:
closes [#8259](https://github.com/activepieces/activepieces/issues/8259)

PR #8287, opened 2025-07-07
feat(messagebird): create list message action in bird

prasanna2000-max:


packages/pieces/community/messagebird/src/lib/actions/list-messages.ts
```diff
@@ -0,0 +1,107 @@
...
+          'Content-Type': 'application/json',
+        },
+      });
+
+      if (response.body.data.results) {
+        allMessages = [...allMessages, ...response.body.data.results];
+      }
+      
+      nextPageToken = response.body.data.nextPageToken;
+    } while (nextPageToken);
```
geekyme-fsmk, line 74: When this runs, the while loop has the potential to fetch all results from bird if the user did not specify a reasonable start time.

The query param limit of `1000` is also local per iteration of the loop. I believe the proper behavior is to stop fetching too many messages from bird, hence tracking a global 1k limit makes more sense.

What's the thinking for doing this? And have you considered the performance implication on the server?

PR #8301, opened 2025-07-08
feat: handle project locking

Louai-Zokerburg:
### Relevant User Scenarios

Fixes # (issue)


packages/server/api/src/app/ee/projects/project-plan/project-plan.service.ts
```diff
@@ -50,23 +51,20 @@ export const projectLimitsService = (log: FastifyBaseLogger) => ({
         if (edition === ApEdition.COMMUNITY) {
             return false
         }
+
+        await checkProjectNotLockedOrThrow(projectId)
+
         try {
-            // TODO (@abuaboud): optmize this by not querying the database
-            const projectPlan = await getOrCreateDefaultPlan(projectId)
```
abuaboud, line 55: Sounds like locked should be in project plan.

We have to be careful here, as minor changes here make the latency of each request high and databas usage.

packages/server/api/src/app/database/migration/postgres/1751878623268-AddLockedColumnToProject.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+
+export class AddLockedColumnToProject1751878623268 implements MigrationInterface {
+    name = 'AddLockedColumnToProject1751878623268'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            ALTER TABLE "project"
```
abuaboud: We need sqlite migration, but if we moved to project_plan we might not to, so check this.

packages/server/api/src/app/ee/platform/platform-plan/platform-plan-helper.ts
```diff
@@ -61,4 +69,178 @@ export async function checkQuotaOrThrow(params: QuotaCheckParams): Promise<void>
             },
         })
     }
+}
+
+export async function checkProjectNotLockedOrThrow(projectId?: string): Promise<void> {
```
abuaboud: Do not export function like this, you should wrap it around const?

for exmaple

export const platformUsageHelper = {

}

abuaboud: It's much easier for people to predict the service then auto complete would choose the name than predict the name directly, and pollute the general name space.

packages/server/api/src/app/ee/platform/platform-plan/platform-plan-helper.ts
```diff
@@ -61,4 +69,178 @@ export async function checkQuotaOrThrow(params: QuotaCheckParams): Promise<void>
...
+    }
+
+    for (const projectId of projectIds) {
+        await projectService.update(projectId, { locked: false })
+    }
+
+    return []
+}
+
+async function handleActiveFlows(
```
abuaboud, line 93: The logic should be like that.

const enabedFlows = ...

const numberToDisable = Math.max(0, enabled.length - newLimit)

const flowsDisable = enabledFlows.splice(0,numberToDisable).map(
  disable function
)

then await Promise.all(disableFlow)

Same for users, same for projects.

PR #8344, opened 2025-07-12
fix: resolve OS-specific path bug in publish-piece

### LINKED ISSUE #8345: [BUG]: publish-piece fails on Windows due to platform-specific path check
**Describe the bug**
In `/tools/scripts/pieces/publish-piece.ts`, the following code threw up an error on Windows after running `npm run publish-piece <piece_name>`:

```typescript
  const distPaths = await findAllPiecesDirectoryInSource()
  const directory = distPaths.find(path => {
    if (path.endsWith(`/${name}`)) {
      return true;
    }
    return false
  })
```
The error thrown was the one triggered by this line:
```typescript
console.error(chalk.red(`[publishPiece] can't find the directory with name ${name}`))
```

**To Reproduce**
1. Run `npm run publish-piece <piece_name>` on a Windows machine on a local build. Issue persists for any piece.
2. Error is triggered

**Expected behavior**
The success log defined by this line must be printed: 
```typescript
 console.info(chalk.green.bold(`[publishPiece] success, name=${name}, version=${version}`))
```
Instead, the aforementioned error is thrown.

prethith:
## What does this PR do?

In `/tools/scripts/pieces/publish-piece.ts`, the following code threw up an error on Windows after running `npm run publish-piece <piece_name>`:

```typescript
  const distPaths = await findAllPiecesDirectoryInSource()
  const directory = distPaths.find(path => {
    if (path.endsWith(`/${name}`)) {
      return true;
    }
    return false
  })
```
The error thrown was the one triggered by this line:
```typescript
console.error(chalk.red(`[publishPiece] can't find the directory with name ${name}`))
```
The problem was that the `/${name}` part caused an error to be thrown on Windows, which uses `\` for its directory paths. 

The fix is to use Node's built-in `path`  module, specifically the `path.basename` method to accomplish this check, to make it platform-independent. 

**The updated code block:**

```typescript
  const directory = distPaths.find(p => path.basename(p) === name)
```

Fixes #8345

### DIFF
--- a/tools/scripts/pieces/publish-piece.ts
+++ b/tools/scripts/pieces/publish-piece.ts
@@ -5,17 +5,13 @@ import { readPackageJson, readProjectJson } from '../utils/files'
 import { findAllPiecesDirectoryInSource } from '../utils/piece-script-utils'
 import { isNil } from '@activepieces/shared'
 import chalk from 'chalk'
+import path from 'node:path'
 
 export const publishPiece = async (name: string): Promise<void> => {
   assert(name, '[publishPiece] parameter "name" is required')
 
   const distPaths = await findAllPiecesDirectoryInSource()
-  const directory = distPaths.find(path => {
-    if (path.endsWith(`/${name}`)) {
-      return true;
-    }
-    return false
-  })
+  const directory = distPaths.find(p => path.basename(p) === name)
   if (isNil(directory)) {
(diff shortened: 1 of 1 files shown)


prethith:
Hi! It looks like this PR needs a label to pass the checks. Could someone with write access add `bug` to it? Thanks!

PR #8496, opened 2025-07-23
feat(piece): add stop-flow piece

### LINKED ISSUE #8495: Feature Request: Add 'Stop Flow' Piece

Currently, in Activepieces, there is no built-in, direct way to intentionally stop a flow's execution from within the flow itself.

This becomes a problem in complex workflows, particularly those using the `For Each` loop. If a user is iterating through thousands of items to find a specific one, the flow will continue to run through the entire list even after the target item has been found. This leads to significant consumption of unnecessary tasks and can make automations slower and less efficient.

I'm always frustrated when I can't exit a loop early after achieving my goal, wasting valuable tasks.

I would like to propose a new community piece named `Stop Flow`.

This piece would provide a single, simple action: `Stop Flow Execution`. When a user adds this action to their flow, it will immediately terminate the run at that point. No subsequent steps will be executed.

This gives users a powerful and accessible tool for flow control, allowing them to build more efficient and intelligent automations. The action would require no inputs or configuration, making it very easy to use.

Describe alternatives you've co

The main alternative is to use complex `Branch` (If/Else) logic. A user can create a branch where one path does nothing, effectively letting the loop continue but skipping actions for non-matching items. However, this does not stop the loop itself; the flow still iterates through all remaining items, which is the core of the problem.

Another alternative is using the `Code` piece to call `context.run.stop()` manually, but this has a higher learning curve and is less accessible for non-developer users. A dedicated piece makes this crucial functionality available to everyone.

**Additional context**

A community piece implementing this feature has already been developed and is ready for submission via a Pull Request. It has no external dependencies and uses the core `context.run.stop()` framework function. This will be a valuable addition to the core flow control tools available in Activepieces.

gadielyossef:
## What does this PR do?

This PR introduces a new community piece named `Stop Flow`.

This piece is designed to provide developers and users with precise control over their flow's execution. It adds a single, straightforward action that allows a flow to be terminated immediately and intentionally when a specific step is reached. This is particularly useful for optimizing flows that involve loops or complex conditional branches, preventing unnecessary task consumption and improving overall efficiency.

The `Stop Flow` piece is straightforward and contains a single action: `Stop Flow Execution`.

When this action is triggered within a flow, it internally calls the `context.run.stop()` function from the Activepieces framework. This call instructs the Activepieces engine to halt the current run immediately. No subsequent steps in the flow will be processed, and the run will conclude with a "Succeeded" status, as the termination was intentional.

The action requires no configuration (inputs) and produces no data (outputs), as its sole purpose is to terminate the execution path.

### Relevant User Scenarios
This feature is valuable in several common automation scenarios:

**Exiting Loops Early:** A user is iterating through a large list of records (e.g., 10,000 users from a Google Sheet) to find the first record that matches a specific criterion (e.g., a specific email address). Once found, the `Stop Flow` piece can be used to immediately exit the loop, saving thousands of unnecessary task operations.

**Conditional Flow Termination based on Input:** A flow is triggered by a webhook. A `Branch` step validates the incoming payload. If the payload is invalid or doesn't meet required conditions (e.g., a missing key), the `Stop Flow` action can terminate the run immediately, preventing errors in subsequent steps that depend on valid data.

**Search-and-Find Operations:** When searching for a file in a series of cloud storage folders or looking for a specific piece of data across multiple API calls, the flow can be stopped as soon as the target is found, making the process much more efficient.

Fixes #8495

### DIFF
--- a/packages/pieces/community/stopflow/README.md
+++ b/packages/pieces/community/stopflow/README.md
@@ -1,0 +1,25 @@
+# pieces-stopflow
+An Activepieces piece to immediately stop the current flow's execution. This library was generated with Nx.
+
+# Action: Stop Flow Execution
+Description: When this action is reached in a flow, it terminates all subsequent execution.
+
+**Inputs:** None.
+
+**Outputs:** None.
+
+# Use Case
+This piece is primarily used to exit a loop, such as a "For Each" step, once a specific condition has been met. This prevents further items in a list from being processed, saving time and tasks.
+
+**Example Flow:**
+
+- A `Trigger` provides a list of items.
+
+- A `For Each` step iterates over the list.
+
+- A `Branch` step checks if the current item meets a condition.
+
+- The `Stop Flow` action is placed in the "True" path of the Branch to halt the entire execution.
+
+# Building
+Run `nx build pieces-stopflow` to build the library.a
--- a/packages/pieces/community/stopflow/package.json
+++ b/packages/pieces/community/stopflow/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "@activepieces/piece-stopflow",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "main": "./src/index.js",
+  "types": "./src/index.d.ts",
+  "dependencies": {
+    "tslib": "^2.3.0"
+  }
+}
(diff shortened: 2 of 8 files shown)


kishanprmr:
Hello @gadielyossef ,

Thanks for the PR! Instead of creating a new piece, you can add this action to the existing [Flow Helper](https://www.activepieces.com/pieces/flow-helper) piece.

gadielyossef:
Hello, thank you for the feedback! I will do that. Thanks!

PR #8545, opened 2025-07-27
fix(subflows): handle failed subflow responses to prevent main flow hanging

### LINKED ISSUE #6231: [BUG]: if a subflow fails and you are waiting for its response in the main flow, your main flow gets stuck forever
**To Reproduce**
1. import this subflow https://cloud.activepieces.com/templates/7kPLsVjTUQ0fsm3fRoML7
2. Import this flow and put the correct subflow https://cloud.activepieces.com/templates/EaL4WfXOeguQoo2RfmrYf

**Expected behavior**
Error should be thrown up to the main caller

hazemadelkhalel:
## What does this PR do?

Fixes #6231


packages/server/api/src/app/webhooks/webhook-controller.ts
```diff
@@ -52,6 +61,8 @@ export const webhookController: FastifyPluginAsyncTypebox = async (app) => {
                 ),
                 flowVersionToRun: WebhookFlowVersionToRun.LOCKED_FALL_BACK_TO_LATEST,
                 execute: true,
+                parentRunId,
```
abuaboud: I would rename it to `...extractOptionsFromRequest(request)` and just make it return the both flag and use it in all functions.

abuaboud (approved):

PR #8575, opened 2025-07-29
fix: use external id in call flow action

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/subflows/src/lib/common.ts
+++ b/packages/pieces/community/subflows/src/lib/common.ts
@@ -14,7 +14,7 @@ export type CallableFlowResponse = {
 
 export const MOCK_CALLBACK_IN_TEST_FLOW_URL = 'MOCK';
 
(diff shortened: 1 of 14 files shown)


packages/pieces/community/subflows/src/lib/common.ts
```diff
@@ -11,4 +12,16 @@ export type CallableFlowResponse = {
     data: unknown;
 }
 
-export const MOCK_CALLBACK_IN_TEST_FLOW_URL = 'MOCK';
\ No newline at end of file
+export const MOCK_CALLBACK_IN_TEST_FLOW_URL = 'MOCK';
+
+export async function listEnabledFlowsWithSubflowTrigger(context: { flows: { list: () => Promise<SeekPage<PopulatedFlow>> } }) {
```
abuaboud: Usually our list endpoints accept `externalids` as filter, we can do backend request based on that instead of listing all flows.

If that is not possible, can we cache the value of id? as listing all flows is very expensive operation

PR #8584, opened 2025-07-30
feat: enhance subflows testing experience

### LINKED ISSUE #7202: Subflow has to be revamped
Things need to be revamped:
- If subflows fails the main flow doesn't get marked as failed, (sometime success also fail to send the callback)
- Testing Experience 
- Better Tracing in Runs (Show Links)
- When trigger sample data is changed, the dropdown stop showing the flow as selected as the value contains the example data.

Image

hazemadelkhalel:
## What does this PR do?

Fixes #7202

### DIFF
--- a/packages/engine/src/lib/handler/context/engine-constants.ts
+++ b/packages/engine/src/lib/handler/context/engine-constants.ts
@@ -55,6 +55,7 @@ export class EngineConstants {
         public readonly httpRequestId: string | null,
         public readonly resumePayload?: ResumePayload,
         public readonly runEnvironment?: RunEnvironment,
+        public readonly returnResponseAction?: string,
     ) {
         if (!publicApiUrl.endsWith('/api/')) {
             throw new Error('Public URL must end with a slash, got: ' + publicApiUrl)
@@ -86,6 +87,7 @@ export class EngineConstants {
             input.httpRequestId ?? null,
             input.executionType === ExecutionType.RESUME ? input.resumePayload : undefined,
             input.runEnvironment,
+            input.returnResponseAction ?? undefined,
         )
     }
 
@@ -109,6 +111,9 @@ export class EngineConstants {
             ProgressUpdateType.NONE,
             null,
             null,
+            undefined,
+            undefined,
+            undefined,
         )
     }
     public static fromExecuteStepInput(input: ExecuteStepOperation): EngineConstants {
@@ -133,6 +138,7 @@ export class EngineConstants {
             input.requestId ?? null,
             undefined,
             input.runEnvironment,
+            input.returnResponseAction ?? undefined,
         )
     }
 
@@ -156,6 +162,9 @@ export class EngineConstants {
             ProgressUpdateType.NONE,
             null,
             null,
+            undefined,
+            undefined,
+            undefined,
         )
     }
 
@@ -179,6 +188,9 @@ export class EngineConstants {
             ProgressUpdateType.NONE,
             null,
             null,
+            undefined,
+            undefined,
+            undefined,
         )
     }
 
--- a/packages/engine/src/lib/handler/piece-executor.ts
+++ b/packages/engine/src/lib/handler/piece-executor.ts
@@ -1,6 +1,6 @@
 import { URL } from 'url'
 import { ActionContext, InputPropertyMap, PauseHook, PauseHookParams, PiecePropertyMap, RespondHook, RespondHookParams, StaticPropsValue, StopHook, StopHookParams, TagsManager } from '@activepieces/pieces-framework'
-import { ActionType, assertNotNullOrUndefined, AUTHENTICATION_PROPERTY_NAME, ExecutionType, FlowRunStatus, GenericStepOutput, isNil, PauseType, PieceAction, RespondResponse, StepOutputStatus } from '@activepieces/shared'
+import { ActionType, assertNotNullOrUndefined, AUTHENTICATION_PROPERTY_NAME, ExecutionType, FlowRunStatus, GenericStepOutput, isNil, PauseType, PieceAction, pieceActionNaming, RespondResponse, StepOutputStatus } from '@activepieces/shared'
 import dayjs from 'dayjs'
 import { continueIfFailureHandler, handleExecutionError, runWithExponentialBackoff } from '../helper/error-handling'
 import { PausedFlowTimeoutError } from '../helper/execution-errors'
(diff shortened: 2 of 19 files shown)


packages/engine/src/lib/handler/context/engine-constants.ts
```diff
@@ -109,6 +111,9 @@ export class EngineConstants {
             ProgressUpdateType.NONE,
             null,
             null,
+            undefined,
+            undefined,
+            undefined,
```
AbdulTheActivePiecer: is it possible to make this a named constructor instead ?

packages/engine/src/lib/handler/piece-executor.ts
```diff
@@ -145,7 +145,8 @@ const executeAction: ActionHandler<PieceAction> = async ({ action, executionStat
                 return url.toString()
             },
         }
-        const runMethodToExecute = (constants.testSingleStepMode && !isNil(pieceAction.test)) ? pieceAction.test : pieceAction.run
+        const constructedActionName = pieceActionNaming.constructActionName(action.settings.pieceName, action.settings.actionName)
+        const runMethodToExecute = (constants.testSingleStepMode && !isNil(pieceAction.test) && constants.returnResponseAction !== constructedActionName) ? pieceAction.test : pieceAction.run
```
AbdulTheActivePiecer: what is constructedActionName? like what is its job ? can you please rename it to explain that? same thing with constructActionName, like why do we need to have `${pieceName}_${actionName}`, maybe rename to constructUniqueActionName

packages/server/api/src/app/flows/test-flow-version/test-flow-version.controller.ts
```diff
@@ -0,0 +1,95 @@
...
+            id: Type.String(),
+        }),
+        body: Type.Object({
+            parentFlowId: Type.String(),
+            parentFlowVersionId: Type.String(),
+            parentFlowDisplayName: Type.String(),
+            returnResponseAction: Type.String(),
+            payload: Type.Unknown(),
+        }),
+    },
```
AbdulTheActivePiecer: The schema should be defined in shared so we can use the type in test method of the subflow piece when sending the request.

packages/server/api/src/app/flows/step-run/step-run-progress.handler.ts
```diff
@@ -0,0 +1,81 @@
...
+function createProgressHandler(context: StepExecutionContext) {
+    return async (progressEvent: FlowRunProgressEvent): Promise<void> => {
+        const { socket, projectId, logger, stepName, requestId } = context
+        
+        if (progressEvent.runId !== context.runId) {
+            return
+        }
+
+        try {
+            const flowRun = await flowRunService(logger).getOneOrThrow({
```
abuaboud: Use `getOnePopulatedOrThrow` then you don't need to load the file.

packages/server/api/src/app/flows/step-run/step-run-progress.handler.ts
```diff
@@ -0,0 +1,81 @@
...
+            })
+            
+            const executionData: ExecutioOutputFile = JSON.parse(fileData.toString('utf8'))
+            const stepOutput = executionData.executionState.steps[stepName]
+            
+            if (isNil(stepOutput) || stepOutput.status === StepOutputStatus.RUNNING) {
+                return
+            }
+            
+            if (stepOutput.status !== StepOutputStatus.SUCCEEDED) {
```
abuaboud: I think we should still send response so the frontend display the error

packages/server/api/src/app/flows/step-run/step-run-progress.handler.ts
```diff
@@ -0,0 +1,81 @@
...
+                success: true,
+                input: executionData.executionState.steps,
+                output: stepOutput.output ?? {},
+                standardError: '',
+                standardOutput: '',
+            }
+            
+            socket.emit(WebsocketClientEvent.TEST_STEP_FINISHED, response)
+        }
+        catch (error) {
```
abuaboud: use exceptionHandler.handle() so it's send in sentry

packages/server/api/src/app/flows/step-run/step-run-progress.handler.ts
```diff
@@ -0,0 +1,81 @@
...
+                return
+            }
+
+            const { data: fileData } = await fileService(logger).getDataOrThrow({
+                fileId: flowRun.logsFileId,
+                projectId: flowRun.projectId,
+            })
+            
+            const executionData: ExecutioOutputFile = JSON.parse(fileData.toString('utf8'))
+            const stepOutput = executionData.executionState.steps[stepName]
```
abuaboud: That's wrong, if the trigger is inside the loop, then it would be nested there. I think we have a function that find the step result inside executionDate in the frontend

abuaboud (changes requested):

packages/server/api/src/app/flows/step-run/step-run-progress.handler.ts
```diff
@@ -0,0 +1,103 @@
...
+        id: testCallbackRequestId,
+        success: isSuccess,
+        input: stepOutput.input ?? {},
+        output: stepOutput.output ?? {},
+        standardError: isSuccess ? '' : (stepOutput.errorMessage as string),
+        standardOutput: '',
+    }
+}
+
+function findStepOutputInNestedStructure(data: unknown, stepName: string): StepOutput | undefined {
```
abuaboud: This is hardcoded, I would believe it should live closer where it belongs otherwise it will break within a month 

https://github.com/activepieces/activepieces/blob/e83261a250afaea0c3d280a5f8acb25e5974fdf3/packages/shared/src/lib/flow-run/execution/step-output.ts#L4

Also I think it should understand the step type and based on that type, It continue searching.

My suggestion to make file in the shared closer to where it live, write a proper function and reuse it in the frontend and here.

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -80,6 +81,10 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
     app.post('/notify-frontend', NotifyFrontendParams, async (request) => {
         const { type, data } = request.body
         app.io.to(request.principal.projectId).emit(type, data)
+        await stepRunProgressHandler.notifyStepFinished({
```
abuaboud: I don't think it's correct place to place notifyStepFinished, It sounds like a side effect for flow run, I think we have place in `flow-run` when run is finished (If I am not wrong, we use it for issues).

In that place we should check If we have to send reply (Maybe it's real run which is more common), If we don't we should do early return, because this put huge burden on app server, since it have to get the run from the database and actual download the whole log from S3.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -62,7 +73,7 @@ export const loopExecutor: BaseExecutor<LoopOnItemsAction> = {
                 return newExecutionContext.upsertStep(action.name, stepOutput.setDuration(performance.now() - stepStartTime))
             }
 
-            if (constants.testSingleStepMode) {
+            if (!isNil(constants.stepNameToTest)) {
```
abuaboud, line 76: It wont ever each here right? Because of the if in the beginning of action?

PR #8593, opened 2025-07-31
fix: install piece from registry timing out

### LINKED ISSUE #8587: [BUG]: Installing piece times out
**Describe the bug**
Installing piece times out

**To Reproduce**
Install a piece via the platform admin pieces page via npm or tar

hazemadelkhalel:
## What does this PR do?

Fixes #8587

### DIFF
--- a/packages/server/worker/src/lib/runner/index.ts
+++ b/packages/server/worker/src/lib/runner/index.ts
@@ -1,5 +1,5 @@
 import { webhookSecretsUtils } from '@activepieces/server-shared'
-import { ActionType, EngineOperation, EngineOperationType, ExecuteFlowOperation, ExecutePropsOptions, ExecuteStepOperation, ExecuteToolOperation, ExecuteTriggerOperation, ExecuteValidateAuthOperation, flowStructureUtil, FlowVersion, isNil, PieceActionSettings, PiecePackage, PieceTriggerSettings, TriggerHookType, TriggerType } from '@activepieces/shared'
+import { ActionType, EngineOperation, EngineOperationType, ExecuteFlowOperation, ExecutePropsOptions, ExecuteStepOperation, ExecuteToolOperation, ExecuteTriggerOperation, ExecuteValidateAuthOperation, flowStructureUtil, FlowVersion, isNil, PackageType, PieceActionSettings, PieceTriggerSettings, TriggerHookType, TriggerType } from '@activepieces/shared'
 import { FastifyBaseLogger } from 'fastify'
 import { pieceWorkerCache } from '../api/piece-worker-cache'
 import { executionFiles } from '../cache/execution-files'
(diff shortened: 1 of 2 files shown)


packages/server/worker/src/lib/runner/index.ts
```diff
@@ -67,7 +67,14 @@ export const engineRunner = (log: FastifyBaseLogger): EngineRunner => ({
     async extractPieceMetadata(engineToken, operation) {
         log.debug({ operation }, '[threadEngineRunner#extractPieceMetadata]')
 
-        const lockedPiece = await pieceEngineUtil(log).resolveExactVersion(engineToken, operation)
+        let lockedPiece: PiecePackage
+        try {
+            lockedPiece = await pieceEngineUtil(log).resolveExactVersion(engineToken, operation)
+        }
+        catch {
+            lockedPiece = await pieceEngineUtil(log).enrichPieceToInstall(engineToken, operation)
```
abuaboud: It's wrong to try something and expect it to throw exception then try to solve it, can we do iit without catch what's the issue there? how is this is related to timeout?

hazemadelkhalel: When you try to install a piece, getPiece throws because the piece metadata does not exist yet.

So it keeps stuck because we didn't reply to extractPieceInformation

packages/server/worker/src/lib/utils/flow-engine-util.ts
```diff
@@ -115,4 +141,12 @@ type LockFlowVersionParams = {
 export type BasicPieceInformation = {
     pieceName: string
     pieceVersion: string
+}
+
+export type EnrichPieceWithArchiveParams = {
```
abuaboud: Delete this

packages/server/worker/src/lib/utils/flow-engine-util.ts
```diff
@@ -59,6 +59,32 @@ export const pieceEngineUtil = (log: FastifyBaseLogger) => ({
             }
         }
     },
+    async enrichPieceToInstall(engineToken: string, pieceData: EnrichPieceWithArchiveParams): Promise<PiecePackage> {
```
abuaboud: Delete this

packages/server/worker/src/lib/runner/index.ts
abuaboud: Use `enrichPieceWithArchive` and for pieceMetadata the parameter make like that.

```
pieceMetadata: Pick<PieceMetadataModel, 'name' | 'version' | 'packageType' | 'pieceType'>
```

abuaboud (changes requested):

packages/server/worker/src/lib/runner/index.ts
```diff
@@ -67,7 +67,14 @@ export const engineRunner = (log: FastifyBaseLogger): EngineRunner => ({
     async extractPieceMetadata(engineToken, operation) {
         log.debug({ operation }, '[threadEngineRunner#extractPieceMetadata]')
 
-        const lockedPiece = await pieceEngineUtil(log).resolveExactVersion(engineToken, operation)
+        let lockedPiece: PiecePackage
+        try {
```
abuaboud: Replace this part with

```
        const lockedPiece = await pieceEngineUtil(log).enrichPieceWithArchive(engineToken, {
            name: operation.pieceName,
            version: operation.pieceVersion,
            packageType: operation.packageType,
            pieceType: operation.pieceType,
        })
        ```

abuaboud (approved):

PR #8711, opened 2025-08-08
feat: Drupal integration

### LINKED ISSUE #8710: Drupal Integration
Drupal, the CMS to be found at https://www.drupal.org, should be integrated with AP. As Drupal sites are individual installations with each of them defining their own data model and available features, a generic approach needs to be dynamic such that the Piece dynamically pulls the available actions from the Drupal site(s) in connects to.

jurgenhaas:
Fixes #8710

### DIFF
--- a/packages/shared/src/lib/flows/flow.ts
+++ b/packages/shared/src/lib/flows/flow.ts
@@ -32,3 +32,12 @@ export const PopulatedFlow = Type.Composite([
 ])
 
 export type PopulatedFlow = Static<typeof PopulatedFlow>
+
+
+export const PopulatedTriggerSource = Type.Composite([
+    TriggerSource,
+    Type.Object({
+        flow: Flow,
+    }),
+])
+export type PopulatedTriggerSource = Static<typeof PopulatedTriggerSource>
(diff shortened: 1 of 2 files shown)


kishanprmr:
@jurgenhaas is this PR ready for review?

jurgenhaas:
@kishanprmr I've just had a long session with Mo yesterday to iron out the latest issues around polling. So, yes, this is ready for review, although I will maybe add some more debugging lines to it in the next few days. But we do have a big conference for Drupal coming up in 2,5 weeks, and we need to have it published by then.

So, if you can, a review would be much appreciated.

kishanprmr:
@jurgenhaas I will review it this week and merge it, Thanks.

kishanprmr:
@jurgenhaas I am testing the triggers and have installed the module using Composer. It returns empty results even when I create a new post or comment. Could you guide me on how to properly test the three triggers?

jurgenhaas:
Hey @kishanprmr, the 3 triggers get handled by the Drupal orchestration module in a generic way. The webhook just gets stored in key-value so that other modules will be able to implement some custom logic with it. The polling triggers are also handled by the orchestration such that it signals them internally, and if one or many custom implementations are available, they will pick that polling request up and add something to the response.

So, that's just infrastructure implemented by orchestration, no actual features are provided ootb because each Drupal site is different and the response will have to be implemented by the Drupal site-builder individually.

If you want to go that far with your tests, I would have to provide you with a test scenario separately.

PR #8726, opened 2025-08-09
fix: return raw webhook payload when trigger fails due to connection errors

kamal:
## What does this PR do?

  Fixes silent webhook failures when connections are in ERROR state (expired OAuth tokens, invalid API keys, etc.). Previously, webhooks would fail silently with no flow runs, no alerts, and no visibility. Now they create flow runs that fail visibly, preserving webhook data and triggering alerts.

  When a webhook trigger encounters a connection error:
  1. Creates a trigger run with new `CONNECTION_ERROR` status
  2. Returns raw webhook payload to create a flow run (instead of failing silently)
  3. Flow run fails at the action step that needs the connection
  4. Issue is created and appears in Issues dashboard
  5. Users see exactly which step needs the connection fixed

  ### Relevant User Scenarios

  - **OAuth expiration**: Slack token expires over weekend → Monday webhooks now visible as failed runs instead of disappearing
  - **API key rotation**: Quarterly key rotation → webhook data preserved in failed runs, can rerun after updating connection
  - **Debugging**: Can see exact webhook payload received when connection failed
  - **No data loss**: All webhooks captured even with bad connections


abuaboud:
Closing in favor proper notifications / alerts for this and other types of error, moving it to the product team.

Thank you @kamal we will revisit it soon.

PR #8803, opened 2025-08-14
feat: Add Private Key authentication for SFTP

prasanna2000-max:
## What does this PR do?
This PR extends the SFTP client connection logic to support private key authentication alongside the existing password authentication.
It also introduces a host key algorithm selection feature for SFTP connections when using private keys, allowing finer control over security and compatibility with different SFTP servers.

### Note
1. **SecretText Property in PieceAuth**
Removed the hardcoded required: true from the SecretText property definition. Previously, all SecretText fields were forced to be required, but this change allows defining optional secrets (SecretText(false)), providing greater flexibility.

2. **StaticMultiSelectDropdownProperty added in CustomAuthProps**
UI chips overlap when multiple selections are made
image

**1. Authentication Options**
- Password Authentication: If password is provided, the client connects using username + password.
- Private Key Authentication: If privateKey is provided, the client connects using username + private key.

**2. Host Key Algorithm Selection (Private Key only)**
- When using a private key, the algorithm array must be provided with at least one valid host key algorithm (ssh-rsa, ssh-ed25519, etc.).
- These algorithms are passed to the underlying SFTP client’s serverHostKey option to ensure compatibility with the remote server.

**3. Validation Rules**
- At least one of password or privateKey is required.
- If privateKey is used without algorithms, the connection is blocked with an error.

Fixes # (issue)


jerbob92:
@prasanna2000-max @kishanprmr I have 3 questions about this PR:
 - Why is `algorithm` required?
 - Why is the `password` field not passed as passphrase in case of encrypted private key?
 - Why is the private key field a single line field and requires you to replace newlines with `\n` to make it work?

AbdulTheActivePiecer:
@jerbob92  we opened this ticket to address the first and second questions, for the third, it is just a cautionary action, that's all, because triggers can also be configured through the API not just the UI. 

https://github.com/activepieces/activepieces/issues/11100

PR #8826, opened 2025-08-17
feat: web search in Universal AI

amrdb:
### Relevant User Scenarios

Fixes # (issue)


packages/pieces/community/common/src/lib/ai/index.ts
```diff
@@ -183,7 +183,114 @@ export const aiProps = <T extends 'language' | 'image'>({ modelType, functionCal
...
+            const provider = propsValue['provider'] as unknown as string;
+
+            if (!webSearchEnabled) {
+                return {};
+            }
+
+            const providerMetadata = SUPPORTED_AI_PROVIDERS.find(p => p.provider === provider);
+            if (isNil(providerMetadata)) {
+                return {};
+            }
```
AbdulTheActivePiecer, line 208: If the providers dropdown only return the supported ai providers why do you check here again ?

packages/pieces/community/text-ai/src/lib/actions/ask-ai.ts
```diff
@@ -100,6 +107,6 @@ export const askAI = createAction({
       await storage.put(conversationKey, conversation);
     }
 
-    return response.text ?? '';
+    return context.propsValue.includeSources ? { text: response.text, sources: response.sources } : response.text;
```
AbdulTheActivePiecer, line 110: Always return the same schema, { result: string, sources: [] }, no need to make it return dynamic result depending on the user inputs.

packages/pieces/community/text-ai/src/lib/actions/summarize-text.ts
```diff
@@ -56,14 +56,11 @@ export const summarizeText: Action = createAction({
...
-      maxOutputTokens: providerName !== 'openai' ? context.propsValue.maxOutputTokens : undefined,
+      maxOutputTokens: context.propsValue.maxOutputTokens,
       temperature: 1,
       providerOptions: {
         [providerName]: {
-          ...(providerName === 'openai' ? {
-            ...(context.propsValue.maxOutputTokens ? { max_completion_tokens: context.propsValue.maxOutputTokens } : {}),
-            reasoning_effort: 'minimal',
-          } : {}),
+          ...(providerName === 'openai' ? { reasoning_effort: 'minimal' } : {}),
```
AbdulTheActivePiecer, line 63: Nit: It might be better to make the reasoning effort a dropdown when the provider is openAI.
image

packages/server/api/src/app/ai/providers/anthropic.ts
```diff
@@ -10,7 +10,7 @@ export const anthropicProvider: AIProviderStrategy = {
     },
 
     calculateUsage: (request: FastifyRequest<RequestGenericInterface, RawServerBase>, response: Record<string, unknown>): Usage => {
-        const apiResponse = response as { usage: { input_tokens: number, output_tokens: number } }
+        const apiResponse = response as { usage: { input_tokens: number, output_tokens: number, server_tool_use: { web_search_requests: number } } }
```
AbdulTheActivePiecer, line 13: why is the response type Record<string,unknown> and then we use as { ... } after ? why not put the type above from the start ? 
You will need to test after doing that, just to make sure all works with fastify

packages/server/api/src/app/ai/providers/openai.ts
```diff
@@ -13,8 +13,13 @@ export const openaiProvider: AIProviderStrategy = {
...
         const body = request.body as { size: string, model: string, n?: string, quality?: string }
         const apiResponse = response as {
+            tools: { type: string }[]
             usage: 
-            | { input_tokens: number, output_tokens: number, input_tokens_details: { text_tokens: number, image_tokens: number } }
+            | { 
+                input_tokens: number
+                output_tokens: number
+                input_tokens_details: { text_tokens: number, image_tokens: number }
+            }
```
AbdulTheActivePiecer, line 22: Same thing for the typing here, and also please define the type outside the method definition to make it easier to read.

packages/shared/src/lib/ai/ai-sdk.ts
```diff
@@ -13,6 +14,7 @@ export function createAIModel<T extends LanguageModelV2 | ImageModel>({
     engineToken,
     baseURL,
     metadata,
+    openaiResponsesModel = false,
```
AbdulTheActivePiecer, line 17: If this flag means to use the responses API from Open AI, rename to useOpenAiResponsesApi, and it might be worth trying to migrate fully to it https://platform.openai.com/docs/guides/migrate-to-responses

PR #8912, opened 2025-08-25
refactor: split inputUiInfo into propertySettings and sampleData inside step settings

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)


packages/engine/src/lib/variables/props-processor.ts
```diff
@@ -16,8 +16,12 @@ export const propsProcessor = {
         props: InputPropertyMap,
         auth: PieceAuthProperty | undefined,
         requireAuth: boolean,
-        dynamaicPropertiesSchema: Record<string, InputPropertyMap> | undefined | null,
+        propertySettings: Record<string, PropertySettings>,
     ): Promise<{ processedInput: StaticPropsValue<PiecePropertyMap>, errors: PropsValidationError }> => {
+        const dynamaicPropertiesSchema: Record<string, InputPropertyMap> = {}
```
abuaboud: Shouldn't this be InputPropertyMap | null | undefined? so we can avoid casting below

Issue #8943, opened 2025-08-27
Bug: Dynamic Valued Array Not Saving


AbdulTheActivePiecer:
If you have an array property in a piece and it has properties, and a defaultValue:[] and you put the (dynamic value) toggle on, and type in something, and go to another step then back to the original step you were on, the values won't be saved.

PR #8948, opened 2025-08-27
fix: toggling dynamic value in a step settings input wasn't saving the new value placed

AbdulTheActivePiecer:


AbdulTheActivePiecer:
closes #8943

PR #9123, opened 2025-09-08
fix: loading some pieces in development wasn't working (open ai)

AbdulTheActivePiecer:


packages/engine/src/lib/helper/piece-loader.ts
```diff
@@ -148,13 +147,12 @@ export const pieceLoader = {
 
 async function loadPieceFromDistFolder(packageName: string): Promise<string | null> {
     const distPath = path.resolve('dist/packages/pieces')
-    const entries = await utils.walk(distPath)
-    for await (const entry of entries) {
+    const entries = (await utils.walk(distPath)).filter((entry) => entry.name === 'package.json')
```
AbdulTheActivePiecer, line 150: It was looking in package-lock.json sometimes which included the piece name so it led to wrong directories being used

PR #9173, opened 2025-09-10
fix: preserve loop iteration selection during flow run updates

### LINKED ISSUE #9165: Loop Items collapse during runs progress
The user trying to view the current loop result while the flow is running, each time there is new update for the flow run, It collapse the step sidebar  
from

To

naaa760:
## What does this PR do?

Fixes loop iterations collapsing when flow runs update.

When you're watching a specific loop iteration (like iteration 5), it used to collapse back to iteration 1 every time new flow data came in. Now it stays on the iteration you selected.

### Relevant User Scenarios

- Watching loop progress in real-time
- Debugging specific iterations
- Keeping track of loop performance

Fixes #9165

### DIFF
--- a/packages/react-ui/src/features/flow-runs/lib/flow-run-utils.ts
+++ b/packages/react-ui/src/features/flow-runs/lib/flow-run-utils.ts
@@ -1,3 +1,6 @@
+/* eslint-disable prettier/prettier */
+/* eslint-disable import/no-unresolved */
+/* eslint-disable import/order */
 import {
   Check,
   CircleCheck,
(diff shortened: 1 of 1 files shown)


abuaboud:
Hi @naaa760 

?How does this solves the issue of collapsing the loop list after each update?

nehaaprasad:
@abuaboud 

- fix: now it remembers which iteration you selected and keeps it there when the flow updates.
- How: changed the code from always using 0 as default to keeping your current selection when it exists.

PR #9250, opened 2025-09-15
fix: move Contiguity to v2 of API

Owlcept:
## What does this PR do?

Move Contiguity piece to be in line with new updates, moved to API v2 with help from Contiguity Dev

Moved over to new api endpoints allowing mms and imessage, inline with Contiguity usage.


s0urfruit:
Contiguity approves this PR ✅

onyedikachi-david (approved):

PR #9322, opened 2025-09-21
fix(store): make run-scope store pieces testable, fixes #9317

### LINKED ISSUE #9317: [BUG]: Testing store pieces with "Run" scope isn't useful
**Describe the bug**
When you have a flow with a put, then a get action and they both have the "run" scope, you will not have any use for testing them because they will always return the same result:

Put X in key Y --returns--> X

Get value in key Y --returns--> undefined

so the user that wants to have "run" scoped steps will have to put it on "flow" finish setting up the flow then switch it back, which can be quite tedious.

uvenkatateja:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/store/src/lib/actions/store-append-action.ts
+++ b/packages/pieces/community/store/src/lib/actions/store-append-action.ts
@@ -44,6 +44,7 @@ export const storageAppendAction = createAction({
       runId: context.run.id,
       key: context.propsValue['key'],
       scope: context.propsValue.store_scope,
+      isTestMode: context.executionType === 'BEGIN' && context.run.id.startsWith('test-'),
     });
     const oldValue = (await context.store.get(key, scope)) || '';
     if (typeof oldValue !== 'string') {
--- a/packages/pieces/community/store/src/lib/actions/common.ts
+++ b/packages/pieces/community/store/src/lib/actions/common.ts
@@ -13,7 +13,9 @@ export function getScopeAndKey(params: Params): { scope: StoreScope, key: string
         case PieceStoreScope.FLOW:
             return { scope: StoreScope.FLOW, key: params.key }
         case PieceStoreScope.RUN:
-            return { scope: StoreScope.FLOW, key: `run_${params.runId}/${params.key}` }
+            // Use a consistent test run ID when testing to allow store operations to work together
+            const runId = params.isTestMode ? 'test-run' : params.runId
+            return { scope: StoreScope.FLOW, key: `run_${runId}/${params.key}` }
     }
 }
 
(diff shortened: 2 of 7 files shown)


packages/pieces/community/store/src/lib/actions/store-append-action.ts
```diff
@@ -44,6 +44,7 @@ export const storageAppendAction = createAction({
       runId: context.run.id,
       key: context.propsValue['key'],
       scope: context.propsValue.store_scope,
+      isTestMode: context.executionType === 'BEGIN' && context.run.id.startsWith('test-'),
```
AbdulTheActivePiecer: Hi thanks a lot for the pr 🙏  

I just have a minor note, instead of having the isTestMode like this, we already have a test() method in the actions, so you can use that instead of having to write the logic manually here, that way the logic is determined by the engine and the piece wouldn't have to worry about it.

basically you would need to move the logic of each run method to functions outside and call them both in the run() and test() methods with a flag (isTesting) and that would solve it. 

afterwards just bump the patch version in package.json of the piece

AbdulTheActivePiecer (dismissed):

packages/pieces/community/store/src/lib/actions/store-append-action.ts
uvenkatateja: Hi @AbdulTheActivePiecer! 👋

Thanks for the feedback! I've implemented the changes you requested:

✅ **Removed manual test mode detection** - No more `context.executionType === 'BEGIN' && context.run.id.startsWith('test-')`
✅ **Extracted shared functions** - Each action now has `executeStorage[Action]()` functions 
✅ **Added test() methods** - All actions now have both [run()](cci:1://file:///c:/Users/uvtej/Desktop/El/activepieces/packages/pieces/community/store/src/lib/actions/store-add-to-list.ts:78:1-80:2) and [test()](cci:1://file:///c:/Users/uvtej/Desktop/El/activepieces/packages/pieces/community/store/src/lib/actions/store-add-to-list.ts:81:1-83:2) methods that call shared functions with `isTestMode` flag
✅ **Bumped version** - Updated to 0.6.9 as requested

The approach now uses the engine-determined [test()](cci:1://file:///c:/Users/uvtej/Desktop/El/activepieces/packages/pieces/community/store/src/lib/actions/store-add-to-list.ts:81:1-83:2) method instead of manual detection, making it cleaner and more maintainable while preserving the core fix for #9317.

AbdulTheActivePiecer:
@coderabbit review

AbdulTheActivePiecer:
@coderabbitai review

AbdulTheActivePiecer:
@allcontributors please add @uvenkatateja  for plugin.

AbdulTheActivePiecer (approved):

PR #9449, opened 2025-09-27
feat: Checking the file cache in the worker prior to requesting `archive` from API

ArvindEnvoy:
At present, the archive for a given piece is fetched every single time that a flow is executed, specifically when the sandbox is prepared. This is not necessary in cases where the archive is cached in the filesystem that the worker has access to. As such, I'm adding a check in `enrichPieceWithArchive` where the worker first checks to see if the archive is present in the filesystem and reads it from there if it exists. If it's not present, then the archive is requested from the API, as is the case today. We have implemented this change in our self-hosted instance of AP, and are seeing latency improvements and a massive reduction in the number of calls the worker has to make to the API server.


ArvindR12:
@abuaboud let me know what you think of this PR, the CI failure looks unrelated to the changes

PR #9577, opened 2025-10-07
refactor: optimize queries for performance 

MrChaker:


packages/server/api/src/app/trigger/trigger-source/trigger-source-service.ts
```diff
@@ -122,7 +136,7 @@ type ExistsByFlowIdParams = {
 
 type GetByFlowIdParams = {
     flowId: string
-    projectId: string
+    projectId?: string
```
abuaboud, line 138: Why is this optional? can we make it not optional

packages/server/api/src/app/trigger/trigger-source/trigger-source-service.ts
```diff
@@ -52,14 +52,28 @@ export const triggerSourceService = (log: FastifyBaseLogger) => {
             })
         },
         async getByFlowId(params: GetByFlowIdParams): Promise<TriggerSource | null> {
```
abuaboud: Is this needed? if not let remove it

packages/server/api/src/app/webhooks/webhook-controller.ts
```diff
@@ -40,11 +40,6 @@ export const webhookController: FastifyPluginAsyncTypebox = async (app) => {
                         flowId: request.params.flowId,
                         async: false,
                         flowVersionToRun: WebhookFlowVersionToRun.LOCKED_FALL_BACK_TO_LATEST,
-                        saveSampleData: await triggerSourceService(request.log).existsByFlowId({
```
abuaboud, line 43: revert

packages/server/api/src/app/webhooks/webhook.service.ts
```diff
@@ -39,19 +38,27 @@ export const webhookService = {
...
+                    span.setAttribute('webhook.saveSampleData', saveSampleData ?? false)
                     return {
                         status: StatusCodes.GONE,
                         body: {},
                         headers: {},
                     }
                 }
 
+                const { flow, ...triggerSource } = triggerSourceResult
+                saveSampleData = saveSampleData ?? true
```
abuaboud: remove since we will keep saveSampleData not optional

packages/server/api/src/app/webhooks/webhook.service.ts
```diff
@@ -162,7 +163,7 @@ export const webhookService = {
 type HandleWebhookParams = {
     flowId: string
     async: boolean
-    saveSampleData: boolean
+    saveSampleData?: boolean
```
abuaboud: Revert

abuaboud (changes requested):

packages/server/api/src/app/trigger/trigger-source/trigger-source-service.ts
MrChaker: in handleWebhook in webhook.service . we don't have projectId . we would need an extra db call to get it

packages/server/api/src/app/trigger/trigger-source/trigger-source-service.ts
MrChaker: it is used in other places

PR #9627, opened 2025-10-09
refactor: unify saving execution state behaviors

amrdb:
## What does this PR do?

Made an endpoint in the engine controller (same behavior as S3) to decouple run metadata from uploading execution state (logs).

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
+++ b/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
@@ -255,9 +255,10 @@ export const flowRunService = (log: FastifyBaseLogger) => ({
         return flowRun
     },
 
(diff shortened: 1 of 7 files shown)


packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -632,16 +635,15 @@ type GetOrCreateLogsFileParams = {
...
-        logsUploadUrl = logsUploadResult.uploadUrl
-        logsFileId = logsUploadResult.fileId
-    }
+    const logsUploadFile = await createLogsUploadFile({
+        flowRun: params.flowRun,
+        projectId: params.flowRun.projectId,
+    }, log)
+    
+    await flowRunRepo().update(params.flowRun.id, {
+        logsFileId: logsUploadFile.fileId,
```
abuaboud: we should not update the log file id, unless we did a first success upload (in the engine) the metadata should do the update

PR #9631, opened 2025-10-09
perf(worker): Use Bloom filter to eliminate disk locks on cache misses

0xSherlokMo:
## What does this PR do?

Cache check operations were experiencing cold start delays due to acquiring disk locks on every cacheCheckState call, even when the cache key definitely didn't exist. This resulted in unnecessary I/O and lock contention, particularly impacting performance during initial worker startup when most cache keys are not yet present.

A Bloom filter is a space-efficient probabilistic data structure that can tell you:
- Definitely NO: An item is definitely not in the set (no false negatives)
- Probably YES: An item might be in the set (with a 1% false positive rate for example, we can control the FP rate by adjusting number of hashing and bit size)
This property is perfect for optimizing cache lookups.

During cold starts or when checking for non-existent pieces, the majority of cacheCheckState calls will hit the "definitely not present" path, eliminating contention on the disk lock and avoiding expensive file I/O operations. The Bloom filter acts as a fast, in-memory guard that protects the slower disk-based cache from unnecessary reads and locks.

### DIFF
--- a/packages/server/worker/src/lib/cache/cache-state.ts
+++ b/packages/server/worker/src/lib/cache/cache-state.ts
@@ -42,7 +42,7 @@ const getBloomFilter = async (folderPath: string): Promise<ScalableBloomFilter> 
 
 export const cacheState = (folderPath: string) => {
     return {
-        async getOrSetIfNotExists(cacheAlias: string, state: string): Promise<string | undefined> {
+        async getOrSetIfNotExists(cacheAlias: string, state: string): Promise<string> {
             const bloomFilter = await getBloomFilter(folderPath)
             const exists = bloomFilter.has(cacheAlias)
             if (!exists) {


packages/server/worker/src/lib/cache/cache-state.ts
```diff
@@ -14,43 +15,87 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
...
-            await saveToCache({}, folderPath)
+            await saveToCache(JSON.stringify({}), folderPath, filePath)
         }
-        cached[folderPath] = await readCache(folderPath)
+        cached[folderPath] = await readCache(filePath) as CacheMap
     }
     const cache = (cached[folderPath] as CacheMap) || {}
     return cache
 }
+const bloomFilters: Record<string, ScalableBloomFilter | null> = {}
```
abuaboud, line 25: I would not use bloom filter, the data we have is enough to fit into the data structure (in-map), what I meant is to barrow the same concept where we do two checks, first we do it in memory, If it's not there then we check the disk.

Majority of the calls will eventually hit cache hit inside the memory.


I would do it with one function something like.

```ts
function cache(
  cacheHit: (key: string) => boolean,
  computeValue: () => unknown
): Promise<void> {

} 
```

The way it works is: the first function checks if the cache is hit or not. If it’s not, the second function calculates the actual result and saves it. Then we use this in all methods.

packages/server/worker/src/lib/cache/package-manager.ts
```diff
@@ -82,7 +82,7 @@ export const packageManager = (log: FastifyBaseLogger) => ({
     },
 
     async init({ path }: InitParams): Promise<PackageManagerOutput> {
-        return memoryLock.runExclusive(`pnpm-init-${path}`, async () => {
+        return fileSystemUtils.runExclusive(path, `pnpm-init`, async () => {
```
abuaboud, line 85: That's a huge bottleneck, we don't do any file locks unless it's cache miss to write, so we should do early return if it's cache hit without locking.

file locks are slow

packages/server/worker/src/lib/cache/cache-state.ts
abuaboud: In that function we kind force the same way of caching and optimization, it might not be exactly like that function but you get the idea

PR #9639, opened 2025-10-10
perf(worker): improve worker performance by checking memory before locking

0xSherlokMo:
### Relevant User Scenarios

Fixes # (issue)


packages/server/worker/src/lib/cache/cache-state.ts
```diff
@@ -22,18 +22,49 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
...
+                    state: key,
+                }
+            }
+            const lockKey = `${folderPath}-${cacheAlias}`
+            return fileSystemUtils.runExclusive(folderPath, lockKey, async () => {
+                const freshKey = cache[cacheAlias]
+                if (freshKey !== undefined && !cacheMiss(freshKey)) {
+                    return { cacheHit: true, state: freshKey }
+                }            
+                await installFn?.()
```
0xSherlokMo: Added this argument as optional so that other users of the function aren’t forced to install anything if it’s cascaded

packages/server/worker/src/lib/cache/cache-state.ts
```diff
@@ -22,18 +22,63 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
...
+
+const isCacheHit = (key: string | undefined, cacheMiss?: (key: string) => boolean): boolean => {
+    if (key === undefined) {
+        return false
+    }
+    return cacheMiss ? !cacheMiss(key) : true
+}
+
+export const cacheState = (folderPath: string): {
+    getOrSetCache: (cacheAlias: string, state: string, cacheMiss?: (key: string) => boolean, installFn?: () => Promise<void>, saveGuard?: (key: string) => boolean) => Promise<CacheResult>
```
abuaboud: I think we should make cacheMiss not optional, and same for installFn.

I don't see saveGuard is used anywhere, so I strongly prefer to remove it and only add it when needed

packages/server/worker/src/lib/cache/cache-state.ts
```diff
@@ -22,18 +22,63 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
...
+                await installFn?.()
+                cache[cacheAlias] = state
+                await saveToCache(cache, folderPath)
+                return {
+                    cacheHit: false,
+                    state,
+                }
             })
         },
         async setCache(cacheAlias: string, state: string): Promise<void> {
```
abuaboud: I think we should not expose this inside export, let's force everywhere to use getOrSet, because it's optimized

abuaboud (changes requested):

packages/server/worker/src/lib/cache/cache-state.ts
0xSherlokMo: i agree on the optional fields, i changed that! 
also, for saveGuard its used inside `flow-worker-cache.ts` on `getFlow` func

packages/server/worker/src/lib/cache/cache-state.ts
0xSherlokMo: there's still a function that uses setCache because it doesn't need to look things up, will see if we can get rid of it!

0xSherlokMo: done deleted

packages/server/worker/src/lib/cache/cache-state.ts
```diff
@@ -22,21 +22,56 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
...
+            const key = cache[cacheAlias]
+            if (!isNil(key) && !cacheMiss(key)) {
+                return {
+                    cacheHit: true,
+                    state: key,
+                }
+            }
+            const lockKey = `${folderPath}-${cacheAlias}`
+            return fileSystemUtils.runExclusive(folderPath, lockKey, async () => {
+                const freshKey = cache[cacheAlias]
```
abuaboud: I don't get how this can become fresh, i suspect it's bug, it's same cache that was check initially

abuaboud (changes requested):

packages/server/worker/src/lib/cache/cache-state.ts
```diff
@@ -22,21 +22,56 @@ const getCache = async (folderPath: string): Promise<CacheMap> => {
     return cache
 }
 
-export const cacheState = (folderPath: string) => {
+type CacheResult = {
+    cacheHit: boolean
+    state: string | null
+}
+
+type CacheStateParams = {
```
abuaboud: Let's remove state, i think it can be what ever value returned in installFn

This can be widely improved by having generic

<T> after word cache state parameter this T type can be used in save guard (instead of always string) and also as return type for install

Then you dont have to worry about moving from/to string

abuaboud (changes requested):

packages/server/worker/src/lib/cache/cache-state.ts
0xSherlokMo: If multiple threads are accessing this function concurrently, the initial cache check happens without locking. During that window, the underlying state may change. Acquiring the lock and re-reading ensures we operate on the most up-to-date cache value.

packages/server/worker/src/lib/cache/cache-state.ts
0xSherlokMo: Yeah, that’s a cleaner approach. I’ll go with this

packages/server/worker/src/lib/cache/cache-state.ts
0xSherlokMo: I think the bug is im checking cache miss using the old value

abuaboud (approved):

PR #9672, opened 2025-10-13
refactor: remove pre handlers 

MrChaker:


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -440,7 +440,9 @@ export const flowService = (log: FastifyBaseLogger) => ({
             }
         })
     },
-
+    async deleteAllByProjectId(projectId: ProjectId): Promise<void> {
```
abuaboud: This has same issue we trying to solve, we need to call delete flow function the individual one since it remove the polling job (the trigger to check new data) from the bullmq using pre delete.

--------

The goal to remove this function, and we always guarantee that there is no stale jobs in bullmq 

packages/server/api/src/app/workers/consumer/pre-handlers/scheduled-job-pre-handler.ts

abuaboud: Please remove all soft delete code as well, as well the system jobs to delete the projects after 30 days
https://github.com/activepieces/activepieces/blob/f58c726837edafcde34cc8011409e98ffc56505a/packages/server/api/src/app/helper/system-jobs/common.ts#L41

packages/server/api/src/app/workers/queue/migration/removeHardDeleteProjectJobs.ts
```diff
@@ -0,0 +1,22 @@
...
+        const redisType = redisConnections.getRedisType()
+        if (redisType !== RedisType.MEMORY) {
+            log.info('[removeHardDeleteProjectJobs] Skipping migration because they are migrated from persistent queue')
+            return
+        }
+        const scheduledJobs = await systemJobsQueue.getJobs();
+        const hardDeleteJobs = scheduledJobs.filter(job => (job.name as string) === 'hard-delete-project')
+
+        for (const job of hardDeleteJobs) {
+            await job.remove()
```
abuaboud: But you need to call delete as well there

packages/server/api/src/app/workers/queue/migration/removeHardDeleteProjectJobs.ts
```diff
@@ -0,0 +1,22 @@
+import { FastifyBaseLogger } from "fastify"
+import { redisConnections } from '../../../database/redis'
+import { RedisType } from "../../../helper/system/system"
+import { systemJobsQueue } from "../../../helper/system-jobs/system-job"
+
+export const removeHardDeleteProjectJobs = (log: FastifyBaseLogger) => ({
+    async run(): Promise<void> {
+        const redisType = redisConnections.getRedisType()
+        if (redisType !== RedisType.MEMORY) {
```
abuaboud: Remove this check, it’s un related 

We need to add migration check, In other migration we maintain boolean to know if we ran this migration or not to avoid it multiple times

Issue #9678, opened 2025-10-14
[BUG]: Medullar Code Piece fails as they changed the API


mllopart:
**Describe the bug**
A small change in Medullar API broke Medullar Code Piece

**To Reproduce**
1.- Add Medullar Piece
2.- Authenticate
3.- Test chat and error will popup.

**Expected behavior**
Chat response is received

PR #9679, opened 2025-10-14
fix: medullar api changes ENG-1096

mllopart:
## What does this PR do?

Fixes Medullar API with the new AI API

Medullar has changed some APIs, and instead of using the Explorer API, we should use the AI API. This PR reroutes all APIS to the new one.

### Bug Fix

The following screenshot shows now the API returning the right information.

CleanShot 2025-10-14 at 11 03 53

Fixes # ENG-1096


kishanprmr:
Fixes #9678

mllopart:
as always, thanks @kishanprmr  :)

PR #9680, opened 2025-10-14
refactor: remove pre handlers

MrChaker:


packages/server/api/src/app/flows/flow/flow.cache.ts
```diff
@@ -0,0 +1,33 @@
+import { FastifyBaseLogger } from "fastify";
+import { redisConnections } from "../../database/redis-connections";
+import { FlowStatus } from "@activepieces/shared";
+
+export enum FlowCacheStatus {
+    ENABLED = FlowStatus.ENABLED,
+    DISABLED = FlowStatus.DISABLED,
+    DELETED = "DELETED",
```
abuaboud: How about we remove that, and we store FlowStatus.

In that case the null means it's deleted, because in all our code we will consider deleted as does not exist

packages/server/api/src/app/flows/flow/flow-service-side-effects.ts
```diff
@@ -37,6 +39,7 @@ export const flowSideEffects = (log: FastifyBaseLogger) => ({
     },
 
     async preDelete({ flowToDelete }: PreDeleteParams): Promise<void> {
+        await flowCache(log).updateStatusCache({ id: flowToDelete.id, status: FlowCacheStatus.DELETED })
```
abuaboud: I would create new function called, deleteStatusCache instead of specific cache status DELETED

abuaboud: The reason is because we don't have two values represent the same status (DELETED , null) means the flow were deleted.

PR #9681, opened 2025-10-14
fix: some instant triggers fail when getting published

AbdulTheActivePiecer:
Before:
Return 410 (GONE) when trigger source isn't found and (404) when flow is disabled.

Now:
Return 404 for trigger source not found and 410 if flow isn't found
+
We create trigger source before calling onEnable for trigger.


packages/server/api/src/app/webhooks/webhook.service.ts
```diff
@@ -45,12 +45,14 @@ export const webhookService = {
                 })
 
                 if (isNil(triggerSourceResult)) {
```
abuaboud: Decouple trigger source query from flow.

Check for flow existence first if not there return gone

Then query trigger source if not there return not found

AbdulTheActivePiecer: done

PR #9708, opened 2025-10-15
feat: add operation team to handbook

hazemadelkhalel:
## What does this PR do?

This PR adds the operations team’s handbook, which includes their mission statement, personnel list, workflow, and issue triage flowcharts.


abuaboud:
Hi @hazemadelkhalel 

That's great initiative, Let's talk about it offline I think we should have centralise team the mission where it boost the team productivity and ability for the company to scale number of people would be great idea.

We can polish it offline first! :D

PR #9711, opened 2025-10-15
feat(imap): add Copy Email, Move Email, and Delete Email actions

simonc:
## What does this PR do?

This PR adds 3 new actions to the IMAP piece:

- Copy Email: Copy an email from one mailbox to another
- Move Email: Move an email from one mailbox to another
- Delete Email: Delete an email

It also makes a change in the way the `New Email`  trigger testing works to ensure it always returns the most recent messages upon retesting, aligning its behavior with what one would expect.

The `Copy Email` action needs a Message UID, a source folder (mailbox) and a target folder. It will copy the message, leaving the original one untouched.

The `Move Email` action needs a Message UID, a source folder and a target folder. It will move the message to the target folder, the original message will be deleted. This action should be used to trash a message.

The `Delete Email` action needs a Message UID and a parent folder. It will permanently delete a message.

The `New Email` trigger test uses a wrapped polling object that will reverse the order of the fetched emails so that when `pollingHelper.test` limit of 5 items apply, the most recent are the ones returned.

### Relevant User Scenarios

These actions are useful for email automations, both simple or LLM-based.

### DIFF
--- a/packages/pieces/community/imap/src/lib/common/imap.ts
+++ b/packages/pieces/community/imap/src/lib/common/imap.ts
@@ -255,7 +255,7 @@ async function performMailboxOperation<T>(
     let lock: MailboxLockObject | null = null;
 
     try {
-      lock = await imapClient.getMailboxLock(mailbox, { readonly: readOnly });
+      lock = await imapClient.getMailboxLock(mailbox, { readOnly });
       return await callback(imapClient);
     } catch (error) {
       detectMissingMailbox(error);


simonc:
@kishanprmr Here's the new PR. After rebasing on top of main I had quite a battle with the translations as Crowdin had a go in the mean time. It seems it does a few things that make it difficult to work with:

1. It changes the order of the keys
2. It un-translates things that are correct
3. It breaks URLs
4. It changes some translations, not always for the best (at least from what I could tell on the French one)

If you look at `packages/pieces/community/imap/src/i18n/fr.json` in commit 32b4faa6d5, you can see that it breaks the URL in the Gmail warning and un-translates "Password".

I understand it's somewhat of an edge case, but for anyone that may have a few ongoing branches, that can quickly become a nightmare 😅

onyedikachi-david:
Thanks for fixing the conflicts. Merging now.

PR #9748, opened 2025-10-17
fix: prevent Helm chart dependencies from being committed

vincz777:
- Remove existing *.tgz files from git tracking
- Add .gitignore in deploy/activepieces-helm/ to ignore charts/*.tgz
- Downloaded Helm dependencies should not be version controlled


abuaboud (approved):

abuaboud:
Thank you!

PR #9752, opened 2025-10-17
fix: pipedrive trigger - updated deal stage bug

jeffrey94:
## What does this PR do?
  Fixes the Pipedrive "Updated Deal Stage" trigger to only fire when a deal's stage is
  actually changed, rather than firing on any update to deals that are already in the
  selected stage.

  **The Problem:**
  The trigger was incorrectly firing whenever ANY field on a deal was updated if that
  deal was in the specified stage. For example, if a user configured the trigger to
  watch for deals moving to "Closed Won", the trigger would fire when updating the deal
  value, title, or any other field on deals already in "Closed Won" stage.

  **The Root Cause:**
  Pipedrive's webhook payload only includes fields that have changed in the `previous`
  object. When comparing `currentDealData.stage_id !== previousDealData.stage_id`, if
  the stage didn't change, `previousDealData.stage_id` would be `undefined`, causing `5
  !== undefined` to evaluate to `true` and incorrectly trigger the flow.

  **The Fix:**
  Added an early return check to verify that `stage_id` exists in the `previousDealData`
   object before comparing values. This ensures the trigger only fires when the stage
  field was actually modified.

  ```typescript
  // Check if previous data exists and has stage_id
  if (!previousDealData || isNil(previousDealData.stage_id)) {
    return [];
  }

### DIFF
--- a/packages/pieces/community/pdf/package.json
+++ b/packages/pieces/community/pdf/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-pdf",
-  "version": "0.2.14",
+  "version": "0.2.15",
   "dependencies": {
     "jimp": "^0.22.12",
     "pdf-parse": "1.1.1",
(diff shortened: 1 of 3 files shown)


danielpoonwj:
just a bit more context, `stage_id` would only be there for the first webhook request on transitioning stages, which correctly triggers the run.

Screenshot 2025-10-17 at 8 22 25 PM

but other updates would not contain the field and would always wrongly trigger runs because it's `undefined`
```javascript
if (currentDealData.stage_id !== previousDealData.stage_id) {
...
}
```

Screenshot 2025-10-17 at 8 22 44 PM

jeffrey94:
@kishanprmr wonder if the item is working on this request in this cycle? Thank you.

kishanprmr (approved):

PR #9769, opened 2025-10-19
fix: user identity reuse after deletion (closes #9669)

### LINKED ISSUE #9669: [BUG] Deleting a user orphans their identity → re-invited user is locked out (self-hosted)
## Summary

Self-hosted: deleting a platform user and re-inviting the same email locks the person out. Signup is rejected ("need an invitation" / "account already exists") and sign-in fails. The bug is edition-agnostic (the delete path is shared by CE and EE); only recovery differs: EE can recover via forgot-password **if SMTP** **is** **configured**, CE has no reset path at all. Multiple reporters (GitHub activepieces/activepieces#9669). Reproduced over HTTP on current build.

## Root cause

Deleting a user removes only the `user` row, never `user_identity` (`userService.delete`, `packages/server/api/src/app/user/user-service.ts` — no path deletes the identity; only Cloud takes the `removeFromPlatform` branch, so both CE and EE self-hosted hit this). The identity's email is globally unique (`user-identity-service.ts:19-28`) and keeps the old password hash, so it blocks re-signup and can't be signed into.

Re-invite makes it worse: clicking the link (`POST /user-invitations/accept` → `provisionUserInvitation`) silently re-creates the `user` on the dead identity and consumes the invite. Accept returns no `registered` field (`user-invitation.module.ts:68-75`), so the frontend always redirects to `/sign-up`, which then throws `INVITATION_ONLY_SIGN_UP` / `EXISTING_USER`.

## Verified repro (ran on CE; delete path identical on EE)

1. Invite + sign up → `user` + `user_identity` created.
2. `DELETE /v1/users/:id` → 204. `user` gone, `user_identity` survives.
3. Re-invite → accept → `user` silently re-created, invite consumed.
4. Sign-up → `INVITATION_ONLY_SIGN_UP`; sign-in → `INVALID_CREDENTIALS`. Locked out.

## Fix

Delete must not orphan the identity: on user delete, also remove `user_identity` when no `user` references it. Also make `/accept` return `registered` so existing identities route to `/sign-in`, not `/sign-up`.

## Workaround

EE: configure `AP_SMTP_*` and have the user reset via forgot-password (note the 10-min OTP throttle). Any edition: `DELETE /v1/users/:id`, then `DELETE FROM user_identity WHERE email = LOWER('<email>')`, then re-invite (verified fixes it).

Ridhim-RR:
## What does this PR do?

 This PR fixes a critical edge case in the user invitation and signup flow where users who have been deleted from a platform cannot rejoin even when they receive a new invitation. The issue occurs because when a user is deleted, only the User entity is removed while the UserIdentity remains in the database, causing authentication conflicts during subsequent signup attempts.
  Additionally, this PR fixes an issue where invitations were being prematurely deleted during the provisioning process, preventing users from properly accepting invitations after being re-invited.
 
 **Previous Behavior:**
  1. User gets invited → UserInvitation created
  2. User accepts invitation → User and UserIdentity entities created
  3. User gets deleted → Only User entity deleted, UserIdentity remains
  4. New invitation sent → UserInvitation created again
  5. User accepts invitation → New User created successfully
  6. Invitation gets deleted immediately → User loses access
  7. User tries to sign up → ERROR: "Sign up is restricted. You need an invitation to join.

**New Behavior:**
  1. User gets invited → UserInvitation created
  2. User accepts invitation → User and UserIdentity entities created
  3. User gets deleted → Only User entity deleted, UserIdentity remains
  4. New invitation sent → UserInvitation created again
  5. User accepts invitation → New User created successfully
  6. Invitation remains active → User maintains access
  7. User tries to sign up → SUCCESS: User can sign up and set new password
  
**Technical Implementation:**
  1. **Fixed Authentication Service:**
  • Modified authenticationService.signUp() to check for existing user identities before creating new ones
  • When an existing UserIdentity is found, the system now reuses it instead of throwing an error
  • Ensures proper user creation for the new platform without conflicts

  2. Fixed Invitation Provisioning:
  • Updated provisionUserInvitation() to only delete invitations when the user already exists for that platform
  • Prevents premature deletion of invitations when UserIdentity is present but User doesn't exist for the platform
  • Maintains invitation integrity during the re-invitation process

**Relevant User Scenarios
  Primary Use Cases:**
  • User Management: When administrators need to remove and re-add users to their platform
  • Account Recovery: When users need to reset their access after being temporarily removed
  • Team Management: When users are moved between different projects or platforms
  • Security Incidents: When users are temporarily removed for security reasons and need to be re-added

Relevant User Scenarios

  Primary Use Cases:
  • User Management: When administrators need to remove and re-add users to their platform
  • Account Recovery: When users need to reset their access after being temporarily removed
  • Team Management: When users are moved between different projects or platforms
  • Security Incidents: When users are temporarily removed for security reasons and need to be re-added

Fixes #9669


abuaboud:

PR #9776, opened 2025-10-20
chore: update trigger run to use redis and store only stats

MrChaker:


packages/server/api/src/app/trigger/trigger-run/trigger-run.service.ts
```diff
@@ -1,100 +1,111 @@
...
-        return triggerRunRepo().findOneByOrFail({ jobId })
+
+    async create({ platformId, pieceName, status }: CreateParams): Promise<void> {
+        const day = apDayjs().format('YYYY-MM-DD')
+        const statusToStore = status === TriggerRunStatus.COMPLETED ? status : TriggerRunStatus.FAILED
+        const redisKey = triggerRunRedisKey(platformId, pieceName, day, statusToStore)
+
+        const redisConnection = await redisConnections.useExisting()
+        await redisConnection.incr(redisKey)
+        await redisConnection.expire(redisKey, 60 * 60 * 24 * 14) // 14 days
```
abuaboud: use apDayJs duration

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -81,8 +81,8 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
     })
 
     app.post('/create-trigger-run', CreateTriggerRunParams, async (request) => {
```
abuaboud: Please delete this endpoint and from there, try to move trigger-run-service into server-shared where it takes redis connection in the paramater, and use it in both places. 

Please remove trigger-run entity (and run TRUNCATE trigger-run) on the tables to free up the database.

PR #9787, opened 2025-10-21
feat: introduce dedicated workers for enterprise customers

amrdb:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/server/api/src/app/workers/queue/job-queue.ts
+++ b/packages/server/api/src/app/workers/queue/job-queue.ts
@@ -1,4 +1,4 @@
-import { apDayjsDuration, AppSystemProp, getPlatformQueueName, QueueName } from '@activepieces/server-shared'
+import { apDayjsDuration, AppSystemProp, getPlatformQueueName, memoryLock, QueueName } from '@activepieces/server-shared'
 import { ApId, getDefaultJobPriority, isNil, JOB_PRIORITY } from '@activepieces/shared'
 import { Queue } from 'bullmq'
 import { BullMQOtel } from 'bullmq-otel'
@@ -15,9 +15,9 @@ const REDIS_FAILED_JOB_RETRY_COUNT = system.getNumberOrThrow(AppSystemProp.REDIS
 const dedicatedWorkersQueues = new Map<string, Queue>()
 
 export const jobQueue = (log: FastifyBaseLogger) => ({
-    async init(): Promise<void> {        
+    async init(): Promise<void> {
         const platformIdsWithDedicatedWorkers = await dedicatedWorkers(log).getPlatformIds()
-        
+
         await Promise.all([
             ...platformIdsWithDedicatedWorkers.map(async (platformId) => {
                 const queueName = await getQueueName(platformId, log)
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/database/migration/postgres/1760998784106-AddDedicatedWorkersToPlatformPlanPostgres.ts
```diff
@@ -0,0 +1,42 @@
...
+            ALTER TABLE "platform_plan"
+            ADD "dedicatedWorkers" jsonb
+        `)
+        await queryRunner.query(`
+            ALTER TABLE "platform_plan"
+            ALTER COLUMN "stripeBillingCycle" DROP DEFAULT
+        `)
+        await queryRunner.query(`
+            ALTER TABLE "flow"
+            ADD CONSTRAINT "fk_flow_project_id" FOREIGN KEY ("projectId") REFERENCES "project"("id") ON DELETE CASCADE ON UPDATE NO ACTION
```
abuaboud: I think this is by mitake

packages/server/worker/src/lib/compute/process/factory/isolate-sandbox-process.ts
```diff
@@ -60,11 +60,15 @@ export const isolateSandboxProcess = (log: FastifyBaseLogger): EngineProcess =>
...
 function getEnvironmentVariables(env: Record<string, string | undefined>, workerId: string): string[] {
+    const executionMode = workerMachine.getSettings().EXECUTION_MODE
+    const isSandboxCodeAndProcess = executionMode === ExecutionMode.SANDBOX_CODE_AND_PROCESS
+    
     return Object.entries({
         ...env,
         AP_BASE_CODE_DIRECTORY: '/codes',
         HOME: '/tmp/',
         WORKER_ID: workerId,
+        ...(isSandboxCodeAndProcess && { AP_EXECUTION_MODE: ExecutionMode.SANDBOX_CODE_ONLY }),
```
abuaboud: Why not pass it as SANDBOX_CODE_AND_PROCESS, You already handling this in code executor.

packages/server/worker/src/lib/consume/job-queue-worker.ts
```diff
@@ -32,7 +32,12 @@ export const jobQueueWorker = (log: FastifyBaseLogger) => ({
             return
         }
         const isOtpEnabled = workerMachine.getSettings().OTEL_ENABLED
-        worker = new Worker<JobData>(QueueName.WORKER_JOBS, async (job, token) => {
+        let queueName: string = QueueName.WORKER_JOBS
+        const platformIdForDedicatedWorker = workerMachine.getSettings().PLATFORM_ID_FOR_DEDICATED_WORKER
```
abuaboud: It's better to have function for this.

const queueName = getWorkerQueueName().

packages/server/api/src/app/workers/machine/machine-service.ts
```diff
@@ -27,12 +31,28 @@ export const machineService = (_log: FastifyBaseLogger) => {
             })
             await workerRepo().delete({ id: request.workerId })
         },
-        async onConnection(): Promise<WorkerSettingsResponse> {
+        async onConnection(platformIdForDedicatedWorker?: string | undefined): Promise<WorkerSettingsResponse> {
+            let executionMode = system.getOrThrow(AppSystemProp.EXECUTION_MODE)
```
abuaboud: This can be function const executionMode = getExecutionMode(platformId: string | undefined).

The check if platform id is null then it should just return the one in environment variable.

You should move the edition check inside getDedicatedWorkerForConfig and make it return null.

packages/server/api/src/app/workers/queue/job-queue.ts
```diff
@@ -1,29 +1,53 @@
...
-export let workerJobsQueue: Queue | undefined = undefined
+const dedicatedWorkersQueues = new Map<string, Queue>()
+export let sharedWorkerJobsQueue: Queue | undefined = undefined
 
 export const jobQueue = (log: FastifyBaseLogger): QueueManager => ({
     async init(): Promise<void> {
-        const queues = Object.values(QueueName).map((queueName) => ensureQueueExists(queueName))
-        await Promise.all(queues)
-        log.info('[redisQueueManager#init] Redis queues initialized')
+        let platformIdsWithDedicatedWorkers: string[] = []
```
abuaboud: same this, the community edition check can be moved inside getPlatformIdsWithDEdicatedWorkers() and make it return empty array.

packages/server/api/src/app/workers/queue/job-queue.ts
```diff
@@ -1,29 +1,53 @@
...
+
+        log.info('[jobQueue#init] Dynamic queue system initialized')
     },
+
     async add(params: AddJobParams<JobType>): Promise<void> {
         const { type, data } = params
 
-        const queue = await ensureQueueExists(QueueName.WORKER_JOBS)
+        const platformId = data.platformId
+        let isDedicatedWorkersEnabled = false
```
abuaboud: All of this should be removed.

We only need getPlatformQueueName(platformId | null), and here you do the check if dedicated workers enabled then return queue name else WORKER_JOBS, the dedicated worker community edition check should be inside isDedicatedWorkersEnabled function

packages/server/api/src/app/workers/queue/job-queue.ts
```diff
@@ -79,13 +120,25 @@ async function ensureQueueExists(queueName: QueueName): Promise<Queue> {
                 count: REDIS_FAILED_JOB_RETRY_COUNT,
             },
         },
+    })
+    
+    await queue.removeGlobalConcurrency()
+    await queue.waitUntilReady()
 
+    if (platformId) {
+        dedicatedWorkersQueues.set(platformId, queue)
```
abuaboud: Why not all of them follow the same structure in the end the queue name is string and value is the queue.

PR #9810, opened 2025-10-23
feat: support maximum concurrent jobs per project

### LINKED ISSUE #9809: Add Maximum Concurrent Jobs Per Project

hazemadelkhalel:
## What does this PR do?

Fixes #9809

### DIFF
--- a/packages/server/worker/src/lib/consume/job-queue-worker.ts
+++ b/packages/server/worker/src/lib/consume/job-queue-worker.ts
@@ -47,9 +47,7 @@ export const jobQueueWorker = (log: FastifyBaseLogger) => ({
                 }
 
                 assertNotNullOrUndefined(jobId, 'jobId')
-                const maxConcurrentJobsPerProject = await getMaxConcurrentJobsPerProject(job.data.projectId)
-
-                const { shouldRateLimit } = await workerJobRateLimiter(log).shouldBeLimited(jobId, job.data, maxConcurrentJobsPerProject)
+                const { shouldRateLimit } = await workerJobRateLimiter(log).shouldBeLimited(jobId, job.data)
                 if (shouldRateLimit) {
                     await job.moveToDelayed(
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/database/migration/postgres/1761245180906-AddMaximumConcurrentJobsPerProject.ts
```diff
@@ -0,0 +1,33 @@
...
+            ADD "maxConcurrentJobs" integer
+        `)
+
+        await queryRunner.query(`
+            UPDATE "project" SET "maxConcurrentJobs" = $1
+        `, [MAX_CONCURRENT_JOBS_PER_PROJECT])
+
+        await queryRunner.query(`
+            ALTER TABLE "project"
+            ALTER COLUMN "maxConcurrentJobs" SET NOT NULL
```
abuaboud: Let's make it nullable so the null one follow the system default

packages/server/worker/src/lib/consume/job-queue-worker.ts
```diff
@@ -46,7 +46,9 @@ export const jobQueueWorker = (log: FastifyBaseLogger) => ({
                 }
 
                 assertNotNullOrUndefined(jobId, 'jobId')
-                const { shouldRateLimit } = await workerJobRateLimiter(log).shouldBeLimited(jobId, job.data)
+                const maxConcurrentJobsPerProject = isNil(job.data.projectId) ? null : Number(await workerDistributedStore.get(getProjectMaxConcurrentJobsKey(job.data.projectId)))
```
abuaboud: Move this to function and in case it null it should the default settings (environment)

hazemadelkhalel: > Move this to function and in case it null it should the default settings (environment)

I think we agreed that shouldBeLimited should skip the limitation when it’s null, right? Because if we do it this way, it means we’re always applying the limit.

PR #9873, opened 2025-10-27
fix: handle deleted flows in metadata queue and preserve pause metadata in engine updates

amrdb:
1. Handle deleted flows in runs metadata queue
   - Added validation to check if flow exists before processing run metadata
   - Skips metadata processing for runs whose flows have been deleted
   - Prevents errors when processing stale job data for deleted flows

2. Preserve pause metadata in flow run updates
   - Added pauseMetadata parameter handling in engine controller


abuaboud:
@allcontributors please add @im-soohyun for security

PR #9879, opened 2025-10-27
feat: add default behaviour for job concurrency based on plan name

hazemadelkhalel:
## What does this PR do?

We give higher priority to the maximum job concurrency configured in the project itself. Otherwise, the limit is determined based on the plan, as different plans have different limits.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2057,6 +2057,24 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "im-soohyun",
+      "name": "Kim SooHyun",
+      "avatar_url": "https://avatars.githubusercontent.com/u/127273427?v=4",
+      "profile": "https://github.com/im-soohyun",
+      "contributions": [
+        "security"
+      ]
+    },
+    {
+      "login": "srimalleswari205",
+      "name": "Sri malleswari",
+      "avatar_url": "https://avatars.githubusercontent.com/u/235991945?v=4",
+      "profile": "https://github.com/srimalleswari205",
+      "contributions": [
+        "plugin"
+      ]
     }
   ],
   "commitType": "docs"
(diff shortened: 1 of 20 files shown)


packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -122,10 +125,40 @@ export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
...
+        return Number(storedValues[getProjectMaxConcurrentJobsKey(projectId)])
+    }
+
+    const platformPlanName = storedValues[getPlatformPlanNameKey(platformId)]
+    if (isNil(platformPlanName)) {
+        return workerMachine.getSettings().MAX_CONCURRENT_JOBS_PER_PROJECT
+    }
+
+    if (workerMachine.getSettings().EDITION === ApEdition.CLOUD && platformPlanName === PlanName.ENTERPRISE) {
+        return 30
```
abuaboud: I think this statement should be removed, Just have it part of PLAN_CONCURRENT_JOBS_LIMITS

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -122,10 +125,40 @@ export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
...
+    [PlanName.APPSUMO_ACTIVEPIECES_TIER1]: 5,
+    [PlanName.APPSUMO_ACTIVEPIECES_TIER2]: 5,
+    [PlanName.APPSUMO_ACTIVEPIECES_TIER3]: 10,
+    [PlanName.APPSUMO_ACTIVEPIECES_TIER4]: 15,
+    [PlanName.APPSUMO_ACTIVEPIECES_TIER5]: 20,
+    [PlanName.APPSUMO_ACTIVEPIECES_TIER6]: 25,
+}
+
+async function getMaxConcurrentJobsPerProject({ projectId, platformId }: GetMaximumConcurrentJovsPerProjectParams): Promise<number> {
+    const keys = [getPlatformPlanNameKey(platformId)]
```
abuaboud, line 169: I would refactor this into two functions.

const concurrentjobsFromProject = .... (function)
const concurrentJobsFromPlan = .... (functiion)

in the second function,  you check if edition is not cloud then return null, otherwise just based on the plan.

Then you have something like that.
if(!isNil(concurrentjobsFromProject)) return concurrentjobsFromProject
if(!isNil(concurrentJobsFromPlan)) return concurrentJobsFromPlan
return workerMachine.getSettings().MAX_CONCURRENT_JOBS_PER_PROJECT

packages/server/api/src/app/ee/platform/platform-plan/platform-plan.service.ts
```diff
@@ -21,14 +21,14 @@ const edition = system.getEdition()
...
+        const platformPlan = await findAndStorePlatformPlan(platformId)
         if (!isNil(platformPlan)) return platformPlan
 
         return distributedLock(log).runExclusive({
             key: `platform_plan_${platformId}`,
             timeoutInSeconds: 60,
             fn: async () => {
-                const platformPlan = await platformPlanRepo().findOneBy({ platformId })
+                const platformPlan = await findAndStorePlatformPlan(platformId)
                 if (!isNil(platformPlan)) return platformPlan
```
abuaboud, line 33: I would remove `findAndStorePlatformPlan` and just have store in that step, for older projects will create an admin endpoint to back fill it.

abuaboud (changes requested):

abuaboud (approved):

PR #9905, opened 2025-10-29
feat(eng-1145): separate pieces dependencies from main

MrChaker:


packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -43,6 +47,66 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
...
+        }
+        catch (e) {
+            return null
+        }
+    }
+
+    async function cachePieceDependencies(pieceName: string, dependencies: Record<string, string>): Promise<void> {
+        let currentCache: Record<string, Record<string, string>> = {}
+        try {
+            currentCache = JSON.parse(readFileSync(pieceDependenciesCacheFilePath, 'utf-8'))
```
amrdb: It is better to use async file operation so we don't block the Node.js event loop. Async operations are already used in this file and imported from 'node:fs/promises'.

packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -43,6 +47,66 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
...
+        }
+        catch (e) {
+            currentCache = {}
+        }
+
+        writeFileSync(pieceDependenciesCacheFilePath, JSON.stringify({
```
amrdb: Async here too.

packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -43,6 +47,66 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
...
+            ...currentCache,
+            [pieceName]: dependencies
+        }, null, 2), 'utf-8')
+    }
+
+    async function getCachedPieceDependencies(pieceName: string): Promise<Record<string, string> | null> {
+
+        let currentCache: Record<string, Record<string, string>> = {}
+        try {
+            currentCache = JSON.parse(readFileSync(pieceDependenciesCacheFilePath, 'utf-8'))
```
amrdb: Async here too.

package.json
```diff
@@ -232,36 +198,25 @@
...
     "node-cron": "3.0.3",
     "nodemailer": "7.0.9",
-    "notion-to-md": "3.1.1",
     "nx-cloud": "19.1.0",
     "object-sizeof": "2.6.3",
-    "openai": "4.67.1",
     "p-limit": "2.3.0",
     "pako": "2.1.0",
     "papaparse": "5.5.3",
-    "pg": "8.11.3",
```
amrdb, line 254: pg is required by typeorm.

packages/pieces/community/agent/package.json
```diff
@@ -1,4 +1,8 @@
 {
   "name": "@activepieces/piece-agent",
-  "version": "0.3.2"
+  "version": "0.3.2",
```
amrdb: Can you bump the patch in the version for each updated piece?

packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -43,6 +47,66 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
...
+        }
+        catch (e) {
+            currentCache = {}
+        }
+
+        writeFileSync(pieceDependenciesCacheFilePath, JSON.stringify({
+            ...currentCache,
+            [pieceName]: dependencies
+        }, null, 2), 'utf-8')
```
amrdb: Pretty printing with `JSON.stringify(data, null, 2)` will the make the file unnecessarily larger since this case is not meant to be read by humans. Just call it like `JSON.stringify(data)`

packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -162,5 +226,10 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
         clearPieceCache,
         getPackageNameFromFolderPath,
         getProjectJsonFromFolderPath,
+        getPieceDependencies,
+        cachePieceDependencies,
+        getCachedPieceDependencies,
+        checkPieceDependenciesUpdated,
```
amrdb: Nit: either inline this logic directly into installPieceDependencies or keep them as private helper functions that aren't exported since they are not used anywhere else.

packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -43,6 +47,66 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
...
+    async function getPieceDependencies(folerPath: string): Promise<Record<string, string> | null> {
+        try {
+            const packageJson = JSON.parse(readFileSync(join(folerPath, 'package.json'), 'utf-8'))
+            if (!packageJson.dependencies) {
+                return null
+            }
+            return packageJson.dependencies
+        }
+        catch (e) {
+            return null
```
amrdb, line 58: Nit: I think it's a good idea to always log the error, see findDirectoryByPackageName.

packages/server/shared/src/lib/pieces/file-pieces-utils.ts
```diff
@@ -43,6 +47,66 @@ export const filePiecesUtils = (packages: string[], log: FastifyBaseLogger) => {
...
+        }
+    }
+
+    async function cachePieceDependencies(pieceName: string, dependencies: Record<string, string>): Promise<void> {
+        let currentCache: Record<string, Record<string, string>> = {}
+        try {
+            currentCache = JSON.parse(readFileSync(pieceDependenciesCacheFilePath, 'utf-8'))
+        }
+        catch (e) {
+            currentCache = {}
```
amrdb: Nit: I think it's a good idea to always log the error, see findDirectoryByPackageName. Since this code expects a not found error maybe do smth like:
if (err.code === 'ENOENT') {
currentCache = {}
} else {
// log error
}

amrdb (approved):

PR #9906, opened 2025-10-29
fix(date-helper): parsing date with formats wasn't working for add/subtract date

onyedikachi-david:
### Relevant User Scenarios

Fixes # (issue)


packages/pieces/community/date-helper/src/lib/actions/add-subtract-date.ts
```diff
@@ -16,6 +17,7 @@ import { z } from 'zod';
 import { propsValidation } from '@activepieces/pieces-common';
 
 dayjs.extend(advancedFormat);
+dayjs.extend(customParseFormat);
 dayjs.extend(utc);
 dayjs.extend(timezone);
```
AbdulTheActivePiecer: Please make sure customParseFormat  doesn't conflict with advancedFormat.

Then create a method in common.ts of this piece and call it "extendDayJs" then call that method in the beginning of each run method, and test to make sure it works, I don't know if it has to be the way it is now or not.

onyedikachi-david:
> Don't merge before these changes, please.

image

Fixed @AbdulTheActivePiecer

AbdulTheActivePiecer:
Can you please test it with all the formats, just to be sure it doesn't break for people who already have flows using this piece?

onyedikachi-david:
> Can you please test it with all the formats, just to be sure it doesn't break for people who already have flows using this piece?

Tested on all formats and its all good. @AbdulTheActivePiecer

PR #9911, opened 2025-10-29
feat: move updating runs to worker and refactor engine<->worker communication

amrdb:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2084,6 +2084,15 @@
       "contributions": [
         "plugin"
       ]
+    },
+    {
+      "login": "Dania02525",
+      "name": "Dania Es",
+      "avatar_url": "https://avatars.githubusercontent.com/u/11240307?v=4",
+      "profile": "http://dania.es",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitType": "docs"
--- a/README.md
+++ b/README.md
@@ -450,6 +450,7 @@ Not into coding but still interested in contributing? Come join our [Discord](ht
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/im-soohyun"><img src="https://avatars.githubusercontent.com/u/127273427?v=4?s=100" width="100px;" alt="Kim SooHyun"/><br /><sub><b>Kim SooHyun</b></sub></a><br /><a href="#security-im-soohyun" title="Security">🛡️</a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/srimalleswari205"><img src="https://avatars.githubusercontent.com/u/235991945?v=4?s=100" width="100px;" alt="Sri malleswari"/><br /><sub><b>Sri malleswari</b></sub></a><br /><a href="#plugin-srimalleswari205" title="Plugin/utility libraries">🔌</a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/SinghaAnirban005"><img src="https://avatars.githubusercontent.com/u/143536290?v=4?s=100" width="100px;" alt="Anirban Singha"/><br /><sub><b>Anirban Singha</b></sub></a><br /><a href="#plugin-SinghaAnirban005" title="Plugin/utility libraries">🔌</a></td>
+      <td align="center" valign="top" width="14.28%"><a href="http://dania.es"><img src="https://avatars.githubusercontent.com/u/11240307?v=4?s=100" width="100px;" alt="Dania Es"/><br /><sub><b>Dania Es</b></sub></a><br /><a href="https://github.com/activepieces/activepieces/commits?author=Dania02525" title="Code">💻</a></td>
     </tr>
   </tbody>
 </table>
--- a/deploy/activepieces-helm/Chart.yaml
+++ b/deploy/activepieces-helm/Chart.yaml
@@ -16,13 +16,13 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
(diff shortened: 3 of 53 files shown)


packages/server/api/src/app/flows/flow.module.ts
```diff
@@ -13,12 +13,29 @@ export const flowModule: FastifyPluginAsyncTypebox = async (app) => {
...
+        return async (data: TestFlowRunRequestBody, principal: Principal): Promise<void> => {
             const flowRun = await flowRunService(app.log).test({
                 projectId: principal.projectId,
                 flowVersionId: data.flowVersionId,
             })
             socket.emit(WebsocketClientEvent.TEST_FLOW_RUN_STARTED, flowRun)
         }
     })
+    websocketService.addListener(PrincipalType.WORKER, WebsocketServerEvent.EMIT_TEST_STEP_PROGRESS, (socket) => {
+        return async (data: StepRunResponse & { projectId: string }, principal: Principal, callback?: (data: unknown) => void): Promise<void> => {
```
abuaboud: Let make this `EmitTestStepProgressRequest` same for below

packages/server/worker/src/lib/compute/flow-runs-queue.ts
```diff
@@ -0,0 +1,43 @@
...
+import { FastifyBaseLogger } from 'fastify'
+import { workerMachine } from '../utils/machine'
+import { workerDistributedStore, workerRedisConnections } from '../utils/worker-redis'
+
+const queue = runsMetadataQueueFactory({ 
+    createRedisConnection: workerRedisConnections.create,
+    distributedStore: workerDistributedStore,
+})
+
+export const runsMetadataQueue = (log: FastifyBaseLogger) => ({
```
abuaboud: I feel we can remove this, It's basically `runsMetadataQueueFactory` with same extra parameters

packages/server/worker/src/lib/flow-worker.ts
```diff
@@ -13,7 +14,45 @@ let workerToken: string
...
+        const timeoutMs = options?.timeoutMs ?? 10000
+        const retries = options?.retries ?? 3
+        const retryDelayMs = options?.retryDelayMs ?? 2000
+
+        for (let attempt = 0; attempt <= retries; attempt++) {
+            try {
+                if (!socket || !socket.connected) {
+                    throw new Error('Socket not connected')
+                }
+                await socket.timeout(timeoutMs).emitWithAck(event, data)
```
abuaboud: Remove all these retries, and use https://socket.io/docs/v4/tutorial/step-8#:~:text=or%20with%20the%20retries%20option%3A

abuaboud (changes requested):

packages/server/worker/src/lib/flow-worker.ts
```diff
@@ -1,19 +1,50 @@
...
 let heartbeatInterval: NodeJS.Timeout
 let socket: Socket
 
-export const flowWorker = (log: FastifyBaseLogger) => ({
+export const runsMetadataQueue = runsMetadataQueueFactory({ 
+    createRedisConnection: workerRedisConnections.create,
+    distributedStore: workerDistributedStore,
+})
+
+export const appSocket = (log: FastifyBaseLogger) => ({
```
abuaboud: I don't we should fail silently, also why is this needed (i see it's only used in one places, can it be moved there)?

packages/engine/src/main.ts
```diff
@@ -8,27 +11,59 @@ import {
...
+export function sendToWorker(type: EngineSocketEvent, data: unknown): void {
+    socket?.emit(type, data)
+}
+
+export const sendToWorkerWithAck = async (
+    type: EngineSocketEvent,
+    data: unknown,
+): Promise<void> => {
+    try {
+        await emitWithAck(socket, type, data, {
```
abuaboud: Do not catch it let it throw and make sure we have generic try catch in upstream to handle it and inline this function

packages/engine/src/lib/services/worker.service.ts
```diff
@@ -0,0 +1,11 @@
+import { EngineSocketEvent, SendFlowResponseRequest, UpdateRunProgressRequest } from '@activepieces/shared'
+import { sendToWorkerWithAck } from '../../main'
+
+export const workerService = {
+    async updateRunProgress(request: UpdateRunProgressRequest): Promise<void> {
+        await sendToWorkerWithAck(EngineSocketEvent.UPDATE_RUN_PROGRESS, request)
+    },
+    async sendFlowResponse(request: SendFlowResponseRequest): Promise<void> {
+        await sendToWorkerWithAck(EngineSocketEvent.SEND_FLOW_RESPONSE, request)
```
abuaboud: can't sendToWorkerWithAck move there?

packages/engine/src/main.ts
```diff
@@ -8,27 +11,59 @@ import {
...
 const WORKER_ID = process.env.WORKER_ID
-const WS_URL = 'ws://127.0.0.1:12345/worker/ws'
+const WS_URL = 'http://127.0.0.1:12345'
 
 
 process.title = `engine-${WORKER_ID}`
-let socket: WebSocket | undefined
+let socket: Socket | undefined
 
+export function sendToWorker(type: EngineSocketEvent, data: unknown): void {
```
abuaboud: I think this is unused

abuaboud (changes requested):

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -125,25 +125,18 @@ const sendUpdateRunRequest = async (updateParams: UpdateStepProgressParams): Pro
...
-            'Content-Type': 'application/json',
-            Authorization: `Bearer ${engineConstants.engineToken}`,
-        },
-        retryDelay: 4000,
-        retries: 3,
-        body: JSON.stringify(request),
-    })
+const sendProgressUpdate = async (request: UpdateRunProgressRequest): Promise<void> => {
+    try {
+        await workerService.updateRunProgress(request)
```
abuaboud: Check utils.tryCatcAndThrowEngineError it can be useful there, espcially if we make progress update error an generic engine error

PR #9925, opened 2025-10-30
fix(e2e-test): install required dotenv package explicitly

amrdb:
### Relevant User Scenarios

Fixes # (issue)


abuaboud:
Added it there https://github.com/activepieces/activepieces/pull/9931

PR #9951, opened 2025-11-02
refactor: enhance engine error handling using tryCatch utility function

hazemadelkhalel:
Fixes # (issue)

### DIFF
--- a/packages/pieces/community/jira-cloud/package.json
+++ b/packages/pieces/community/jira-cloud/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-jira-cloud",
-  "version": "0.1.17",
+  "version": "0.1.18",
   "dependencies": {
     "@atlaskit/adf-schema": "50.4.0",
     "@atlaskit/editor-json-transformer": "8.27.2",
--- a/packages/pieces/community/jira-cloud/src/lib/actions/create-issue.ts
+++ b/packages/pieces/community/jira-cloud/src/lib/actions/create-issue.ts
@@ -6,11 +6,27 @@ import {
 	issueTypeIdProp,
 	createPropertyDefinition,
 	transformCustomFields,
+	isFieldAdfCompatible,
 } from '../common/props';
 import { jiraApiCall, jiraPaginatedApiCall } from '../common';
 import { IssueFieldMetaData, VALID_CUSTOM_FIELD_TYPES } from '../common/types';
 import { HttpMethod } from '@activepieces/pieces-common';
 import { isNil } from '@activepieces/shared';
+
+async function getFields(auth: JiraAuth, projectId: string, issueTypeId: string): Promise<IssueFieldMetaData[]> {
+	const fields = await jiraPaginatedApiCall<IssueFieldMetaData, 'fields'>({
+		auth: auth,
+		method: HttpMethod.GET,
+		resourceUri: `/issue/createmeta/${projectId}/issuetypes/${issueTypeId}`,
+		propertyName: 'fields',
+	});
+
+	if (!fields || !Array.isArray(fields)) {
+		return [];
+	}
+
+	return fields;
+}
 
 export const createIssueAction = createAction({
 	name: 'create_issue',
@@ -32,14 +48,10 @@ export const createIssueAction = createAction({
 				const props: DynamicPropsValue = {};
 
 				const authValue = auth as JiraAuth;
-				const fields = await jiraPaginatedApiCall<IssueFieldMetaData, 'fields'>({
-					auth: authValue,
-					method: HttpMethod.GET,
-					resourceUri: `/issue/createmeta/${projectId}/issuetypes/${issueTypeId}`,
-					propertyName: 'fields',
-				});
+				const projectIdValue =  projectId as unknown as string;
+				const issueTypeIdValue =  issueTypeId as unknown as string;
 
-				if (!fields || !Array.isArray(fields)) return {};
+				const fields = await getFields(authValue, projectIdValue, issueTypeIdValue);
 
 				for (const field of fields) {
 					// skip invalid custom fields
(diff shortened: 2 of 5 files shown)


packages/engine/src/main.ts
```diff
@@ -23,7 +24,12 @@ process.title = `engine-${WORKER_ID}`
 let socket: WebSocket | undefined
 
 async function executeFromSocket(operation: EngineOperation, operationType: EngineOperationType): Promise<void> {
-    const result = await execute(operationType, operation)
+    const { data: result, error: resultError } = await tryCatch(execute(operationType, operation))
```
abuaboud: If we don't have to catch it then just do not use the function

packages/engine/src/lib/handler/code-executor.ts
```diff
@@ -34,29 +37,11 @@ const executeAction: ActionHandler<CodeAction> = async ({ action, executionState
         type: FlowActionType.CODE,
         status: StepOutputStatus.RUNNING,
     })
+    
+    const { data: executionStateResult, error: executionStateError } = await tryCatch(prepareAndExecuteCodeAction({ action, executionState, constants, stepOutput, resolvedInput, stepStartTime }))
```
abuaboud: I think you can inline function then you do need a type for the param

packages/engine/src/lib/handler/piece-executor.ts
```diff
@@ -38,159 +41,172 @@ const executeAction: ActionHandler<PieceAction> = async ({ action, executionStat
...
 
-    try {
-        assertNotNullOrUndefined(action.settings.actionName, 'actionName')
-        const { pieceAction, piece } = await pieceLoader.getPieceAndActionOrThrow({
-            pieceName: action.settings.pieceName,
-            pieceVersion: action.settings.pieceVersion,
-            actionName: action.settings.actionName,
-            pieceSource: constants.piecesSource,
-        })
+    const { data: executionStateResult, error: executionStateError } = await tryCatch(prepareAndExecutePieceAction({ action, executionState, constants, stepOutput, stepStartTime }))
```
abuaboud: There is no indicator here that tryCatch actually throw the engine error.

abuaboud: Try to inline prepareAndExecutePieceAction, it might look better and no need for function

packages/engine/src/lib/helper/execution-errors.ts
```diff
@@ -96,3 +96,15 @@ export class FetchError extends ExecutionError {
         super('FetchError', formatMessage(`Failed to fetch from ${url}`), ExecutionErrorType.ENGINE, cause)
     }
 }
+
+export class AssertionError extends ExecutionError {
```
abuaboud: Can we have Something like EngineGenricError(message) and has type ENGINE, instead of two errors

packages/engine/src/lib/helper/piece-helper.ts
```diff
@@ -35,72 +37,10 @@ export const pieceHelper = {
...
-                    return {
-                        type: PropertyType.DROPDOWN,
-                        options,
-                    }
-                }
-            }
-        }
-        catch (e) {
-            console.error(e)
+        const { data: executePropsResult, error: executePropsError } = await tryCatch(prepareAndExecuteProps({ property, params, executionState, constants, searchValue }))
```
abuaboud: Inline this function

packages/engine/src/lib/helper/trigger-helper.ts
```diff
@@ -21,7 +23,7 @@ type Listener = {
 export const triggerHelper = {
     async executeOnStart(trigger: FlowTrigger, constants: EngineConstants, payload: unknown) {
         const { pieceName, pieceVersion, triggerName, input, propertySettings } = (trigger as PieceTrigger).settings
-        assertNotNullOrUndefined(triggerName, 'triggerName is required')
+        assertEngineNotNullOrUndefined(triggerName, 'triggerName is required')
```
abuaboud: The naming is wrong because what we are checking is not the engine wether it's null or undefiend, I would remove assertEngineNotNullORUndefined and just make it if(isNil(triggerNam){ throw new GenericEngineError)

packages/engine/src/lib/helper/trigger-helper.ts
```diff
@@ -139,94 +141,102 @@ export const triggerHelper = {
...
-                    }
-                    if (!params.webhookSecret) {
-                        throw new Error(`Webhook secret is not available for piece name ${pieceName}`)
-                    }
 
-                    try {
-                        const verified = piece.events?.verify({
+                    const { data: verified, error: verifiedError } = await tryCatch((async () => {
+                        if (!params.appWebhookUrl) {
+                            throw new Error(`App webhook url is not available for piece name ${pieceName}`)
```
abuaboud: I think these must be a EngineGenricError (of type engine) instead of normal error

packages/engine/src/lib/helper/try-catch.ts
```diff
@@ -0,0 +1,28 @@
+import { ExecutionError, ExecutionErrorType } from './execution-errors'
```
abuaboud: Remove this file, and in each function should use just normal shared tryCatch let's have if (isEngine(error)) {
 throw new error
}

packages/engine/src/lib/variables/processors/date-time.ts
```diff
@@ -1,14 +1,17 @@
...
 import { ProcessorFn } from './types'
 
 export const dateTimeProcessor: ProcessorFn = (_property, value) => {
     dayjs.extend(utc)
     dayjs.extend(timezone)
     const dateTimeString = value
     try {
-        if (!dateTimeString) throw Error('Undefined input')
+        if (!dateTimeString) {
+            throw new EngineGenericError('DateTimeStringUndefinedError', 'Date time string is undefined')
```
abuaboud: Revert this, it doesn't have to be EngineGenricError, we just catch it and return undefined.

packages/engine/src/lib/utils.ts
```diff
@@ -17,36 +19,31 @@ export const utils = {
...
+                    await walkRecursive(fullPath)
                 }
             }
-            catch (error) {
-                // Skip directories that can't be read
-            }
         }
 
-        await walkRecursive(dirPath)
+        await tryCatchAndThrowEngineError(walkRecursive(dirPath))
```
abuaboud: Why return data and error? I think we are skipping it right?

packages/engine/src/lib/utils.ts
```diff
@@ -17,36 +19,31 @@ export const utils = {
             return JSON.parse(file)
         }
         catch (e) {
-            throw Error((e as Error).message)
+            throw new EngineGenericError('ParseJsonFileError', `Failed to parse JSON file: ${filePath}`)
```
abuaboud: I think the whole file parseJsonFile is unused, It should be deleted.

packages/engine/src/lib/helper/try-catch.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+  
+  type Failure<E> = {
+      data: null
+      error: E
+  }
+  
+  type Result<T, E = Error> = Success<T> | Failure<E>
+  
+export async function tryCatchAndThrowEngineError<T, E = Error>(
```
abuaboud: If you are going with this approach, delete this file, move this function to utils and use tryCatch (in shared) and check if the returned error is engineError throw it.

packages/engine/src/lib/handler/piece-executor.ts
```diff
@@ -38,28 +39,31 @@ const executeAction: ActionHandler<PieceAction> = async ({ action, executionStat
...
         })
-
+    
         stepOutput.input = censoredInput
-
+    
         const { processedInput, errors } = await propsProcessor.applyProcessorsAndValidators(resolvedInput, pieceAction.props, piece.auth, pieceAction.requireAuth, action.settings.propertySettings)
         if (Object.keys(errors).length > 0) {
-            throw new Error(JSON.stringify(errors, null, 2))
+            throw new EngineGenericError('PieceActionExecutionError', `Piece action execution error: ${JSON.stringify(errors, null, 2)}`)
```
abuaboud: This is wrong, that happens when the user enters invalid validation, we watch this error in the piece and consider it user error.

And we depend on the string be json because we display this inside the piece step.

PR #9986, opened 2025-11-05
feat: full refactor for installing pieces 

MrChaker:
This PR includes:
- replacing pnpm with bun for installing pieces
- replacing pnpm tsc with bun build for compiling code
- remove the pieces file mode and just override whats defined in the dev pieces.
- major refactor

### DIFF
--- a/packages/server/worker/src/lib/consume/job-queue-worker.ts
+++ b/packages/server/worker/src/lib/consume/job-queue-worker.ts
@@ -93,16 +93,16 @@ export const jobQueueWorker = (log: FastifyBaseLogger) => ({
                 )
             }
         },
-            {
-                connection: await workerRedisConnections.create(),
-                telemetry: isOtpEnabled
-                    ? new BullMQOtel(QueueName.WORKER_JOBS)
-                    : undefined,
-                concurrency: workerMachine.getSettings().WORKER_CONCURRENCY,
-                autorun: true,
-                stalledInterval: 30000,
-                maxStalledCount: 5,
-            },
+        {
+            connection: await workerRedisConnections.create(),
+            telemetry: isOtpEnabled
+                ? new BullMQOtel(QueueName.WORKER_JOBS)
+                : undefined,
+            concurrency: workerMachine.getSettings().WORKER_CONCURRENCY,
+            autorun: true,
+            stalledInterval: 30000,
+            maxStalledCount: 5,
+        },
         )
         await worker.waitUntilReady()
         log.info({


packages/server/worker/src/lib/compute/process/factory/isolate-sandbox-process.ts
```diff
@@ -70,11 +69,15 @@ function getEnvironmentVariables(env: Record<string, string | undefined>, worker
 
 async function getDirsToBindArgs(flowVersionId: string | undefined, customPiecesPath: string, reusable: boolean): Promise<string[]> {
     const etcDir = path.resolve('./packages/server/api/src/assets/etc/')
+    const basePath = path.resolve(__dirname.split('/dist')[0])
 
     const dirsToBind = [
         '--dir=/usr/bin/',
         `--dir=/etc/=${etcDir}`,
         `--dir=/root=${path.resolve(GLOBAL_CACHE_COMMON_PATH)}`,
+        `--dir=${path.join(basePath, '.pnpm')}=/${path.join(basePath, '.pnpm')}:maybe`,
```
abuaboud: Can we do that only if there is dev pieces > 0, as security prevention

packages/server/worker/src/lib/cache/pieces/development/local-piece-manager.ts
```diff
@@ -1,33 +1,51 @@
...
 import { FastifyBaseLogger } from 'fastify'
+import { PieceManager } from '..'
 import { workerMachine } from '../../../utils/machine'
 import { cacheState, NO_SAVE_GUARD } from '../../cache-state'
 import { packageManager } from '../../package-manager'
 import { CacheState } from '../../worker-cache'
-import { PieceManager } from '../piece-manager'
 
-export class LocalPieceManager extends PieceManager {
+export const localPieceManager = (log: FastifyBaseLogger): PieceManager => ({
```
abuaboud: Do we still need for that after installing directly in the piece-builder?

packages/server/api/src/app/pieces/piece-sync-service.ts
```diff
@@ -39,10 +34,6 @@ export const pieceSyncService = (log: FastifyBaseLogger) => ({
         })
     },
     async sync(): Promise<void> {
-        if (syncMode !== PieceSyncMode.OFFICIAL_AUTO) {
```
abuaboud, line 42: Revert this and above line I think we still need for piece sync mode, as some users might not want to sync pieces from the cloud.

packages/engine/src/lib/helper/piece-loader.ts
```diff
@@ -119,17 +119,15 @@ export const pieceLoader = {
         })
     },
 
-    getPiecePath: async ({ packageName, pieceSource }: GetPiecePathParams): Promise<string> => {
+    getPiecePath: async ({ packageName, devPieces }: GetPiecePathParams): Promise<string> => {
         let piecePath = null
```
abuaboud: I think this can be inlined 

const piecePath = firstCondition ? firstPath : secondPath

packages/server/worker/src/lib/cache/pieces/development/local-piece-manager.ts
MrChaker: Yes . this actually adds node_modules in dist folder of the piece . otherwise engine will fail to execute pieces . we could do some post-build installation but I think this is better because it only installs when flow runs

packages/server/api/src/app/pieces/piece-sync-service.ts
```diff
@@ -10,19 +10,15 @@ import { system } from '../helper/system/system'
...
+import { PieceMetadataEntity } from './metadata/piece-metadata-entity'
+import { pieceMetadataService } from './metadata/piece-metadata-service'
 
 const CLOUD_API_URL = 'https://cloud.activepieces.com/api/v1/pieces'
 const piecesRepo = repoFactory(PieceMetadataEntity)
 const syncMode = system.get<PieceSyncMode>(AppSystemProp.PIECES_SYNC_MODE)
 
 export const pieceSyncService = (log: FastifyBaseLogger) => ({
     async setup(): Promise<void> {
-        if (syncMode !== PieceSyncMode.OFFICIAL_AUTO) {
```
abuaboud: Revert

abuaboud (dismissed):
LGTM

abuaboud (approved):

PR #9989, opened 2025-11-05
feat: allow optional name arg for build piece command

Dania02525:
## What does this PR do?

This allows passing the `name` argument in order to bypass the prompt for the `build-piece` command in the CLI.  The argument is optional; if not provided, the prompt will request the piece name.

Build a piece in a non-interactive way:
```bash
npm run build-piece my-piece-name
```
Build a piece using the interactive prompt as before:
```bash
npm run build-piece
```

### Relevant User Scenarios

This is useful for automating the piece build process in things like GitHub Actions.  The piece can be built non-interactively and then installed on a server through the Activepieces API

Fixes # (issue)


abuaboud:
@allcontributors please add @Dania02525 for code

abuaboud (approved):

PR #9991, opened 2025-11-05
chore: change the security chain for different authn and authz handlers

hazemadelkhalel:
## What does this PR do?

Fixes # (issue)

### DIFF
--- a/packages/ee/shared/src/lib/project-members/project-member-request.ts
+++ b/packages/ee/shared/src/lib/project-members/project-member-request.ts
@@ -23,6 +23,13 @@ export type AcceptProjectResponse = Static<typeof AcceptProjectResponse>
 
 export const UpdateProjectMemberRoleRequestBody = Type.Object({
     role: Type.String(),
+    projectId: Type.String(),
 })
 
 export type UpdateProjectMemberRoleRequestBody = Static<typeof UpdateProjectMemberRoleRequestBody>
+
+export const GetCurrentProjectMemberRoleRequestQuery = Type.Object({
+    projectId: Type.String(),
+})
+
+export type GetCurrentProjectMemberRoleRequestQuery = Static<typeof GetCurrentProjectMemberRoleRequestQuery>
--- a/packages/react-ui/src/features/flows/lib/flows-api.tsx
+++ b/packages/react-ui/src/features/flows/lib/flows-api.tsx
@@ -4,6 +4,7 @@ import { Button } from '@/components/ui/button';
 import { toast, UNSAVED_CHANGES_TOAST } from '@/components/ui/use-toast';
 import { api } from '@/lib/api';
 import {
+  CountFlowsRequest,
   CreateFlowRequest,
   ErrorCode,
   FlowOperationRequest,
@@ -84,7 +85,7 @@ export const flowsApi = {
   delete(flowId: string) {
     return api.delete<void>(`/v1/flows/${flowId}`);
   },
-  count() {
-    return api.get<number>('/v1/flows/count');
+  count(request: CountFlowsRequest) {
+    return api.get<number>('/v1/flows/count', request);
   },
 };
--- a/packages/react-ui/src/lib/authentication-api.ts
+++ b/packages/react-ui/src/lib/authentication-api.ts
@@ -1,6 +1,7 @@
 import { api } from '@/lib/api';
 import {
   CreateOtpRequestBody,
+  GetCurrentProjectMemberRoleRequestQuery,
   ResetPasswordRequestBody,
   VerifyEmailRequestBody,
 } from '@activepieces/ee-shared';
@@ -35,8 +36,8 @@ export const authenticationApi = {
       providerName,
     });
   },
-  getCurrentProjectRole() {
-    return api.get<ProjectRole | null>('/v1/project-members/role');
+  getCurrentProjectRole(request: GetCurrentProjectMemberRoleRequestQuery) {
+    return api.get<ProjectRole | null>('/v1/project-members/role', request);
   },
   claimThirdPartyRequest(request: ClaimTokenRequest) {
     return api.post<AuthenticationResponse>(
--- a/packages/server/api/src/app/app.ts
+++ b/packages/server/api/src/app/app.ts
@@ -85,8 +85,8 @@ import { webhookModule } from './webhooks/webhook-module'
 import { engineResponseWatcher } from './workers/engine-response-watcher'
 import { queueMetricsModule } from './workers/queue/metrics/queue-metrics.module'
 import { migrateQueuesAndRunConsumers, workerModule } from './workers/worker-module'
-import { authenticationMiddleware } from './core/security/authentication-middleware'
-import { authorizationMiddleware } from './core/security/authorization-middleware'
+import { authenticationMiddleware } from './core/security/authn/authentication-middleware'
+import { authorizationMiddleware } from './core/security/authz/authorization-middleware'
 
 export const setupApp = async (app: FastifyInstance): Promise<FastifyInstance> => {
 
--- a/packages/server/api/src/app/authentication/authentication-utils.ts
+++ b/packages/server/api/src/app/authentication/authentication-utils.ts
@@ -187,14 +187,20 @@ export const authenticationUtils = {
     async extractUserIdFromPrincipal(
         principal: UserPrincipal | ServicePrincipal | EnginePrincipal,
     ): Promise<string> {
-        if (principal.type === PrincipalType.USER) {
-            return principal.id
+        switch (principal.type) {
+            case PrincipalType.USER:
+                return principal.id
+            case PrincipalType.SERVICE:{
+                const platform = await platformService.getOneOrThrow(principal.platform.id)
+                return platform.ownerId
+            }
+            case PrincipalType.ENGINE:{
+                const project = await projectService.getOneOrThrow(principal.projectId)
+                return project.ownerId
+            }
+            default:
+                throw new Error(`Unsupported principal type`)
         }
-
-        const projectId = principal.type === PrincipalType.SERVICE ? principal.project.id : principal.projectId
-        // TODO currently it's same as api service, but it's better to get it from api key service, in case we introduced more admin users
-        const project = await projectService.getOneOrThrow(projectId)
-        return project.ownerId
     },
 }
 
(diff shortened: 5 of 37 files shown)


packages/shared/src/lib/common/security/authorization.ts
```diff
@@ -0,0 +1,56 @@
...
+export type PlatformAuthorization = {
+    type: AuthorizationType.PLATFORM
+    allowedPrincipals: (PrincipalType.USER | PrincipalType.ENGINE | PrincipalType.SERVICE)[]
+}
+
+export type ProjectAuthorization = {
+    type: AuthorizationType.PROJECT
+    allowedPrincipals: (PrincipalType.USER | PrincipalType.ENGINE | PrincipalType.SERVICE)[]
+    project: {
+        projectId: (request: unknown) => string
```
abuaboud: Can we make this strongly typed? I think we can pass the request using generics

packages/server/api/src/app/mcp/mcp-server/mcp-server-controller.ts
```diff
@@ -75,16 +79,26 @@ export const mcpServerController: FastifyPluginAsyncTypebox = async (app) => {
...
     config: {
-        allowedPrincipals: [PrincipalType.USER] as const,
-        permissions: [Permission.WRITE_MCP],
+        security: {
+            kind: RouteKind.AUTHENTICATED,
+            authorization: {
+                type: AuthorizationType.PROJECT,
+                allowedPrincipals: [PrincipalType.USER],
+                project: {
+                    projectId: async (request) => {
```
abuaboud: I think we should make request strongly typed without need to define `ProjectAuthRouteOptions<{ Body: CreateMcpRequestBody }> `

If you inline this directlyt into the controller and try to use preHandler function built-in it's strongly typed based on schame you define, why here we need to define the request type.

packages/server/api/src/app/core/security/security-handler.ts
```diff
@@ -14,3 +14,14 @@ export abstract class BaseSecurityHandler implements SecurityHandler {
     protected abstract canHandle(request: FastifyRequest): Promise<boolean>
     protected abstract doHandle(request: FastifyRequest): Promise<void>
 }
+
+export abstract class BaseAuthzHandler<TRequest extends FastifyRequest> implements SecurityHandler {
```
abuaboud: I think this is overkill, we don't need this abstraction and above abstraction at all.

packages/server/api/src/app/core/security/authn/platform-api-key-authn-handler.ts
```diff
@@ -152,7 +151,7 @@ export class PlatformApiKeyAuthnHandler extends BaseSecurityHandler {
     }
 
     private async extractProjectIdFromResource(
```
abuaboud: We decided that we need to remove everything related to project id (Extracting in the api key).

This place should be much simpler after removing all related logic

packages/server/api/src/app/core/security/security-handler.ts
abuaboud: // This logic can be simplified into a single middleware:
//
// if (publicEndpoint) {
//     skip;
// }
//
// if (isApiKey()) {
//     return apiKeyPrincipal;
// }
//
// if (isBearerToken()) {
//     // handle bearer token
// }
//
// return anonymous;
//
// All of this can be done cleanly in one file.

packages/server/api/src/app/core/security/authz/project-authz-handler.ts
```diff
@@ -1,55 +1,56 @@
...
-            routerPath,
-        )
-        return Promise.resolve(requestMatches)
-    }
+import { BaseAuthzHandler } from '../security-handler'
+import { AuthenticatedFastifyRequest } from '../../../../../types/fastify'
+import { projectMemberService } from '../../../ee/projects/project-members/project-member.service'
+import { AuthorizationType, ProjectAuthorization } from '@activepieces/server-shared'
+
+export class ProjectAuthzHandler extends BaseAuthzHandler<ProjectAuthorizedRequest> {
```
abuaboud: I think the authorization also can be simpler, one entry to the code.

export const securityAuthorization= {
    authorize()
}

and it read which type and validate based on that instead of classes and securty chain handler

packages/shared/src/lib/authentication/model/principal.ts
```diff
@@ -25,7 +27,9 @@ export type ServicePrincipal = {
 export type UserPrincipal = {
     id: ApId
     type: PrincipalType.USER
-    projectId: ProjectId
+    project: {
```
abuaboud: This should be removed also from Service Principal too, check my other comment in fastify.d

packages/server/api/types/fastify.d.ts
```diff
@@ -27,8 +29,10 @@ declare module 'fastify' {
...
         RequestType = unknown,
     > {
-        principal: ContextConfig extends { allowedPrincipals: infer P extends readonly PrincipalType[] }
-            ? PrincipalForTypes<P>
+        principal: ContextConfig extends { security: AuthenticatedRoute }
+            ? ContextConfig['security']['authorization'] extends { allowedPrincipals: readonly (infer P extends PrincipalType)[] }
+                ? PrincipalForType<P>
+                : Principal
             : Principal
         rawBody?: string | Buffer
```
abuaboud, line 43: Suggestion:

we should have `project` object here and it should be infered depend if the security authorization type is project otherwise it will be never or undefined.


Reason: 
Check my other comment about side effects of having project.id inside the principal and why it's confusing.

packages/server/api/src/app/core/security/authn/api-key-authn.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    return apiKeyValue
+}
+
+const createPrincipal = (apiKey: ApiKey): Principal => {
+    return {
+        id: apiKey.id,
+        type: PrincipalType.SERVICE,
+        project: {
+            id: 'ANONYMOUS_' + nanoid(),
```
abuaboud: If I read this code, I’d assume that project.id will always be ANONYMOUS when using an API key, which confuse me why it should exists in first place If just random id.

But that’s not true for project authorization routes — in those cases, it will be the actual project ID.

Another issue is that I might use this inside a platform API endpoint and mistakenly believe the project ID is real, when it’s actually a fake one.

packages/server/api/src/app/core/security/authentication-middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+        return
+    }
+
+    if (apiKeyAuthn.isApiKey(request)) {
+        request.principal = await apiKeyAuthn.authenticate(request)
+        return
+    }
+
+    if (bearerTokenAuthn.isBearerToken(request)) {
+        request.principal = await bearerTokenAuthn.authenticate(request)
```
abuaboud: It's unclear to me when I read this what will happen if the token is invalid, there is no indicator that this function throw error, I would imagine it also can be null.

I think what you did in the engine is great candidate here, where this function expect to return authentication error or data (principal), and rename function to something like verify(request)

First I would understand two things:
- That function is actually verifying the principal.
- In casse it fails It return an error that we will the row again.

packages/server/api/src/app/core/security/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        if (isNil(projectId)) {
+            throw new ActivepiecesError({
+                code: ErrorCode.AUTHORIZATION,
+                params: {
+                    message: 'project id is not available for this route',
+                },
+            })
+        }
+        return projectId
```
abuaboud: This return is unused, The code is more readable and would have same behaviour if you just remove `projectId`  and do early return and remove the check and always throw at the end.

packages/server/api/src/app/core/security/utils.ts
```diff
@@ -0,0 +1,53 @@
+import { AuthorizationType, RouteKind, ProjectResourceType } from "@activepieces/server-shared"
+import { ActivepiecesError, ErrorCode, isNil } from "@activepieces/shared"
+import { FastifyRequest } from "fastify"
+import { requestUtils } from "../request/request-utils"
+
+export const securityUtils = {
+    async getProjectIdFromRequest(request: FastifyRequest): Promise<string> {
+        const security = request.routeOptions.config?.security
```
abuaboud: The check if it's project authorization should be done where you decide the authorization, so coming to this function the code already know there is project authorization.

This can be enforced instead of sending request: FastifyRequest, you could send the `request.authorization.project` directly intoi the parameter 

Then the function would be small enough and only used in one place that check the project authorization, so if it's only used in one place, maybe it's good to move it there.

packages/server/api/src/app/ee/helper/project-authorization.ts
```diff
@@ -0,0 +1,19 @@
+import { ActivepiecesError, ErrorCode, ProjectId, UserId } from "@activepieces/shared";
+import { projectMemberService } from "../projects/project-members/project-member.service";
+import { FastifyBaseLogger } from "fastify";
+
+export const projectMustBeAccessibleByCurrentUser = async (projectId: ProjectId, userId: UserId, log: FastifyBaseLogger): Promise<void> => {
```
abuaboud, line 5: I don't think that's correct, because there is platform admin which is not part of the project members, I am not sure how we check it today in our code base so maybe re use the same logic.

packages/server/api/src/app/core/security/authorization-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+export const authorizationMiddleware = async (request: FastifyRequest): Promise<void> => {
+    const security = request.routeOptions.config?.security
+    
+    if (security.kind === RouteKind.PUBLIC) {
+        return
+    }
+
+    await principalTypeAuthz.authorize(request, security)
+    
+    if (security.authorization.type === AuthorizationType.PROJECT && (request.principal.type === PrincipalType.USER || request.principal.type === PrincipalType.SERVICE)) {
```
abuaboud: Why do we have to check for request.princpal.type , isn't it already enforced at the type level when authorizationType is project?

packages/server/api/src/app/core/security/authorization-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+import { PrincipalType } from '@activepieces/shared'
+
+export const authorizationMiddleware = async (request: FastifyRequest): Promise<void> => {
+    const security = request.routeOptions.config?.security
+    
+    if (security.kind === RouteKind.PUBLIC) {
+        return
+    }
+
+    await principalTypeAuthz.authorize(request, security)
```
abuaboud: In case of failed authorization, I can't tell wether this will throw error or return a boolean that we forgot to check it.

PR #9993, opened 2025-11-06
feat(slack): allow multi channel filter for new message trigger

AdamSelene:
## What does this PR do?

- Allow selecting 0, 1 or more channels to filter new messages (in the new message trigger)
- Deprecate the "new message in channel" trigger

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/slack/package.json
+++ b/packages/pieces/community/slack/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-slack",
-  "version": "0.10.15",
+  "version": "0.10.16",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


Blightwidow (approved):

kishanprmr:
@AdamSelene Instead of removing the existing one, we can add a multi-channel dropdown to the `New Message in Channel ` trigger. What do you think?

AdamSelene:
@kishanprmr I think it's much simpler to have a single trigger for 1, N, or all channels
we want to avoid "all channels" triggers followed by a branch checking the origin channel 😬

PR #10016, opened 2025-11-09
refactor: active flows addon logic

Louai-Zokerburg:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/react-ui/src/app/components/sidebar/sidebare-usage-limits.tsx
+++ b/packages/react-ui/src/app/components/sidebar/sidebare-usage-limits.tsx
@@ -12,6 +12,7 @@ import { cn, formatUtils } from '@/lib/utils';
 import { ApEdition, ApFlagId, isNil } from '@activepieces/shared';
 
 import { FlagGuard } from '../flag-guard';
+import { platformHooks } from '@/hooks/platform-hooks';
 
 const getTimeUntilNextReset = (nextResetDate: number) => {
   const now = dayjs();
@@ -46,6 +47,7 @@ const getTimeUntilNextReset = (nextResetDate: number) => {
 
 const SidebarUsageLimits = React.memo(() => {
   const { project, isPending } = projectHooks.useCurrentProject();
+  const { platform } = platformHooks.useCurrentPlatform();
 
   const { data: edition } = flagsHooks.useFlag<ApEdition>(ApFlagId.EDITION);
 
@@ -101,8 +103,8 @@ const SidebarUsageLimits = React.memo(() => {
           />
           <UsageProgress
             name={t('Active Flows')}
-            value={12}
-            max={30}
+            value={platform.usage?.activeFlows!}
+            max={platform?.plan.activeFlowsLimit!}
             variant={'success'}
           />
         </div>


packages/server/api/src/app/ee/platform/platform-plan/stripe-helper.ts
```diff
@@ -44,6 +44,95 @@ export const stripeHelper = (log: FastifyBaseLogger) => ({
...
+                
+                    for (let i = 1; i < relevantSchedules.length; i++) {
+                        await stripe.subscriptionSchedules.release(relevantSchedules[i].id)
+                    }
+                }
+                else {
+                    await createSubscriptionSchedule({ stripe, subscription, logger: log, extraActiveFlows, isFreeDowngrade })
+                }
+            }
+            return `/platform/setup/billing/success?action=${isUpgrade ? 'upgrade' : 'downgrade'}`
```
abuaboud: Nit: the backend should not be aware from the frontend routes, It should return just success / error / status what ever the raw information the redirect should be handled in the frontend.

packages/server/api/src/app/ee/platform/platform-plan/platform-plan.controller.ts
```diff
@@ -138,6 +138,43 @@ export const platformPlanController: FastifyPluginAsyncTypebox = async (fastify)
         })
     })
 
+    fastify.post('/create-subscription', CreateSubscriptionRequest, async (request) => {
```
abuaboud: Why this is handled here? shouldn't we just redirect to the stripe and listen for the webhook when subscription is created for that customer id?

packages/server/api/src/app/ee/platform/platform-plan/platform-plan.controller.ts
```diff
@@ -138,6 +138,43 @@ export const platformPlanController: FastifyPluginAsyncTypebox = async (fastify)
...
+        const extraActiveFlows = Math.max(0, newActiveFlowsLimit - baseActiveFlowsLimit)
+
+        return stripeHelper(request.log).startSubscription({
+            platformId: platformPlan.platformId,
+            customerId,
+            extraActiveFlows,
+        })
+    })
+
+    fastify.post('/update-active-flows-limit', UpdateActiveFlowsLimitRequest, async (request) => {
```
abuaboud: Nit: this should be just update for platform_plan object and activeFlows is just in the body.

packages/server/api/src/app/ee/platform/platform-plan/stripe-helper.ts
```diff
@@ -44,6 +44,95 @@ export const stripeHelper = (log: FastifyBaseLogger) => ({
...
+            )
+
+            if (isUpgrade) {
+                for (const schedule of relevantSchedules) {
+                    await stripe.subscriptionSchedules.release(schedule.id)
+                }
+
+                await updateSubscription({ stripe, subscription, extraActiveFlows })
+            }
+            else {
```
abuaboud: Nit: I would use early return inside isUpgrade than else

packages/server/api/src/app/ee/platform/platform-plan/stripe-helper.ts
```diff
@@ -44,6 +44,95 @@ export const stripeHelper = (log: FastifyBaseLogger) => ({
...
+                if (relevantSchedules.length > 0) {
```
abuaboud: I don't understand what this condition means, I would rename it to something like that.

`const hasScheduledUpdate = ....`

then use the variable.

packages/server/api/src/app/ee/platform/platform-plan/platform-plan.controller.ts
Louai-Zokerburg: This is what we are basically doing, this start subscription route redirected to checkout session that creates a subscription and we listen for that even in the webhooks route

If u think that the route is badly named tell me i would call it, start-checkout-session instead of current name

packages/server/api/src/app/ee/platform/platform-plan/platform-plan.controller.ts
Louai-Zokerburg: I wouldn't do that coz this routes isnt only to update platform_plan instead this is a special route that directly effects the subscription, I would rename it to 'update-active-flows-addon'_

PR #10039, opened 2025-11-12
feat: replace pnpm & tsc with bun 

MrChaker:


packages/server/worker/src/lib/cache/pieces/production/registry-piece-manager.ts
```diff
@@ -179,3 +179,35 @@ const getArchiveAndSaveToDisk = async (
     await writeFile(archivePath, piece.archive as Buffer)
 }
 
+createRootPackageJson
```
abuaboud: I think this is mistake

PR #10082, opened 2025-11-16
fix: handle empty/null JSON bodies in webhook endpoints

### LINKED ISSUE #10067: Webhook should able to accept empty body
<br>The 3P provider are sending the application/json with complete empty body, we should be able to accept it even if they don't send {}<br>

akramcodez:
## What does this PR do?

Enables webhooks to accept empty JSON bodies from third-party providers, preventing crashes and processing failures.

Third-party providers sometimes send `Content-Type: application/json` with empty bodies (not even `{}`). This PR:

1. **Overrides Fastify's JSON parser** to handle empty bodies
2. **Treats empty/null/whitespace as `{}`** (empty object)
3. **Fixes critical null-safety bug** (checks null before `.trim()`)
4. **Maintains compatibility** with valid JSON payloads

**Before:** Empty body → crash/failure  
**After:** Empty body → processed as `{}`

### Relevant User Scenarios

- **Event webhooks**: Providers sending notifications without payload data
- **Status updates**: Services with event type in headers, empty body
- **Health checks**: Monitoring pings with JSON content-type, no body

All tested and verified:

- [x] null/undefined/empty/whitespace → `{}`  
- [x] valid JSON → normal parsing  
- [x] malformed JSON → proper error  

Fixes #10067

### DIFF
--- a/docs/developers/piece-reference/triggers/webhook-trigger.mdx
+++ b/docs/developers/piece-reference/triggers/webhook-trigger.mdx
@@ -35,13 +35,3 @@ To make your webhook accessible from the internet, you need to configure the bac
 Once you have completed these configurations, you are good to go!
 
 </Warning>
-
-## JSON Body Handling
-
-Activepieces webhooks handle various content types including `application/json`. For JSON payloads:
-
-- **Valid JSON**: Normal parsing and processing
-- **Empty Body**: Third-party providers that send `Content-Type: application/json` with an empty body are supported and will be processed as an empty object `{}`
-- **Malformed JSON**: Will result in a parsing error and webhook failure
-
-This ensures compatibility with providers that send empty JSON payloads without body content.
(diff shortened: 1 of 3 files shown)


akramcodez:
PTAL @abuaboud, thanks.

docs/developers/piece-reference/triggers/webhook-trigger.mdx
```diff
@@ -35,3 +35,13 @@ To make your webhook accessible from the internet, you need to configure the bac
 Once you have completed these configurations, you are good to go!
 
 </Warning>
+
+## JSON Body Handling
```
amrdb: Can you remove it from docs? seems unrelated.

amrdb:
Hi there,
I moved the parser to the webhook module so it would only accept empty bodies for webhooks and made it return 400 in case of validation error. Also, I removed the part from the docs because it's unrelated.

Thanks for your contribution!

PR #10089, opened 2025-11-16
feat(piece-cache): refresh the local pieces cache every 15min

Yasser-Belatreche:
## What does this PR do?

in local pieces cache, it was using the database for cache validation check, now it is using redis first, then the database

- introduce redis in the local pieces list cache management, now intead of calling the db everytime to check if we have a new update, we store the pieces state in redis, and get it from there directly
- update the redis cache everytime a piece is created/updated or deleted
- add an extra redis cache update when we update the local caches array, just in case the service write methods never get called. 

 **NOTE: i assumed that the only places where the pieces are updated are inside the create, updateUsage & bulkDelete method, that's where i call the method to update the redis cache, not sure if there are other places**

### DIFF
--- a/packages/react-ui/src/app/components/project-layout/index.tsx
+++ b/packages/react-ui/src/app/components/project-layout/index.tsx
@@ -70,7 +70,7 @@ export function ProjectDashboardLayout({
             {!hideHeader && (
               <>
                 <ProjectDashboardLayoutHeader />
-                <Separator className="mb-6" />
+                <Separator className="mb-5" />
               </>
             )}
             <div className="px-4"> {children} </div>
--- a/packages/react-ui/src/app/components/project-layout/project-dashboard-layout-header.tsx
+++ b/packages/react-ui/src/app/components/project-layout/project-dashboard-layout-header.tsx
@@ -127,7 +127,7 @@ export const ProjectDashboardLayoutHeader = () => {
     });
 
   return (
-    <div className="flex flex-col px-4 gap-3">
+    <div className="flex flex-col px-4 gap-1">
       <ProjectDashboardPageHeader title={project?.displayName} />
       <Tabs>
         <TabsList variant="outline">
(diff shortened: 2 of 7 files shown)


packages/server/api/src/app/pieces/metadata/local-piece-cache.ts
```diff
@@ -9,45 +6,33 @@ import { PieceMetadataEntity, PieceMetadataSchema } from './piece-metadata-entit
...
-            const updatedRequiredSecondCheck = await requireUpdate()
-            if (!updatedRequiredSecondCheck) {
-                return cache
-            }
-            log.info('Syncing pieces from database')
+
+        return cache
+    },
+
+    async refreshPiecesCache(): Promise<void> {
```
abuaboud: I think this can be written in better way, because using cache inside look like side effect, It's hard to trace cache in a lot of places.

I would write it as:

async fetchLatestPieces(): Promise<PieceMetadata[]>, and assign the cache variable inside setup function from the caller.

abuaboud:
```ts
async function fetchPieces(): Promise<PieceMetadata[]> {
  try {
    const piecesFromDatabase = await repo().find();

    return piecesFromDatabase.sort((a, b) => {
      if (a.name !== b.name) {
        return a.name.localeCompare(b.name);
      }
      return semVer.rcompare(a.version, b.version);
    });
    
  } catch (error) {
    log.error({ error }, 'Error fetching local pieces');
    return [];
  }
}
```

abuaboud: We can also check the database for most recent as it's very fast query, If there is need for update we can update otherwise we can skip the update.

packages/server/api/src/app/pieces/metadata/local-piece-cache.ts
```diff
@@ -9,45 +6,33 @@ import { PieceMetadataEntity, PieceMetadataSchema } from './piece-metadata-entit
...
+        try {
             const result = await repo().find()
```
abuaboud: Nit: We start adapting `Go style` of error handling, where you can use tryCatch (in shared package), It takes the function and return `data` and `error` and if there is error you can just print the error, it look nicer.

abuaboud (dismissed):

Issue #9484, opened 2025-09-30
Improve queue rate limiter


amrdb:

Yasser-Belatreche:
Hello guys, so for this one, the current implementation doesn't enforce the SLA, it only pushes back the throttled jobs to the queue and run them after sometime wishing that the next run will have a space to run

what i propose a different method, without that much changes in the existing one, and i believe it will do what we are looking for

we can introduce a new set in redis for each project, and store the waiting flows in it,
so basically we replace

`
await job.changePriority({
    priority: JOB_PRIORITY[RATE_LIMIT_PRIORITY],
})
`
by
`redis.sadd(waiting_jobs_set:projectId, jobData)`

then inside the `onCompleteOrFailedJob` method, at the end after updating the active jobs set, we can check the waiting jobs set from redis, if it has values, we pull the first one, add it to our old job workers queue (the bullmq one), and finally remove it from the set, the bullmq worker will pick it up again and repeat the same flow

like that we are sure that we always start a new waiting job directly when one is completed, and we will mostly fulfill the SLA

Now there are some challenges:

- if the worker crash before calling the `onCompleteOrFailedJob` then the jobs will stay in the set forever ---> to fix that we can add a cron task that will check the jobs that were in the waiting queue for X amount of time, and push them to bullmq again
- if the worker crash inside the `onCompleteOrFailedJob`, after sending it to bullmq again, and before removing the job from the waiting set, in that case we may send it again by the cron task ---> to fix that we can change the flow a little bit, and inside the worker method, after making sure that we can pass it, we remove it from the redis waiting set if it was there, instead of removing it inside the `onCompleteOrFailedJob` we can even add a boolean to the worker data, `wasWaiting`, and if that's the case we remove it after passing the shouldRateLimit check, not sure about this one though, maybe it's not needed

PR #10095, opened 2025-11-16
feat(worker-job-rate-limiter): change the jobs rate limit mechanism

Yasser-Belatreche:
implemenation of 
https://github.com/activepieces/activepieces/issues/9484#issuecomment-3538968668

### DIFF
--- a/packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
+++ b/packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
@@ -75,7 +75,6 @@ export const workerJobRateLimiter = (log: FastifyBaseLogger) => ({
         const redis = await workerRedisConnections.useExisting()
         const projectId = castedJob.projectId
 
-        await redis.srem(projectActiveSetKey(projectId), jobId)
         const waitingSetKey = projectWaitingSetKey(projectId)
         const waitingSize = await redis.zcard(waitingSetKey)
         if (waitingSize === 0) {
@@ -98,8 +97,15 @@ export const workerJobRateLimiter = (log: FastifyBaseLogger) => ({
                     message: '[workerJobRateLimiter] Promoting next job',
                     jobId: nextJobId,
                 })
+                await redis.multi()
+                    .srem(projectActiveSetKey(projectId), jobId)
+                    .zadd(projectActiveSetKey(projectId), nextJobId)
+                    .zrem(waitingSetKey, nextJobId)
+                    .exec()
+
                 await nextJob.promote()
-                await redis.zrem(waitingSetKey, nextJobId)
+
+
             },
         })
     },
@@ -130,6 +136,11 @@ export const workerJobRateLimiter = (log: FastifyBaseLogger) => ({
                 local setKey = KEYS[1]
                 local jobId = ARGV[1]
                 local maxConcurrent = tonumber(ARGV[2])
+                
+                -- If the job is already in the set, do not rate limit
+                if redis.call('SISMEMBER', setKey, jobId) == 1 then
+                    return 0 -- Job is already counted, should not rate limit
+                end
                 
                 local size = redis.call('SCARD', setKey)
                 if size >= maxConcurrent then


packages/server/api/src/app/workers/queue/throttled-job-queue.ts
```diff
@@ -0,0 +1,107 @@
...
+import { BullMQOtel } from 'bullmq-otel'
+import { FastifyBaseLogger } from 'fastify'
+import { redisConnections } from '../../database/redis-connections'
+import { dedicatedWorkers } from '../../ee/platform/platform-plan/platform-dedicated-workers'
+import { system } from '../../helper/system/system'
+import { AddJobParams, JobType } from './queue-manager'
+
+const dedicatedWorkersQueues = new Map<string, Queue>()
+
+export const throttledJobQueue = (log: FastifyBaseLogger) => ({
```
abuaboud: The dedicated queue is a feature we sell for the enterprise on the cloud, where we completely separate their queue and workers that consume the jobs.


Given that, we don't need a throttle queue for each platform and rather one queue.

I think this file should be deleted and replaced with the following inside the worker rate limiter file.

```ts
   const isOtpEnabled = system.getBoolean(AppSystemProp.OTEL_ENABLED)
            const throttledQueue = new Queue("throttledJobs", {
                telemetry: isOtpEnabled ? new BullMQOtel(queueName) : undefined,
                connection: await redisConnections.create(),
            })
```

And it can be used directly to add.

packages/server/worker/src/lib/consume/job-queue-worker.ts
```diff
@@ -47,21 +48,21 @@ export const jobQueueWorker = (log: FastifyBaseLogger) => ({
...
+                    const redis = await workerRedisConnections.useExisting()
+                    const key = projectWaitingSetKey(job.data.projectId!)
+
+                    const member = JSON.stringify({ id: job.id!, data: job.data })
+                    await redis.zadd(key, Date.now(), member)
+
+                    await throttledJobQueue(log).add({
+                        id: job.id!,
+                        data: job.data,
+                    })
```
abuaboud: This place is designed for jobQueueWorker and this logic is part of workerJobLimits, so this logic should be moved to worker job limiter file, and should be strongly encapsulated.

I imagine this place all it does, acknowledge the job so it's get removed.

packages/server/worker/src/lib/consume/job-queue-worker.ts
```diff
@@ -47,21 +48,21 @@ export const jobQueueWorker = (log: FastifyBaseLogger) => ({
...
-                    const randomFactor = 0.6 + Math.random() * 0.4
-                    const delayInSeconds = Math.round(baseDelay * randomFactor)
-                    await job.moveToDelayed(
-                        dayjs().add(delayInSeconds, 'seconds').valueOf(),
-                        token,
-                    )
+                    const redis = await workerRedisConnections.useExisting()
+                    const key = projectWaitingSetKey(job.data.projectId!)
+
+                    const member = JSON.stringify({ id: job.id!, data: job.data })
```
abuaboud: We already adding the job to the throttledQueue, there is no need to store all the data, we should just store the `jobId` and when we pop it out we can fetch the data from the queue and insert it in the main queue.

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -22,26 +25,46 @@ export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
...
-            `,
-        1,
-        setKey,
-        jobId,
+            
+            return 1
+                `,
+            1,
+            setKey,
+            jobId,
```
abuaboud: Lua script are bad for redis performance in the general, but for this script It's doing O(N) search, I think projectSetKey should be renamed to projectAcitveJobKeys, where it's a normal set and you just remove the `id`

There is lua script below, we should also remove it, I think we no longer need to expire it, as we are sure this place will get called

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -22,26 +25,46 @@ export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
...
         )
+
+        const projectKey = projectWaitingSetKey(castedJob.projectId)
+        const redis = await workerRedisConnections.useExisting()
+
+        // zpopmin returns the [member, score] tuple
+        const popped = await redis.zpopmin(projectKey)
```
abuaboud: What if we popped the new job id and we failed to add it?

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -1,14 +1,38 @@
...
-const projectSetKey = (projectId: string): string => `active_jobs_set:${projectId}`
+const projectActiveSetKey = (projectId: string): string => `active_jobs_set:${projectId}`
+const projectWaitingSetKey = (projectId: string): string => `waiting_jobs_set:${projectId}`
 
 export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
+    async throttleJob(id: string, data: JobData): Promise<void> {
+        const redis = await workerRedisConnections.useExisting()
+        const key = projectWaitingSetKey(data.projectId!)
+
+        await redis.zadd(key, Date.now(), id)
```
abuaboud: If we are going by the time, It's better to use queue as It's O(1) fore pop, but I would suggest we go by the priority there and time or plain queue.

You choose.

https://redis.io/glossary/redis-queue/

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -19,29 +43,49 @@ export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
...
+        const waitingSetKey = projectWaitingSetKey(castedJob.projectId)
+        const redis = await workerRedisConnections.useExisting()
+
+        const members = await redis.zrange(waitingSetKey, 0, 0)
+        const nextJobId = members[0] as string | undefined
+        if (!nextJobId) {
+            return
+        }
+
+        const nextJobData = await throttledJobQueue(_log).getJobById(nextJobId)
```
abuaboud: Nit: make getJobById returns the Job instead of just the data, also copy the priority from there.

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
```diff
@@ -19,29 +43,49 @@ export const workerJobRateLimiter = (_log: FastifyBaseLogger) => ({
...
+                `,
+            1,
+            activeSetKey,
+            jobId,
         )
+
+        const waitingSetKey = projectWaitingSetKey(castedJob.projectId)
+        const redis = await workerRedisConnections.useExisting()
+
+        const members = await redis.zrange(waitingSetKey, 0, 0)
```
abuaboud: This is wrong, because we have race condition there, the job might be added twice

abuaboud (changes requested):

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
Yasser-Belatreche: i ended up keeping the sorted sets, the only problem with using the queues is that they are not unique, so to avoid any duplication issue for any reason of the reasons, i kept the set
Even though i don't think the duplicated jobs in the waiting set will happens, but i prefer to stay on the safe area, the set per project will be small anyways so O(log n) will pretty much be equal to O(1)

packages/server/worker/src/lib/consume/worker-job-rate-limiter.ts
Yasser-Belatreche: i have added another sorted set named it pre-requeue-jobs, this one is global, and has items like projectId:jobId
and i have added a cron task that runs every minute and check for stalled jobs, i have used a 1 minute window to decide if the job is stalled, i guess it should enough

PR #10246, opened 2025-11-26
feat: display trigger publish errors

MrChaker:


packages/engine/src/lib/worker-socket.ts
```diff
@@ -12,7 +12,7 @@ import {
     isNil,
 } from '@activepieces/shared'
 import { io, type Socket } from 'socket.io-client'
-import { EngineGenericError } from './helper/execution-errors'
+import { EngineGenericError } from '../../../shared/src/lib/engine/execution-errors'
```
abuaboud: I think this should be @activepieces/shared

packages/engine/tsconfig.lib.json
```diff
@@ -5,6 +5,6 @@
     "declaration": true,
     "types": ["node"]
   },
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*.ts", "../shared/src/lib/engine/execution-errors.ts"],
```
abuaboud: revert

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -603,6 +529,54 @@ export const flowService = (log: FastifyBaseLogger) => ({
...
+                data: {
+                    ...data,
+                    preUpdateDone: false,
+                },
+                jobId: `update-flow-status-${data.id}`,
+                
+            },
+            schedule: {
+                type: 'one-time',
+                date: apDayjs().add(1, 'second'),
```
abuaboud: why not instant?

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -287,20 +286,26 @@ export const flowService = (log: FastifyBaseLogger) => ({
...
                     }
 
                     case FlowOperationType.CHANGE_STATUS: {
-                        await this.updateStatus({
+                        await this.addUpdateStatusJob({
                             id,
                             projectId,
                             newStatus: operation.request.status,
                         })
+                        await flowRepo().update(id, {
```
abuaboud: This should happen before the adding to the queue as there could be a race condition here, ideally I would recommend this to be inside the queue job.

Same for above

packages/react-ui/src/features/flows/lib/flows-utils.tsx
```diff
@@ -29,9 +29,43 @@ const zipFlows = async (flows: PopulatedFlow[]) => {
...
+    typeof executionError.message !== 'string' ||
+    typeof executionError.name !== 'string'
+  ) {
+    return null;
+  }
+  if (executionError.type !== ExecutionErrorType.USER) return null;
+
+  return executionError;
+};
+
```
abuaboud: I would just make the output of the engine a more structured way, maybe we can ask the engine to output the code and params or certain format the error and we can just use it directly without this parsing logic.

packages/react-ui/src/features/flows/lib/flows-hooks.ts
```diff
@@ -228,4 +240,40 @@ export const flowsHooks = {
...
+          });
+        }
+        callback(FlowOperationStatus.NONE, updatedFlow?.status);
+        socket.off(WebsocketClientEvent.FLOW_STATUS_UPDATED, onUpdateFinish);
+      };
+      socket.on(WebsocketClientEvent.FLOW_STATUS_UPDATED, onUpdateFinish);
+    };
+
+    return updateStatusListener;
+  },
```
abuaboud: Having three nested function inside each is unreadable, this can be written in much better way.

packages/server/api/src/app/flows/flow/flow.service.ts
MrChaker: I;ll add them above . because if I add them inside the job . ui won't show updating status after this request ends

packages/react-ui/src/features/flows/lib/flows-utils.tsx
MrChaker: I did refactor this part a bit 
but still the engine puts error in standardError as string . I think if we want to change this we do it in another scope

PR #10255, opened 2025-11-27
Meenu lekhapremakumar/q 1763 pdf support password encrypted files

meenulekha-premakumar:
## What does this PR do?

Adds a new `Remove Password` action to the PDF piece that allows users to unlock password-protected PDF files and save them without encryption

It uses `pdf-lib` to load the encrypted PDF with the provided password and saves it without encryption. The resulting PDF can be used without password restrictions.

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/pdf/package.json
+++ b/packages/pieces/community/pdf/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-pdf",
-  "version": "0.3.0",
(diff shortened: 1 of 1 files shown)


sanket-a11y:
@meenulekha-premakumar  [pdf-lib](https://www.npmjs.com/package/pdf-lib) does not currently support encrypted documents.

PR #10260, opened 2025-11-27
Fix/handle agent updates

Louai-Zokerburg:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/react-ui/src/features/flow-runs/lib/flow-runs-api.ts
+++ b/packages/react-ui/src/features/flow-runs/lib/flow-runs-api.ts
@@ -80,10 +80,17 @@ export const flowRunsApi = {
         socket.off('error', handleError);
         reject(error);
       };
+
       socket.on(WebsocketClientEvent.TEST_STEP_FINISHED, handleStepFinished);
       socket.on('error', handleError);
+
       if (onProgress) {
-        socket.on(WebsocketClientEvent.TEST_STEP_PROGRESS, onProgress);
+        const handleOnProgress = (response: StepRunResponse) => {
+          if (response.runId === stepRun.id) {
+            onProgress(response);
+          }
+        }
+        socket.on(WebsocketClientEvent.TEST_STEP_PROGRESS, handleOnProgress);
       }
     });
   },


jyroscoped:
*   I would recommend making the `agentResult` prop required in `AgentTestStep` — however, this creates a breaking change since the component is now receiving `agentResult` as potentially `null` from the parent. You should either handle the null case in the parent before passing it, or keep the prop optional and add a null check inside the component.
*   The state management refactoring in `TestSampleDataViewer` moves `agentResult` to a prop, but you're passing `liveAgentResult ?? (sampleData as AgentResult)` which could cause type issues. If `sampleData` is not an `AgentResult`, this cast could lead to runtime errors. Consider adding explicit type guards or validation.
*   In `test-action-section.tsx`, the `isRunAgent` function is duplicated — it already exists in `test-sample-data-viewer.tsx`. Extract this to a shared utility file to avoid duplication and potential inconsistencies if the logic needs to change.
*   The refactoring in `flow-runs-api.ts` simplifies the type unions, but the old `isForTodo` check is replaced with checking `onProgress` directly. This means you're now relying on the absence/presence of a callback rather than explicit type information. Consider keeping both pieces of information or creating a more explicit discriminator union if this pattern expands in the future.
*   In `test-webhook-dialog.tsx`, you're explicitly setting `onProgress: undefined`. This is unnecessary — just omit the property entirely to let TypeScript's strict checks work more effectively. Explicitly setting undefined can mask bugs where the property should actually have a callback.
*   The socket event listener cleanup in `AgentTimeline` was removed, which is good for simplification. However, make sure the parent component (`TestSampleDataViewer`) is properly managing the lifecycle of `liveAgentResult`. If multiple test runs happen in sequence, you need to ensure the state is reset or updated correctly between runs.

jyroscoped:
*   I would recommend making the `agentResult` prop required in `AgentTestStep` — however, this creates a breaking change since the component is now receiving `agentResult` as potentially `null` from the parent. You should either handle the null case in the parent before passing it, or keep the prop optional and add a null check inside the component.
*   The state management refactoring in `TestSampleDataViewer` moves `agentResult` to a prop, but you're passing `liveAgentResult ?? (sampleData as AgentResult)` which could cause type issues. If `sampleData` is not an `AgentResult`, this cast could lead to runtime errors. Consider adding explicit type guards or validation.
*   In `test-action-section.tsx`, the `isRunAgent` function is duplicated — it already exists in `test-sample-data-viewer.tsx`. Extract this to a shared utility file to avoid duplication and potential inconsistencies if the logic needs to change.
*   The refactoring in `flow-runs-api.ts` simplifies the type unions, but the old `isForTodo` check is replaced with checking `onProgress` directly. This means you're now relying on the absence/presence of a callback rather than explicit type information. Consider keeping both pieces of information or creating a more explicit discriminator union if this pattern expands in the future.
*   In `test-webhook-dialog.tsx`, you're explicitly setting `onProgress: undefined`. This is unnecessary — just omit the property entirely to let TypeScript's strict checks work more effectively. Explicitly setting undefined can mask bugs where the property should actually have a callback.
*   The socket event listener cleanup in `AgentTimeline` was removed, which is good for simplification. However, make sure the parent component (`TestSampleDataViewer`) is properly managing the lifecycle of `liveAgentResult`. If multiple test runs happen in sequence, you need to ensure the state is reset or updated correctly between runs.

Louai-Zokerburg:
> * I would recommend making the `agentResult` prop required in `AgentTestStep` — however, this creates a breaking change since the component is now receiving `agentResult` as potentially `null` from the parent. You should either handle the null case in the parent before passing it, or keep the prop optional and add a null check inside the component.
> * The state management refactoring in `TestSampleDataViewer` moves `agentResult` to a prop, but you're passing `liveAgentResult ?? (sampleData as AgentResult)` which could cause type issues. If `sampleData` is not an `AgentResult`, this cast could lead to runtime errors. Consider adding explicit type guards or validation.
> * In `test-action-section.tsx`, the `isRunAgent` function is duplicated — it already exists in `test-sample-data-viewer.tsx`. Extract this to a shared utility file to avoid duplication and potential inconsistencies if the logic needs to change.
> * The refactoring in `flow-runs-api.ts` simplifies the type unions, but the old `isForTodo` check is replaced with checking `onProgress` directly. This means you're now relying on the absence/presence of a callback rather than explicit type information. Consider keeping both pieces of information or creating a more explicit discriminator union if this pattern expands in the future.
> * In `test-webhook-dialog.tsx`, you're explicitly setting `onProgress: undefined`. This is unnecessary — just omit the property entirely to let TypeScript's strict checks work more effectively. Explicitly setting undefined can mask bugs where the property should actually have a callback.
> * The socket event listener cleanup in `AgentTimeline` was removed, which is good for simplification. However, make sure the parent component (`TestSampleDataViewer`) is properly managing the lifecycle of `liveAgentResult`. If multiple test runs happen in sequence, you need to ensure the state is reset or updated correctly between runs.

Hello there, I just wanna tell u that this pr is just a experimental PR between me and my colleague to share some knowledge and this one won't be  pushed to prod, the changes in this branch will be refactored and pushed to prod in an other PR. thank you very much for you excellent points definitely helped me in the new PR

PR #10294, opened 2025-12-02
fix: persist run step expansion state to prevent collapse

### LINKED ISSUE #9165: Loop Items collapse during runs progress

akramcodez:
Fixes #9165

The Issue Loop items in the run details sidebar were collapsing automatically whenever the flow run updated. This happened because the expansion state was stored in local useState, which reset every time the component re-rendered due to flow progress updates.

The Fix I moved the expansion state to the global BuilderState context.
- Added expandedSteps to the global store to persist visibility across re-renders.
- Updated setRun to preserve this state when new run data arrives.
- Updated RunStepCardItem to read/write from this global state instead of local state.

Result Users can now expand or collapse loop items while a flow is running, and the UI will maintain that state even as new steps complete.

### DIFF
--- a/packages/react-ui/src/app/builder/builder-hooks.ts
+++ b/packages/react-ui/src/app/builder/builder-hooks.ts
@@ -99,6 +99,7 @@ export type BuilderState = {
   outputSampleData: Record<string, unknown>;
   inputSampleData: Record<string, unknown>;
   loopsIndexes: Record<string, number>;
+  expandedSteps: Record<string, boolean>;
   run: FlowRun | null;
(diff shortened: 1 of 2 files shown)


akramcodez:
PTAL @abuaboud, thanks.

abuaboud:
Hi @akramcodez 

Sorry for late reply, this interface will be removed in couple of weeks, so i am closing the PR and the ticket.

PR #10309, opened 2025-12-03
feat(ai-proxy): add the needed new endpoints, add the models calls

Yasser-Belatreche:
Still WIP

### DIFF
--- a/packages/server/api/src/app/ai/providers/ai-provider.ts
+++ b/packages/server/api/src/app/ai/providers/ai-provider.ts
@@ -1,12 +1,6 @@
-import { AIProviderConfig } from '@activepieces/common-ai'
+import { AIProviderConfig, AIProviderModel } from '@activepieces/common-ai'
 
 export type AIProviderStrategy<T extends AIProviderConfig> = {
     name(): string;
-    listModels(config: T): Promise<ProviderModel[]>;
-}
-
-export type ProviderModel = {
-    id: string;
-    name: string;
-    type: 'image' | 'text';
+    listModels(config: T): Promise<AIProviderModel[]>;
 }
--- a/packages/server/api/src/app/ai/providers/azure-provider.ts
+++ b/packages/server/api/src/app/ai/providers/azure-provider.ts
@@ -1,14 +1,13 @@
-import { AIProviderStrategy, ProviderModel } from './ai-provider';
+import { AIProviderStrategy } from './ai-provider';
 import { httpClient, HttpMethod } from '@activepieces/pieces-common';
-import { AzureProviderConfig } from '@activepieces/common-ai';
-
+import { AzureProviderConfig, AIProviderModel, AIProviderModelType } from '@activepieces/common-ai';
 
 export const azureProvider: AIProviderStrategy<AzureProviderConfig> = {
     name() {
         return 'Azure OpenAI';
     },
 
-    async listModels(config: AzureProviderConfig): Promise<ProviderModel[]> {
+    async listModels(config: AzureProviderConfig): Promise<AIProviderModel[]> {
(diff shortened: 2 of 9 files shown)


packages/server/api/src/app/ai/providers/ai-provider.ts
```diff
@@ -0,0 +1,19 @@
...
+import { FastifyRequest, RawServerBase, RequestGenericInterface } from 'fastify'
+
+
+export type AIProviderStrategy<T extends object> = {
+    name(): string;
+    listModels(config: T): Promise<ProviderModel[]>;
+    authHeaders(config: T): Record<string, string>;
+}
+
+export type ConfigSchema<T> = { attribute: keyof T; label: string; type: 'string' | 'number' }[];
```
abuaboud: Please Remove

packages/server/api/src/app/ai/providers/ai-provider.ts
```diff
@@ -0,0 +1,19 @@
+import { AIProvider, AIUsageMetadata } from '@activepieces/common-ai'
+import { SeekPage } from '@activepieces/shared'
+import { Type } from '@sinclair/typebox';
+import { FastifyRequest, RawServerBase, RequestGenericInterface } from 'fastify'
+
+
+export type AIProviderStrategy<T extends object> = {
+    name(): string;
+    listModels(config: T): Promise<ProviderModel[]>;
+    authHeaders(config: T): Record<string, string>;
```
abuaboud: I don't think we need that as that will be specified by common ai, please remove authHeader to remove any friction in adding ai providers

packages/server/api/src/app/ai/providers/azure-provider.ts
```diff
@@ -0,0 +1,47 @@
+import { ActivepiecesError, ErrorCode } from '@activepieces/shared';
+import { AIProviderStrategy, ProviderModel } from './ai-provider';
+import { httpClient, HttpMethod } from '@activepieces/pieces-common';
+
+export type AzureProviderConfig = {
```
abuaboud: I would move this to shared, where the CreateProviderRequest is there and this should be typebox object and you should infer it's type

packages/server/api/src/app/ai/providers/index.ts
```diff
@@ -1,15 +1,24 @@
...
+import { anthropicProvider, AnthropicProviderConfig } from './anthropic-provider'
+import { googleProvider, GoogleProviderConfig } from './google-provider'
+import { openaiProvider, OpenAIProviderConfig } from './openai-provider'
+import { AIProviderStrategy } from './ai-provider'
+import { openRouterProvider, OpenRouterProviderConfig } from './openrouter-provider'
+import { azureProvider, AzureProviderConfig } from './azure-provider'
 
-export const aiProvidersStrategies: Record<string, AIProviderStrategy> = {
+export type AiProviderConfig = {
+    openai: OpenAIProviderConfig;
```
abuaboud: I think the keys here should be enum and define them in the shared

packages/server/api/src/app/ai/providers/index.ts
```diff
@@ -1,15 +1,24 @@
...
+    anthropic: AnthropicProviderConfig;
+    openrouter: OpenRouterProviderConfig;
+    azure: AzureProviderConfig;
+    google: GoogleProviderConfig;
+}
+
+export const aiProviders: Record<string, AIProviderStrategy<any>> = {
```
abuaboud: You should not use any and inside the AIProviderStrategy

You should the generic type that's union of all configs (Check the comment about creating typebox object for each provider and infer it type) as the default value.

abuaboud (changes requested):

PR #10332, opened 2025-12-04
fix: added functionality for file inputs in http form data

### LINKED ISSUE #8702: Support Sending Files in HTTP Piece
The user wants to send a PDF file (available through a URL) to an external server using a form submission.

---

### **Current Behavior**

* The form data currently supports only simple key/value pairs.
* It does **not** allow sending files as values.
* As a result, it's not possible to send the PDF file directly in the form.

---

### **Expected Behavior**

* The form data should support sending **files** (e.g., PDF) as values, not just text.
* This would allow the user to attach and send a PDF file (retrieved via URL) to the external server as part of the form submission.

Daniel-Klippa:
## What does this PR do?

Added new functionality to the Object prop to support files as values and send these using the http form data

### Relevant User Scenarios

Fixes #8702


kishanprmr:
Closing this PR as file input support has already been added in PR #10479.

PR #9795, opened 2025-10-22
feat(piece): adds couchbase piece

chedim:
## What does this PR do?

This PR adds a piece to query a Couchbase cluster using Query Service API. It supports basic CRUD operations, as well as running SQL++ and Vector queries using Composite Vector Indexes. 

### Explain How the Feature Works

https://drive.google.com/file/d/17aYcOJlsQRH-JxqKVyIpLeKPUD4XEoxD/view?usp=drive_link

### Relevant User Scenarios

This feature would be valuable to users that may want to fetch and store data on a Couchbase cluster.

Thank you!

PR #10353, opened 2025-12-06
feat: add couchbase piece

onyedikachi-david:
What does this PR do?
This PR adds support for Couchbase cluster using Query Service API.

This is a maintainer copy of PR https://github.com/activepieces/activepieces/pull/9795 by @chedim.

All credit goes to the original contributor.
Fixes https://github.com/activepieces/activepieces/pull/9795


chedim:
@onyedikachi-david Thank you for merging the PR, but this completely re-wrote the PR to use the SDK client instead of Rest API, any specific reason for that? The original PR used HTTP API to avoid external dependencies as well as to prevent the piece from loading bucket maps and off-load query routing to query service or our new [stellar gateway](https://github.com/couchbase/stellar-gateway) available as [Data API](https://docs.couchbase.com/cloud/data-api-reference/index.html#section/Introduction) on Capella so that the piece itself is as lightweight as possible.

onyedikachi-david:
> @onyedikachi-david Thank you for merging the PR, but this completely re-wrote the PR to use the SDK client instead of Rest API, any specific reason for that? 

For maintenance purpose, if there is an official SDK to use for a piece, its preferred, most especially if it is typescript based.

> and off-load query routing to query service or our new [stellar gateway](https://github.com/couchbase/stellar-gateway) available as [Data API](https://docs.couchbase.com/cloud/data-api-reference/index.html#section/Introduction)

The SDK doesn't do this? 

@chedim

chedim:
@onyedikachi-david 
The SDK does exactly that, but it also makes it... a bit large :) 
In our practice we ran into issues, for example, when using it in lambdas or similar deployments because maintaining vbucket map inside the SDK requires additional resources that may be needed for other dependencies.

Besides that, establishing a connection using the SDK is a pretty hefty operation and doing it every time a CB piece is used *will* affect both piece and cluster performance. 

All this comes from our "master-less" cluster architechture (well, one may say that the SDK acts as master node).

Data API addresses both of these issues — it doesn't need to maintain vbuckets or connect/disconnect to the cluster because it uses REST APIs on cluster nodes, pushing all that logic onto the cluster itself. That's why since Data API became available we recommend using it instead of the SDK in projects that cannot cache cluster connections and may need to  connect/reconnect to the cluster at a high rate (we may potentially implement a lightweight SDK for this at some point). If I understand AP architecture correctly, that's exactly how pieces work and that's why the original PR was implemented via HTTP. 

Maybe it would make sense for me to prepare a PR that adds HTTP pieces separately from SDK pieces? That way, people who use self-hosted clusters and don't want to use data api can use the SDK pieces and people that, for example, connect to Capella can use HTTP pieces?

Thank you!

PR #10374, opened 2025-12-07
feat(platform-admin): make branding images uploadable instead of url …

MrChaker:


packages/react-ui/src/lib/file-utils.ts
```diff
@@ -0,0 +1,23 @@
...
+    const uint8Array = new Uint8Array((file.data as any).data);
+    let binary = '';
+    for (let i = 0; i < uint8Array.length; i++) {
+      binary += String.fromCharCode(uint8Array[i]);
+    }
+    const base64 = btoa(binary);
+    assertNotNullOrUndefined(file.metadata?.mimetype, 'Mimetype is required');
+    return `data:${file.metadata.mimetype};base64,${base64}`;
+  },
+};
```
abuaboud: This is no longer needed, please delete

abuaboud (changes requested):

packages/server/api/src/app/file/file.controller.ts
```diff
@@ -0,0 +1,93 @@
+import { ActivepiecesError, ALL_PRINCIPAL_TYPES, ErrorCode, FileCompression, isMultipartFile, PrincipalType, UploadFileRequestBody } from '@activepieces/shared'
+import { FastifyPluginAsyncTypebox, Type } from '@fastify/type-provider-typebox'
+import { StatusCodes } from 'http-status-codes'
+import { projectService } from '../project/project-service'
+import { fileService } from './file.service'
+
+export const fileController: FastifyPluginAsyncTypebox = async (app) => {
```
abuaboud: Discussed offline

PR #10377, opened 2025-12-07
fix: csv export flattening logic for single-column sheets

### LINKED ISSUE #10370: Google sheets export CSV produces TSV result
The Google Sheets Export Sheet produces tsv even when export format is set to csv.

To reproduce create a google sheet with rows like so and use the piece in "return as text" mode:
test test
example example

The ending result (when selecting csv) results in tsv instead:

"test test example example"

**Issue occurs only when there is a single header and single column of values**

Athaxv:
## What does this PR do?

This PR fixes an issue in the Google Sheets “Export Sheet” action where exporting a sheet as CSV produces the same TSV format for single-column sheets. The updated logic now correctly converts the response into comma-separated values only for single-column cases, while multi-column CSV and TSV outputs remain unchanged.

CSV output for single-column values
```
{
  "text": "hi,hola,amigos,theek ,ho ,ji",
  "format": "csv"
}
```

CSV output for multi-column values
```
{
  "text": "hi,hola,amigos,theek ,ho ,ji\nhello ji,,,,,",
  "format": "csv"
}
```

Fixes #10370

### DIFF
--- a/packages/pieces/community/google-sheets/src/lib/actions/export-sheet.ts
+++ b/packages/pieces/community/google-sheets/src/lib/actions/export-sheet.ts
@@ -59,7 +59,29 @@ export const exportSheetAction = createAction({
       });
 
       if (returnAsText) {
-        const textData = Buffer.from(response.body).toString('utf-8');
+        let textData = Buffer.from(response.body).toString('utf-8');
+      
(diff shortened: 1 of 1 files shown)


abuaboud:

PR #10420, opened 2025-12-10
feat: add migration for missing errorHandling Options in legacy steps

### LINKED ISSUE #10117: [BUG]: Old steps that don't have "errorHandlingOptions" in their values will fail to update
When you update a step that doesn't have the "errorHandlingOptions" property in its "settings" property the API will return BAD_REQUEST because the schema isn't matching. 

We need to fix this by adding a migration that checks if this property exists or not and add it (make sure it works with old templates too)

akramcodez:
## What does this PR do? 

Adds a migration to fix legacy flows that fail to update due to missing `errorHandlingOptions` property in step settings. The migration automatically backfills all existing flows and templates with safe defaults that preserve original behavior.

When `errorHandlingOptions` was introduced, existing flows didn't have this property, causing BAD_REQUEST errors on update. This migration:

1. Traverses all flow steps (including nested actions)
2. Adds missing `errorHandlingOptions` with defaults: `continueOnFailure=false`, `retryOnFailure=false`
3. Processes both flow versions and templates
4. Runs idempotently on server startup

**Result:** Legacy flows can now be updated without validation errors.

### Relevant User Scenarios 

1. **User editing old flows** - Can now modify legacy flows without BAD_REQUEST errors
2. **Using old templates** - Templates created before the feature work seamlessly
3. **API integrations** - Programmatic flow updates no longer fail validation
4. **Enterprise migrations** - Large batches of legacy flows automatically fixed

**Performance:** Processes 100 flows per second, logs progress, completes in 1-30 minutes depending on database size.

Fixes #10117


akramcodez:
PTAL @abuaboud @AbdulTheActivePiecer

PR #10425, opened 2025-12-10
feat(cloudflare-openai-compatible-providers): add 2 extra ai provider -  cloudflare/openai

Yasser-Belatreche:
# Pieces Migrations:
The PR contains a piece migration, it's related to the ai-piece
Basically, for the actions that have a "provider" input, before we were using a slug (eg: openai, google...), now we are using the database unique id for that value, in order for us to be able to support multiple versions of the same provider type
So what the migration is doing, it is checking if the provider value is an old value, i replace it with the id from the "ai_provider" table, otherwise i skip

### DIFF
--- a/packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
+++ b/packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
@@ -1,5 +1,12 @@
 import { t } from 'i18next';
-import { Plus, Pencil, Trash2, ImageIcon, TextIcon, AlertCircle } from 'lucide-react';
+import {
+  Plus,
+  Pencil,
+  Trash2,
+  ImageIcon,
+  TextIcon,
+  AlertCircle,
+} from 'lucide-react';
 import { useState } from 'react';
 import { UseFormReturn, useFieldArray } from 'react-hook-form';
 
@@ -231,7 +238,9 @@ export const UpsertProviderConfigForm = ({
                 <AlertCircle className="h-8 w-8 mx-auto" />
               </span>
               <p className="text-sm text-muted-foreground">
-                {t('This provider does not support listing models via API, please add models manually.')}
+                {t(
+                  'This provider does not support listing models via API, please add models manually.',
+                )}
               </p>
             </div>
           ) : (
--- a/packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
+++ b/packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
@@ -225,34 +225,44 @@ const createFormSchema = (provider: AIProviderName, editMode: boolean) => {
     return Type.Object({
       provider: Type.Literal(AIProviderName.AZURE),
       config: AzureProviderConfig,
-      auth: editMode ? Type.Optional(AzureProviderAuthConfig) : AzureProviderAuthConfig,
+      auth: editMode
+        ? Type.Optional(AzureProviderAuthConfig)
+        : AzureProviderAuthConfig,
     });
   }
   if (provider === AIProviderName.CLOUDFLARE_GATEWAY) {
     return Type.Object({
       provider: Type.Literal(AIProviderName.CLOUDFLARE_GATEWAY),
       config: CloudflareGatewayProviderConfig,
-      auth: editMode ? Type.Optional(CloudflareGatewayProviderAuthConfig) : CloudflareGatewayProviderAuthConfig,
+      auth: editMode
+        ? Type.Optional(CloudflareGatewayProviderAuthConfig)
+        : CloudflareGatewayProviderAuthConfig,
     });
   }
   if (provider === AIProviderName.OPENAI_COMPATIBLE) {
     return Type.Object({
       provider: Type.Literal(AIProviderName.OPENAI_COMPATIBLE),
       config: OpenAICompatibleProviderConfig,
-      auth: editMode ? Type.Optional(OpenAICompatibleProviderAuthConfig) : OpenAICompatibleProviderAuthConfig,
+      auth: editMode
+        ? Type.Optional(OpenAICompatibleProviderAuthConfig)
+        : OpenAICompatibleProviderAuthConfig,
     });
   }
   return Type.Object({
     provider: Type.Literal(provider),
-    auth: editMode ? Type.Optional(Type.Union([
-      AnthropicProviderAuthConfig,
-      GoogleProviderAuthConfig,
-      OpenAIProviderAuthConfig,
-    ])) : Type.Union([
-      AnthropicProviderAuthConfig,
-      GoogleProviderAuthConfig,
-      OpenAIProviderAuthConfig,
-    ]),
+    auth: editMode
+      ? Type.Optional(
+          Type.Union([
+            AnthropicProviderAuthConfig,
+            GoogleProviderAuthConfig,
+            OpenAIProviderAuthConfig,
+          ]),
+        )
+      : Type.Union([
+          AnthropicProviderAuthConfig,
+          GoogleProviderAuthConfig,
+          OpenAIProviderAuthConfig,
+        ]),
     config: Type.Union([
       AnthropicProviderConfig,
       GoogleProviderConfig,


packages/pieces/community/ai/src/index.ts
```diff
@@ -23,7 +23,5 @@ export const ai = createPiece({
   triggers: [],
 });
 
-
-export * from './lib/common/types';
```
AbdulTheActivePiecer, line 27: make sure you bump the patch version of the piece.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
```diff
@@ -0,0 +1,390 @@
...
+            name="config.gatewayId"
+            render={({ field }) => (
+              <FormItem className="grid space-y-3">
+                <FormLabel htmlFor="gatewayId">{t('Gateway ID')}</FormLabel>
+                <FormControl>
+                    <Input
+                      {...field}
+                      required
+                      id="gatewayId"
+                      placeholder={t('your-gateway-id')}
```
AbdulTheActivePiecer: no need to add t('') for placeholder values as they don't differ between languages.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
```diff
@@ -14,187 +14,279 @@ import {
...
+                            />
+                            {p.name}
+                          </div>
+                        </SelectItem>
+                      ))}
+                    </SelectContent>
+                  </Select>
+                  <FormMessage />
+                </FormItem>
+              )} */}
```
AbdulTheActivePiecer: please remove commented code, if we need it we can go back in the file history and get it .

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
```diff
@@ -14,187 +14,279 @@ import {
...
+  defaultDisplayName = '',
 }: UpsertAIProviderDialogProps) => {
   const [open, setOpen] = useState(false);
+  const [selectedProvider, setSelectedProvider] = useState<AIProviderName>(
+    provider
+  );
+
+  const { data: config, isLoading: isLoadingConfig } =
+    aiProviderHooks.useConfig(
+      providerId ?? '',
```
AbdulTheActivePiecer: I am not sure about this, I think the provider should be passed from the outside and have the loading experience done there, what if the provider ID is undefined or is wrong ? will the UI break ? I can't test it atm because on (community edition/dev mode) there is a broken migration 

```
column "displayName" of relation "ai_provider" does not exist 
```

Yasser-Belatreche: it is passed from the outside, but the component support both the insert and update, and before inserting we don't have an id to retrieve the existing config
it won't throw any error since i am disabling the call when the id is empty
let me know what you think

packages/server/api/src/app/ai/ai-provider-controller.ts
```diff
@@ -10,7 +10,13 @@ export const aiProviderController: FastifyPluginAsyncTypebox = async (app) => {
     })
     app.get('/:id/config', GetAIProviderConfig, async (request) => {
         const platformId = request.principal.platform.id
-        return aiProviderService(app.log).getConfig(platformId, request.params.id)
+
+        const config = await aiProviderService(app.log).getConfig(platformId, request.params.id)
+        if (request.principal.type === PrincipalType.USER && config.apiKey) {
```
amrdb: I think it is better to depend on the list endpoint for getting data for the update logic in UI.

packages/server/api/src/app/ai/ai-provider-controller.ts
```diff
@@ -21,6 +27,11 @@ export const aiProviderController: FastifyPluginAsyncTypebox = async (app) => {
         await aiProviderService(app.log).upsert(platformId, request.body)
         return reply.status(StatusCodes.NO_CONTENT).send()
     })
+    app.patch('/:id', UpdateAIProvider, async (request, reply) => {
```
amrdb: We use POST for updates.

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -76,34 +80,74 @@ export const aiProviderService = (log: FastifyBaseLogger) => ({
             })
         }
 
+        // if the user update a part of the config, and he kept the api key unchanged (it will be sent again in this format ****)
+        if (request.config.apiKey.includes('*')) {
```
amrdb: I suggest making the API key field empty in UI, so if it is undefined or empty then the user didn't update it.

This way if the API key is not sent, you can fetch the saved config, merge it with new config and then upsert. Otherwise, upsert directly. (maybe use deepMergeAndCast from shared for future proofing)

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -76,34 +80,74 @@ export const aiProviderService = (log: FastifyBaseLogger) => ({
...
-            platformId,
-            provider: providerId,
-        })
-        return encryptUtils.decryptObject<AIProviderConfig>(aiProvider.config)
+
+        let aiProvider: AIProviderSchema;
+
+        // this first check is for backward compatibilty, because at first we were returning the provider name as the id
+        if (Object.values(AIProviderName).includes(providerId as AIProviderName)) {
+            const aiProvidersFound = await aiProviderRepo().find({
```
amrdb: Do you think we should make another migration to eliminate backward compatibilty code?

For now, I think this can be simplified by using a single findOne like:
```typescript
const isLegacyProviderId = Object.values(AIProviderName).includes(providerId as AIProviderName);

const whereClause = isLegacyProviderId
    ? { platformId, provider: providerId as AIProviderName }
    : { platformId, id: providerId };

const aiProvider = await aiProviderRepo().findOne({
    where: whereClause,
    order: { created: 'ASC' },
});
```

packages/server/api/src/app/database/migration/postgres/1765757655723-add-display-name-to-ai-providers.ts
```diff
@@ -0,0 +1,26 @@
...
+    name = 'AddDisplayNameToAiProviders1765757655723'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            DROP INDEX "public"."idx_ai_provider_platform_id_provider"
+        `);
+        await queryRunner.query(`
+            ALTER TABLE "ai_provider" ADD "displayName" character varying;
+            UPDATE "ai_provider" SET "displayName" = "provider";
+            ALTER TABLE "ai_provider" ALTER COLUMN "displayName" SET NOT NULL;
```
amrdb, line 17: The migration is failing because it is trying to run three queries in one `.query()` call. Separate them.

packages/pieces/community/ai/src/lib/common/ai-sdk.ts
```diff
@@ -1,12 +1,13 @@
 import { anthropic, createAnthropic } from '@ai-sdk/anthropic'
 import { createGoogleGenerativeAI, google } from '@ai-sdk/google'
 import { createOpenAI, openai } from '@ai-sdk/openai'
+import { createOpenAICompatible } from '@ai-sdk/openai-compatible'
```
amrdb, line 4: The package is not installed in package.json.

packages/server/api/src/app/ai/ai-provider-controller.ts
Yasser-Belatreche: that end point is called by the piece as well, so we preferred to return minimal data and have a separate endpoint for the config

packages/server/api/src/app/ai/ai-provider-service.ts
Yasser-Belatreche: yes i think that would be much better

packages/server/api/src/app/flows/flow-version/migrations/migrate-v10-ai-pieces-provider-id.ts
```diff
@@ -0,0 +1,77 @@
...
+            FROM flow f
+            JOIN project p ON p.id = f.projectId    
+            WHERE f.id = $1
+        `, [flowVersion.flowId])
+        const aiProviders = await db.query<{ id: string; provider: string }[]>(`
+            SELECT id, provider
+            FROM ai_provider 
+            WHERE "platformId" = $1 
+            ORDER BY "created" ASC
+        `, [platform.platformId])
```
AbdulTheActivePiecer: please don't use db connection to get the ai providers or platform, instead please use the services for each entity to get the values you need, this is not a db migration, it is a method that gets called in the application layer.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
```diff
@@ -0,0 +1,392 @@
...
+              <Button type="submit">
+                {initialData ? t('Update') : t('Add')}
+              </Button>
+            </div>
+          </form>
+        </div>
+      </PopoverContent>
+    </Popover>
+  );
+};
```
AbdulTheActivePiecer: Please move this to separate file

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
```diff
@@ -0,0 +1,392 @@
...
+        </div>
+      )}
+      </div>
+  )
+}
+
+
+type ModelFormPopoverProps = {
+  initialData?: CloudflareGatewayProviderConfig['models'][0];
+  onSubmit: (model: CloudflareGatewayProviderConfig['models'][0]) => void;
```
AbdulTheActivePiecer: try to not use CloudflareGatewayProviderConfig['models'][0] --> extract the type to a separate type and use it directly.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
```diff
@@ -0,0 +1,407 @@
...
+                  {t('Add your first model')}
+                </Button>
+              </ModelFormPopover>
+            </div>
+          ) : (
+            <div className="space-y-3">
+              {fields.map((field, index) => (
+                <div
+                  key={field.id}
+                  className="flex items-center justify-between p-4 border rounded-lg hover:bg-accent/50 transition-colors"
```
AbdulTheActivePiecer: unclickable items shouldn't have hovering styling.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-config-form.tsx
```diff
@@ -0,0 +1,407 @@
...
+                      onClick={() => handleRemoveModel(index)}
+                      disabled={isLoading}
+                    >
+                      <Trash2 className="h-4 w-4 text-destructive" />
+                      <span className="sr-only">{t('Delete')}</span>
+                    </Button>
+                  </div>
+                </div>
+              ))}
+            </div>
```
AbdulTheActivePiecer, line 257: Extract the card component to its own separate file to make things more readable please.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
```diff
@@ -14,187 +14,246 @@ import {
...
+      } else if (!providerId) {
+        form.reset({
+          provider,
+          displayName: currentProviderDef.name,
+          config: {} as any,
+        });
+        setSelectedProvider(provider);
+      }
+    }
+  }, [open, config, defaultDisplayName, form]);
```
AbdulTheActivePiecer, line 103: Don't use useEffect to reset the value of the form, use the key property on the component itself to make it reconstruct the component when something changes.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
```diff
@@ -14,187 +14,246 @@ import {
...
+                form={form}
+                provider={selectedProvider}
+                apiKeyRequired={!config}
+                isLoading={isPending}
+              />
+
+              {form.formState.errors.root?.serverError && (
+                <FormMessage>
+                  {form.formState.errors.root.serverError.message}
+                </FormMessage>
```
AbdulTheActivePiecer: I don't see us setting the server error message in the onError method in the mutation.

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
```diff
@@ -14,187 +14,246 @@ import {
...
-      provider: Type.Literal(provider),
+      provider: Type.Literal(selectedProvider),
       config: Type.Union([
         AnthropicProviderConfig,
         GoogleProviderConfig,
         OpenAIProviderConfig,
       ]),
     });
-  }, [provider]);
+  }, [selectedProvider]);
```
AbdulTheActivePiecer: please move this to a function outside the component, in the same file, call it createFormSchema and do the logic there, you don't need to use useMemo.

packages/react-ui/src/features/platform-admin/lib/ai-provider-hooks.ts
```diff
@@ -0,0 +1,16 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { aiProviderApi } from './ai-provider-api';
+
+export const aiProviderHooks = {
+  useConfig: (provider: string, enabled: boolean = true) => {
```
AbdulTheActivePiecer: Always use named params instead, like this ```useConfig: ({provider,enabled}: {provider: string, enabled?:boolean})```

packages/react-ui/src/app/routes/platform/setup/ai/universal-pieces/upsert-provider-dialog.tsx
```diff
@@ -14,187 +14,246 @@ import {
...
-  configured,
-  displayName,
-  markdown,
+  providerId,
+  defaultDisplayName = '',
 }: UpsertAIProviderDialogProps) => {
   const [open, setOpen] = useState(false);
+  const [selectedProvider, setSelectedProvider] = useState<AIProviderName>(
+    provider
+  );
```
AbdulTheActivePiecer: this state won't be needed once you use the key property on this component inside its parent, you will be able to directly use the provider prop you pass.

packages/server/api/src/app/flows/flow-version/migrations/migrate-v10-ai-pieces-provider-id.ts
amrdb: Also, this can be a single join query instead of two queries.

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -76,34 +80,58 @@ export const aiProviderService = (log: FastifyBaseLogger) => ({
...
+                platformId,
+                provider: request.provider,
+            })
+            if (record) {
+                const config = await encryptUtils.decryptObject<AIProviderConfig>(record.config)
+                request.config.apiKey = config.apiKey
+            }
+            else {
+                throw new ActivepiecesError({
+                    code: ErrorCode.AI_REQUEST_NOT_SUPPORTED,
```
amrdb: `ErrorCode.VALIDATION` makes more sense here.

packages/server/api/src/app/flows/flow-version/migrations/migrate-v10-ai-pieces-provider-id.ts
```diff
@@ -0,0 +1,73 @@
...
+        const db = databaseConnection()
+
+        const aiProviders = await db.query<{ id: string, provider: string }[]>(`
+            SELECT ap.id, ap.provider
+            FROM flow f
+            JOIN project p ON p.id = f.projectId
+            JOIN ai_provider ap ON ap.platformId = p.platformId
+            WHERE f.id = $1
+            ORDER BY ap.created ASC
+        `, [flowVersion.flowId])
```
amrdb: Let's use a repo instead of raw sql, it is cleaner and less error prone.

abuaboud (approved):

PR #10428, opened 2025-12-10
fix: fixed json processor so invalid json doesn't pass anymore

Daniel-Klippa:
## What does this PR do?

Modified the JSON processor so that invalid JSON values don't pass the validation

The JSON validation didn't work because the value was set to undefined when the JSON wasn't valid. For props that are not required this failed to catch them as they are allowed to be nullable and are optional. By setting the value to false when the JSON isn't valid this won't happen anymore

### Relevant User Scenarios

Fixes # (issue)


abuaboud:
https://github.com/activepieces/activepieces/compare/fix/json?expand=1

PR #10506, opened 2025-12-16
feat(auth): use publicAccess and publicPlatformAccess

MrChaker:


packages/server/api/types/fastify.d.ts
```diff
@@ -33,12 +33,12 @@ declare module 'fastify' {
         ContextConfig extends { allowedPrincipals: infer P extends readonly PrincipalType[] }
             ? PrincipalForTypes<P>
             : ContextConfig['security'] extends { authorization: { type: AuthorizationType.ENGINE } }
-                ? PrincipalForType<PrincipalType.ENGINE>
+                ? PrincipalForTypeV2<PrincipalType.ENGINE>
```
MrChaker: Principal V2 doesn't have projectId . this helps with typing to get errors on endpoints that use principle.projectId

packages/server/api/src/app/ee/flow-template/platform-flow-template.module.ts
```diff
@@ -132,7 +131,7 @@ const ListFlowParams = {
 const DeleteParams = {
     config: {
         allowedPrincipals: [PrincipalType.USER, PrincipalType.SERVICE] as const,
-        scope: EndpointScope.PLATFORM,
+        // security: publicPlatformAccess([PrincipalType.USER, PrincipalType.SERVICE]),
```
abuaboud: Why is this commented?

packages/server/api/src/app/ee/flow-template/platform-flow-template.module.ts
```diff
@@ -105,8 +105,7 @@ async function resolveTemplatesPlatformId(principal: Principal): Promise<string
 
 const GetParams = {
     config: {
-        allowedPrincipals: ALL_PRINCIPAL_TYPES,
-        scope: EndpointScope.PLATFORM,
+        security: publicAccess(),
```
abuaboud: Sadly these are not the same, publicAcess actually doesn't have authenticated, It keeps the prinicpal null.

In older case It does authenticate but doesn't authorize

packages/server/api/src/app/trigger/app-event-routing/app-event-routing.module.ts
```diff
@@ -55,7 +55,7 @@ export const appEventRoutingController: FastifyPluginAsyncTypebox = async (
         {
             config: {
                 rawBody: true,
-                allowedPrincipals: ALL_PRINCIPAL_TYPES,
+                security: publicAccess(),
```
abuaboud, line 58: Same cas here as above, I think there is many more

PR #10567, opened 2025-12-22
fix: optimize canvas rendering to prevent full re-renders

### LINKED ISSUE #10562: [BUG]: if anything is edited in a step all of the canvas gets re-rendered
Because almost every component we have in the builder relies on the flow version one way or another, every time a step is edited we edited the flow version causing all components to re-render.

akramcodez:
### What does this PR do?

Optimizes canvas rendering performance by preventing unnecessary re-renders when editing individual steps in the flow builder.

### The Problem

Previously, editing any single step triggered a complete re-render of the entire canvas because:
- All canvas components subscribed to the full `flowVersion` object
- Any change to `flowVersion` (triggered by step edits) caused all subscribers to re-render
- Components computed derived values using `useMemo` with `flowVersion` as a dependency

This resulted in 50+ component re-renders for every single step modification, causing performance degradation especially in flows with many steps.

### The Solution

Implemented selective state subscription pattern:

1. **Added equality function parameter** to `useBuilderStateContext` hook to support custom comparison logic
2. **Applied shallow comparison** using Zustand's `shallow` helper to all canvas component subscriptions
3. **Moved computed values into selectors** - calculations like `stepIndex`, `isSkipped`, and `isStepValid` now happen inside selectors rather than in components
4. **Extracted only required primitives** - components now receive only the specific primitive values they need instead of entire objects

### Technical Changes

- Modified `useBuilderStateContext` to accept optional `equalityFn` parameter
- Refactored 11 canvas-related components to use shallow equality comparison
- Eliminated `useMemo` hooks that depended on `flowVersion`
- Changed component props from complex objects to primitive values

### Performance Impact

- Reduced re-renders from approximately 50+ to 1-5 per step edit
- Only components directly affected by a change now re-render
- Editing a step in a branch no longer re-renders unrelated branches

### Files Modified

- Core: `builder-hooks.ts`
- Nodes: `step-node/index.tsx`, `step-node-status.tsx`, `big-add-button-node.tsx`
- Edges: `add-button.tsx`, `branch-label.tsx`
- Widgets: `incomplete-settings-widget.tsx`, `test-flow-widget/index.tsx`, `widgets/index.tsx`
- Other: `flow-drag-layer.tsx`, `pieces-selector/index.tsx`

Fixes #10562

### DIFF
--- a/packages/react-ui/src/app/builder/builder-hooks.ts
+++ b/packages/react-ui/src/app/builder/builder-hooks.ts
@@ -67,11 +67,12 @@ export const BuilderStateContext = createContext<BuilderStore | null>(null);
 
 export function useBuilderStateContext<T>(
   selector: (state: BuilderState) => T,
+  equalityFn?: (left: T, right: T) => boolean,
 ): T {
   const store = useContext(BuilderStateContext);
   if (!store)
     throw new Error('Missing BuilderStateContext.Provider in the tree');
-  return useStore(store, selector);
+  return useStore(store, selector, equalityFn);
 }
 
 export enum LeftSideBarType {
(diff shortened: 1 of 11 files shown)


abuaboud:

PR #10600, opened 2025-12-23
feat: added new azure blob storage piece

Daniel-Klippa:
## What does this PR do?

Added a new piece for the Azure Blob Storage feature

This new piece uses the azure blob storage NodeJS SDK to do the following actions:
**Container actions:**
- List Containers
- Create Container
- Delete Container

**Blob Actions:**
- List Blobs
- Create Blob
- Delete Blob
- Read Blob
- Add Tags to Blob
- Find Blobs

### Relevant User Scenarios

Fixes # (issue)


Daniel-Klippa:
I could not figure out how to add the logo, so please help me there

PR #10619, opened 2025-12-25
feat(loginbypass): bypass login

khaledsallam:
### Relevant User Scenarios

Fixes # (issue)


abuaboud:
Hi @khaledsallam 

Thank you for opening PR, I think the code will require maintenance outweighs the problem of logging in the development environment, for example the request will be done always regardless dev / prod environment .

PR #10651, opened 2025-12-29
feat(aicredit new billing): add max auto topup monthly + some UI fixes

Yasser-Belatreche:
### Relevant User Scenarios

Fixes # (issue)


packages/server/api/src/app/ee/platform/platform-plan/platform-ai-credits.service.ts
Louai-Zokerburg, line 189: ` if (creditsRemaining > plan.aiCreditsAutoTopUpThreshold!) {` (line 189

This place here might lead to a broken logic if the `plan.aiCreditsAutoTopUpThreshold` is undefined for what ever reason so i would suggest using an assertion line before the condition and not rely on the ! operator

packages/server/api/src/app/ee/platform/platform-plan/platform-ai-credits.service.ts
```diff
@@ -193,6 +194,21 @@ async function tryAutoTopUpPlan(plan: PlatformPlan, apiKeyHash: string, log: Fas
     assertNotNullOrUndefined(plan.aiCreditsAutoTopUpCreditsToAdd, 'Auto Topup Credits To add is not set')
     assertNotNullOrUndefined(plan.aiCreditsAutoTopUpThreshold, 'Auto Topup Threashold is not set')
 
+    if (!isNil(plan.maxAutoTopUpCreditsMonthly) && plan.maxAutoTopUpCreditsMonthly > 0) {
+        const totalAmountThisMonth = await stripeHelper(log).getAutoTopUpInvoicesTotalThisMonth(plan.stripeCustomerId, plan.platformId)
+        const totalCreditsThisMonth = totalAmountThisMonth * CREDIT_PER_DOLLAR
+
+        if (totalCreditsThisMonth + plan.aiCreditsAutoTopUpCreditsToAdd > plan.maxAutoTopUpCreditsMonthly) {
```
Louai-Zokerburg: please extract the condition in a variable because doing + operation in the if statement is confusing

packages/server/api/src/app/ee/platform/platform-plan/stripe-helper.ts
```diff
@@ -280,6 +280,39 @@ export const stripeHelper = (log: FastifyBaseLogger) => ({
...
+            collection_method: 'charge_automatically',
+            limit: 100,
+        })
+
+        for await (const invoice of invoices) {
+            if (
+                invoice.metadata?.platformId === platformId &&
+                invoice.metadata?.type === StripeCheckoutType.AI_CREDIT_AUTO_TOP_UP
+            ) {
+                totalCents += invoice.total ?? 0
```
Louai-Zokerburg: NIT: use invoice.amount_paid its more precise than total because total can include unpaid amounts + stripe recommends using amount_paid anyways

packages/ee/shared/src/lib/billing/index.ts
```diff
@@ -45,6 +45,7 @@ export const UpdateAICreditsAutoTopUpParamsSchema = Type.Union([
         state: Type.Literal(AiCreditsAutoTopUpState.ENABLED),
         minThreshold: Type.Number(),
         creditsToAdd: Type.Number(),
+        maxMonthlyLimit: Nullable(Type.Number()),
```
Louai-Zokerburg, line 48: we usually use Nullable for Db Sachemas so if u don't have any good reason please use Type.Optional

packages/shared/src/lib/platform/platform.model.ts
```diff
@@ -58,6 +58,7 @@ export const PlatformPlan = Type.Object({
     aiCreditsAutoTopUpState: Type.Enum(AiCreditsAutoTopUpState),
     aiCreditsAutoTopUpThreshold: Type.Optional(Type.Number()),
     aiCreditsAutoTopUpCreditsToAdd: Type.Optional(Type.Number()),
+    maxAutoTopUpCreditsMonthly: Type.Optional(Nullable(Type.Number())),
```
Louai-Zokerburg: this should be only: ` maxAutoTopUpCreditsMonthly: Nullable(Type.Number())`

packages/ee/shared/src/lib/billing/index.ts
Yasser-Belatreche: i am using nullable because i'm sending null from l frontend when the user wants to unset the maxMonthlyLimit, so that typeorm receive null and do the update, if you go for optional, it won't be sent, and the update won't happen

PR #10657, opened 2025-12-30
chore(db): implement healthcheck for db

### LINKED ISSUE #10499: Healtcheck Pass If DB is down
If the application starts successfully and the database goes down later, the health endpoint still reports **healthy** even though the database is unavailable.

The health check should continuously verify database connectivity and report unhealthy when the database is down.

Ridhim-RR:
## What does this PR do?
Fixes #10499

This PR implement the health check to the application’s health endpoint.

**Problem**
Previously, if the application started successfully and the database became unavailable later, the health endpoint would still report the service as healthy.
This happened because database connectivity was not part of the health check logic

**Solution**
The health check now actively verifies database connectivity by executing a lightweight query .
If the database is unavailable or the connection is not initialized, the health endpoint reports an unhealthy status

On both health check requests:
- The application verifies that the database connection is initialized
- Executes a lightweight query (SELECT 1) to confirm connectivity

If the check fails, the health endpoint returns unhealthy

### Relevant User Scenarios

 -Application starts with a healthy database → health endpoint is healthy
-Application is running and database goes down → health endpoint becomes unhealthy
-Database is restored → health endpoint becomes healthy again

The health check should continuously verify database connectivity and report unhealthy when the database is down.

Fixes # (issue)


packages/server/api/src/app/health/health.service.ts
```diff
@@ -10,21 +11,38 @@ export const healthStatusService = (log: FastifyBaseLogger) => ({
...
+            log.warn({ error }, 'Database health check failed')
+            return false
+        }
+    },
+    isHealthy: async (): Promise<boolean> => {
         if (system.isWorker()) {
             return workerHealthStatus
         }
-        return true
+        const databaseHealthy = await healthStatusService(log).checkDatabaseHealth()
```
amrdb: This will not run in case of WORKER_AND_APP mode. I would check both `system.isWorker()` and `system.isApp()` separately in `isHealthy()` and perform the appropriate checks for each mode (e.g. `return workerHealthy && databaseHealthy`).

Ridhim-RR: Looking into this

Ridhim-RR: I have modified the function :

```
 isHealthy: async (): Promise<boolean> => {
        let workerHealthy = true
        let databaseHealthy = true

        if (system.isWorker()) {
            workerHealthy = workerHealthStatus
        }
        if (system.isApp()) {
            databaseHealthy = await healthStatusService(log).checkDatabaseHealth()
        }
        return  workerHealthy && databaseHealthy
    },
```
Now It can handle all the three scenarios.
@amrdb .

amrdb (approved):

amrdb:
Thanks for your contribution!

Issue #10194, opened 2025-11-24
[BUG]: OOM when syncing pieces


lyuboto:
**Describe the bug**
Since 09:00 AM 24 November 2025, starting activepieces containers fails with OOM errors. The OOM event happens right after the `Starting piece synchronization` message.

```
14 | 2025-11-24T11:50:20.710Z | {"level":30,"time":1763985020625,"pid":11,"hostname":"9fa348cf19fb","msg":"Starting piece synchronization"}
-- | -- | --
15 | 2025-11-24T11:50:30.069Z | <--- Last few GCs --->
16 | 2025-11-24T11:50:30.069Z | [11:0x14020fe0] 15619 ms: Mark-Compact 503.0 (514.5) -> 502.3 (514.8) MB, 420.51 / 0.00 ms (average mu = 0.147, current mu = 0.016) allocation failure; scavenge might not succeed
17 | 2025-11-24T11:50:30.069Z | [11:0x14020fe0] 16206 ms: Mark-Compact 503.5 (515.0) -> 502.5 (515.1) MB, 580.98 / 0.00 ms (average mu = 0.072, current mu = 0.010) allocation failure; scavenge might not succeed
18 | 2025-11-24T11:50:30.069Z | <--- JS stacktrace --->
19 | 2025-11-24T11:50:30.069Z | FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
20 | 2025-11-24T11:50:30.069Z | ----- Native stack trace -----
21 | 2025-11-24T11:50:30.098Z | 1: 0xb71e94 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
22 | 2025-11-24T11:50:30.098Z | 2: 0xeaa7dc v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
23 | 2025-11-24T11:50:30.098Z | 3: 0xeaa9ac v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
24 | 2025-11-24T11:50:30.098Z | 4: 0x10b263c [node]
25 | 2025-11-24T11:50:30.098Z | 5: 0x10c8b14 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
[...]

PR #10668, opened 2025-12-30
feat: add piece sync parallel limit env var parameter

lyuboto:
## What does this PR do?
Adds a new environment variable AP_PIECES_SYNC_PARALLEL_LIMIT to configure the number of concurrent requests when syncing pieces from the Activepieces cloud. This helps users with memory-constrained environments avoid OOM (Out of Memory) errors during piece synchronization on startup

When PIECES_SYNC_MODE is set to OFFICIAL_AUTO, the server fetches piece metadata from the cloud in parallel. Previously, this was hardcoded to 20 concurrent requests. This PR makes it configurable via the AP_PIECES_SYNC_PARALLEL_LIMIT environment variable.
- Default value: 20 (preserves existing behavior)
- Usage: Set AP_PIECES_SYNC_PARALLEL_LIMIT=5 to reduce memory usage during sync

### Relevant User Scenarios
- Users running Activepieces on low-memory containers, experiencing OOM crashes during startup when piece sync runs
- Fresh installations where the initial sync needs to download metadata for hundreds of pieces simultaneously
- Users who want faster sync on high-resource systems by increasing the limit

Fixes # (issue)
https://github.com/activepieces/activepieces/issues/10194


abuaboud:
Hi @lyuboto,

Thank you for bringing this up. I think it’s better to install in batches rather than using plimit. We now select only the required columns from the database, and we sync the latest version of each piece first, then fill the gaps for templates, 5 is more than enough now IMO.

https://github.com/activepieces/activepieces/pull/10676

PR #10672, opened 2025-12-30
feat: setup better with with es module build

MrChaker:


packages/engine/src/lib/variables/processors/date-time.ts
```diff
@@ -1,6 +1,6 @@
 import dayjs from 'dayjs'
-import timezone from 'dayjs/plugin/timezone'
-import utc from 'dayjs/plugin/utc'
```
MrChaker, line 3: @abuaboud these import changes to support esm build instead of cjs which doesn't work with better auth new versions

packages/ee/shared/src/lib/authn/enterprise-local-authn/requests.ts
```diff
@@ -1,14 +1,20 @@
-import { ApId, SignUpRequest } from '@activepieces/shared'
```
MrChaker, line 1: @abuaboud otp module is removed some of its needed types are moved heere

packages/server/api/src/app/authentication/better-auth/better-auth-service.ts
```diff
@@ -0,0 +1,43 @@
+import { User } from "better-auth/types";
```
MrChaker: This service used in betterauth options just to be cleaner

packages/server/api/src/app/authentication/user-identity/user-identity-entity.ts
```diff
@@ -1,19 +1,32 @@
...
-export const UserIdentityEntity = new EntitySchema<UserIdentity>({
+type UserIdentitySchema = UserIdentity & {
+    name: string,
+    image?: string,
+}
+
+export const UserIdentityEntity = new EntitySchema<UserIdentitySchema>({
     name: 'user_identity',
     columns: {
-        ...BaseColumnSchemaPart,
```
MrChaker, line 8: changed baseSchema types to match better-auth fields, 
Ideally we would not want a user_identity Entity at all . but we can't because future migrations would try to delete it . and also we need it for typeorm getRepository("user_identity")

PR #10738, opened 2026-01-06
fix: prevent oom during sync

lyuboto:
## What does this PR do?

This PR optimizes the piece synchronization process to prevent "Out of Memory" (OOM) crashes. It removes the forced memory cache refresh that occurred immediately after syncing with the cloud registry, decoupling the "download" phase from the "load into memory" phase.

Currently, the pieces-sync job (which runs hourly) forces a localPieceCache.refresh() immediately after it finishes downloading updates. This is a memory-intensive operation that parses metadata for ~6,000+ pieces.
The issue is that this forced refresh often overlaps with the scheduled 15-minute cron job that also refreshes the cache. When these two heavy operations run concurrently (or in close succession on limited resources), the Node.js heap limit is exceeded, causing the container to crash.

The Fix:
We removed the explicit refresh calls from the sync service. The system will now rely solely on the scheduled localPieceCache cron job (every 15 mins) to load new pieces into memory. This ensures that memory spikes are predictable and non-concurrent.

### Relevant User Scenarios

Users running Activepieces on instances with standard memory limits will no longer experience crashes at round hours (e.g., 12:00, 13:00) when the sync job aligns with the cache refresh cron.

Fixes # (issue)
https://github.com/activepieces/activepieces/issues/10194

### DIFF
--- a/packages/server/api/src/app/pieces/metadata/local-piece-cache.ts
+++ b/packages/server/api/src/app/pieces/metadata/local-piece-cache.ts
@@ -23,7 +23,7 @@ type State = {
 export const localPieceCache = (log: FastifyBaseLogger) => ({
     async setup(): Promise<void> {
         await updateCache(log)
-        cron.schedule('*/15 * * * *', () => {
+        cron.schedule('15,30,45 * * * *', () => {
             log.info('[localPieceCache] Refreshing pieces cache via cron job')
             rejectedPromiseHandler(updateCache(log), log)
         })


abuaboud:
Hi @lyuboto 

I think we should move it to this operation to be on the disk, I can see the total size of metadata is growing to be around 70 mb

```
activepieces=> SELECT
    pg_size_pretty(pg_total_relation_size('piece_metadata')) AS total_size;
 total_size
------------
 69 MB
(1 row)
```

lyuboto:
I have noticed that the OOM crash happens when there has been an update to a piece in the piece registry within the last hour.

I think we should remove the one redundant local-piece-cache refresh, because currently:
- local-piece-cache refresh triggers every 15 minutes - at :00, :15, :30, :45
- piece-sync job triggers every round hour and after it completes, it triggers another local-piece-cache refresh

Meaning that we get 2 local-piece-cache refreshes at around :00.

We can remove the scheduled local-piece-refresh at :00 and leave the one triggered by the piece-sync job. This way we make sure that the local-piece-cache refresh runs after the sync-piece job and we don't have a redundant refresh.

If this doesn't help, we can think of ways to move the operation to work on disk.

lyuboto:
Hey @abuaboud, did you have the chance to look into this? Do you think it could help?

abuaboud:
Hi @lyuboto I am waiting for @amrdb I think he proposed to change it to be based on local sqlite / database instead.

abuaboud:
It should be fixed in https://github.com/activepieces/activepieces/pull/10953

PR #10749, opened 2026-01-06
feat: add admin route for increasing platform ai credits

Louai-Zokerburg:
### Relevant User Scenarios

Fixes # (issue)


packages/server/api/src/app/ee/platform/admin/admin-platform.controller.ts
```diff
@@ -97,6 +101,15 @@ const ApplyLicenseKeyByEmailRequest = {
     },
 }
 
+const IncreaseAICreditsForPlatformRequest = {
+    schema: {
+        body: IncreaseAICreditsForPlatformRequestBody,
+    },
+    config: {
+        security: securityAccess.platformAdminOnly([PrincipalType.USER, PrincipalType.ENGINE, PrincipalType.SERVICE]),
```
abuaboud: This is will allow normal users to use it, match other endpoints in same file

abuaboud (changes requested):

PR #10760, opened 2026-01-07
fix(sftp): don't require password for anonymous auth

Daniel-Klippa:
## What does this PR do?
Removed the password/private key requirement for the sftp piece. This is not always needed in anonymous servers, now users can still create credentials even if their server doesn't require a password/private key

### Relevant User Scenarios

Fixes # (issue)


abuaboud:
Please bump the version of the piece

sanket-a11y:
@Daniel-Klippa 
possible error:  When only privateKey is provided, the code may try to connect twice (inside the condition and again outside), which can cause errors. Connection should happen only once.

Daniel-Klippa:
@sanket-a11y Fixed the double authentication issue

sanket-a11y:
@Daniel-Klippa 
Since SFTP typically requires authentication, if the goal is to support server-specific “anonymous” setups (no password or private key), add an “Anonymous connection” checkbox.
When this checkbox is not selected, we should continue validating that either a password or a private key is provided.

This would help avoid confusion and make the behavior more intentional.

Daniel-Klippa:
@sanket-a11y Changed it to require a checkbox to be checked to allow for anonymous authentication

PR #10768, opened 2026-01-07
feat(ui): added warning to connection delete confirmation dialog (#10758)

### LINKED ISSUE #10758: [Design Debt] Show warning on deleting connection
### **Problem:**

**There is no warning that active flows that using this connection will be broken once the connection is deleted.**

### Suggested Solution

We need to add warning box that mentions

```  
Any flows currently using these connections **will break immediately**. Please proceed with caution.
```

vishakha1411:
## What does this PR do?
This PR adds a clear warning to the delete confirmation dialog when users try to delete connections. Fixes #10758

### Why this approach?
**Instead of hardcoding the warning inside the delete message, the dialog was extended with an optional warning prop. This keeps the component reusable, avoids duplicating warning UI, and ensures a consistent look and behavior wherever delete warnings are needed.**

Screenshot 2026-01-08 015445

### DIFF
--- a/packages/react-ui/src/app/routes/connections/index.tsx
+++ b/packages/react-ui/src/app/routes/connections/index.tsx
@@ -88,7 +88,7 @@ function AppConnectionsPage() {
   const pieceName = searchParams.get('pieceName') ?? undefined;
   const displayName = searchParams.get('displayName') ?? undefined;
 
-  const { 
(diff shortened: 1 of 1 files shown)


AbdulTheActivePiecer:
@allcontributors please add @vishakha1411  for code

PR #10776, opened 2026-01-08
feat: add attachment & data residency support to Sendgrid piece

Thijs-Attenza:
## What does this PR do?

This PR adds attachment support to the Sendgrid piece, and allows the use of [EU data residency](https://www.twilio.com/en-us/blog/products/data-residency-for-email-eu).

### Relevant User Scenarios

- Allows users with a EU account to use this piece without having to create all "Custom API Call" steps.
- Allows users to add attachments to mails send with Sendgrid, both for Regular as Dynamic Template emails.

### Note
This is the first time modifying a Piece for me, so please let me know if there are things that can be done better!

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5899,7 +5899,7 @@
     },
     "packages/pieces/community/smartsheet": {
       "name": "@activepieces/piece-smartsheet",
-      "version": "0.1.4",
+      "version": "0.1.3",
       "dependencies": {
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",


jerbob92:
@sanket-a11y  I see this is merged now, but won't this break auth on existing pieces?

sanket-a11y:
Hey @jerbob92   the minor version bump to `0.5.0` covers this. Since the piece is still under `1.0.0`, a minor bump is treated as a breaking change per the [versioning policy](https://www.activepieces.com/docs/build-pieces/piece-reference/piece-versioning). Existing published flows stay locked to the previous version and won't be affected.

jerbob92:
@sanket-a11y does that also work with piece auth though? What if I make a new flow with an existing auth with the old format?

sanket-a11y:
@jerbob92  In the new flow, you’ll need to create a new auth connection.

sanket-a11y:
@jerbob92  We have released multi-auth support for SendGrid, so the new flow will also work with existing connections.

jerbob92:
@sanket-a11y thank you, much appreciated for making this compatible with existing connections

PR #10935, opened 2026-01-20
feat: send template telemetry events to template manager

hazemadelkhalel:
## What does this PR do?

Integrate the UI with template telematry to collect some data for improvements


abuaboud (dismissed):

packages/server/api/src/app/trigger/trigger-source/trigger-source-service.ts
```diff
@@ -140,6 +151,14 @@ export const triggerSourceService = (log: FastifyBaseLogger) => {
                 projectId,
             })
             log.info('[triggerSourceService#disable] Soft deleted trigger source')
+            if (templateId) {
+                const templateTelemetryEvent: TemplateTelemetryEvent = {
```
abuaboud: inline the const

abuaboud (approved):

PR #11042, opened 2026-01-27
chore: add flow run failed status for log size exeeding limit

MrChaker:


packages/server/engine/src/lib/handler/flow-executor.ts
```diff
@@ -39,6 +40,7 @@ export const flowExecutor = {
         const trigger = input.flowVersion.trigger
         if (input.executionType === ExecutionType.BEGIN) {
             await triggerHelper.executeOnStart(trigger, constants, input.triggerPayload)
+            await checkLogsSizeExceeded(executionState, trigger, constants)
```
abuaboud: Why check here the flow just got started

PR #11046, opened 2026-01-28
feat(shopify): implement OAuth2 authorization

sanket-a11y:
### Relevant User Scenarios

Fixes # (issue)


VishalYadavCF:
Why is this an issue? Does workflow execution does not support version compatibility?
@kishanprmr

PR #11069, opened 2026-01-30
feat(asknews): news action added

sanket-a11y:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/telegram-bot/package.json
+++ b/packages/pieces/community/telegram-bot/package.json
@@ -1,4 +1,4 @@
 {
   "name": "@activepieces/piece-telegram-bot",
-  "version": "0.4.2"
+  "version": "0.5.0"
 }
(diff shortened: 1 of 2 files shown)


packages/pieces/community/asknews/src/lib/common/auth.ts
```diff
@@ -0,0 +1,7 @@
+import { PieceAuth } from '@activepieces/pieces-framework';
+
+export const asknewsAuth = PieceAuth.SecretText({
+  displayName: 'API Key',
+  description: 'Provide your AskNews API key here',
```
kishanprmr: Please add steps to get API key.

packages/pieces/community/asknews/src/lib/common/auth.ts
```diff
@@ -0,0 +1,7 @@
...
+  required: true,
```
kishanprmr, line 14: You can use [Get User Profile](https://docs.asknews.app/en/reference#get-/v1/profiles/me) endpoint for auth validation.

PR #11148, opened 2026-02-06
fix(sftp): password as passphrase, optional algorithm, handle key new…

### LINKED ISSUE #11100: [BUG]: SFTP piece password input is not passed as a paraphrase when a private key is assigned
**Describe the bug**
SFTP piece password input is not passed as a paraphrase when a private key is assigned.

Also please check why the algorithm dropdown is required for SFTP configuration. 
it is not required in the library that we use, https://github.com/mscdex/ssh2#user-content-client-methods (This option allows you to explicitly override the default transport layer algorithms used for the connection), it looks like it was added to support ssh-dss , which is not in the default list, maybe it should not be a required property.

Another issue is with pasting the private key, please check that it works when the user copy pastes it in the connection dialog with endlines, currently users have to manually replace new lines with \n for it to work.

onyedikachi-david:
…lines

Fixes #11100:
1. Pass password as passphrase when private key is provided
2. Make host key algorithm dropdown optional (not required by ssh2)
3. Handle private key newlines - reconstruct PEM format when browser strips newlines

The fix allows users to:
- Use password field as passphrase for encrypted private keys
- Leave algorithm field empty (uses ssh2 defaults)
- Paste private keys with or without newlines (code handles both cases)

Bump version: 0.4.8 → 0.4.9

### DIFF
--- a/packages/pieces/community/sftp/package.json
+++ b/packages/pieces/community/sftp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-sftp",
-  "version": "0.4.9",
+  "version": "0.4.10",
   "dependencies": {
     "basic-ftp": "5.0.5",
     "ssh2-sftp-client": "9.1.0"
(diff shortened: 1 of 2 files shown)


sanket-a11y (dismissed):

jerbob92:
Won't this change cause that the private key is not stored as a secret anymore?

sanket-a11y (approved):

PR #11159, opened 2026-02-07
fix(cli): fix assertPieceExists func in case of missing piece

murex971:
## What does this PR do?

When One tries to create an action with cli command 
`npm run cli actions create` and the piece is not there, error message displays is as follows:

image

The `assertPieceExists` function is modified to handle such case:

image

It shows the correct error message now, the modified ss is pasted above.

### Relevant User Scenarios

Fixes # (issue)


kishanprmr:
Thanks for the contribution! The check is already based on `pieceFolder`, not `pieceName`. Closing for now.

PR #11182, opened 2026-02-09
feat(slack): add reactions:write user scope

AdamSelene:
## What does this PR do?

Add missing scope to add a reaction on behalf of a user

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/slack/package.json
+++ b/packages/pieces/community/slack/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-slack",
-  "version": "0.12.1",
+  "version": "0.12.2",
   "dependencies": {
     "@slack/web-api": "7.9.0",
     "slackify-markdown": "4.4.0"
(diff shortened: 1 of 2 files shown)


sanket-a11y:
@AdamSelene  Since this adds a new scope, we should bump the minor version.

AdamSelene:
@sanket-a11y but it does bump it - or you mean major version, i.e. should bump to 0.13? In which case I see your point but
bumping the major vsn won't solve it: pre-existing connections need to be "reconnected" since they're not versioned (unless I missed something)

onyedikachi-david:
> @sanket-a11y but it does bump it - or you mean major version, i.e. should bump to 0.13? In which case I see your point but bumping the major vsn won't solve it: pre-existing connections need to be "reconnected" since they're not versioned (unless I missed something)

@AdamSelene Yes he meant bumping it to 0.13.0, its a breaking change, since users will need to reconnect to make use of the reaction toggle, also, a new PR has been opened that both PR's you raised into one. Thanks for your contributions once again

AbdulTheActivePiecer:
@AdamSelene you are right, we should add a way for users to know when a connection is stale and needs to be reconnected, I opened this issue https://github.com/activepieces/activepieces/issues/11255, I think we should add something like a SCHEMA_VERSION for OAuth2 connections in pieces, and use that in the connection definition so we can let the user know their connection needs to be reconnected.

kishanprmr:
Thanks for the contribution! Closing this PR due to merge conflicts and lack of edit access. The changes from this PR have been included in #11673.

PR #11196, opened 2026-02-10
feat: sort projects and folders alphabetically

AdamSelene:
## What does this PR do?

Sort projects and folders by displayName

### Relevant User Scenarios

Fixes # (issue)


AbdulTheActivePiecer:
Hi @AdamSelene, I spoke to our product team and we won't be merging this because we intend to keep showing the newest folders/projects at the top like we do in all other places with pagination.

The fix for your issues would be to add sorting to our pagination API. https://github.com/activepieces/activepieces/issues/11715

PR #11201, opened 2026-02-10
fix: pass password as passphrase for SFTP private key authentication

### LINKED ISSUE #11100: [BUG]: SFTP piece password input is not passed as a paraphrase when a private key is assigned

themavik:
## Summary

Fixes #11100

Three issues with the SFTP piece's private key authentication:

### 1. Password not used as passphrase for encrypted private keys

When both `privateKey` and `password` were provided, the code checked `if (password)` first and used password-only authentication, completely ignoring the private key. This made it impossible to use passphrase-protected private keys.

**Fix:** Prioritize private key authentication when a private key is provided. If a password is also present, pass it as the `passphrase` option to decrypt the key.

### 2. Algorithm field unnecessarily required for private key auth

The code threw an error if no algorithm was selected when using private key auth. The ssh2 library has sensible defaults (`ssh-ed25519`, `ecdsa-sha2-nistp256`, etc.) and the algorithm field should only be needed for special cases like `ssh-dss`.

**Fix:** Make algorithm optional — only pass `algorithms.serverHostKey` when the user explicitly selects one or more algorithms.

### 3. Updated password field description

Clarified that when used alongside a private key, the password field serves as the passphrase to decrypt the key.

## Test Plan

- [ ] Connect with password-only authentication (no change in behavior)
- [ ] Connect with private key without passphrase (no algorithm required)
- [ ] Connect with passphrase-protected private key (password used as passphrase)
- [ ] Connect with private key + explicit algorithm override
- [ ] Verify pasted private keys with literal `
` are still converted correctly

### DIFF
--- a/packages/pieces/community/sftp/src/index.ts
+++ b/packages/pieces/community/sftp/src/index.ts
@@ -28,27 +28,31 @@ export async function getClient<T extends Client | FTPClient>(auth: { protocol: 
   if (protocolBackwardCompatibility === 'sftp') {
     const sftp = new Client();
 
-    if (password) {
+    if (privateKey) {
+      // Private key authentication — use password as passphrase if provided
+      const connectOptions: Client.ConnectOptions = {
+        host,
+        port,
+        username,
+        privateKey: privateKey.replace(/\\n/g, '\n').trim(),
+        timeout: 10000,
+      };
+      if (password) {
+        connectOptions.passphrase = password;
+      }
+      if (algorithm && algorithm.length > 0) {
+        connectOptions.algorithms = {
+          serverHostKey: algorithm,
+        } as Client.ConnectOptions['algorithms'];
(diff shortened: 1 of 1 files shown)


themavik:
Recreating with fresh base

PR #11210, opened 2026-02-11
feat(notion): add support for simple access token auth

AdamSelene:
## What does this PR do?

We want to be able to use simple integration tokens instead of OAuth

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/notion/package.json
+++ b/packages/pieces/community/notion/package.json
@@ -1,12 +1,9 @@
 {
   "name": "@activepieces/piece-notion",
-  "version": "0.5.4",
+  "version": "0.6.0",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
-    "@activepieces/pieces-common": "workspace:*",
-    "@activepieces/pieces-framework": "workspace:*",
-    "@activepieces/shared": "workspace:*",
     "@notionhq/client": "2.2.14",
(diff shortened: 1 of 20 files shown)


packages/server/api/src/app/trigger/app-event-routing/app-event-routing.module.ts
```diff
@@ -30,7 +30,7 @@ import { JobType } from '../../workers/queue/queue-manager'
 import { triggerSourceService } from '../trigger-source/trigger-source-service'
 import { appEventRoutingService } from './app-event-routing.service'
 
-const appWebhooks: Record<string, Piece<PieceAuthProperty | undefined>> = {
+const appWebhooks: Record<string, Piece<PieceAuthProperty | PieceAuthProperty[] | undefined>> = {
```
onyedikachi-david: This change is out of scope, please undo.

AdamSelene: Done!

onyedikachi-david:
Lint is failing @AdamSelene, please use this locally: `npx nx lint pieces-notion`

AdamSelene:
> Lint is failing @AdamSelene, please use this locally: `npx nx lint pieces-notion`

@onyedikachi-david should be good now! (but I think `nx` is no longer used?)

PR #11213, opened 2026-02-11
feat(gmail): add service account auth

AdamSelene:
## What does this PR do?
We want to be able to use service accounts (like for `google-sheets`)

### Relevant User Scenarios

Fixes # (issue)


AhmadTash:
We’ve moved the changes from this PR to a new one as some updates were required: #11656

PR #11273, opened 2026-02-16
feat: new automations table instead of (flow + tables) tables

Louai-Zokerburg:
### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/server/api/src/app/database/migration/postgres/1769638834372-add-folder-column-to-table.ts
+++ b/packages/server/api/src/app/database/migration/postgres/1769638834372-add-folder-column-to-table.ts
@@ -8,15 +8,9 @@ export class AddFolderColumnToTable1769638834372 implements MigrationInterface {
             ALTER TABLE "table"
             ADD "folderId" character varying(21)
         `)
-        await queryRunner.query(`
-            CREATE INDEX "idx_cell_record_id" ON "cell" ("recordId")
-        `)
     }
 
     public async down(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`
-            DROP INDEX "public"."idx_cell_record_id"
-        `)
         await queryRunner.query(`
             ALTER TABLE "table" DROP COLUMN "folderId"
         `)


packages/server/api/src/app/database/migration/postgres/1771336996554-DropCellRecordIdIndex.ts
```diff
@@ -0,0 +1,18 @@
...
+        `)
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            CREATE INDEX "idx_cell_record_id" ON "cell" ("recordId")
+        `)
+    }
+
+}
```
AbdulTheActivePiecer: This was already done on main you probably haven't synced in a long while, I updated the branch, please delete this and test just to be extra sure.

packages/server/api/src/app/tables/table/table.service.ts
```diff
@@ -37,16 +40,36 @@ export const recordRepo = repoFactory(RecordEntity)
...
+        return (await flowFolderService(log).upsert({
+            projectId,
+            request: {
+                projectId,
+                displayName: folderName,
+            },
+        })).id
+    }
+    return null
+}
```
AbdulTheActivePiecer: This is duplicated logic from flows service, please ensure you unify them in one method and call it in both places.

packages/server/api/src/app/tables/table/table.service.ts
```diff
@@ -263,17 +300,24 @@ export const tableService = {
...
-        await tableRepo().update({ id, projectId }, { 
+
+        const updateData: Record<string, unknown> = {
             ...spreadIfDefined('name', request.name),
             ...spreadIfDefined('trigger', request.trigger),
             ...spreadIfDefined('status', request.status),
-        })
+        }
+
+        updateData.folderId = request.folderId
```
AbdulTheActivePiecer: why is this not added to the updateData record above and put on a separate line?

packages/react-ui/src/features/pieces/components/piece-icon.tsx
```diff
@@ -43,7 +43,7 @@ const pieceIconVariantsWithPadding = cva('', {
       lg: 'p-2',
       md: 'p-1.75',
       sm: 'p-1.25',
-      xs: '',
+      xs: 'p-[5px]',
```
AbdulTheActivePiecer: can't you use p-1.5 ?

packages/react-ui/src/app/guards/permission-guard.tsx
```diff
@@ -9,10 +9,12 @@ export const RoutePermissionGuard = ({
   children,
 }: {
   children: ReactNode;
-  permission: Permission;
+  permission: Permission | Permission[];
```
AbdulTheActivePiecer: rename this to allowedPermissions so people know this is an any check not an all check.

packages/react-ui/src/app/routes/automations/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+  useEffect(() => {
+    resetPagination();
+  }, [
+    filters.searchTerm,
+    filters.typeFilter,
+    filters.statusFilter,
+    filters.connectionFilter,
+    filters.ownerFilter,
+    resetPagination,
+  ]);
```
AbdulTheActivePiecer: I really discourage using effects for something like this, instead I think we should use the react keys property to reset things, useEffect is really tricky to debug and can lead to unexpected calls, instead if we restructure the code to use keys (split into different components and add keys to them), and every time the key changes we reinitialize the component it would be better.

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+      flows: flowsResult.data,
+      tables: tablesResult.data,
+      flowsNextCursor: flowsResult.next,
+      tablesNextCursor: tablesResult.next,
+    });
+  });
+  return map;
+}
+
+const STALE_TIME = 30_000;
```
AbdulTheActivePiecer: Nit: always put any consts or types or helper functions at the end of the file.

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+  const prevProjectIdRef = useRef(projectId);
+  useEffect(() => {
+    if (prevProjectIdRef.current !== projectId) {
+      prevProjectIdRef.current = projectId;
+      setRootPage(0);
+      setExpandedFolders(new Set());
+      setFolderVisibleCounts(new Map());
+      setLoadingFolders(new Set());
+    }
+  }, [projectId]);
```
AbdulTheActivePiecer: this can be solved with keys just like I mentioned above https://github.com/activepieces/activepieces/pull/11273/changes#r2825985319

if the hook is always called in reinitialized component, it will never need the use effect to be reinitalized

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+        folders.flatMap((folder) => [
+          flowsApi.count({ projectId, folderId: folder.id }),
+          tablesApi.count({ projectId, folderId: folder.id }),
+        ]),
+      );
+      const map = new Map<string, number>();
+      folders.forEach((folder, i) => {
+        map.set(folder.id, results[i * 2] + results[i * 2 + 1]);
+      });
+      return map;
```
AbdulTheActivePiecer: Please don't use this way to extract data from a result, instead use a clearer mapping like this:
 ```
   const folders = foldersQuery.data!;
      const [folderFlowCounts, folderTableCounts] = await Promise.all([
        Promise.all(folders.map(({id}) => flowsApi.count({ projectId, folderId: id }))),
        Promise.all(folders.map(({id}) => tablesApi.count({ projectId, folderId: id }))),
      ]);
      const folderTotalCounts = folderFlowCounts.map((count, index) => count + folderTableCounts[index]);
      return new Map(folderTotalCounts.map((count, index) => [folders[index].id, count]));
```

File: packages/react-ui/src/features/automations/hooks/use-automations-data.ts [hunk 10]
```diff
@@ -0,0 +1,402 @@
...
+    refetchOnMount: 'always',
+  });
+
+  const folderContentsQuery = useQuery<FolderContentsMap>({
+    queryKey: ['all-folder-contents', projectId, folderIds],
+    queryFn: async () => {
+      const folders = foldersQuery.data!;
+      if (folders.length > PARALLEL_FOLDER_THRESHOLD) {
+        return new Map();
+      }
```
AbdulTheActivePiecer: what is this check for ? like you want to close all folders if the user opens more than 50 folders?

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+          }),
+          tablesApi.list({
+            projectId,
+            folderId: folder.id,
+            limit: FOLDER_PAGE_SIZE,
+            cursor: undefined,
+          }),
+        ]),
+      );
+      return buildFolderContentsMap(folders, results);
```
AbdulTheActivePiecer: Same thing as before:
```
      const [flows, tables] = await Promise.all([
        Promise.all(folders.map(({id}) => flowsApi.list({ projectId, folderId: id, limit: FOLDER_PAGE_SIZE, cursor: undefined }))),
        Promise.all(folders.map(({id}) => tablesApi.list({ projectId, folderId: id, limit: FOLDER_PAGE_SIZE, cursor: undefined }))),
      ]);
      const folderResults = folders.map(({id}, index) => ({
        folderId: id,
        flowsResult: flows[index],
        tablesResult: tables[index],
      }));

      return buildFolderContentsMap(folderResults);
```

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+    queryFn: () =>
+      flowsApi.list({
+        projectId,
+        folderId: isFiltered ? undefined : UncategorizedFolderId,
+        limit: 1000,
+        cursor: undefined,
+        name: filters.searchTerm || undefined,
+        status:
+          filters.statusFilter.length > 0
+            ? (filters.statusFilter as any)
```
AbdulTheActivePiecer: Please resolve the typing issue here and do not use ` as any`

packages/web/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+      );
+      return buildFolderContentsMap(folders, results);
+    },
+    enabled: !!foldersQuery.data && foldersQuery.data.length > 0,
+    staleTime: STALE_TIME,
+    refetchOnMount: 'always',
+  });
+
+  const skipFlows =
+    filters.typeFilter.length > 0 && !filters.typeFilter.includes('flow');
```
AbdulTheActivePiecer, line 113: this can be better improved by making the typeFilter: `{ 'flow': boolean, 'table': boolean}` or something similar, it is better than now having a generic type like 'string'

packages/web/src/features/automations/hooks/use-automations-data.ts
```diff
@@ -0,0 +1,402 @@
...
+      });
+
+      setLoadingFolders((prev) => {
+        const next = new Set(prev);
+        next.delete(folderId);
+        return next;
+      });
+    },
+    [folderContentsQuery.data, projectId, folderIds, queryClient],
+  );
```
AbdulTheActivePiecer, line 252: I thought we all agreed that folders don't have so many flows or tables so it is okay to fetch all of the folder content in one go by making the limit for the fetch queries for the flows/tables 9999.

packages/web/src/components/custom/image-with-color-background.tsx
```diff
@@ -35,15 +35,11 @@ const ImageWithColorBackground = ({
...
-      className={cn(
-        'relative inline-block h-full w-full rounded-lg',
-        className,
-        {
-          'bg-background': backgroundColor === null,
-          'border border-border/50 dark:bg-foreground/10':
-            backgroundColor === null && props.border,
-        },
-      )}
+      className={cn('relative inline-block h-full w-full', className, {
```
AbdulTheActivePiecer, line 56: Nit; please check the place this component is used, maybe you should add a flag for rounded-lg instead of fully removing it and affecting existing components.

packages/web/src/features/automations/components/automations-empty-state.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const handleTemplateSelect = (template: Template) => {
+    navigate(`/templates/${template.id}`);
+  };
+
+  const handleViewAllTemplates = () => {
+    navigate('/templates');
+  };
+
+  const topTemplates = templates?.slice(0, 3) || [];
+  const hasTemplates = topTemplates.length > 0;
```
AbdulTheActivePiecer, line 232: topTemplates means these will change whenever the templates change, as in, they aren't static.
If the user has no templates, they will get inifinite loading skeletons, maybe instead, we should copy three templates from our current templates repo as a json, and import them here and use them, just so the result is always same as expected. 

At least when the user has no templates after the fetch query finished.

packages/react-ui/src/features/automations/components/automations-filters.tsx
```diff
@@ -0,0 +1,481 @@
...
+        options.push({
+          value: member.userId,
+          label: `${member.user.firstName} ${member.user.lastName}`,
+        });
+        seenIds.add(member.userId);
+      }
+    });
+
+    return options;
+  })();
```
AbdulTheActivePiecer: why is this a function that is always called in rendering ? why not just create another separate hook and make that hook rely on the projectMembers hooks to fetch them.

packages/web/src/features/automations/components/automations-filters.tsx
```diff
@@ -0,0 +1,481 @@
...
+      (p) => p.name === connection.pieceName,
+    )?.logoUrl;
+    return {
+      value: connection.externalId,
+      label: connection.displayName,
+      icon: pieceIcon ? (
+        <img src={pieceIcon} alt="" className="h-4 w-4 object-contain" />
+      ) : undefined,
+    };
+  });
```
AbdulTheActivePiecer, line 129: Nit (you can do it or open an issue to do later): if used like this, when the user opens the connections filter while the fetch query hasn't finished , the user will see an empty connection dropdown, thinking they have none, we need a loading state, same for other filters.

packages/web/src/features/automations/components/automations-table-row.tsx
```diff
@@ -0,0 +1,220 @@
...
+          }}
+        >
+          <ArrowDown className="h-4 w-4" />
+          <span>
+            {t('Load {count} more items...', { count: item.loadMoreCount })}
+          </span>
+        </div>
+      </div>
+    );
+  }
```
AbdulTheActivePiecer, line 113: same question about the folders limit (no load more in folders as we fetch all flows/tables)

packages/react-ui/src/features/automations/components/automations-table-row.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  const getItemIcon = () => {
+    if (item.type === 'folder') {
+      return <Folder className="h-4 w-4 text-gray-400 fill-gray-400" />;
+    }
+    if (item.type === 'flow') {
+      return <Workflow className="h-4 w-4 text-primary" />;
+    }
+    return <Table2 className="h-4 w-4 text-emerald-500" />;
+  };
```
AbdulTheActivePiecer: make it a separate component called AutomationItemIcon and pass it the item type

packages/react-ui/src/features/automations/components/automations-table-row.tsx
```diff
@@ -0,0 +1,220 @@
...
+            id={flow.ownerId}
+            includeAvatar={true}
+            includeName={true}
+            size="small"
+          />
+        );
+      }
+    }
+    return <span className="text-muted-foreground">-</span>;
+  };
```
AbdulTheActivePiecer: Separate components as well at the end of the file.

packages/react-ui/src/features/automations/components/automations-filters.tsx
```diff
@@ -0,0 +1,481 @@
...
+};
+
+const MultiSelectFilter = ({
+  label,
+  icon,
+  options,
+  selectedValues,
+  onChange,
+  searchable = false,
+}: MultiSelectFilterProps) => {
```
AbdulTheActivePiecer: Put component in a separate file for filters, check packages/react-ui/src/components/ui/data-table

packages/web/src/features/automations/components/import-flow-dialog.tsx
```diff
@@ -43,8 +43,8 @@ import {
 } from '@activepieces/shared';
 
 import { FormError } from '../../../components/ui/form';
-import { flowHooks } from '../lib/flow-hooks';
-import { templateUtils } from '../lib/template-parser';
+import { flowHooks } from '../../flows/lib/flow-hooks';
+import { templateUtils } from '../../flows/lib/template-parser';
```
AbdulTheActivePiecer: because these hooks and files were moved inside flows feature folder, now all components that will use them outside the flows feature folder will have to get them as props drilled into them or we break the linting rule of (features shouldn't import from each other, i.e templates feature can't import from flows feature. 

Please move back these files to lib as they were.

packages/web/src/features/automations/components/create-folder-dialog.tsx
```diff
@@ -29,20 +29,21 @@ import {
...
-  refetchFolders: (
+  refetchFolders?: (
     options?: RefetchOptions,
   ) => Promise<QueryObserverResult<FolderDto[], Error>>;
   className?: string;
+  open?: boolean;
+  onOpenChange?: (open: boolean) => void;
 };
 
 const CreateFolderFormSchema = Type.Object({
```
AbdulTheActivePiecer, line 32: same nit as before, move types or consts to the bottom of the component file.

packages/react-ui/src/features/automations/components/create-folder-dialog.tsx
```diff
@@ -58,8 +59,19 @@ export const CreateFolderDialog = ({
...
+  const [internalOpen, setInternalOpen] = useState(false);
+  const isControlled = open !== undefined;
+  const isDialogOpen = isControlled ? open : internalOpen;
+  const setIsDialogOpen = (value: boolean) => {
+    if (isControlled) {
+      onOpenChange?.(value);
+    } else {
+      setInternalOpen(value);
+    }
+  };
```
AbdulTheActivePiecer: what is this internalOpen, I checked where the component is used and we defined open and openChanged methods always, can you please simplify it by make these to params required and remove isControlled and isInternalOpen and isDialogOpen

packages/react-ui/src/features/automations/components/create-folder-dialog.tsx
```diff
@@ -81,7 +93,7 @@ export const CreateFolderDialog = ({
       form.reset();
       setIsDialogOpen(false);
       updateSearchParams(folder.id);
-      refetchFolders();
+      refetchFolders?.();
```
AbdulTheActivePiecer: Refetch folders is never defined in params where the component is used, please check packages/react-ui/src/app/routes/automations/index.tsx and see, in there we invalidate all queries each time the user closes the dialog, instead we should only invalidate when a new folder is created.

packages/web/src/features/automations/hooks/use-automations-dialogs.ts
```diff
@@ -0,0 +1,62 @@
...
+  selectedItems,
+}: DialogsDeps) {
+  const [isFolderDialogOpen, setIsFolderDialogOpen] = useState(false);
+  const [isImportFlowDialogOpen, setIsImportFlowDialogOpen] = useState(false);
+  const [isImportTableDialogOpen, setIsImportTableDialogOpen] = useState(false);
+  const [moveToDialogOpen, setMoveToDialogOpen] = useState(false);
+  const [moveToFolderId, setMoveToFolderId] = useState<string>('');
+  const [renameDialogOpen, setRenameDialogOpen] = useState(false);
+  const [newName, setNewName] = useState('');
+  const [itemToRename, setItemToRename] = useState<TreeItem | null>(null);
```
AbdulTheActivePiecer, line 23: instead of having so many useState, create only one with one json and call it the state and use it, and every time you edit something you just soft clone the json, set the changed property and reset the json.

AbdulTheActivePiecer: if you ask me this could even be a context and used accordingly.

packages/react-ui/src/features/automations/hooks/use-automations-filters.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (prevProjectIdRef.current !== projectId) {
+      prevProjectIdRef.current = projectId;
+      setSearchInput('');
+      setSearchTerm('');
+      setTypeFilter([]);
+      setStatusFilter([]);
+      setConnectionFilter([]);
+      setOwnerFilter([]);
+    }
+  }, [projectId]);
```
AbdulTheActivePiecer: same concept with the key property, if project id is the key for the component , you won't need to do this useEffect.

packages/react-ui/src/features/automations/hooks/use-automations-mutations.ts
```diff
@@ -0,0 +1,196 @@
...
+  invalidateFolder: (folderId: string) => void;
+  clearSelection: () => void;
+  flows: PopulatedFlow[];
+};
+
+export function useAutomationsMutations(deps: MutationDeps) {
+  const [isDeleting, setIsDeleting] = useState(false);
+  const [isMoving, setIsMoving] = useState(false);
+  const [isRenaming, setIsRenaming] = useState(false);
+  const [isExportingTables, setIsExportingTables] = useState(false);
```
AbdulTheActivePiecer: these are all flags that should be coming from mutations not useState, check the code below you are using callbacks with promises to do api calls to do mutations, instead use useMutation.

packages/react-ui/src/features/automations/hooks/use-automations-mutations.ts
```diff
@@ -0,0 +1,196 @@
...
+        deps.invalidateAll();
+      } catch {
+        toast.error(t('Failed to rename item'));
+      } finally {
+        setIsRenaming(false);
+      }
+    },
+    [deps],
+  );
+
```
AbdulTheActivePiecer: these should all be written with useMutation not useCallback.

packages/web/src/features/automations/hooks/use-automations-selection.ts
```diff
@@ -0,0 +1,126 @@
...
+    const map = new Map<string, TreeItem[]>();
+    treeItems.forEach((item) => {
+      if (item.parentId && item.type !== 'load-more-folder') {
+        const list = map.get(item.parentId) ?? [];
+        list.push(item);
+        map.set(item.parentId, list);
+      }
+    });
+    return map;
+  }, [treeItems]);
```
AbdulTheActivePiecer, line 20: use reduce instead of forEach and no need for useMemo

packages/react-ui/src/features/automations/hooks/use-automations-selection.ts
```diff
@@ -0,0 +1,126 @@
+import { useCallback, useMemo, useState } from 'react';
+
+import { TreeItem } from '../lib/types';
+import { getItemKey } from '../lib/utils';
+
+export function useAutomationsSelection(treeItems: TreeItem[]) {
+  const [selectedItems, setSelectedItems] = useState<Set<string>>(new Set());
```
AbdulTheActivePiecer: Don't rely on prefixes in a string, it is not strongly typed, like in getSelectedIdsByType i.e ''flow-" or "table-", please change the type of the set to actually be  something like 
```
type AutomationSelectedItem = {
  type: 'flow' | 'table' | 'folder';
  id: string;
}
```
never rely on string prefixes existing in a string type.

packages/react-ui/src/features/automations/lib/utils.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  tables.forEach((table) => {
+    items.push({
+      id: table.id,
+      type: 'table',
+      name: table.name,
+      data: table,
+      depth: 0,
+      parentId: null,
+    });
```
AbdulTheActivePiecer: shouldn't parentId be the folderId?

packages/react-ui/src/features/automations/components/automations-table-row.tsx
```diff
@@ -0,0 +1,220 @@
...
+          maxNumberOfIconsToShow={3}
+          size="xs"
+          circle={false}
+        />
+      );
+    }
+    if (item.type === 'table') {
+      return <span className="text-muted-foreground">-</span>;
+    }
+    return null;
```
AbdulTheActivePiecer: this is the empty folder state, it doesn't tell the user that their folder is empty, it should

Image

packages/react-ui/src/features/automations/components/automations-table-row.tsx
```diff
@@ -0,0 +1,220 @@
...
+          trigger={flow.version.trigger}
+          maxNumberOfIconsToShow={3}
+          size="xs"
+          circle={false}
+        />
+      );
+    }
+    if (item.type === 'table') {
+      return <span className="text-muted-foreground">-</span>;
+    }
```
AbdulTheActivePiecer: I prefer a switch.

packages/react-ui/src/app/components/flow-actions-menu.tsx
```diff
@@ -102,6 +104,38 @@ const FlowActionMenu: React.FC<FlowActionMenuProps> = ({
   const { projectMembers } = projectMembersHooks.useProjectMembers();
   const hasProjectMembers = projectMembers && projectMembers.length > 0;
 
+  const [isRenameOpen, setIsRenameOpen] = useState(false);
+  const [renameValue, setRenameValue] = useState(flowVersion.displayName);
+  const [isMoveOpen, setIsMoveOpen] = useState(false);
+  const [moveFolderId, setMoveFolderId] = useState('');
```
AbdulTheActivePiecer: rename to folderToMoveId

packages/web/src/app/components/flow-actions-menu.tsx
```diff
@@ -133,24 +167,44 @@ const FlowActionMenu: React.FC<FlowActionMenuProps> = ({
...
-        onCloseAutoFocus={(e) => e.preventDefault()}
-      >
-        {!readonly && (
-          <>
-            {insideBuilder && (
-              <PermissionNeededTooltip
-                hasPermission={userHasPermissionToUpdateFlow}
-              >
+    <>
+      <DropdownMenu open={open} onOpenChange={setOpen}>
```
AbdulTheActivePiecer, line 170: Something has changed here that I might not be aware of or is a bug:
comparing the pr to cloud -->
cloud:

Image


In pr:

Image

Duplicate flow, change owner, export ..etc are all not showing anymore for flows.

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
Louai-Zokerburg: that threshold controls how many parallel fetch requests depending on how many folders the project have.

data flow is as follows:
- fetch folders 
- fetch first page data for each folder in parallel (default 50 parallel if the project have less than 50 folders if they have more it gets fetch in next go
- fetch first page of root level data

packages/web/src/features/automations/hooks/use-automations-data.ts
Louai-Zokerburg: nope not really, some folders have up to 1k+ items inside of them

packages/server/api/src/app/tables/table/table.service.ts
Louai-Zokerburg: that is because if adding it to the updateData using the ...spreadIfDefinded wont work. because sometimes folderId is undefined and ii want it to set the the value as undefined and NOT spreadIfDefault

packages/react-ui/src/features/pieces/components/piece-icon.tsx
Louai-Zokerburg: nope, p-1.5 is away to far than 5px and i tried multiple values it worked with 5px and not any other value

packages/react-ui/src/app/guards/permission-guard.tsx
Louai-Zokerburg: I think it should be requiredPermissions not allowedPermissions, and its not an any check is a "all check" means all permissions in the array must be there to access the page

packages/web/src/features/automations/components/automations-empty-state.tsx
Louai-Zokerburg: I made sure that if u don't have templates it wont show to u at all (not infinite loading) and for others lets keep it as it is now, we can make it hard coded templates for everyone in a next iteration

packages/web/src/features/automations/components/automations-table-row.tsx
Louai-Zokerburg: I answered that above already some folders have 1k+ items and the idea of fetching their data all at once is kinda wrong

packages/react-ui/src/features/automations/components/create-folder-dialog.tsx
Louai-Zokerburg: this is needed because the open state of the folder isn't always controlled by the click (in search we need all folders to be open by default) s

packages/web/src/features/automations/components/import-flow-dialog.tsx
Louai-Zokerburg: there is no such a rule saying u cant import from other features, the actual rule says u cant import from the /app folder, also the lint is working fine no errors are shown

packages/react-ui/src/features/automations/components/create-folder-dialog.tsx
AbdulTheActivePiecer: can't there be a way to pass that from the outside instead of having two open states ?

packages/react-ui/src/features/automations/hooks/use-automations-data.ts
AbdulTheActivePiecer: Let's discuss this in vc, I am not really getting to imagine it in comments.

packages/react-ui/src/features/automations/components/automations-table-row.tsx
Louai-Zokerburg: Not really u already have the details column saying pretty clearly 0 files, plus such behaviour/ux/design idk how to call it is already common and not something new, I dont think its a good idea to have an empty state for a folder like u expand to see empty state :|

packages/web/src/features/automations/components/import-flow-dialog.tsx
AbdulTheActivePiecer: great, sorry I mistook it for the rule you mentioned.

packages/react-ui/src/app/guards/permission-guard.tsx
AbdulTheActivePiecer: Array.some is used down below if it is an all check then it should be Array.every.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

But either way you should be able to view the automations route if you have the flow read permission OR the tables read permission, that's why some is correctly used below and why it should be allowedPermissions.

packages/react-ui/src/features/automations/components/automations-table-row.tsx
AbdulTheActivePiecer: The details column is not very visible and when someone sees an expandable accordion they expect there to be an item below it, it can be something very simple as, "This folder is empty", or remove the accordion expand button

packages/web/src/app/routes/automations/index.tsx
```diff
@@ -0,0 +1,294 @@
...
+        userHasPermissionToWriteFolder={userHasPermissionToWriteFolder}
+        onCreateFlow={mutations.createFlow}
+        onCreateTable={() => mutations.createTable(t('New Table'))}
+        onCreateFolder={() => dialogs.setIsFolderDialogOpen(true)}
+        onImportFlow={() => dialogs.setIsImportFlowDialogOpen(true)}
+        onImportTable={() => dialogs.setIsImportTableDialogOpen(true)}
+        onClearAllFilters={clearAllFilters}
+        hasActiveFilters={filtersActive}
+        isCreatingFlow={mutations.isCreateFlowPending}
+        isCreatingTable={mutations.isCreatingTable}
```
AbdulTheActivePiecer, line 192: no need for params you can use them inside the component through context, not all params maybe but most of them

packages/web/src/features/automations/components/automations-table-row.tsx
AbdulTheActivePiecer: for now just at least increase the load more to 50 and the minimum fetched items.

packages/react-ui/src/features/automations/hooks/use-automations-mutations.ts
```diff
@@ -0,0 +1,266 @@
...
+  const { mutateAsync: deleteItem } = useMutation({
+    mutationFn: async (item: TreeItem) => {
+      if (item.type === 'flow') {
+        await flowsApi.delete(item.id);
+      } else if (item.type === 'table') {
+        await tablesApi.delete(item.id);
+      } else if (item.type === 'folder') {
+        await foldersApi.delete(item.id);
+      }
```
AbdulTheActivePiecer: nit: swtich

packages/react-ui/src/app/routes/automations/index.tsx
AbdulTheActivePiecer: Since the keys trick here won't work you can just call the function itself (resetPagination) after each time one of the deps change, declaratively.

packages/web/src/features/automations/hooks/use-automations-dialogs.ts
Louai-Zokerburg: I llike it this way :)

packages/react-ui/src/features/automations/components/automations-table-row.tsx
Louai-Zokerburg: might a good future improvement

packages/web/src/app/routes/automations/index.tsx
Louai-Zokerburg: I researched the idea of using a context, the idea of shrinking code isn't really true coz same code removed from index.tsx and child components as props will be traded as context.ts file.

second point is that current implementation allows some level of independency between components in re-renders but if they all have one shared context a change in one value consumed by A component will lead to a re-render to a  X component that have no relationship with that changed value

finally current prop passing is only one level so its no harmful at all 

I decide to leave it like this

PR #11324, opened 2026-02-21
fix: pass password as passphrase for SFTP private key authentication

themavik:
When using private key auth with SFTP, the password field should be passed as the passphrase to decrypt the private key.


themavik:
Recreating from upstream

PR #11330, opened 2026-02-21
refactor(turbo): major change replace nx with turbo

abuaboud:
# Summary                      
                                                                                                                                                                 
  - Migrate build orchestration from NX to Turbo (turbo.json, dependsOn: ["^build"] for dependency-aware builds)                                               
  - Move shared packages (@activepieces/shared, @activepieces/pieces-framework, @activepieces/pieces-common) to local ./dist output so workspace symlinks resolve
   to compiled JS at runtime instead of raw TypeScript source                                                                                                    
  - Add cp package.json to all 621 piece build scripts so package.json is available in the dist output folder                                                    
  - Remove bun link based dev piece linking (dev-pieces-installer.ts) — no longer needed since workspace resolution now correctly loads compiled output          
  - Simplify dev pieces builder to use turbo filters directly without post-build linking steps


OsamaHaikal:
that is huge

PR #11395, opened 2026-02-25
fix(google-drive): fix pagination bugs

AdamSelene:
## What does this PR do?

The list-files action could only retrieve the first page (100 items) because the pagination token was not returned by the API 🤦 
Added pagination to the search action and to both triggers too

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/google-drive/src/lib/common/index.ts
+++ b/packages/pieces/community/google-drive/src/lib/common/index.ts
@@ -103,15 +103,23 @@ export const common = {
         ).format()}'`
       );
     q.push(`trashed = false`);
-    const response = await drive.files.list({
-      q: q.concat("mimeType!='application/vnd.google-apps.folder'").join(' and '),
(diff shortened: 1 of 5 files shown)


packages/pieces/community/google-drive/src/lib/common/index.ts
```diff
@@ -103,15 +103,23 @@ export const common = {
...
-      orderBy: order ?? 'createdTime desc',
-      supportsAllDrives: true,
-      includeItemsFromAllDrives: search?.includeTeamDrive,
-    });    
+    const allFiles: any[] = [];
+    let pageToken: string | undefined = undefined;
+    do {
+      const listParams: Record<string, any> = {
+        q: q.concat("mimeType!='application/vnd.google-apps.folder'").join(' and '),
+        fields: 'nextPageToken, files(id, name, mimeType, webViewLink, kind, createdTime)',
```
AdamSelene, line 111: `createdTime` was missing here

PR #11401, opened 2026-02-25
feat: add AP_MAX_WEBHOOK_PAYLOAD_SIZE_MB environment variable

### LINKED ISSUE #11267: Add an environment variable to determine the max size for webhooks payload
As things stand the fastify app body size limit is determined relying on the MAX_FILE_SIZE_MB and MAX_FLOW_RUN_LOG_SIZE_MB variables, we should extract this to its own separate variable : AP_MAX_WEBHOOK_PAYLOAD_SIZE

Vaibhavee89:
## Summary
Adds a dedicated environment variable `AP_MAX_WEBHOOK_PAYLOAD_SIZE_MB` to control webhook payload size limits independently from file upload size and flow run log size limits.

## Problem
Previously, the Fastify app's `bodyLimit` was calculated from `MAX_FILE_SIZE_MB` and `MAX_FLOW_RUN_LOG_SIZE_MB`:
```typescript
bodyLimit: Math.max(fileSizeLimit + 4, flowRunLogSizeLimit + 4, 25) * 1024 * 1024
```

This coupling made it impossible to configure webhook payload limits independently, which could lead to:
- Unnecessarily large body limits when only file sizes or log sizes needed to be increased
- Inability to restrict webhook payloads separately from other concerns

## Solution
Introduced a new dedicated environment variable `AP_MAX_WEBHOOK_PAYLOAD_SIZE_MB` that directly controls the HTTP body size limit:

```typescript
const webhookPayloadSizeLimit = system.getNumberOrThrow(AppSystemProp.MAX_WEBHOOK_PAYLOAD_SIZE_MB)
bodyLimit: webhookPayloadSizeLimit * 1024 * 1024
```

### Changes Made:
1. **Added system property**: `MAX_WEBHOOK_PAYLOAD_SIZE_MB` to `AppSystemProp` enum
2. **Set default value**: `25 MB` (maintains current behavior)
3. **Added validator**: Number validator for the new property
4. **Updated server config**: Uses new dedicated variable instead of calculating from other limits

## Configuration

Users can now set:
```bash
AP_MAX_WEBHOOK_PAYLOAD_SIZE_MB=50
```

To control webhook payload size independently of:
- `AP_MAX_FILE_SIZE_MB` - File uploads in steps/triggers
- `AP_MAX_FLOW_RUN_LOG_SIZE_MB` - Flow run log size

## Backward Compatibility
✅ **Fully backward compatible**
- Default value of `25 MB` matches previous effective default
- Existing deployments without this variable will continue working with the same limits
- No breaking changes to existing configurations

## Testing
- ✅ Default value set to 25 MB
- ✅ Validator added for type checking
- ✅ Clean separation from file size and log size limits
- ✅ Follows existing environment variable patterns

Fixes #11267


kishanprmr:
Completed in https://github.com/activepieces/activepieces/commit/60d3098f38.

PR #11516, opened 2026-03-03
feat: Allow Skyvern BASE_URL to be specified

peblair:
## What does this PR do?

This PR extends the Skyvern integration to support a custom `BASE_URL` parameter, which can facilitate integration with self-hosted Skyvern deployments.

Users can specify a base url by setting the `SKYVERN_BASE_URL` environment variable. This will override the default `'https://api.skyvern.com/v1'` URL.

### Relevant User Scenarios

This would be useful for people with custom or local deployments of Skyvern with which they would like to integrate.

### DIFF
--- a/packages/pieces/community/skyvern/src/lib/common/client.ts
+++ b/packages/pieces/community/skyvern/src/lib/common/client.ts
@@ -35,7 +35,7 @@ export async function skyvernApiCall<T extends HttpMessageBody>({
 
 	const request: HttpRequest = {
 		method,
-		url: BASE_URL + resourceUri,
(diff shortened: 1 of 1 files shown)


kishanprmr:
Thanks for the contribution! Using `process.env` inside a piece isn't the right approach. The correct way is to add a baseUrl field to the piece auth so users can configure it themselves. Please update the auth to include an optional base URL field and use that instead.

kishanprmr:
Closing due to inactivity. Please add a base URL as the connection field to support self-hosted domains and create a PR.

PR #11529, opened 2026-03-04
fix(google-slides): missing client id error in google slides generate from template piece

bertrandong:
### The Problem
When users execute the **Generate from template** action using the preconfigured connection OAuth2 credential, the piece crashes with a `400 Bad Request` error:
> `error_description: 'Could not determine client ID from request.'`
Screenshot 2026-03-05 at 4 36 53 PM

Interestingly, the action succeeds if a user provides their own personal Client ID and Client Secret. 

### Root Cause
The bug occurs due to a conflict between how the Activepieces engine handles Platform Auth and how the `google-auth-library` SDK behaves:

1. In a preconfigured Auth scenario, it seems like the Activepieces core engine handles token refreshing securely in the background. Before executing the piece, it strips `client_id` and `client_secret` from `context.auth` to prevent leaking platform secrets, but it still passes the `refresh_token`.
2. The original code passed the entire `context.auth` object to the Google SDK: `authClient.setCredentials(context.auth)`.
3. The Google SDK saw the `refresh_token` and assumed it was responsible for managing the token lifecycle. When it inevitably attempted to hit the `oauth2.googleapis.com/token` endpoint, it crashed because the `client_id` was intentionally missing from the payload.

### Fix
Updated the `OAuth2Client` initialization to **only** receive the `access_token`. 

### Notes
While this setCredentials(context.auth) pattern exists in other Google pieces (like Drive or Docs), the only affected piece is this generate from template action.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 1,
   "workspaces": {
     "": {
       "name": "activepieces",
@@ -472,7 +471,7 @@
     },
     "packages/pieces/community/ai": {
       "name": "@activepieces/piece-ai",
-      "version": "0.1.16",
+      "version": "0.1.17",
       "dependencies": {
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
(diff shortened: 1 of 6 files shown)


onyedikachi-david:
Thanks for the fix @bertrandong

PR #11543, opened 2026-03-04
fix: simplify form-data fields to single value input

### LINKED ISSUE #11263: Improve handling formdata for http steps
For the http piece when sending a request with form data, Instead of having both a file input and a text input and a select for the file type, we should simplify the UI 

Image

akramcodez:
### What does this PR do?

Simplifies the form data UI for HTTP steps by replacing the three-field setup (Field Type dropdown + Text Field Value + File Field Value) with a single **Value** field that automatically handles both text strings and file inputs based on what is passed in.

When building an HTTP request with `form_data` body type, each form entry now only requires two fields:
- **Field Name** - the form key
- **Value** - accepts either typed text or a file passed from a previous step

The runtime detects the value type (`typeof value === 'string'`) and appends it to the FormData accordingly. No explicit type selection needed.

Changes applied consistently across all three HTTP action implementations: `helpers/index.ts`, `send-http-request-action.ts`, and `send-oauth2-http-request.ts`.

### Relevant User Scenarios

- User wants to POST a file upload via form data — previously had to select "File" from a dropdown, now just passes the file directly into Value
- User wants to send a text field via form data — previously had to select "Text" and fill a separate text input, now just types into Value
- Multi-agent / automation builders who wire file outputs from previous steps into HTTP form fields with less configuration friction

fixes #11263

### DIFF
--- a/packages/pieces/common/src/lib/helpers/index.ts
+++ b/packages/pieces/common/src/lib/helpers/index.ts
@@ -231,24 +231,10 @@ i.e ${getBaseUrlForDescription(baseUrl, auth)}/resource or /resource`,
                     displayName: 'Field Name',
                     required: true,
                   }),
-                  fieldType: Property.StaticDropdown({
-                    displayName: 'Field Type',
-                    required: true,
-                    options: {
-                      disabled: false,
-                      options: [
-                        { label: 'Text', value: 'text' },
-                        { label: 'File', value: 'file' },
-                      ],
-                    },
-                  }),
-                  textFieldValue: Property.LongText({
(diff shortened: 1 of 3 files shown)


akramcodez:
PTAL @AbdulTheActivePiecer

kishanprmr:
Thanks for the contribution! `Property.File` has a file processor at the engine level (variables/processors/file.ts) that converts the value to either an `ApFile` or null before it reaches the action, plain text strings get processed as base64 or a URL, and if neither matches they become null. So the `typeof value === 'string'` check will never be true and text values will silently be dropped. Closing this one.

PR #11581, opened 2026-03-05
chore: secret manager OAuth2 integration

MrChaker:
Previously we had problem exposing endpoint for resolving secrets . 
Right now we don't do that we build the popup url in the backend and return it .


packages/server/api/src/app/ee/secret-managers/secret-manager-providers/aws-provider.ts
MrChaker, line 1: changes on providers below are just for  better error messages

MrChaker:
Closing in favour of refactor/secret-managers branch with updated scope, cache improvements, and docs.

PR #11666, opened 2026-03-10
feat: add global search v1

Louai-Zokerburg:
### Relevant User Scenarios

Fixes # (issue)


packages/web/src/features/automations/hooks/use-automations-filters.ts
```diff
@@ -46,6 +46,11 @@ export function useAutomationsFilters() {
     searchParams.getAll(FOLDER_PARAM),
   );
 
+  const folderParamStr = searchParams.getAll(FOLDER_PARAM).join('\0');
+  useEffect(() => {
+    setFolderFilterState(folderParamStr ? folderParamStr.split('\0') : []);
+  }, [folderParamStr]);
+
```
AbdulTheActivePiecer, line 50: please remove this use effect, try to use a more declarative way for the change, (either try to use the key property or see if there is a listener or a hook that exists already that listens to params changes, or even if this state is just extracted and you don't need to use useState.)

AbdulTheActivePiecer (approved):
Please make opus review it, and also make sure that the generic components you edit  work everywhere they are used, i.e command.tsx
also make sure that it works on embed, either from the sdk-demo, or on embed.activepieces.com

PR #11548, opened 2026-03-04
feat(pieces): add webhook triggers to Baserow piece

bst1n:
## Summary

- Add three new **webhook triggers** to the Baserow piece: **Row Created**, **Row Updated**, and **Row Deleted**
- Webhooks are automatically registered/unregistered via the Baserow webhooks API (`POST /api/database/webhooks/table/{table_id}/` and `DELETE /api/database/webhooks/{webhook_id}/`)
- Uses the recommended batch event types (`rows.created`, `rows.updated`, `rows.deleted`) instead of the deprecated single-row events
- Extends `BaserowClient` with `createWebhook()` and `deleteWebhook()` methods

## Changes

| File | Description |
|---|---|
| `src/lib/common/client.ts` | Added `createWebhook()` and `deleteWebhook()` methods |
| `src/lib/triggers/row-created.ts` | New trigger for `rows.created` events |
| `src/lib/triggers/row-updated.ts` | New trigger for `rows.updated` events |
| `src/lib/triggers/row-deleted.ts` | New trigger for `rows.deleted` events |
| `src/index.ts` | Registered the three new triggers |

## Test plan

- [ ] Configure Baserow auth with API URL and Database Token
- [ ] Create a flow with the "Row Created" trigger, set a table ID, and verify a webhook is created in Baserow
- [ ] Add a row in Baserow and verify the flow triggers with the correct payload
- [ ] Test "Row Updated" trigger by modifying a row
- [ ] Test "Row Deleted" trigger by deleting a row
- [ ] Disable a flow and verify the webhook is removed from Baserow

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11697, opened 2026-03-13
fix(pieces/baserow): fix triggers, add Find Row and Clean Row actions

### LINKED ISSUE #7503: [BUG]: Baserow : Update Row does not authorize blank for multiple select field
**Describe the bug**
Baserow's Update Row action does not tolerate a null value for multiple select fields (whereas the Create Row action does).

**To Reproduce**
1. Create a table with different fields and a multiple select field
2. Update a row without modifying the multiple select field, leave blank

**Expected behavior**
The flow should work without having to choose for an option for the multiple select field

bst1n:
## Summary

Fixes issues flagged in the [Greptile review of PR #11548](https://github.com/activepieces/activepieces/pull/11548#issuecomment-4039122662) (confidence 2/5), fixes long-standing bug #7503, and adds two new actions.

## Bug fixes

**Fix triggers returning raw webhook body** — `rows.created` and `rows.updated` webhooks send `{ items: [...] }` but triggers were returning the entire body as a single event. Now returns each item individually so the flow fires once per row (including batch operations). The previous payload structure was unusable (users had to navigate `body.items[0]` to access fields). These triggers were added days ago in PR #11548 with virtually no existing flows built on them — this is a bug fix, not a breaking change.

**Fix row-deleted trigger** — Returns individual `{ id }` objects from `row_ids` instead of the raw body.

**Fix row-updated trigger** — Returns `{ row, previous }` as two separate objects instead of merging `old_item` into the row (avoids field name collision).

**Fix Update Row with blank select fields (fixes #7503)** — Update Row now skips fields that are `null`, `undefined`, `''`, or `[]` instead of sending them to the API. This fixes the long-standing issue where leaving single/multiple select fields blank caused API errors, and prevents unintentional clearing of multi-select and link fields. Note: sending `""` to a text field is safe (Baserow ignores it), but select fields reject it with a 400 error — filtering applies to all field types as a safe no-op for text, and a necessary fix for selects.

**Fix Create Row with blank select/link fields** — Applied the same `null`/`undefined`/`''`/`[]` filtering as Update Row. Prevents a crash when fx mode expressions return empty values for link-type fields (calling `.map()` on `''`).

**Fix sampleData** — Updated across all three triggers to reflect the per-row structure users will actually map fields from.

**Fix tableId description** — Changed "Select the table to watch" to "Select the table" since this dropdown is shared by actions and triggers.

## New actions

**Find Row** — Searches for a row by selecting a field from a dropdown and providing a value (exact match via `filter__field_{id}__equal`). Returns `{ found: true/false, row }`. Field dropdown filters out unsupported types (link_row, multiple_select, multiple_collaborators, file, date, rollup, lookup, count, last_modified, last_modified_by, created_on, created_by). Includes guard for missing field_name/field_value.

**Clean Row** — Same form as Update Row, but empty values (`null`, `''`, `[]`) are sent through to the API, allowing users to intentionally clear fields. Use Update Row for safe partial updates (untouched fields stay unchanged). Use Clean Row to explicitly clear fields.

## Other improvements

- Increase rowId dropdown limit from 100 to 200 rows

Bumps piece version from 0.2.0 → 0.2.1.

## Test plan
- [ ] Create a row → verify Row Created trigger fires with row fields directly accessible (not nested under `body.items[0]`)
- [ ] Create multiple rows at once → verify the flow fires once per row
- [ ] Update a row → verify Row Updated trigger fires with `{ row, previous }` structure
- [ ] Delete a row → verify Row Deleted trigger fires with `{ id }`
- [ ] Find Row: select a table and field, enter a value → verify correct row returned
- [ ] Find Row with no match → verify `{ found: false, row: null }` returned
- [ ] Update Row: leave a single select field empty (dropdown mode) → verify field is NOT cleared
- [ ] Update Row: leave a single select field empty (fx mode) → verify no API error
- [ ] Clean Row: leave a select field empty → verify field is cleared in Baserow
- [ ] Create Row: leave a link field empty in fx mode → verify no crash
- [ ] Verify rowId dropdown shows up to 200 rows

### DIFF
--- a/packages/pieces/community/baserow/src/lib/actions/find-row.ts
+++ b/packages/pieces/community/baserow/src/lib/actions/find-row.ts
@@ -40,6 +40,7 @@ export const findRowAction = createAction({
           BaserowFieldType.DATE,
           BaserowFieldType.LAST_MODIFIED,
           BaserowFieldType.CREATED_ON,
+          BaserowFieldType.DURATION,
         ];
         const filterable = fields.filter((f) => !unsupportedTypes.includes(f.type));
         return {


bst1n:
Thanks for the review! All points have been addressed:

**✅ Find Row — field_name guard**: Added a guard for falsy `field_name` (commit 9fdef6e).

**✅ Find Row — unsupported field types**: The dropdown now filters out field types that don't support the `equal` filter (link_row, multiple_select, multiple_collaborators, file) (commit 9fdef6e).

**Re: Update Row empty-string trade-off**: This is intentional. We tested extensively and confirmed that Baserow's PATCH API actually ignores empty strings on text fields (they don't clear the value). The `''` filter is needed specifically for select fields (which reject `''` with a 400 error — this is the original #7503 bug). Filtering `''` for all types is a safe no-op for text fields and fixes the select issue.

**Re: old_item key collision in row-updated**: Acknowledged. The chance of a Baserow column being named exactly `old_item` is very low. We'll keep the current naming for simplicity — renaming to something like `__old_item__` would make it less readable for users in the flow builder.

bst1n:
Addressed the remaining review comments:

**Re: empty array guard (update-row.ts lines 41-43)** — We keep the guard in Update Row to protect against unintentional field clearing. For users who need to intentionally clear fields, we added a new **Clean Row** action (commit cf73457) that sends all values through, including empty arrays and null.

- **Update Row** → safe by default, untouched fields stay unchanged
- **Clean Row** → sends everything, empty values clear the fields

**Re: old_item key collision (row-updated.ts)** — Fixed in commit cf73457. Instead of renaming the key, the trigger now returns two separate objects:
```typescript
return { row: item, previous: old_items[i] }
```
No collision possible, cleaner for users in the flow builder.

bst1n:
Addressed all new review comments:

**✅ Find Row — computed field types**: Added ROLLUP, LOOKUP, COUNT, LAST_MODIFIED_BY, CREATED_BY to the unsupported types list.

**✅ Update Row description**: Updated to mention that empty values are skipped and points users to Clean Row for clearing fields.

**Re: breaking change on triggers**: These triggers were just added in PR #11548 (merged days ago). They are brand new — there are virtually no existing flows using them yet. The payload fix is necessary because the previous structure was incorrect (returning the raw webhook body instead of individual rows).

bst1n:
Thanks for the review!

**✅ Date fields excluded from Find Row** — Added `DATE`, `LAST_MODIFIED`, and `CREATED_ON` to the `unsupportedTypes` list (commit 858c5c2). These require `date_equal` instead of `equal`.

**Re: breaking change on triggers** — These triggers were added in PR #11548 (merged recently). Very few users have had time to build flows with them. The previous payload structure was incorrect (returning the raw webhook body instead of individual rows), so this is a bug fix, not a breaking change.

**Re: empty string/array skip in Update Row** — Intentional trade-off, addressed by the new Clean Row action. The description already mentions this: *"Empty or blank values are skipped (leaving those fields unchanged). To explicitly clear a field, use the Clean Row action."*

bst1n:
Closing in favor of a cleaner PR with refactored code (shared helper, no duplication).

PR #11699, opened 2026-03-13
fix(pieces/airtable): fix Update Record clearing fields and add Clean Record action

### LINKED ISSUE #10344: [BUG]: Airtable "Update Record" Action Clears Single select Field When Not Provided
Describe the bug
When updating a record with a dropdown field column, if it is not provided and left empty, it removes existing record data from the column. 

To Reproduce

Create a flow with the "Update Record" action in Airtable.
Select the table containing a single select field column. Ensure the record has this a single select  column filled.
Change values in other columns as required, and leave the a single select  field empty.
Test the action and observe the result in the a single select  column.
Expected behavior
If the value for a single select  field is not provided, it should not override the existing value in the column.

Screenshots

Additional context

PS : 
Bug is very similar to this one : https://github.com/activepieces/activepieces/issues/4990

bst1n:
## Summary

Fix the **Update Record** action clearing single select and multi-select fields when they are left empty by the user, and add a new **Clean Record** action for intentionally clearing fields.

## Bug fix — Update Record
- The existing code already filters empty strings (`''`), but `null` values (from unselected dropdowns in integrated mode) and empty arrays `[]` (from unselected multi-selects) were still sent to the Airtable API, causing those fields to be cleared unexpectedly
- Added `null`, `undefined`, and empty array `[]` filtering alongside the existing `''` filter
- Renamed variable from `fieldsWithoutEmptyStrings` to `fieldsWithoutEmptyValues` to reflect the broader filtering

## New action — Clean Record
- Same form as Update Record, but empty values are sent through to the API, allowing users to intentionally clear fields
- Added `allowEmpty` parameter to `createNewFields` to bypass empty array filtering when needed
- Use **Update Record** for safe partial updates (untouched fields stay unchanged)
- Use **Clean Record** when you need to explicitly clear fields

Bumped version from 0.6.5 to 0.6.6.

## Test plan
- [ ] Configure an Update Record action on a table with single select and multi-select fields
- [ ] Leave select fields unselected (integrated mode) → verify they are not cleared
- [ ] Leave select fields empty in fx mode → verify they are not cleared
- [ ] Provide actual values for select fields → verify they are updated correctly
- [ ] Configure a Clean Record action → leave a select field empty → verify field is cleared in Airtable
- [ ] Clean Record with a value → verify it updates normally

Fixes #10344

### DIFF
--- a/docs/build-pieces/building-pieces/create-action.mdx
+++ b/docs/build-pieces/building-pieces/create-action.mdx
@@ -94,9 +94,9 @@ export const gelato = createPiece({
 
 # Testing
 
-By default, the development setup only builds specific components. Open the file `packages/server/api/.env` and include "gelato" in the `AP_DEV_PIECES`.
+By default, the development setup only builds specific components. Open the file `.env.dev` and include "gelato" in the `AP_DEV_PIECES`.
 
-For more details, check out the [Piece Development](./development-setup) section.
+For more details, check out the [Piece Development](/build-pieces/building-pieces/development-setup#pieces-development) section.
 
 Once you edit the environment variable, restart the backend. The piece will be rebuilt. After this process, you'll need to **refresh** the frontend to see the changes.
 
--- a/docs/build-pieces/building-pieces/create-trigger.mdx
+++ b/docs/build-pieces/building-pieces/create-trigger.mdx
@@ -45,14 +45,14 @@ import {
   pollingHelper,
 } from '@activepieces/pieces-common';
 import {
-  PiecePropValueSchema,
   TriggerStrategy,
   createTrigger,
+  AppConnectionValueForAuthProperty
 } from '@activepieces/pieces-framework';
 import dayjs from 'dayjs';
 
 const polling: Polling<
-  PiecePropValueSchema<typeof gelatoAuth>,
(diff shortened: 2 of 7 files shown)


bst1n:
Addressed review comment:

**✅ allowEmpty sending undefined fields**: Fixed. The condition now checks `fields[key] !== undefined` when `allowEmpty` is true, so only explicitly provided values (including empty arrays `[]`) are passed through. Untouched fields are no longer sent.

bst1n:
Thanks for the thorough review! All 3 points have been addressed in commit 51d454e:

**✅ multipleAttachments not handled for allowEmpty** — Fixed. When `allowEmpty = true` and the field is empty/null, we now send `[]` instead of letting it fall through to the generic else branch.

**✅ else branch sends null for untouched fields** — Fixed. Added an early `if (!(key in fields)) return;` guard so only fields explicitly present in the input object are processed. Untouched fields are now skipped entirely.

**✅ allowEmpty passes undefined for multipleRecordLinks/multipleSelects** — Fixed. The condition now simply checks `allowEmpty` (truthy) without the `fields[key] !== undefined` part, since the early `key in fields` guard already handles this.

PR #12027, opened 2026-03-18
fix: in dev mode, actions/triggers run methods were not being hot reloaded on change

hazemadelkhalel:
## Summary
- **Integrated file watcher**: Replace the standalone `tools/watch-dev-pieces.js` process and `setup-dev.js` detached spawn with a chokidar-based watcher inside the API server (`dev-piece-watcher.ts`). This eliminates redundant build processes and gives the watcher direct access to socket.io for frontend refresh.
- **Per-piece debounce** (300ms): Coalesces rapid file saves into a single rebuild per piece. Changes arriving during an active build are queued and trigger one rebuild after completion.
- **Pre-build on setup**: `setup-dev.js` still pre-builds dev pieces so `dist/` exists before the server starts, but no longer spawns a background watcher.
- **Dead code cleanup**: Remove unused `dev-pieces.ts` and `file-pieces-utils.ts` from the worker package, along with their dead dependencies (`chalk`, `clear-module`).
- **Stream UI**: Switch turbo UI to `stream` for better log readability.

## Test plan
- [ ] Run `npm start` with `AP_DEV_PIECES` set — confirm "Watching for changes: \<piece\>" appears in server logs
- [ ] Edit a piece source file — confirm debounced rebuild triggers, followed by "Changes are ready" log
- [ ] Rapidly save multiple files — confirm only one build triggers (debounce working)
- [ ] Edit during an active build — confirm queued rebuild fires after current build completes
- [ ] Verify no detached watcher process is spawned by `setup-dev.js`


tools/watch-dev-pieces.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+    buildPiece(pieceName);
+  }, 300));
+});
+
+watcher.on('error', (err) => {
+  console.error(`[piece-watcher] Watcher error:`, err.message);
+});
+
+console.log(`[piece-watcher] Watching for changes in: ${pieceNames.join(', ')}`);
```
AbdulTheActivePiecer: this file is pretty similar to what we have in packages/server/worker/src/lib/cache/pieces/dev-pieces.ts, shouldn't we just fix the issue in that file ?

PR #12091, opened 2026-03-23
feat(piece-tally): auto-register webhooks via API + named field keys

bst1n:
## Problem

The Tally piece had two major usability issues:
1. **Manual webhook setup** — users had to copy-paste a URL into Tally's dashboard, which is error-prone and doesn't clean up on disable
2. **Numbered fields** — form responses were exposed as a positional array (`data.fields[0]`, `data.fields[1]`...), making field mapping impossible without guessing the index

### API integration
- Added `PieceAuth.SecretText` for Tally API key (Settings > API keys in Tally dashboard)
- Added a **form dropdown** that fetches all non-deleted forms via `GET /forms`
- `onEnable`: automatically registers the webhook via `POST /webhooks` and stores the webhook ID
- `onDisable`: automatically deletes the webhook via `DELETE /webhooks/{id}`
- New `src/lib/common/index.ts` with the API client and forms dropdown

### Named field keys
- `data.fields` is now an object keyed by field label (`data.fields["Your Email"]`) instead of a positional array
- Multiple choice / dropdown values are resolved from option IDs to human-readable text
- Duplicate field labels get a numeric suffix (`"Name"`, `"Name (2)"`)
- Added sample data so the flow builder shows field names without a real submission

## Test plan

- [ ] Connect with a valid Tally API key — form dropdown should populate
- [ ] Enable the trigger — verify webhook appears in Tally dashboard automatically
- [ ] Disable the trigger — verify webhook is removed from Tally dashboard
- [ ] Submit the form — verify `data.fields` is keyed by label, not index
- [ ] Test a multiple choice field — verify value shows option text, not option ID
- [ ] Test two fields with the same label — verify second gets `(2)` suffix

### DIFF
--- a/packages/pieces/community/tally/src/index.ts
+++ b/packages/pieces/community/tally/src/index.ts
@@ -1,14 +1,22 @@
 import { createPiece, PieceAuth } from '@activepieces/pieces-framework';
 import { PieceCategory } from '@activepieces/shared';
 import { tallyFormsNewSubmission } from './lib/triggers/new-submission';
+
+export const tallyAuth = PieceAuth.SecretText({
+  displayName: 'API Key',
+  description:
+    'Your Tally API key. Go to **Settings > API keys** in your Tally dashboard to create one.',
+  required: true,
+});
+
 export const tally = createPiece({
   displayName: 'Tally',
   description: 'Receive form submissions from Tally forms',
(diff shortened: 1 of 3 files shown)


bst1n:
Closing in favour of a clean branch — the previous head included unrelated Outseta files due to fork divergence.

PR #12124, opened 2026-03-24
feat(mindee): update to Mindee v2 inference API with async enqueue+poll

### LINKED ISSUE #8229: Update Mindee Piece to Use v2 Inference API
New v2 Endpoint:

https://api-v2.mindee.net/v2/inferences/enqueue

However, in the current Mindee piece in Activepieces, we’re still using the older v1 format:

[https://api.mindee.net/v1/products/${account_name}/${api_name}/predict](https://api.mindee.net/v1/products/$%7Baccount_name%7D/$%7Bapi_name%7D/predict)

Docs for Reference:
Mindee v2 API: [API Reference | Mindee](https://docs.mindee.com/integrations/api-reference)

Harmatta:
## What this PR does

Updates the existing **Mindee** piece to use the [Mindee v2 inference API](https://api-v2.mindee.net).

### Problem

The current implementation uses the synchronous v1 API (`POST /v1/products/{account_name}/{api_name}/predict`), which is being superseded by the v2 async API.

### Changes

**`predict-document.ts`** — rewritten with v2 async flow:
1. **Enqueue**: `POST /v2/inferences/enqueue` (multipart/form-data)
2. **Poll**: `GET /v2/inferences/{inference_id}` every 2 seconds until `status = Processed` or `Failed` (max 45 attempts / ~90s)
3. Returns the full inference result

**Field changes**:
- `account_name` + `api_name` → replaced by `model_id` (v2 model UUID)
- Supports file URL, local file path, or base64 string input (same as before)

**`index.ts`**:
- Updated `createCustomApiCallAction` `baseUrl` from `https://api.mindee.net/v1` → `https://api-v2.mindee.net`
- Added `Harmatta` to authors

### Build & Lint

- `npx nx run pieces-mindee:build` ✅ exit 0
- TypeScript compiles cleanly

Closes #8229

### DIFF
--- a/packages/pieces/community/mindee/src/lib/actions/predict-document-v2.ts
+++ b/packages/pieces/community/mindee/src/lib/actions/predict-document-v2.ts
@@ -70,7 +70,7 @@ export const mindeePredictDocumentV2Action = createAction({
 
     const body: Record<string, unknown> = { model_id };
 
-    if (isHttpUrl(file)) {
+    if (isHttpsUrl(file)) {
       body['url'] = file;
     } else {
       body['file_base64'] = file;
@@ -97,7 +97,7 @@ export const mindeePredictDocumentV2Action = createAction({
     const job = enqueueResponse.body?.job;
     if (!job?.polling_url) {
       throw new Error(
-        `Mindee enqueue did not return a polling URL. Response: ${JSON.stringify(enqueueResponse.body)}`
+        `Mindee enqueue did not return a polling URL. Response: ${JSON.stringify(enqueueResponse.body)}`,
       );
     }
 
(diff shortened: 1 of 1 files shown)


Harmatta:
I noticed the v2 migration raises a backward-compatibility question for existing workflows. I’ve preserved the existing  action and added a separate  action locally to avoid breaking current flows. For the remaining compatibility surface: would you prefer that I keep the existing Custom API Call on the legacy v1 base URL and introduce a separate v2-specific custom API action, or is changing the existing Custom API Call to v2 acceptable for a community piece? My default bias is to preserve existing workflows unless maintainers explicitly want the migration path instead.

Harmatta:
I noticed the v2 migration still has a backward-compatibility decision point for existing workflows.

I have now preserved the existing `mindee_predict_document` action, added a separate `mindee_predict_document_v2` action, and reverted the existing Custom API Call base URL back to v1 to avoid silently breaking current users.

Question for maintainers: would you prefer that I keep Custom API Call on v1 permanently and add a separate v2-specific custom API action, or do you want the existing Custom API Call migrated to v2 later with an explicit migration path?

My default bias is to preserve existing workflows unless maintainers explicitly want the migration.

PR #12225, opened 2026-03-26
fix: improve error handling in update flow status

MrChaker:


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -671,8 +671,9 @@ export const flowService = (log: FastifyBaseLogger) => ({
                 date: apDayjs(),
             },
             customConfig: {
+                attempts: 5,
```
abuaboud: I think 2 times is enough and we will do same for the flows

PR #12231, opened 2026-03-26
fix(tables): use field externalId in find-records

hazemadelkhalel:
## Summary
- Fixed `find-records` action storing internal `field.id` instead of `field.externalId` in dropdown values, which broke git sync and project releases across environments
- Fixed `createFieldValidations` keying by `field.id` instead of `field.externalId`, causing silent validation mismatch with form data
- Added runtime resolution of `externalId` back to `field.id` for the records API

## Test plan
- [ ] Create a table with fields, use find-records in a flow with filters, verify field externalIds are stored in the flow JSON
- [ ] Sync the flow via git to another environment and verify field references resolve correctly

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/shared",
-  "version": "0.49.1",
+  "version": "0.50.0",
   "type": "commonjs",
   "sideEffects": false,
   "main": "./dist/src/index.js",


packages/server/api/src/app/flows/flow-version/migrations/migrate-v18-tables-find-records-field-ids.ts
```diff
@@ -0,0 +1,114 @@
...
+    const fieldIds: string[] = []
+
+    flowStructureUtil.getAllSteps(flowVersion.trigger).forEach((step) => {
+        if (step.type !== FlowActionType.PIECE || step.settings.pieceName !== TABLES_PIECE_NAME) {
+            return
+        }
+        if (step.settings.actionName !== FIND_RECORDS_ACTION) {
+            return
+        }
+
```
AbdulTheActivePiecer, line 29: Check the piece version as well I don't know if previous versions of this action can be automatically upgraded to this most recent version or not

packages/server/api/src/app/flows/flow-version/migrations/migrate-v18-tables-find-records-field-ids.ts
```diff
@@ -0,0 +1,114 @@
...
+                    input: {
+                        ...input,
+                        filters: {
+                            ...filters,
+                            filters: migratedFilters,
+                        },
+                    },
+                },
+            }
+        })
```
AbdulTheActivePiecer, line 122: You are forgetting to upgrade the piece version to 3.0

PR #12355, opened 2026-04-02
fix(server): bypass concurrent-edit guard for service principal calls

aguerin:
## Summary

- `assertThatFlowIsNotBeingUsed` raised a 409 FLOW_IN_USE error when the
  same API key updated a flow twice within one minute, because the service
  principal has no `userId` (`null`), which never equals `flow.version.updatedBy`
- The guard is designed to protect against *different human users* editing
  concurrently; it is meaningless for API key calls where `userId` is always `null`
- Skip the check entirely when `request.principal.type === PrincipalType.SERVICE`,
  consistent with line 119 which already sets `userId: null` for service principals

## Test plan

- [ ] Update a flow via API key twice in quick succession — no 409 returned
- [ ] Two different human users editing the same flow within 1 minute — 409 still returned
- [ ] Single human user editing their own flow — no 409 returned

### DIFF
--- a/packages/server/api/src/app/flows/flow/flow.controller.ts
+++ b/packages/server/api/src/app/flows/flow/flow.controller.ts
@@ -113,7 +113,9 @@ export const flowController: FastifyPluginAsyncZod = async (app) => {
                 PlatformUsageMetric.ACTIVE_FLOWS,
             )
         }
-        await assertThatFlowIsNotBeingUsed(flow, userId)
+        if (request.principal.type !== PrincipalType.SERVICE) {
+            await assertThatFlowIsNotBeingUsed(flow, userId)
+        }
         const updatedFlow = await flowService(request.log).update({
             id: request.params.id,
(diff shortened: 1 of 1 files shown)


aguerin:
> [CLA assistant check](https://cla-assistant.io/activepieces/activepieces?pullRequest=12355) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/activepieces/activepieces?pullRequest=12355) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/activepieces/activepieces?pullRequest=12355) it.

It was signed few days ago

image

abuaboud:
We have revamped the feature in the latest version, It's now real time and based on persons instead of fragile update timestamp.

Where you can see who is editing and you can take over the lock from the UI.

Same thing applies for the table

image

AbdulTheActivePiecer:
solved in https://github.com/activepieces/activepieces/pull/12500

PR #12385, opened 2026-04-04
refactor(promotekit): consolidate 5 triggers into single New Event trigger

bst1n:
## Summary

Refactors the PromoteKit piece to replace 5 separate event-specific triggers with a single unified "New Event" trigger that handles all webhook event types.

## Changes

- **Deleted triggers:**
  - `affiliate-approved.ts`
  - `new-commission.ts`
  - `new-referral.ts`
  - `referral-converted.ts`

- **New trigger:**
  - `new-event.ts` - Universal trigger supporting all event types:
    - `affiliate.created`
    - `affiliate.approved`
    - `referral.created`
    - `referral.converted`
    - `commission.created`

- **Fixes:**
  - Fixed auth reference: changed `auth.secret_text` to `auth as string` in common helpers
  - Added `event_type` field to sample data for event identification
  - Updated instructions to explain multi-event configuration

## Benefits

- Simplified UI with single trigger instead of five
- Reduced code duplication (307 lines removed)
- Better maintainability with centralized event handling
- Users can configure all events from a single webhook endpoint


kishanprmr:
Thanks for the contribution! 

The main issue is the test() method, it always fetches /affiliates regardless of which event the user configured, so 4 out of 5 event types will show completely wrong sample data in the builder. The old per-trigger approach was correct because each trigger owned its own test() implementation that matched its event type.

PR #12388, opened 2026-04-04
fix: update Node.js version check to require v22+ in setup-dev.js

### LINKED ISSUE #12290: [BUG]: Dev setup allows Node 18 but API crashes on startup
**Describe the bug**
[Docs](https://www.activepieces.com/docs/build-pieces/building-pieces/development-setup#prerequisites) says Node 18 is supported, but the API crashes during startup `npm start` when using Node 18

**To Reproduce**
1. Use Node v18.x
1. Run `npm start`
1. Wait for `api:serve` to start
1. See crash error: `TypeError [ERR_INVALID_ARG_TYPE]: The "original" argument must be of type function. Received undefined`

**Expected behavior**
If Node 18 is shown as supported, the API should start without crashing.
Or, if Node 18 is not supported anymore, setup should clearly fail early with a clear message.

**Additional context**
- Crash happens in `file-compressor.ts` when calling `promisify` on `zstdCompress` / `zstdDecompress` from `node:zlib`
- Those APIs exist only from Node v22.15.0 onward (see [Node 22 zlib docs](https://nodejs.org/docs/latest-v22.x/api/zlib.html#class-zlibzstdcompress))
- `tools/setup-dev.js` currently accepts v18, v22, and v24, so setup passes even though runtime fails on v18

Vitalini:
## Summary

Fixes #12290

The dev setup script `tools/setup-dev.js` was accepting Node v18 and v20, but the API server crashes at runtime due to missing `zstd` compression support.

## Root cause

`file-compressor.ts` calls `promisify(zstdCompress)` and `promisify(zstdDecompress)` from `node:zlib`. These APIs are only available from **Node v22.15.0+** (see [Node.js docs](https://nodejs.org/docs/latest-v22.x/api/zlib.html#class-zlibzstdcompress)).

With Node 18 or v20, startup crashes with:
```
TypeError [ERR_INVALID_ARG_TYPE]: The 'original' argument must be of type function. Received undefined
```

## Changes

- Updated `requiredVersions` from `['v18', 'v20']` to `['v22', 'v24']`
- Added a clearer error message explaining why Node 18/20 are no longer supported
- Added a comment documenting the reason for the version requirement

## Testing

Verified that setup correctly fails early with a helpful message when running on Node 18 or v20.

### DIFF
--- a/tools/setup-dev.js
+++ b/tools/setup-dev.js
@@ -4,8 +4,10 @@ const { execSync } = require('child_process');
 const fs = require('fs');
 
 // Check Node.js version
+// Note: Node 18 and v20 are no longer supported. Node v22+ is required because
+// zstdCompress/zstdDecompress from node:zlib are only available from Node v22.15.0+
 const nodeVersion = execSync('node --version').toString().trim();
-const requiredVersions = ['v18','v20'];
+const requiredVersions = ['v22', 'v24'];
 
 // Check operating system
 const os = process.platform;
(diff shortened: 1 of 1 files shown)


abuaboud:
Merged in another branch :D Thanks!

PR #12390, opened 2026-04-04
fix: prevent trigger badge from overlapping step name text

### LINKED ISSUE #12354: [Bug] Trigger badge overlaps step name text in flow canvas
```markdown
**Describe the bug**
In the flow canvas, the "Trigger" badge overlaps the step name text. The name starts behind the badge instead of to its right, making it partially hidden.

**To Reproduce**
1. Open any flow in the canvas editor
2. Observe the trigger step card (first node in the canvas)
3. Notice that the step name text begins underneath the "Trigger" badge rather than after it

**Expected behavior**
The step name should start to the right of the "Trigger" badge, with both elements fully visible and non-overlapping.

**Screenshots**
[attach screenshot]

**Additional context**
The step name container likely lacks a `padding-left` or `margin-left` accounting for the badge width. Affects at minimum the Tally trigger ("Nouvelle soumission") — not tested on other integrations.
```

Image

Vitalini:
## Problem

The "Trigger" badge on the flow canvas overlaps the step name text and logo, making the step name partially hidden behind the badge.

Screenshot

## Root Cause

Two issues:

1. **Missing z-index on TriggerWidget** — The badge is absolutely positioned above the step card via `-translate-y-[26px]`, but without a z-index, step card content can visually overlap it.

2. **Invalid Tailwind class** — The step node container uses `overflow-show` which is not a valid Tailwind utility. This means overflow behavior falls back to the browser default (`visible`), but the intent should be explicit.

## Solution

- Added `z-10` to the TriggerWidget component so it always renders above the step card content
- Replaced invalid `overflow-show` with `overflow-visible` on the step node container to explicitly allow the badge (positioned outside the card bounds) to remain visible

## Testing

- Verified the trigger badge no longer overlaps with step name text
- The badge correctly sits above the card border in the tab-like position
- No visual regression on non-trigger step nodes

Fixes #12354

### DIFF
--- a/packages/web/src/app/builder/flow-canvas/nodes/step-node/index.tsx
+++ b/packages/web/src/app/builder/flow-canvas/nodes/step-node/index.tsx
@@ -130,7 +130,7 @@ const ApStepCanvasNode = React.memo(
         }}
         onContextMenu={(e) => handleContextMenu(e)}
         className={cn(
-          'transition-all border-box rounded-md border border-solid border-border relative overflow-show  group',
+          'transition-all border-box rounded-md border border-solid border-border relative overflow-visible  group',
           {
             'border-primary': isSelected,
             'bg-background': !isDragging,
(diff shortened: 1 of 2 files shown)


AbdulTheActivePiecer:
@allcontributors please add @jerboa88 for code

PR #12444, opened 2026-04-07
feat: formula & data manipulation functions in flow builder

Louai-Zokerburg:
## Summary

This PR introduces a formula/function system to the Activepieces flow builder, letting users apply built-in data transformations directly inside text input fields — no code required.

- **Function registry** (`packages/shared`) — 80+ built-in functions across five categories: Text, Number, Date, List, and Logic (e.g. `uppercase`, `combine`, `round`, `format_date`, `filter_list`, `if_empty`)
- **Formula editor** (TipTap-based) — functions appear as visual badges with structured argument slots (`function_start` / `function_sep` / `function_end` inline atoms); supports `{{variable}}` interpolation alongside formula calls
- **Function search popover** — `/` opens a searchable, categorized list of all functions with syntax, description, and live example
- **Hover popover** — hovering a function badge shows its signature, description, and argument types inline
- **Type checker** — validates argument count and types at edit time, surfacing errors inline before the flow runs
- **Engine evaluation** — formulas are resolved server-side via `expr-eval` during flow execution; `{{variables}}` are pre-resolved (deduplicated — one async resolution per unique token) before the expression is evaluated
- **Tests** — evaluator, type checker, and serializer round-trip tests in `packages/shared/test/formula/` and `packages/server/engine/test/variables/`

## How it works

1. User types `/` in a text input → function search popover opens
2. Selecting a function inserts it as a structured node with argument slots; `;` separates arguments
3. On save the editor serializes the formula to a string wrapped with explicit markers: `ap-formula-v1::{uppercase({{trigger.name}})}::ap-formula-v1` (mirrored close marker so a literal `}` inside a string arg can never close the wrapper early)
4. At runtime the engine sees the wrapper, pre-resolves any `{{variable}}` references (deduplicating repeated tokens), then evaluates the expression
5. The result replaces the field value before the step executes

## Test plan

- [ ] Type `/` in a text input field and verify the function search popover opens
- [ ] Insert a text function (e.g. `uppercase`) and confirm the badge renders with argument slots
- [ ] Hover the badge and confirm the signature popover appears
- [ ] Run a flow with a formula field and verify the output is correctly transformed
- [ ] Test with a `{{variable}}` inside a formula (e.g. `uppercase({{trigger.name}})`)
- [ ] Test with the same `{{variable}}` repeated in one formula and confirm it resolves once
- [ ] Confirm type errors surface for wrong argument counts/types at edit time
- [ ] Confirm a literal `}` inside a quoted string arg does not break wrapper parsing


packages/shared/src/lib/formula/function-evaluator.ts
```diff
@@ -0,0 +1,813 @@
...
+    const parsed = dayjs(String(d ?? ''))
+    return parsed.isValid() ? parsed.toISOString() : ''
+}
+
+parser.functions.filter_list = (list: unknown, field: unknown, value: unknown) =>
+    toArray(list).filter(
+        (item) =>
+            typeof item === 'object' &&
+            item !== null &&
+            (item as Record<string, unknown>)[String(field)] == value,
```
MrChaker: @Louai-Zokerburg Are you sure this shouldn't be strict equal ( === )

packages/server/engine/src/lib/variables/props-resolver.ts
```diff
@@ -98,6 +99,27 @@ function extractReferencedStepNames(input: unknown, stepNames: string[]): Set<st
  */
 async function resolveInputAsync(params: ResolveInputInternalParams): Promise<unknown> {
     const { input, currentState, engineToken, projectId, apiUrl, censoredInput } = params
+
```
MrChaker, line 104: @Louai-Zokerburg I would suggest that we prefix these function with something like ap-formula-v1::{func()} .
1. it will help us in future for debugging/checking database . 
2. versioned prefix will allow for future migrations if ever needed

2. this is considered as breaking change . there is very high change users already have inputs that the engine will resolve as formulas 
3. I'm not sure how you're resolving the syntax of a formula inside input . but with prefix + {} wrap it makes it easy and straightforwad . no guess work

packages/shared/src/lib/formula/function-evaluator.ts
```diff
@@ -0,0 +1,813 @@
...
+parser.functions.is_equal = (a: unknown, b: unknown) => a == b
+// and/or/not are reserved operators in expr-eval — register under prefixed names
+// and normalizeExpression() rewrites them in the expression before evaluation
+parser.functions.ap_and = (a: unknown, b: unknown) => Boolean(a) && Boolean(b)
+parser.functions.ap_or = (a: unknown, b: unknown) => Boolean(a) || Boolean(b)
+parser.functions.ap_not = (a: unknown) => !a
+parser.functions.coalesce = (...args: unknown[]) =>
+    args.find((a) => a !== '' && a != null) ?? ''
+
+export function evaluateExpression({ expression, sampleData }: EvaluateExpressionParams): EvaluateExpressionResult {
```
MrChaker: I would separate ( line 290 and down ) to another file . and keep our style with ``` const formulaEvaluator = { ... } ```

packages/shared/src/lib/formula/formula-evaluator.ts
```diff
@@ -0,0 +1,517 @@
...
+
+function wrap(expression: string): string {
+    return `${FORMULA_PREFIX}${expression}${FORMULA_SUFFIX}`
+}
+
+function containsWrapper(input: string): boolean {
+    return input.includes(FORMULA_PREFIX)
+}
+
+function unwrap(template: string): string {
```
MrChaker, line 23: @Louai-Zokerburg I would just use regex matching here

packages/shared/src/lib/formula/formula-evaluator.ts
```diff
@@ -0,0 +1,517 @@
...
+            continue
+        }
+        const { result, error } = evaluateSingleFormula({ expression: seg.value, sampleData })
+        if (error) return { result: null, error }
+        parts.push(result != null ? (typeof result === 'object' ? JSON.stringify(result) : String(result)) : '')
+    }
+    return { result: parts.join(''), error: null }
+}
+
+function tokenizeFormulaTemplate(template: string): Segment[] {
```
MrChaker, line 51: @Louai-Zokerburg same here ?

packages/shared/src/lib/formula/formula-evaluator.ts
```diff
@@ -0,0 +1,517 @@
...
+    const emptyArgError = validateFunctionArgs(trimmed)
+    if (emptyArgError) return { result: null, error: emptyArgError }
+
+    const { processed, vars } = preprocessExpression({ expression: trimmed, sampleData })
+    try {
+        // expr-eval's published `Values` type is narrower than what the library
+        // accepts at runtime — arrays, null, and mixed-type objects all flow
+        // through fine. Ignore the library-type mismatch here rather than
+        // forcing a misleading cast like `as Record<string, number>`.
+        // @ts-expect-error narrow third-party type; runtime accepts unknown values
```
MrChaker: @Louai-Zokerburg if there any why to remove // @ts-expect-error would be cool . if not ok

packages/shared/src/lib/formula/formula-evaluator.ts
MrChaker: @Louai-Zokerburg I think in general . after adding the prefix. lot of things got be simplified . so just tell claude to re-think knowing that you added prefix+suffix

MrChaker: @Louai-Zokerburg I think suffix can't be just '}' . because the funtion args could contain a string that has '}' and it will break things . so maybe make the suffix same as prefix format '}:ap-formula-v1'

packages/server/api/src/app/mcp/tools/ap-add-step.ts
```diff
@@ -45,7 +45,7 @@ export const apAddStepTool = (mcp: McpServer, log: FastifyBaseLogger): McpToolDe
             parentStepName: z.string().describe('The step name to insert after/into (e.g. "trigger", "step_1"). Use ap_flow_structure to get valid values.'),
             stepLocationRelativeToParent: z.enum(Object.values(StepLocationRelativeToParent) as [string, ...string[]]).describe('Where to place the step: AFTER = after the parent, INSIDE_LOOP = first action inside a loop, INSIDE_BRANCH = first action inside a router branch'),
             branchIndex: z.number().optional().describe('Branch index (required when stepLocationRelativeToParent is INSIDE_BRANCH)'),
-            stepType: z.enum([FlowActionType.CODE, FlowActionType.PIECE, FlowActionType.LOOP_ON_ITEMS, FlowActionType.ROUTER]).describe('The type of step to add. Prefer PIECE over CODE — only use CODE when no piece exists for the task.'),
+            stepType: z.enum([FlowActionType.CODE, FlowActionType.PIECE, FlowActionType.LOOP_ON_ITEMS, FlowActionType.ROUTER]).describe('The type of step to add'),
```
AbdulTheActivePiecer: revert this

AhmadTash (approved):

PR #12488, opened 2026-04-09
fix(subflows): use internal resume URL only for internal callbacks

### LINKED ISSUE #12386: [BUG]: Test panel stuck on "Testing..." / "Agent running..." despite backend completing successfully (fix found: comments)
**Describe the bug**

When testing AI steps in the flow builder, the test panel gets permanently stuck — even though the backend executes successfully and delivers results via WebSocket.

- **AI Agent steps**: UI shows "Agent running..." / "Agent is thinking" forever
- **Other AI steps** (e.g. Classify Text): UI shows "Testing..." forever

Image

Image

In both cases, the WebSocket `UPDATE_RUN_PROGRESS` event arrives in the browser with a correct `"status": "SUCCEEDED"` (or `"FAILED"` with a proper error). The frontend simply never processes these messages into the UI. There is a disconnect between the backend and the frontend.

**Steps to reproduce**

1. Create a flow with any AI piece step (Agent, Classify Text, etc.)
2. Click "Test" / "Generate Sample Data"
3. Observe: the backend executes, the WebSocket delivers the result (visible in DevTools > Network > WS tab)
4. The UI stays stuck on "Testing..." or "Agent running..." indefinitely

**What I've tested / ruled out**

- Multiple models: gpt-4o-mini, o4-mini, gpt-5.4, gpt-5.4-mini — same result on all
- Accessed via **HTTPS** through Tailscale Funnel (reverse proxy) — stuck
- Accessed via **direct HTTP** (`http://<ip>:8080`, still via tailscale) — also stuck
- Could not test `localhost` as I'm remote, but the WebSocket is clearly connected and delivering messages regardless of access method
- The WebSocket shows both successful and failed runs — the data is there, the UI just doesn't pick it up

It might just be that my Tailscale-connection is to blame here, but everything else seems to work flawlessly - just testing these workflows don't work for some reason, evidently just a disconnect between UI and backend.

**WebSocket evidence**

Via `ws://100.125.116.102:8080/api/socket.io/?EIO=4&transport=websocket`
Image

The socket.io connection receives the correct `UPDATE_RUN_PROGRESS` payload. Example of a successful run:

```json
[
  "UPDATE_RUN_PROGRESS",
  {
    "step": {
      "name": "step_3",
      "output": {
        "type": "PIECE",
        "status": "SUCCEEDED",
        "output": "relevant_user",
        "duration": 540.93
      }
    },
    "flowRun": {
      "status": "RUNNING",
      "environment": "TESTING"
    }
  }
]
```

The UI never updates despite this message arriving.

**Environment**

- Activepieces version: **0.81.3** (also reproduced on 0.81.0)
- Deployment: Docker (self-hosted)
- `AP_EXECUTION_MODE`: `UNSANDBOXED`
- Browser: Chrome
- piece-ai version: `@activepieces/piece-ai@0.3.0`

**Bonus: separate bug in Classify Text**

While debugging this, I also found that the "Classify Text" action (`piece-ai@0.3.0`, `classify-text.js:48`) has a broken prompt — it tells the model to "respond with only the selected category as a single word", then does a strict exact match against the full category string. Multi-word categories like "Likely a customer" will always fail with `Error: Unable to classify the text into the provided categories`. Using single-word category labels works around this. I can create another ticket on this as well - my guess is that if the WebSocket messages got through to the UI, it would tell me/the user what the issue is, and I'd see "Oh I need to make it a single word", but thought it'd be worth mentioning.

---

Thanks for this awesome software and all the work you put in to creating it!

cafeco09:
## What does this PR do?

This PR keeps `generateResumeUrl()` public by default, and adds an optional `visibility` parameter so internal-only callback paths can explicitly request an internal URL.

It then updates the sub-flow `callFlow` action to use `visibility: 'internal'` for its internal callback URL when `waitForResponse` is enabled.

`generateResumeUrl()` now accepts:

- `visibility: 'public'` (default)
- `visibility: 'internal'`

Existing approval and externally shared resume links continue using the public URL by default.

The sub-flow callback path is the internal-only caller that now opts into `visibility: 'internal'`, so it can resume within containerised/self-hosted deployments where internal and public URLs differ.

### Relevant User Scenarios

- Sub-flows using **Wait for Response** in self-hosted/containerised deployments
- Internal callback flows where the public URL is not reachable from inside the container
- Approval links and other externally shared resume URLs continue to work unchanged because the default remains public

Fixes #12386

### DIFF
--- a/packages/pieces/core/subflows/src/lib/actions/call-flow.ts
+++ b/packages/pieces/core/subflows/src/lib/actions/call-flow.ts
@@ -126,7 +126,8 @@ export const callFlow = createAction({
       body: {
         data: payload,
         callbackUrl: context.propsValue.waitForResponse ?  context.generateResumeUrl({
-          queryParams: {}
+          queryParams: {},
+          visibility: 'internal'
         }) : undefined,
       },
     });
--- a/packages/pieces/framework/src/lib/context/index.ts
+++ b/packages/pieces/framework/src/lib/context/index.ts
@@ -206,7 +206,8 @@ type BaseActionContext<
   run: RunContext;
   generateResumeUrl: (params: {
     queryParams: Record<string, string>,
-    sync?: boolean
+    sync?: boolean,
+    visibility?: 'public' | 'internal'
   }) => string;
 };
 
--- a/packages/server/engine/src/lib/handler/piece-executor.ts
+++ b/packages/server/engine/src/lib/handler/piece-executor.ts
@@ -149,7 +149,9 @@ const executeAction: ActionHandler<PieceAction> = async ({ action, executionStat
                 const visibility = params.visibility ?? 'public'
                 const baseApiUrl =
                     visibility === 'internal'
-                        ? (constants.internalApiUrl ?? constants.publicApiUrl)
+                const baseApiUrl =
+                    visibility === 'internal'
+                        ? constants.internalApiUrl
                         : constants.publicApiUrl
 
                 const path = `v1/flow-runs/${constants.flowRunId}/requests/${executionState.pauseRequestId}${params.sync ? '/sync' : ''}`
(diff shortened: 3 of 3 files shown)


abuaboud:

PR #12502, opened 2026-04-09
feat(promotekit): replace single webhook trigger with 3 entity-specific triggers

bst1n:
## Summary

Replaces the generic `new_event` trigger with 3 dedicated entity-specific triggers, each with a multi-select dropdown and a `test()` function that fetches sample data from the correct endpoint:

- **New Affiliate Event** — `affiliate.created` / `affiliate.approved`, `test()` → `/affiliates`
- **New Referral Event** — `referral.created` / `referral.converted`, `test()` → `/referrals`
- **New Commission Event** — `commission.created`, `test()` → `/commissions`

The multi-select filter is optional: leave empty to catch all events for that entity. Also fixes all i18n gaps — strings were previously mixed EN/FR in the UI.

## Breaking change note

`new_event` is removed. This is intentional and safe: the PromoteKit piece was introduced in a prior PR by the same author (this account) and has never been used in production. No existing flows depend on `new_event`.

## Test plan

- [ ] Open each trigger in a French workspace → all labels, descriptions and setup instructions are in French
- [ ] For **New Affiliate Event**: select 0, 1 or 2 event types → "Load Sample Data" returns affiliate records tagged with the correct `type`
- [ ] For **New Referral Event**: same check with referral records
- [ ] For **New Commission Event**: "Load Sample Data" returns commission records tagged `commission.created`
- [ ] Send a real webhook from PromoteKit → trigger fires and output contains the correct `type` field


kishanprmr:
@bst1n The event filter dropdown is required to fetch sample data (last 5–10 records) from Promptekit.

bst1n:
> @bst1n The event filter dropdown is required to fetch sample data (last 5–10 records) from Promptekit.

@kishanprmr Thanks for the review. I respectfully disagree with the dropdown being required, for the following reasons:

1. PromoteKit does not expose a programmatic webhook management API. Unlike pieces such as Stripe or Shopify where Activepieces can register a filtered webhook for the user, here the user has to manually create the endpoint in the PromoteKit dashboard. This is fundamentally the same setup as the Senja piece, where the same constraint applies.

2. Because webhook creation is manual, the user already selects the event types they want directly in the PromoteKit UI when they create the endpoint. PromoteKit only sends the events the user checked — so the payload reaching Activepieces is always already filtered at the source. A second filter on the Activepieces side duplicates that choice and creates a silent misconfiguration trap: if the user enables an event in PromoteKit but forgets to re-select it in the Activepieces dropdown (or vice versa), the trigger never fires with no visible error.

3. Each Activepieces flow gets its own unique webhook URL (/api/v1/webhooks/{flowId}). So if a user wants different flows for different events, they create separate flows and add separate endpoints in PromoteKit, each filtered to the desired events. No need for an Activepieces-side filter to route events.

4. The dropdown is not required to fetch sample data. Senja, which has the exact same manual-webhook constraint, implements test() by calling a default endpoint (/testimonials) without consulting any dropdown value. This PR follows the same approach: test() fetches /affiliates and returns them tagged as affiliate.created. Users can refine their flow by triggering an actual webhook from PromoteKit if they need sample data for another event type.

Happy to reconsider if I'm missing something specific to PromoteKit that doesn't apply to Senja, but as it stands, imposing the dropdown here feels inconsistent with how other pieces handle the same "no programmatic webhook management" constraint.

kishanprmr:
> The dropdown is not required to fetch sample data. Senja, which has the exact same manual-webhook constraint, implements test() by calling a default endpoint (/testimonials) without consulting any dropdown value. This PR follows the same approach: test() fetches /affiliates and returns them tagged as affiliate.created. Users can refine their flow by triggering an actual webhook from PromoteKit if they need sample data for another event type.

Previously, each trigger had its own endpoint to fetch sample data, so users didn’t need to create dummy data in their live app just to test the trigger.

If the API provides a list endpoint for each entity that can have a trigger, then a `test()` method should be implemented accordingly.

In the current approach, if a user selects any trigger event other than affiliates, the trigger still returns test data for affiliates, which is incorrect.

To determine the correct endpoint, a dropdown is necessary so we know which specific entity’s trigger is being tested.

bst1n:
> > If the API provides a list endpoint for each entity that can have a trigger, then a `test()` method should be implemented accordingly.
> 
> In the current approach, if a user selects any trigger event other than affiliates, the trigger still returns test data for affiliates, which is incorrect.
> 
> To determine the correct endpoint, a dropdown is necessary so we know which specific entity’s trigger is being tested.

You're right — I was overcomplicating this. After thinking it through, your rationale is correct: since PromoteKit exposes list endpoints for each entity, test() should leverage them to avoid forcing users to create dummy events in their live account, which is the standard Activepieces convention. The dropdown is the natural way to drive that.

I've pushed a revision that:

Keeps the dropdown and uses it to route test() to the correct endpoint per entity (/affiliates, /referrals, /commissions)
Fixes the i18n issue that started this PR (mixed EN/FR labels on the dropdown)
Adds a type field to the trigger output so flows know which event fired
Ready for another look.

kishanprmr (approved):

PR #12509, opened 2026-04-09
feat: Think + Work — AI chat experience

hazemadelkhalel:
## Summary

AI Chat where users interact with an automation assistant powered by
Claude Code running inside E2B cloud sandboxes. The agent has direct
access to project tools via MCP — it can list flows, build automations,
manage tables, query data, and troubleshoot issues through natural
conversation.

### Key Features

- **Chat as project tab** — accessible from the sidebar with Beta badge
- **Cloud + Self-hosted** — cloud uses OpenRouter directly, self-hosted
  uses direct Anthropic API key
- **File attachments** — drag-and-drop images, CSV, JSON, Markdown,
  plain text (10 MB limit, MIME allowlist)
- **Rich markdown** — headings, lists, tables, code blocks with syntax
  highlighting, source link pills
- **Structured agent output** — automation proposal cards, connection
  cards with inline auth dialog, quick-reply chips
- **Full streaming** — thinking timer, tool call groups (chain of
  thought), plan progress, cancel with session destroy
- **Conversation management** — sidebar list, cursor-based pagination,
  per-user scoping, delete with session cleanup

### Agent Runtime (E2B + sandbox-agent)

- Claude Code runs in [E2B](https://e2b.dev) cloud sandboxes via the
  [sandbox-agent](https://www.npmjs.com/package/sandbox-agent) SDK
- **No shared SDK cache** — each operation creates its own SandboxAgent
  for full isolation between users (scales to 1000+ concurrent users)
- Sessions persist via custom Postgres driver (keyset pagination)
- Concurrency semaphore (max 20 concurrent sandbox creations) with
  30s timeout to prevent E2B API exhaustion
- Auto-approve permissions via `onPermissionRequest` handler
- Model: `opus[1m]` (Claude Opus with 1M context)

### OpenRouter Integration (Cloud Path)

Cloud users route through the Activepieces provider (OpenRouter):
- `ANTHROPIC_BASE_URL` set to `openrouter.ai/api` directly
- No proxy needed — OpenRouter handles model routing natively
- Self-hosted path: direct Anthropic API key, no base URL override

### SSE Streaming Pipeline

```
E2B Sandbox → session.onEvent → historyReplayFilter → streamWriter → UIMessageStream → SSE → Client
```

- **25s keepalive** heartbeat prevents reverse-proxy idle timeouts
- **History replay filter** — state machine (detecting → suppressing →
  passthrough) that strips replayed events on session resume
- **Stream writer** — maps sandbox events to AI SDK stream protocol
  (text, reasoning, tool calls, plans, usage, session title)
- **Title update** — deferred to stream end, skipped on cancel
- **Client disconnect** — resolves streaming promise, logs at debug
  level (not error)

### Cancel Flow

When user clicks Stop:
1. Frontend calls `stop()` (closes SSE) + `POST /cancel`
2. Server destroys the E2B sandbox session via SDK
3. Clears `sandboxSessionId` on conversation + deletes session/event
   DB rows
4. Running tool calls shown with pause icon (not spinner)
5. `SandboxDestroyedError` logged at debug level

### Structured Agent Output

| Block | Renders as |
|---|---|
| `` ```automation-proposal `` | Card with title, steps, "Build this" CTA |
| `` ```connection-required `` | Piece icon + "Connect" button (inline dialog) |
| `` ```quick-replies `` | Clickable choice chips |

Structured blocks parse during streaming (not just after) so they
appear as soon as the closing fence arrives.

### Scalability

- **No shared sandbox** — each conversation gets isolated E2B sandbox
- **History bypasses E2B** — reads events directly from Postgres via
  async generator (no sandbox slot consumed)
- **AI config cached** — per-platform with 5min TTL (eliminates 2 DB
  queries per message)
- **30s SDK timeout** — orphaned sandboxes cleaned up if timeout wins
- **Concurrency semaphore** — max 20 simultaneous sandbox creations
- **DB cleanup** — session + event rows deleted on cancel/delete
- **Keyset pagination** — for both events and conversation list

### Security

- `securityAccess` on every endpoint (WRITE_CHAT / READ_CHAT)
- Conversation ownership enforced (projectId + userId on all queries)
- System prompt injection prevention via `sanitizeProjectName`
- File MIME type allowlist with size limits
- `chatEnabled` feature flag gated per platform plan
- Response header filtering on external API calls

### Database

- `chat_conversation` — per-user conversations with sandbox session
  link
- `sandbox_session` + `sandbox_event` — E2B session persistence
- Composite index `(projectId, userId, created DESC, id DESC)` for
  conversation listing
- Unique index on `mcp_server.token` (CONCURRENTLY with PGlite guard)
- `chatEnabled` column on `platform_plan` (default false for CE)
- System prompt loaded from markdown file via `__dirname` path

## Test Coverage

**91 server unit tests** across 3 test files:

- `ai-event-utils` — content extraction, replay detection, tool output
- `stream-adapter` — stream writer event mapping, history replay filter
  state machine
- `sandbox-agent` — session update type constants

## Test Plan

- [ ] Cloud: send message via OpenRouter, verify streaming response
- [ ] Self-hosted: configure Anthropic API key, verify direct path
- [ ] File upload: attach image + CSV, verify agent receives them
- [ ] Streaming: verify thinking timer, tool call groups, plan cards
- [ ] Conversation list: create, switch, delete conversations
- [ ] Connection flow: agent requests connection, user connects inline
- [ ] Quick replies: verify chips render and send on click
- [ ] Cancel: click stop, verify sandbox destroyed and tools show pause
- [ ] Long session: verify SSE stays alive past 60s (keepalive)
- [ ] History reload: switch conversation, verify messages load
- [ ] Feature gate: verify chat hidden when chatEnabled is false


hazemadelkhalel (commented):
## Greptile Review Comments — All Resolved

| # | Issue | Status |
|---|-------|--------|
| 1 | Filename collision silently drops file content | **Resolved** — `writeFilesToDisk` removed. Text files inlined, binary skipped. |
| 2 | `initPromises` never cleaned up | **Resolved** — replaced Maps with single `cachedSdk`/`cachedSdkPromise`. |
| 3 | History replay filter leaks early chunks | **Resolved** — filter now buffers and suppresses ALL event types during detection. |
| 4 | MCP server runs as `project.ownerId` | **N/A** — agent controller removed. Pre-existing behavior in OAuth controller. |
| 5 | Binary files decoded as UTF-8 | **Resolved** — only text MIME types decoded; binary types skipped. |
| 6 | Stop button doesn't abort during creation | **Resolved** — uses `useChat`'s `stop()` which properly aborts fetch. |
| 7 | Anthropic API key never reaches agent | **Resolved** — passed via `envs: { ANTHROPIC_API_KEY }` to E2B sandbox. |
| 8 | History replay only suppresses text chunks | **Resolved** — all event types suppressed during detection/suppression states. |
| 9 | DB pool not closed on dispose | **Resolved** — persist driver uses shared TypeORM DataSource. No separate pool. |
| 10 | Suppression exits after first non-marker chunk | **Resolved** — requires 3+ non-replay chunks before transitioning to passthrough. |
| 11 | TOCTOU race in `ensureSession` | **Resolved** — atomic `WHERE sandboxSessionId IS NULL` conditional update. |

hazemadelkhalel:
**Re: listSessions full-table scan** — `listSessions()` implements the `SessionPersistDriver` interface required by the `sandbox-agent` library. The library's interface doesn't accept tenant context, so we can't add a WHERE clause. This method is **never called** in our code path — the library only calls `getSession(id)`, `updateSession()`, `insertEvent()`, and `listEvents(sessionId)`. The `listSessions` implementation exists purely to satisfy the interface contract. Access control is enforced at the conversation layer (all queries filter by `projectId` + `userId`).

hazemadelkhalel:
**Re: Unauthenticated proxy endpoint** — This endpoint is `public()` by necessity: the E2B sandbox (running Claude Code) sends requests with standard Anthropic headers (`x-api-key`, `anthropic-version`). Claude Code's HTTP client cannot be configured to send custom auth headers.

**Mitigations in place:**
- Hostname hardcoded to `openrouter.ai` — cannot be used for SSRF
- Path hardcoded to `/api/v1/messages` — cannot reach other endpoints
- API key validated (401 on missing `x-api-key`)
- Response headers filtered to allowlist (no upstream info leak)
- Callers must supply their own valid OpenRouter API key — no credential leakage

The endpoint functions as a model-name rewriter, not an open relay. An attacker with their own API key gains nothing by proxying through us vs calling OpenRouter directly.

hazemadelkhalel:
**Re: destroySession creates a new sandbox** — The `sandbox-agent` SDK's `SandboxAgent.start()` with the E2B provider uses the `connect` callback to reconnect to existing sandboxes via the `sandbox_id` stored in the persist driver's session record. When `destroySession(sessionId)` is called, the SDK reads the session from postgres, finds its `sandbox_id`, connects to that sandbox via the `connect` callback, and destroys it. It does NOT create a brand-new sandbox — the `create` callback is only used when `createSession` needs a new sandbox.

hazemadelkhalel:
**Re: Post-replay response head silently dropped** — This is a known trade-off in the replay filter design. The `shouldSuppress` API is a pure boolean — it can't retroactively unsuppress events that were already returned as `true`. The ~200 chars are a detection window to distinguish replay text from real content.

In practice, this only affects session resume, and the lost prefix is a small portion of the agent's continuation. The full content is persisted in the event store and loads correctly on history reload.

Fixing this properly requires changing the filter's return type from `boolean` to `{ suppress: boolean, flush?: Event[] }` so it can emit buffered events on transition. This is a worthwhile improvement but beyond the scope of this PR — tracked for follow-up.

PR #12575, opened 2026-04-14
fix: strip CORS headers from webhook response.headers to prevent duplicates

madhukar-tradaisdigital:
Calling a webhook's `/sync` endpoint from a browser via `fetch()` fails with:

> The 'Access-Control-Allow-Origin' header contains multiple values '*, *', but only one is allowed.

The response contains two separate `Access-Control-Allow-Origin` headers. The global Fastify CORS plugin sets one, and the webhook controller sets another by forwarding `response.headers` from the flow execution directly to the reply — which may include CORS headers if the flow itself made HTTP requests that returned them.

Fastify's `.headers()` appends rather than replaces, so you end up with both. Browsers concatenate duplicate headers as `*, *` and reject the response even though the server returns 200.

Reproduction (self-hosted, no CORS headers in nginx):

```
curl -sI -X POST \
  'https://<host>/api/v1/webhooks/<flow-id>/sync' \
  -H 'Origin: https://your-frontend.app'
```

Response:
```
access-control-allow-origin: *       <- global CORS plugin
access-control-expose-headers: *
Access-Control-Allow-Origin: *       <- forwarded from response.headers
Access-Control-Allow-Methods: GET, POST, OPTIONS
```

The fix adds a `filterCorsHeaders` helper that strips all CORS-related headers from the flow's response before passing them to `reply.headers()`. CORS is already handled globally by the plugin — the webhook controller does not need to forward them.

### DIFF
--- a/packages/server/api/src/app/webhooks/webhook-controller.ts
+++ b/packages/server/api/src/app/webhooks/webhook-controller.ts
@@ -1,4 +1,3 @@
-
 import {
     RAW_PAYLOAD_HEADER,
     WebhookUrlParams,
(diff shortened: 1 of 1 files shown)


madhukar-tradaisdigital:
@AbdulTheActivePiecer @abuaboud would appreciate a review on this when you get a chance.

abuaboud:

PR #12607, opened 2026-04-14
feat: add env var to deactivate personal projects

AdamSelene:
## What does this PR do?

We have a dual staging + production setup, and we want to disable personal projects on staging
This adds an env var to hide all personal projects from the UI / API

### Relevant User Scenarios

Fixes # (issue)


majewskibartosz:
decision has been made that we'll handle this internally, thanks for contribution nonetheless.

PR #12645, opened 2026-04-16
feat(baserow): auto-create select options, Row Event trigger, and real sample data for triggers

bst1n:
## Summary

This PR bundles three improvements to the Baserow piece:

### 1. Auto-create missing select options (Create/Update Row actions)

A new **"Create missing select options"** toggle (disabled by default) on the **Create Row** and **Update Row** actions. When enabled, values sent for single/multi select fields that aren't yet defined as options are automatically created on the field before the row is inserted/updated.

Baserow's API strictly rejects unknown select values (confirmed by a Baserow developer — unlike Airtable's `typecast: true`, there is no native API parameter). The only workaround is a separate `PATCH /api/database/fields/{field_id}/` call first, which this feature handles transparently.

### 2. New **Row Event** combined trigger

A unified trigger that reacts to **any** row change (created, updated, or deleted) in a single flow. Users who need to filter by event type continue to use the dedicated Row Created / Row Updated / Row Deleted triggers — the combined trigger has no event filter prop, to avoid UI redundancy.

Payload shape: `{ event_type, row, previous_row }`.

### 3. Real sample data via `async test()` on all triggers

All 7 webhook triggers (3 single-row + 3 batch + Row Event) now implement `async test()`, which fetches the 10 most recent rows from the selected table via the REST API. Clicking "Test trigger" in the flow builder now returns actual data matching the user's table structure, instead of the previous hardcoded placeholder (`{ id: 1, Name: 'Example row' }`) that made field mapping unusable.

## Why

- Unknown select values are a common friction point in automations where input data is dynamic
- Users with flows that should react to any row change had to duplicate their logic across three flows or use a generic Activepieces webhook
- The test/mapping UX for triggers was broken — placeholder data didn't reflect real table structure, forcing users to manually trigger events in Baserow to capture a real payload

## Implementation details

**Actions**
- `BaserowClient` gains `updateFieldSelectOptions({ fieldId, newOptions, existingOptions })` which batches all missing options for a field into a single PATCH
- Exported helper `ensureSelectOptionsExist({ fields, payload, client })` in `common/index.ts` detects missing values per field and delegates to the client, with `tryCatch` for go-style error handling and `console.error` on failure
- `BaserowSelectOption` extracted as a named interface in `types.ts`

**Triggers**
- New `row-event.ts` registering a webhook for all 3 Baserow row events, with payload normalization handling the different shape Baserow sends for `rows.deleted` (`row_ids` vs `items`)
- `async test(context)` added to all 7 triggers, calling `client.listRows(table_id, 1, 10)` and shaping the result to match each trigger's `run()` output

**i18n**
- New keys for the Row Event trigger added to all 13 language files (ca, de, es, fr, hi, id, ja, nl, pt, ru, translation, vi, zh), with full translations for locales that have translated descriptions and English fallbacks for the rest

**Version**: bumped `@activepieces/piece-baserow` from 0.4.1 to 0.5.0 (minor — new trigger + new action feature).

## Test plan

**Select options auto-create**
- [ ] Create Row with an unknown single select value + toggle enabled → row is created, option appears on the field
- [ ] Create Row with multiple unknown multi select values for the same field → row is created, all options appear (single PATCH, no duplicated IDs)
- [ ] Toggle disabled → Baserow rejects unknown values as before
- [ ] Option creation fails → row operation is not aborted, error is logged

**Row Event trigger**
- [ ] Creating a row fires the trigger with `event_type: 'rows.created'`
- [ ] Updating a row fires with `event_type: 'rows.updated'` and `previous_row` populated
- [ ] Deleting a row fires with `event_type: 'rows.deleted'` and `row: { id }`
- [ ] Clicking "Test" returns 10 real rows from the table

**Test sample data**
- [ ] Each trigger's Test button returns real rows from the selected table, in the expected shape


bst1n:
Closing this PR — it's based on the JWT auth architecture that was removed in #12759, so it has unresolvable conflicts with main. I'll re-open a fresh, smaller PR with the same features adapted to the new architecture (auth array pattern à la Umami).

PR #12669, opened 2026-04-18
feat(pieces-framework): add AI metadata fields to pieces to help mcps

ibrahim-abuznaid:
…and properties

Adds five new optional fields so piece authors can enrich actions and triggers for LLM/agent use without breaking any existing pieces.

New fields on ActionBase/TriggerBase: descriptionForLLM, tags, difficulty
New field on ActionBase only: outputSchema
New field on BasePropertySchema: example
New exported type: ActionResult<T> for standardized run() return shape

All fields are optional — zero breaking changes to existing 643+ pieces.

### Relevant User Scenarios

Fixes # (issue)


packages/pieces/framework/src/lib/action/action.ts
```diff
@@ -30,6 +30,10 @@ type CreateActionParams<PieceAuth extends PieceAuthProperty | PieceAuthProperty[
   auth?: PieceAuth
   displayName: string
   description: string
+  descriptionForLLM?: string
+  outputSchema?: Record<string, unknown>
+  tags?: string[]
+  difficulty?: 'easy' | 'medium' | 'hard'
```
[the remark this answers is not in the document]
AbdulTheActivePiecer: @ibrahim-abuznaid please make it an enum and use it in all places that you need it

packages/pieces/framework/src/lib/action/action.ts
```diff
@@ -43,6 +47,10 @@ export class IAction<PieceAuth extends PieceAuthProperty | PieceAuthProperty[] |
     public readonly name: string,
     public readonly displayName: string,
     public readonly description: string,
+    public readonly descriptionForLLM: string | undefined,
+    public readonly outputSchema: Record<string, unknown> | undefined,
+    public readonly tags: string[] | undefined,
+    public readonly difficulty: 'easy' | 'medium' | 'hard' | undefined,
```
AbdulTheActivePiecer: If all of these fields are for llm, please encapsulate them in a single object and call it something like infoForLLM and add a jsdoc explaining it and why we need it.


And please think about outputSchema, sometimes you need to tell the LLM it is dynamic, I.e the column names of a table, maybe making it a string is better so you can do both tell it the output is dynamic when needed and give it an example of how it can possibly look, and when static you can also just give it a stringified string.

AbdulTheActivePiecer (changes requested):

packages/pieces/framework/src/lib/action/action.ts
ibrahim-abuznaid: Both make sense — agreed on both. I'll update this PR rather than close + reopen, since the change is iterating on the same contract you've been
  reviewing.

  infoForLLM encapsulation — will nest descriptionForLLM, outputSchema, tags, and difficulty under a single optional object on actions and triggers. 

  outputSchema as string — you're right. Walking the pieces I've already enriched locally, the dynamic case breaks Record<string, unknown> exactly as you
  described — HTTP responses, Google Sheets rows, Postgres queries don't have a shape known at design time, so a typed object would be lying to the LLM.

ibrahim-abuznaid:
Closing — superseded by a broader Phase 1 PR coming next that lands the full AI-ready framework slice (audience, sampleData, infoForLLM with description + outputSchema + idempotent, ap_proxy_request MCP tool, codemods, scoped lint, aiReady opt-in).

PR #12695, opened 2026-04-20
fix: ignore canvas shortcuts when focus is inside an input field

### LINKED ISSUE #12535: [BUG]: ctrl + E while editing steps shouldn't skip them
**Describe the bug**
if someone presses ctrl + E while editing a step inputs they will skip the step.

octo-patch:
Fixes #12535

## Problem

Canvas keyboard shortcuts (e.g. **Ctrl+E** to skip a step, **Ctrl+C** to copy, **Shift+Delete** to delete) fire even when the user is actively typing inside a step's input field, textarea, or any `contenteditable` element.

This means pressing Ctrl+E while editing a text input unexpectedly skips the selected step, and can also trigger `ENTITY_NOT_FOUND` errors.

## Solution

Added an early return guard at the top of the `handleKeyDown` callback in `shortcuts.ts` that skips all shortcut processing when the keyboard event originates from an editable element (`INPUT`, `TEXTAREA`, or `contenteditable`).

```ts
const target = e.target as HTMLElement;
const isTyping =
  target.tagName === 'INPUT' ||
  target.tagName === 'TEXTAREA' ||
  target.isContentEditable;
if (isTyping) {
  return;
}
```

## Testing

1. Open a flow in the builder and select a step.
2. Click into one of the step's input fields and type something.
3. Press **Ctrl+E** — the step should **not** be skipped.
4. Click outside the input to return focus to the canvas, then press **Ctrl+E** — the step should be skipped as expected.

### DIFF
--- a/packages/web/src/app/builder/shortcuts.ts
+++ b/packages/web/src/app/builder/shortcuts.ts
@@ -36,6 +36,15 @@ export const useHandleKeyPressOnCanvas = () => {
 
   const handleKeyDown = useCallback(
     (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement;
+      const isTyping =
+        target.tagName === 'INPUT' ||
+        target.tagName === 'TEXTAREA' ||
+        target.isContentEditable;
+      if (isTyping) {
+        return;
+      }
+
       const insideSelectionRect =
         e.target instanceof HTMLElement &&
         e.target.classList.contains(


AbdulTheActivePiecer:
Hi @octo-patch, thanks for your pr, this fixes the issue, but why not just make sure that when the shortcut is about to run that the focused element is the flow canvas tab and not something else? that would be better, because let's say someone selects the text in the description and hits a ctrl +e they would still skip the step.

abuaboud:

PR #12712, opened 2026-04-20
feat: add AWS Bedrock as a platform AI provider

Louai-Zokerburg:
Wires Bedrock into the Universal AI provider system so platform admins can expose Bedrock-hosted models (Claude, Llama, Titan, Mistral, etc.) through the AI piece and agent model picker. Credentials are stored as an access key / secret pair in auth with the AWS region in config (mirroring the Azure apiKey+resourceName split). Models are listed dynamically via ListFoundationModelsCommand; the existing standalone amazon-bedrock piece is unchanged.

Auth narrowing in ai-sdk.ts is updated to destructure apiKey via a BaseAIProviderAuthConfig cast per case, since the widened union now includes Bedrock's non-apiKey shape. The update-mutation auth check in the upsert dialog and ai-provider-hooks is generalized from apiKey?.length > 0 to a provider-agnostic 'any string field filled' helper so editing Bedrock without re-entering secrets works.

### Relevant User Scenarios

Fixes # (issue)


packages/pieces/community/ai/package.json
```diff
@@ -1,13 +1,14 @@
...
-  "version": "0.3.5",
+  "version": "0.3.6",
   "type": "commonjs",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
     "@activepieces/pieces-common": "workspace:*",
     "@activepieces/pieces-framework": "workspace:*",
     "@activepieces/shared": "workspace:*",
+    "@ai-sdk/amazon-bedrock": "^3.0.0",
```
AbdulTheActivePiecer: remove the ^ please, don't include any wild cards in pices

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -68,7 +68,7 @@ export const aiProviderService = (log: FastifyBaseLogger) => ({
     async listModels(platformId: PlatformId, provider: AIProviderName): Promise<AIProviderModel[]> {
         const { config, auth } = await this.getConfigOrThrow({ platformId, provider })
 
-        const cacheKey = `${provider}-${auth.apiKey}`
+        const cacheKey = `${provider}-${Object.values(auth).join('|')}`
```
AbdulTheActivePiecer: this can be problematic in the future without us knowing, i.e having an auth having lots of properties and now the cache key is too long, make a factory function with a switch on the provider type and for aws bedrock, join the two auth properties it has and keep "-" as the joiner.

packages/server/api/package.json
```diff
@@ -10,6 +10,7 @@
     "@activepieces/pieces-framework": "workspace:*",
     "@activepieces/server-utils": "workspace:*",
     "@activepieces/shared": "workspace:*",
+    "@ai-sdk/amazon-bedrock": "^3.0.0",
```
AbdulTheActivePiecer: same thing here

packages/pieces/community/ai/package.json
AbdulTheActivePiecer: make sure you check the other places where you added deps to have no wild cards, and make sure you use the version that's pinned in bun.lock so we don't have inconsistencies.

AbdulTheActivePiecer (changes requested):

packages/pieces/community/ai/package.json
Louai-Zokerburg: :ok_hand:

packages/server/api/src/app/ai/ai-provider-service.ts
Louai-Zokerburg: :ok_hand:

packages/server/api/package.json
Louai-Zokerburg: :ok_hand:

AbdulTheActivePiecer (approved):

Issue #12704, opened 2026-04-20
[BUG]: Router step incorrectly marked as incomplete (converted to project)


AbdulTheActivePiecer:
This issue has been converted to a project [[BUG]: Router step incorrectly marked as incomplete](https://linear.app/activepieces/project/bug-router-step-incorrectly-marked-as-incomplete-817b8df79e2d/overview)

---

### Problem 1: False "Incomplete" error

* The Router step shows an "Incomplete" error and cannot be tested, even though a condition/branch has been added
* After multiple hard refreshes, the step eventually works and can be tested successfully
* It should not error on the first try

### Problem 2: Slow first test after hard refresh

* After a hard refresh, testing the Router step for the first time loads extremely slowly
* After another refresh/retry, it works fine

PR #12714, opened 2026-04-21
fix: router step was always invalid after refreshing and selecting step

AbdulTheActivePiecer:


AbdulTheActivePiecer:
fixes #12704

PR #12748, opened 2026-04-23
feat(rate-limiter): promote-on-release waitlist for concurrent pools

abuaboud:
## Summary

Replaces the exponential-backoff retry loop in `rate-limiter-interceptor` with an event-driven **promote-on-release waitlist**. When a pool is full, new runs are pushed to a Redis sorted-set waitlist instead of being rejected with a doubling delay. The moment a running job finishes, atomic Lua pops the next waiter and BullMQ's `Job.promote()` wakes it immediately. A bounded safety-net delay re-dispatches any waiter the normal path misses.

**Why:** the previous design rejected rate-limited runs with delays that doubled each retry (20s → 40s → … → 10min). A freshly-freed slot sat idle until the next retry timer fired. The new path promotes waiters within milliseconds of a release on the happy path.

## What changed

- `concurrency-pool-redis.ts` *(new)*: two atomic Lua scripts — `ACQUIRE_OR_ENQUEUE_LUA` and `RELEASE_AND_POP_LUA` — plus a plain `ZREM`-backed `dropPromotedWaiter` helper. The acquire script also opportunistically drains the waitlist into the active set whenever it finds capacity (this is what handles a pool that became wedged with stale entries — there is no separate self-heal primitive). Helpers: `encodeMember` / `decodeMember`. `forPool(...)` returns the public `tryAcquire` / `release` API and owns the cascade-on-failed-promote loop.
- `rate-limiter-interceptor.ts`: replaces inline Lua + exponential delay with the new primitives. Rate-limited runs get a fixed **safety-net delay** (flow timeout + 120s) that only fires if the normal promote path is missed. On `release`, the popped waiter is promoted via `jobQueue.promoteJob`; if `Job.promote()` returns false or throws, `dispatchPromoted` calls `releaseSlotAndPopWaiter` for the failed waiter — dropping it from the active set and popping the next waiter (the failed run still has its safety-net delay in BullMQ, so it is not lost).
- `jobQueue.promoteJob({ jobId, platformId })`: new method wrapping BullMQ's `Job.promote()`, returns `false` on missing job or promote failure.
- `keys.ts`: added `getConcurrencyPoolWaitlistKey`. (The active set continues to use `getConcurrencyPoolSetKey`.)
- `benchmark/smoke-rate-limiter.sh` *(new)*: docker e2e harness — fires N async webhooks (default 1000) at a published flow with the rate-limiter enabled, monitors RUNNING via `count-by-status`, asserts no concurrency-limit violations, and waits for all runs to reach `SUCCEEDED`.
- `benchmark/docker-compose.yml`: parameterizes `AP_PROJECT_RATE_LIMITER_ENABLED` and `AP_DEFAULT_CONCURRENT_JOBS_LIMIT` so the smoke harness can flip them on without forking the compose file.
- Docs: new Architecture page at `docs/install/architecture/concurrent-pools.mdx` and feature doc at `.agents/features/concurrent-pools.md`.

## Race-safety guarantees

- **Double acquire** impossible: the Lua script serializes on Redis's single-threaded event loop; exactly one caller sees `ZCARD < maxJobs`.
- **Double pop** impossible: `ZPOPMIN` inside the same Lua returns one element to one caller.
- **Duplicate acquire by same jobId**: a `ZSCORE` guard against the active set short-circuits idempotent re-dispatch; a second `ZSCORE` against the waitlist short-circuits a re-queue.
- **Promote failure**: `dispatchPromoted` cascades — `releaseSlotAndPopWaiter` removes the failed waiter from the active set and pops the next waiter into the freed slot. The failed waiter still carries its BullMQ safety-net delay, so it is re-attempted later via that path (order is **not** preserved across a failed promote — the failed run goes to the tail of the line on its next `preDispatch`).
- **Wedged pool** (active full of stale entries from crashed workers): the next call to `acquireSlotOrEnqueue` runs `ZREMRANGEBYSCORE active -inf (now - timeoutMs)` as its first step and then drains the waitlist into the freed slots in the same Lua script. **If no new acquire arrives**, recovery happens via the safety-net delay (flow timeout + 120s ≈ 12 min) when BullMQ re-dispatches each waiter.
- **Sustained saturation**: `RELEASE_AND_POP_LUA` and `ACQUIRE_OR_ENQUEUE_LUA` both refresh the TTL on `activeKey` and `waitlistKey` on every successful operation, so the keys never expire mid-processing in a long-running pool.

## Tests

**30 unit tests** in `test/unit/app/workers/job-queue/interceptors/concurrency-pool-redis.test.ts` (all passing):

- **Suite A — single-threaded correctness (18 tests, A1–A18):** acquire / overflow / FIFO order / release-empty / release-promotes / idempotent acquire / stale-entry sweep / partial stale cleanup / TTL on active and waitlist / re-queue idempotency / `dropPromotedWaiter` reuse and no-requeue / orphan promotion on next acquire / invalid-member sweeping / no over-pop into a saturated pool / `forPool` cascades to the next waiter when `promote()` returns false.
- **Suite B — concurrent / race-condition correctness (8 tests, B1–B8):** acquire storm, release storm, interleaved acquire+release, thundering-herd release, duplicate-member race, pool isolation, clock-skew safety (future-scored entries not swept), full-pipeline fuzz.
- **Suite D — crash recovery (2 tests):** D4 stale active entry does not block new acquires indefinitely; D5 500-op acquire+release soak ends with empty state.
- **Smoke (2 tests):** promote-on-release happy path; release refreshes waitlist + active TTLs under sustained saturation.

**E2E smoke** (`benchmark/smoke-rate-limiter.sh`): verified locally with `TOTAL=1000 CONCURRENCY_LIMIT=10 WORKER_REPLICAS=4` — 1000 runs SUCCEEDED in 20s, peak waitlist 776 queued, no concurrency-limit violations.

## Test plan

- [x] `vitest run test/unit/app/workers/job-queue/interceptors/concurrency-pool-redis.test.ts` — 30/30 pass
- [x] Pre-push `turbo run lint && test-ce && test-ee && test-cloud` all green
- [x] `benchmark/smoke-rate-limiter.sh` end-to-end with TOTAL=1000 — pass

## Docs

- `docs/install/architecture/concurrent-pools.mdx` — user-facing Architecture page with mermaid diagrams explaining the pool, the lifecycle, the limit-resolution order, and the recovery timers (flow timeout 600s, stale slot sweep 660s, safety-net delay 720s).


abuaboud:
Greptile pass triage on latest:

- **3187793090** (release pop without checking ZREM) — **valid, fixed in 27c8cc8d66.** `RELEASE_AND_POP_LUA` now early-returns when `ZREM activeKey member == 0`, so a stale-swept long-running job's eventual release no longer over-promotes a waiter into a refilled slot. Added A17 covering the saturated-pool scenario.
- **3129955190, 3130340465, 3130405739, 3181794079** — **stale.** They reference `rollbackPromotion`, `LPUSH`/`RPUSH`/`LLEN`/`LPOP`, `waitlistMembersKey`, and `parseMember` — none of which exist in the current code. They were posted against an earlier list-based revision; the current implementation uses a single ZSET waitlist with FIFO scores, refreshes both keys' TTLs on release, and self-heals invalid members via `dropPromotedWaiter` rather than re-queuing.

abuaboud:
Latest pass: cascading the dispatch loop on failed promotion is also valid — fixed in 921eaf6d84. dispatchPromoted now calls releaseSlotAndPopWaiter (instead of dropPromotedWaiter) when promote() returns false, so the freed slot immediately admits the next waiter rather than waiting for the safety-net delay. Added A18 covering the cascade. dropPromotedWaiter stays in raw for the unrelated stale-active-set sweep path that A14 documents.

PR #12750, opened 2026-04-23
fix(pieces/baserow): unify trigger auth with piece-level auth to prev…

onyedikachi-david:
…ent duplicate connections

Triggers were using a separate `baserowJwtAuth` definition instead of the shared `baserowAuth` array, forcing users to create two separate connections — one for actions and one for triggers.

All six triggers now use `baserowAuth` so a single JWT connection covers the whole piece. `onEnable` guards against Database Token auth with a clear error, since Baserow's webhook API requires JWT.

Removes unused `baserowJwtAuth`, `BaserowJwtAuthValue`, and `makeJwtClient`.

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/baserow/package.json
+++ b/packages/pieces/community/baserow/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-baserow",
-  "version": "0.4.1",
+  "version": "0.5.0",
   "main": "./dist/src/index.js",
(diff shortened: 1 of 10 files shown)


kishanprmr:
@onyedikachi-david bump the piece version

kishanprmr (approved):

PR #12759, opened 2026-04-23
fix(pieces/baserow): remove JWT auth and unify to database token only

kishanprmr:
## Summary

- Removes the Email & Password (JWT) authentication option from the Baserow piece
- Renames `databaseTokenAuth` to `baserowAuth` as the single exported auth
- Cleans up `isDatabaseTokenAuth` type guard, `BaserowClient.getJwtToken` static method, and all JWT branching logic in `makeClient`, webhook triggers, and the custom API call `authMapping`

## Test plan

- [x] Connect a Baserow account using Database Token auth — verify connection validates successfully
- [x] Run a Create Row / Get Row action — verify it works end-to-end
- [x] Enable a Row Created trigger — verify webhook registers in Baserow and fires correctly

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1029,7 +1029,7 @@
     },
     "packages/pieces/community/baserow": {
       "name": "@activepieces/piece-baserow",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "dependencies": {
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
--- a/packages/pieces/community/baserow/package.json
+++ b/packages/pieces/community/baserow/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-baserow",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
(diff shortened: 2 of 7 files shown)


sanket-a11y (approved):

bst1n:
Hi @kishanprmr, jumping in here as a co-author of the original JWT PR (#12226).

I want to acknowledge that our initial design had a real UX flaw — we exposed `baserowJwtAuth` as a separate auth, which forced users to create two connections. That was the root cause of everything that followed: #12750 trying to unify it, this PR removing JWT entirely, and #12912 having to revert to manual webhook setup.

However, removing JWT also removed the only way to register webhooks programmatically (Baserow's API restricts webhook management to JWT auth — Database Tokens cannot do it). The result is that all triggers now require manual copy-paste setup, which is a UX regression compared to most other pieces in Activepieces.

Would you be open to a different approach: keep a **single** `baserowAuth` (no duplicate connection), but expose two auth modes inside it via a property (Database Token *or* Email/Password JWT). Triggers would then auto-register webhooks when JWT is selected, and fall back to manual setup with clear instructions when Database Token is selected. Users with 2FA enabled stay on Database Token (since JWT doesn't support 2FA), while users without 2FA get a smoother auto-webhook experience.

Happy to put up a PR if you're aligned with this direction. I just want to discuss before submitting to avoid another cycle. Thanks!

kishanprmr:
@bst1n Multi-authentication for a piece only supports a single type of authentication. When jwtAuth was added, there were two customAuth for the piece, which broke it.

bst1n:
@kishanprmr Thanks for clarifying. Got it — single CustomAuth is the constraint.

Would the following pattern work within that constraint? A single `PieceAuth.CustomAuth` containing an `authMode` `StaticDropdown` (Database Token / Email & Password), with all credential fields as props, and the `validate` function branching on `authMode`. Triggers would then read `auth.authMode` in `onEnable` to decide between auto-webhook (JWT) and manual setup (token).

The one piece I'm not sure about: can props in a CustomAuth be conditionally shown based on a dropdown value? If yes, the UX would be clean. If not, all fields would be visible at once which is uglier but still functional. Do you know if conditional rendering of auth props is supported?

PR #12797, opened 2026-04-25
feat(google-docs): enrich with 7 actions, 1 trigger, UX dropdowns, bug fix

ibrahim-abuznaid:
## Summary

Significant enrichment of the Google Docs piece — 7 new actions, 1 new polling trigger, searchable Drive-backed dropdowns, conditional field groups, flattened outputs, and a fix for a destructive template-edit bug.

**Actions (5 → 12):**

- ✨ **Copy Document** — duplicate via Drive `files.copy`, optional new title + folder.
- ✨ **Find and Replace Text** — replace all occurrences with `matchCase`; returns `occurrencesChanged`.
- ✨ **Insert Text** — `Where to Insert` semantic dropdown (end / beginning / after specific text / before specific text / advanced index). `DynamicProperties` shows the right follow-up field per choice.
- ✨ **Insert Image** — same position dropdown plus an **Image Size** preset (Original / Small 100 pt / Medium 300 pt / Large 500 pt / Custom). Custom reveals width + height.
- ✨ **Replace Image** — `imageObjectId` is a **dynamic dropdown** that lists images in the selected document with labels like `Image 1 — 80×80`.
- ✨ **Share Document** — user / group / domain / anyone-with-link, role (viewer/commenter/editor), notification toggle. `DynamicProperties` only shows email/domain/notification fields when relevant.
- ✨ **Export Document** — PDF / DOCX / ODT / RTF / TXT / HTML / EPUB. Returns an Activepieces file URL.
- 🐛 **Create Document from Template** (renamed from "Edit template file") — was destructively editing the template in place. Now copies first via `files.copy`, then replaces placeholders on the copy. New optional fields: `copyTemplate` (default `true`, opt-out for legacy behavior), `newDocumentTitle`, `folderId`. Placeholder format prop reordered before variables for clearer UX.
- 🪄 **Read / Append / Existing actions** — Document ID raw ShortText replaced with a searchable Drive-backed dropdown across all callers. Outputs flattened to `{id, title, url, plainText?, revisionId, raw}`. `Append Text` now has a `Start on a new line` toggle. `Read Document` extracts plain text from paragraphs + table cells. OAuth2 description filled in with self-hosted setup guidance.

**Trigger (1 → 2):**

- ✨ **Updated Document** — polling. Watch a specific doc, restrict to a folder, or leave both empty to watch every accessible doc. Uses `modifiedTime` for dedupe.

**Shared infra:**

- `documentIdProp` + `documentImageIdProp` searchable dropdowns in `common/props.ts`.
- `findTextLocation` helper to anchor inserts before/after specific text.
- `flattenDoc`, `flattenDriveFile`, `buildDocUrl`, `extractPlainText`, `moveFileToFolder` helpers.
- Version bumped `0.4.1 → 0.5.0`.
- i18n `translation.json` expanded with new strings.

**Suggested labels:** `pieces` + `feature` (couldn't apply via API from fork — please add on the PR).

## Test plan

- [x] `npx turbo run build --filter=@activepieces/piece-google-docs` — clean build.
- [x] `npx eslint 'src/**/*.ts'` on the piece — 0 errors (1 pre-existing warning untouched).
- [x] **Live API verification** of every action/trigger via `POST /v1/sample-data/test-step` and `POST /v1/test-trigger` against a real Google account: 12/12 actions and 2/2 triggers green; `replace_image` with a fake `imageObjectId` surfaces a clean user-legible error.
- [x] Position dropdown variants on `insert_text` exercised: end ✓ / beginning ✓ / after anchor ✓ / before anchor ✓ / missing-anchor returns clean error ✓ / custom index ✓.
- [x] Size presets on `insert_image` exercised: small ✓ / medium ✓ / custom width+height ✓.
- [x] `share_document` dynamic recipient: `anyone` (no email field) ✓; `user` (email + notification toggle visible) ✓.
- [x] `replace_image` end-to-end via the dropdown: insert → read → pick from dropdown → replace → re-read confirmed contentUri changed for the same objectId.


packages/pieces/community/google-docs/src/lib/actions/copy-document.ts
```diff
@@ -0,0 +1,53 @@
...
+	displayName: 'Copy Document',
+	description: 'Duplicate an existing Google Doc and optionally move the copy to a folder.',
+	props: {
+		documentId: documentIdProp('Document to Copy', 'The Google Doc to duplicate.'),
+		newTitle: Property.ShortText({
+			displayName: 'New Document Title',
+			description: 'Title of the copy. Leave empty to keep the default "Copy of …" title.',
+			required: false,
+		}),
+		folderId: folderIdProp,
```
kishanprmr: `folderIdProp` currently uses the same description in all actions where it appears. It should be refactored like `documentId `to allow the description to be passed as an input.

kishanprmr (changes requested):

packages/pieces/community/google-docs/src/lib/actions/copy-document.ts
ibrahim-abuznaid: Refactored in a6a7a5e10f — `folderIdProp` is now `(displayName, description, required = false)`, matching `documentIdProp`. Each caller now passes its own copy:

- `find-document` → "Search only inside this folder. Leave empty to search across all of Drive."
- `create-document` → "Place the new document inside this folder. Leave empty to use My Drive."
- `create-document-based-on-template` → "Place the new document inside this folder. Leave empty to keep it next to the template."
- `copy-document` → "Destination Folder" / "Move the copy into this folder. Leave empty to keep it next to the original."
- `new-document` trigger → "Watch only documents created inside this folder. Leave empty to watch across Drive."
- `updated-document` trigger → "Folder (optional)" / "Watch only documents inside this folder. Leave empty to watch across Drive."

kishanprmr (approved):

PR #12919, opened 2026-04-27
feat: friendly view for step outputs and the data selector

AhmadTash:
## What does this PR do?

Introduces **Output Schema** — a way for piece authors to describe how a step's output should be presented in the builder, instead of users having to read raw JSON.

Pieces can now declare a structured view of their output: a flat list of fields, each with an optional label, an optional value-path override, an optional format (`email`, `url`, `date`, `datetime`, `number`, `boolean`, `image`, `html`, `currency`, `filesize`, `duration`), and optional `children` / `listItems` for nested objects and array-of-record shapes. A `dynamicKey: true` flag covers map-shaped values where the keys are not known statically. This is a non-breaking, opt-in addition to the framework — pieces without an output schema continue to render exactly as before.

Two surfaces consume the schema:

1. **Smart Output Viewer** (test-step output, run details). Two tabs:
   - **Friendly view** — labelled fields with type icons, expandable nested values, automatic table view for arrays of records, formatted images, clickable emails, safe outbound URLs (`http(s)` only — `javascript:`, `data:`, `file:`, etc. are rejected), pretty-printed file sizes, durations, dates, and currencies. Webhook & HTTP outputs (where the shape is fully user-defined) get a generic field-list view that adapts to whatever shape the data has.
   - **Raw JSON** — the existing JSON viewer, unchanged.

   It integrates with the platform's refactored test panel: the friendly/raw output renders in the Output pane for successful step output, while errors continue to use the existing friendly-error view.

2. **Data Selector (variable picker)** — within the existing **Data** tab, a **Friendly / Advanced** toggle lets users switch between the schema-driven view and the raw tree. When a piece declares a schema, the **Friendly** view shows meaningful labels and value previews (with the same friendly formatting), making it much easier to find and insert the right variable into downstream steps. Both modes share a single, unified node UI (same rows, indentation, expand/collapse, and **per-field type icons** in both views); the icon is driven by the field's declared `format` and falls back to value inference. Values render in the brand purple, empty fields show a purple `—`, and long values surface the full text in the design-system tooltip (black background, animated) on hover, replacing the OS-native `title=` tooltip. The platform's **Variables** tab is unchanged.

Other notable bits in this PR:

- New `@activepieces/pieces-framework` exports: `OutputSchema`, `OutputSchemaField`, `FieldFormat` — **plain TypeScript types** (no Zod schema, no runtime helper). The schema is provided as a plain object on `outputSchema`; it is fully type-checked through the `createAction` / `createTrigger` parameter type (or `satisfies OutputSchema` when extracted into a standalone constant). It is embedded into the piece metadata via `z.custom`, mirroring the existing `handshakeConfiguration` field. Bumps the framework to `0.30.0`.
- Hardening of attacker-controlled output (HTTP / webhook pieces): URL scheme allow-list, capped recursion depth, escaped bracket-path keys, header-injection guard on `mailto:`.
- Defensive value coercion in the viewer:
  - `boolean` format treats `"false" / "0" / "no" / "off"` as **No** rather than using JavaScript truthiness on `unknown`.
  - `currency` format catches invalid ISO codes (e.g. `'US'` instead of `'USD'`) and falls back to plain number formatting instead of crashing the render tree via `Intl.NumberFormat`.
  - The path resolver preserves intentional `null` at the requested path rather than silently falling through to a wrapper-key envelope (`data.*`, `body.*`, …) and returning unrelated data.
- Tabular-array detection scans up to the rendered row cap (100 rows) when checking shape consistency — divergent rows past row 5 now correctly disable the table view instead of leaving missing cells.
- React Query–backed schema fetching (shares the existing `['piece', name, version]` cache, no extra network calls), memoised tree builds, full keyboard accessibility, and full i18n coverage for new strings.
- Unit-test coverage for `pathUtils`, `schemaUtils` (resolve-schema), `format-value`, `output-table-view`, and `utils-schema`.

A piece author opts in by passing `outputSchema` to `createAction` / `createTrigger`:

```ts
import { createAction } from '@activepieces/pieces-framework';

export const getEmail = createAction({
  name: 'get_email',
  displayName: 'Get email',
  // ...
  outputSchema: {
    fields: [
      { key: 'subject',  label: 'Subject' },
      { key: 'from',     label: 'From',     format: 'email' },
      { key: 'date',     label: 'Received', format: 'datetime' },
      { key: 'snippet',  label: 'Preview' },
      { key: 'labelIds', label: 'Labels' },
      {
        key: 'attachments',
        label: 'Attachments',
        listItems: [
          { key: 'fileName' },
          { key: 'size', format: 'filesize' },
        ],
      },
    ],
  },
  run: async (ctx) => { /* ... */ },
});
```

The inline `outputSchema` object is type-checked against `OutputSchema` via the `createAction` / `createTrigger` parameter type (including nested `children` / `listItems`). When extracting the schema into its own constant, annotate it with `satisfies OutputSchema` to keep the same checking.

At runtime:
- The **Smart Output Viewer** reads the schema from the piece metadata and renders the field list. Field paths can be dot/bracket notation (`thread.data.messages[0].subject`) and resolve through the actual sample/run output, with a small wrapper-key fallback (`data.*`, `body.*`, `payload.*`) so common API envelopes "just work" — without overriding values that are intentionally `null`.
- The **Data Selector** does the same when the user is looking at that step in the variable picker, and produces fully-qualified mention paths (e.g. `step_1["thread"]["data"]["messages"][0]["subject"]`) when the user clicks **Insert**.
- For pieces that don't declare a schema, both surfaces behave exactly as today (generic friendly view / raw JSON in the viewer; existing tree in the data selector).

### Relevant User Scenarios

- **Gmail / email pieces** — instead of scanning raw JSON for `payload.headers[*].value`, users see Subject, From, Date, Snippet, and an attachments list with human-readable file sizes.
- **Lookup / "Get X by ID" actions** (HubSpot, Salesforce, Airtable, Notion, etc.) — the fields that matter are labelled, formatted, and surfaced directly; nested objects expand inline.
- **List/search actions returning arrays of records** (rows, contacts, messages, orders) — the viewer auto-detects tabular shape and renders a real table.
- **Payment / billing pieces** (Stripe, PayPal) — `currency` fields render as `€19.99` / `¥1,200` rather than `1999` cents when the piece supplies a currency code, and survive bad codes without crashing the render.
- **File-handling pieces** (Drive, Dropbox, S3) — file sizes render as `2.3 MB`, durations as `1m 23s`, image URLs as inline thumbnails.
- **Webhook / HTTP / Schedule pieces** (output shape is user-defined) — get a sensible generic view instead of a raw JSON dump, including expandable nested objects.
- **Variable picker UX** — when wiring a downstream step, users no longer have to guess which JSON path holds the value they want; meaningful labels and previews are right there, with the full value on hover via the design-system tooltip.

Fixes # (issue)


amrdb (commented):
A couple of questions on the new hints types/schema in `packages/pieces/framework/src/lib/output-display-hints.ts`:

### 1. Why `defineHints` + a Zod schema instead of plain TS types?

`createAction`/`createTrigger` already type `outputDisplayHints` as `OutputDisplayHints`, so an inline hints literal gets full type + **excess-property** checking from TS — including nested `children`/`listItems`, since contextual typing flows down the recursive type:

```ts
outputDisplayHints: {
  fields: [
    { key: 'from', format: 'emial' },   // ❌ TS error already
    { key: 'subject', lable: 'Subject' }, // ❌ TS error already (excess prop)
  ],
}
```

Given that:
- `defineHints` is a non-generic identity function, so it adds nothing over the parameter type (or `satisfies OutputDisplayHints`) — there's no inference to preserve.
- The Zod schemas (`HintField`, `OutputDisplayHints`, and by extension the `ActionBase`/`TriggerBase`/`PieceMetadata` schemas) are **never `.parse()`d / `.safeParse()`d anywhere** in the repo — `piece.metadata()` returns the raw object and the DB read is a typecast. They're used purely as `z.infer` type sources.

So the runtime Zod schema + `defineHints` don't currently buy any validation. Could hints just be plain TS types? Keeping Zod (and a validating `defineHints`) only pays off if we actually intend to validate untrusted input somewhere — e.g. server-side for custom/third-party pieces. If we don't, dropping `defineHints` (or replacing it with `satisfies`) trims surface for the same safety.

### 2. `OutputDisplayHints.label` looks unused

```ts
export const OutputDisplayHints = z.object({
  fields: z.array(HintField),
  label: z.string().optional(),   // read anywhere?
});
```

Neither consumer reads it — the Smart Output Viewer and the data selector both derive the step heading from the step's `displayName`. Either wire it up or drop it so authors don't set it expecting an effect.

### 3. (resolves itself)

The `dynamicKey: z.literal(true)` vs `boolean` concern I raised earlier is moot — if `defineHints`/Zod goes away per #1, `dynamicKey` just becomes an optional `boolean` on the TS type, which the consumer's `=== true` check already handles. No action needed.

amrdb (approved):

PR #12964, opened 2026-04-27
feat(baserow): restore JWT auth + Row Event trigger + auto-create select options

bst1n:
## Why

The Baserow piece has been through churn over the last weeks:

- **#12226** (mine) introduced JWT auth + automatic webhook registration but as a **separate, second auth** — forcing users to maintain two duplicate Baserow connections. That was a design mistake on my part.
- **#12759** removed the JWT auth entirely.
- **#12912** was forced to revert the auto-webhook feature to manual setup once the JWT was gone — which removed the main user benefit of the original change.

This PR restores the JWT auth **the right way** (without breaking existing Database Token connections) and brings back auto-webhook on top.

## Auth design

After @onyedikachi-david's review feedback, the framework rejects `[CustomAuth, CustomAuth]` arrays (validation: "Auth properties must be unique by type"). Both Baserow auth methods need a custom `apiUrl` field, so neither can be reduced to `SecretText` — ruling out the multi-type pattern used by Umami.

The accepted approach is a **single `CustomAuth` with an `authType` dropdown** that selects between Database Token and Email & Password (JWT). All credential fields are optional at the schema level; the runtime branches on `authType` and validates accordingly.

**Backward compatibility:** existing Database Token connections (with no `authType` field) continue to work unchanged — the helper falls back to detecting the mode from which fields are filled.

### 1. Restore JWT auth (final shape after review)
- Single `CustomAuth` with `authType` dropdown (`database_token` | `jwt`), `apiUrl`, and optional `token` / `email` / `password`.
- `validate()` branches per mode and tests the right endpoint.
- `makeClient` and `buildCustomApiAuthHeader` resolve the right auth header (`JWT …` vs `Token …`) with explicit errors when credentials for the chosen mode are missing.
- One Baserow connection per workspace; users keep multiple connections (different scopes, different modes) side by side as before.

### 2. Conditional auto-webhook on the 6 existing triggers
- `createWebhookTriggerHooks` no-ops in Database Token mode (Baserow's API forbids webhook creation with that token type) and registers/cleans up via the API in JWT mode.
- Markdown setup instructions kept, framed as "only follow these if you authenticated with Database Token".
- Reintroduces the `table_id` prop on every trigger — required to register the webhook in JWT mode and to fetch real sample data.

### 3. Auto-create missing select options on Create/Update Row
- New opt-in `Create missing select options` toggle.
- When enabled, single/multi-select values that don't yet exist are batched into **one** PATCH on `/api/database/fields/{id}/` per field (existing options + new options) — never one PATCH per option, since Baserow replaces the full `select_options` array.
- Go-style `tryCatch` from `@activepieces/shared`. On failure, the row write still succeeds and we log to `console.error`.

### 4. `async test()` on every trigger
- Replaces hardcoded `sampleData` placeholders by fetching the latest 5 rows from the configured table.
- The shape returned by `test()` matches `run()` exactly per trigger, so flow mapping shows realistic data.

### 5. New combined `Row Event` trigger
- Single trigger that fires on create / update / delete in one webhook (saves a slot in JWT mode).
- `run()` normalises every event into a flat `{ event_type, row, previous_row }` so flows can branch on `event_type` without touching Baserow's underlying payload shape (which differs — `row_ids` vs `items`).
- Dedicated single-event triggers remain.

### 6. i18n harmonisation
- French: every `(Batch)` trigger display name now reads `par lot`, matching the action names. Same harmonisation in DE (`(Stapel)`), ES (`(lote)`, also fixes a typo `(atch)`), JA (`(一括)`).
- Auth description rewritten to cover both modes in a single block.
- New keys (`Authentication`, `Authentication Method`, `Row Event`, `Create missing select options`, descriptions) added to all 13 locale files. Translated for FR; English fallback for the 11 others.

## Test plan

- [ ] Create a Baserow connection with **Database Token** mode: actions still work, triggers ask for manual webhook setup.
- [ ] Create a Baserow connection with **Email & Password (JWT)** mode: enabling a trigger registers a webhook automatically; disabling deletes it.
- [ ] Existing Database Token connections (created before this change) keep working without re-edition.
- [ ] `Row Event` trigger fires on create / update / delete with the same `event_type` discriminator field.
- [ ] `test()` on each trigger returns real rows shaped like the production `run()` output.
- [ ] `Create missing select options` on Create Row: a single-select value not in the field is added (one PATCH) before the row is written.
- [ ] French UI: trigger picker shows \"Lignes créées par lot\" / \"Lignes mises à jour par lot\" / \"Lignes supprimées par lot\".


onyedikachi-david:
Hello @bst1n, thanks for your contributions, we do support multi auth but it must be of different types; we can't have this: **[CustomAuth1, CustomAuth2]**, which your PR added, it won't work, the 2 PR that removed the jwt auth and the auto webhook creation was so to fix that and then **gradually migrate** solely to jwt auth alone, because database table operations can still be done with either jwt auth or db api key. 

I was already working on migrating solely to jwt and reinstating the auto webhook implementation. But since you've created a PR already, refactor your current implementation such that previous actions now make use of jwt custom auth type. Thanks for your contributions again.

bst1n:
Hi @onyedikachi-david, thanks for the clarification — I've dug into the framework code and confirmed the constraint: `[CustomAuth, CustomAuth]` throws "Auth properties must be unique by type" at piece registration, so that approach is indeed broken.

Before aligning on JWT-only though, I want to raise two concerns:

**1. 2FA incompatibility** — `/api/user/token-auth/` rejects accounts with 2FA enabled. JWT-only would lock out any Baserow user who has 2FA activated, which is a real regression.

**2. Access scoping** — Database Tokens support granular per-table, per-operation permissions. JWT gives full user-level access across all databases. These are not equivalent from a security standpoint, and some users specifically rely on token scoping.

My intent with JWT was never to replace Database Tokens — it was an opt-in workaround for users willing to trade 2FA support for automatic webhook registration. Worth noting: the Baserow founder has mentioned that programmatic webhook management for Database Tokens is on their roadmap, so JWT-only would likely need revisiting once that lands anyway.

Given the `[CustomAuth, CustomAuth]` constraint, would a **single `CustomAuth` with an `authType` dropdown** (`database_token` | `jwt`) be acceptable? The user picks their method, the code branches accordingly. All fields appear simultaneously which is slightly cluttered, but both auth methods are preserved without breaking 2FA or scoping. Is that a pattern you'd accept in a PR, or is there a better approach you'd recommend?

onyedikachi-david:
That's a great option. You may go ahead with the drop down option!

bst1n:
@onyedikachi-david the refactor is in — pushed as commit `4d7e327` on top of the existing branch.

The piece now exposes a single `CustomAuth` with an `authType` dropdown (`Database Token` | `Email & Password (JWT)`), `apiUrl`, and optional `token` / `email` / `password` fields. The runtime branches on `authType` and `validate()` tests the right endpoint depending on the chosen mode.

Existing Database Token connections keep working unchanged — they have no `authType` field, the helper falls back to detecting the mode from which credentials are filled, so no production flows break.

The PR description has been updated to reflect the final design. Ready for another look whenever you have a minute. Thanks again for the guidance.

onyedikachi-david:
You don't need to add translation files to your PR, they are auto generated by one of our CI. Please take note of this next time. @bst1n

onyedikachi-david (approved):

PR #13065, opened 2026-05-01
fix(Gamma): upgrade gamma piece to v1.0 from v0.2 and add folder selection

aryakdangol:
## What does this PR do?

Gamma no longer support the v0.2 version that the this repo was working with. This PR upgrades the gamma api to v1.0 and also adds a folder selection dropdown.

There is no video, but I have attached screenshots to explain this feature.

**Previously we would get this error while working with the v0.2 Gamma API piece:**

Screenshot 2026-05-01 at 11 52 43

**By upgrading the api to v1.0 the Gamma piece works correctly. There is also an option to select folder which the v1.0 API supports:**

Screenshot 2026-05-01 at 14 01 48

### Relevant User Scenarios

- This allows users to use Gamma piece which was not usable before. I have attached the image in the above section on the error encountered while using older version of Gamma API

Fixes # (issue)


onyedikachi-david (approved):

onyedikachi-david:
Thanks for contributing @aryakdangol

PR #13073, opened 2026-05-02
feat: variables manager (encrypted project secrets)

Louai-Zokerburg:
## What does this PR do?

Adds **Variables** — encrypted project-scoped secret values (API keys, tokens, opaque strings) that users create once and reference inside any flow input via a mention pill:

```
{{variables['STRIPE_API_KEY']}}
```

The value is stored AES-encrypted at rest, decrypted only by the engine at flow execution (via a worker endpoint) or by an authorized user via the explicit reveal action. Every reveal is audit-logged.

> History: this branch was originally `credentials-manager` and piggy-backed on `app_connection` via a `kind` discriminator and nullable `pieceName`. After team review that coupling was scrapped in favour of a clean, dedicated module — the `app_connection` table is back to its pre-branch shape and Variables live in their own table, service, controller, permissions, and audit events.

### Architecture

**New `variable` table** (`packages/server/api/src/app/variable/`)
- Columns: `id, created, updated, name, projectId, platformId, ownerId, value (jsonb, encrypted), metadata`
- Unique index on `(projectId, name)`; SET NULL FK to `user`
- Migration: `1793000000000-AddVariableTable.ts`

**Routes**
| Method | Path | Auth | Permission |
|---|---|---|---|
| `POST` | `/v1/variables` | USER + SERVICE | `WRITE_VARIABLE` (upsert by name; fires `VARIABLE_UPSERTED`) |
| `GET` | `/v1/variables` | USER + SERVICE | `READ_VARIABLE` (paginated; filters by `name` substring) |
| `GET` | `/v1/variables/owners` | USER + SERVICE | `READ_VARIABLE` (distinct project members owning at least one variable) |
| `POST` | `/v1/variables/:id/reveal` | USER only | `WRITE_VARIABLE` (fires `VARIABLE_VALUE_REVEALED`) |
| `DELETE` | `/v1/variables/:id` | USER + SERVICE | `WRITE_VARIABLE` (fires `VARIABLE_DELETED`) |
| `GET` | `/v1/worker/variables/:name` | ENGINE only | (engine principal carries projectId) |

**Engine** — `packages/server/engine/src/lib/piece-context/variable-resolver.ts` mirrors the connection resolver. `props-resolver` adds a `variables` branch alongside `connections`; censored input renders `**REDACTED**`.

**Permissions** — new `READ_VARIABLE` / `WRITE_VARIABLE` added to the project ACL. ADMIN + EDITOR get write; VIEWER reads only. The reveal endpoint is USER-principal only.

**Audit events** — `VARIABLE_UPSERTED`, `VARIABLE_DELETED`, `VARIABLE_VALUE_REVEALED`. Each carries `{ variable: { id, name } }`.

### Web UI

- **`/variables` page** with paginated table, owner filter, search by name, bulk delete, per-row dropdown (Edit / Copy reference / Copy value / Delete).
  - **Name** · **Owner** · **Used in** (flow count, plural-aware) · **Last updated** · **Actions**.
- **Variable dialog** for create / rotate value, with the eye toggle.
- **Builder data-selector tab** to insert `{{variables['NAME']}}` mentions.
- **Mention chip** renders the lucide `Variable` icon plus `Variable · <name>`.
- Audit-log rows and event-destination labels learn the three new event types.

### Permission summary (UI)

| Role | List page | New / Edit / Rotate / Delete | Copy reference | Copy value (reveal) |
|---|---|---|---|---|
| ADMIN | ✓ | ✓ | ✓ | ✓ |
| EDITOR | ✓ | ✓ | ✓ | ✓ |
| VIEWER | ✓ | – | ✓ | – |

Copy reference is gated only by `READ_VARIABLE` — it writes the public mention syntax to the clipboard, never the plaintext.

### Manual cleanup for the preview env

The earlier `1791000000001-RelaxAppConnectionPieceFields` migration ran on the preview DB but is now deleted. Run this once against the preview env to restore the original `app_connection` schema:

```sql
DELETE FROM "app_connection" WHERE "pieceName" IS NULL;
DROP INDEX IF EXISTS "idx_app_connection_piece_name";
ALTER TABLE "app_connection" ALTER COLUMN "pieceName" SET NOT NULL;
ALTER TABLE "app_connection" ALTER COLUMN "pieceVersion" SET NOT NULL;
DELETE FROM "migrations" WHERE "name" = 'RelaxAppConnectionPieceFields1791000000001';
```

Fresh installs (and prod once this lands) are unaffected — the migration registry never sees it.

## Test plan

- [ ] **CRUD** — Create `STRIPE`, edit/rotate the value, delete it from `/variables`. Confirm rows appear/disappear and the `variable` table reflects them.
- [ ] **Reveal + audit** — Copy a variable's value; confirm the plaintext matches. A `VARIABLE_VALUE_REVEALED` row appears at `/platform/security/audit-logs`. Repeat for `VARIABLE_UPSERTED` and `VARIABLE_DELETED`.
- [ ] **Mention rendering** — Drop a variable into an HTTP step's Authorization header via the data-selector. The pill renders `Variable · STRIPE`; saved value is `{{variables['STRIPE']}}`.
- [ ] **Flow execution** — Run a flow that references a variable. The engine resolves it via `/v1/worker/variables/:name` and the external service receives the correct value.
- [ ] **Used-in column** — Reference the same variable in 3 flows; the "Used in" cell shows `3 flows` and updates to `No flows` after removing the references.
- [ ] **Owner filter** — Create variables as two different project members; the owner filter dropdown shows both, multi-select narrows the table.
- [ ] **Permission gating** — As a VIEWER, list works but New / Edit / Rotate / Copy value / Delete are disabled. As an EDITOR, all actions work. `POST /:id/reveal` returns 403 for VIEWER.
- [ ] **Connections page unchanged** — Connections list, replace, reconnect, rename all still pass. No `kind` query param, no `pieceName IS NULL` rows.
- [ ] **MCP** — `ap_list_connections` returns piece connections only.
- [ ] **Docs** — `.agents/features/variables.md` exists; credentials sections removed from `app-connections.md`, `global-connections.md`, `mcp.md`.


Louai-Zokerburg:
On the recurring "missing SQLite migration" P1 from Greptile:

SQLite is **deprecated** in this codebase — `packages/server/api/src/app/database/sqlite-connection.ts` carries the JSDoc:

> `@deprecated SQLite3 is deprecated and only exists for migration purposes. Use PGLite instead.`

The SQLite DataSource is invoked exactly once, by `MigrateSqliteToPglite1765308234291`, which copies rows from a legacy SQLite file into PGLite via `INSERT` statements and then destroys the SQLite session. Schema changes on the source SQLite database after that migration are **never executed** — every community deployment that has booted past `1765308234291` runs the Postgres migration list against PGLite from then on.

This is why every Postgres-only migration since `1765264096034` (Dec 2025) — `DropChatTokenColumns`, `AddChatCompactionColumns`, `AddSsoDomainVerification` (which landed alongside this PR), `RelaxAppConnectionPieceFields` here — ships without a SQLite counterpart. Adding a SQLite migration here would be dead code on a path that's already been retired, so I'm leaving this as Postgres-only to match the convention.

The other P2 (no audit log on reveal) is addressed in da21e05737 — `CONNECTION_VALUE_REVEALED` is emitted whenever the reveal endpoint is hit, so we can answer "who pulled credential X and when".

packages/web/src/app/builder/piece-properties/text-input-with-mentions/text-input-utils.ts
```diff
@@ -167,6 +167,14 @@ function parseLabelFromMention(
   stepsMetadata: (StepMetadataWithDisplayName | undefined)[],
 ) {
   const { stepName, path } = parseStepAndNameFromMention(mention);
+  if (stepName === 'connections') {
+    const externalId = path[0] ?? '';
+    return {
+      displayText: `Credential · ${externalId}`,
```
AbdulTheActivePiecer: shouldn't show external id, show display name of the connection instead

packages/server/api/src/app/variable/variable.service.ts
```diff
@@ -0,0 +1,262 @@
...
+    projectId: ProjectId,
+    variables: VariableSchema[],
+): Promise<Map<string, string[]>> {
+    const flowIdsByName = new Map<string, string[]>()
+    if (variables.length === 0) {
+        return flowIdsByName
+    }
+    const variableNames = variables.map((v) => v.name)
+
+    const flowsPage = await flowService(log).list({
```
MrChaker: I think this might be not optimal . fetching all flows with variableNames . if we don't really need this info ( "flows refrencing the variable ) let's just remove it

MrChaker (approved):

PR #13074, opened 2026-05-03
feat(error-handling): add error handlers to steps in flows

AbdulTheActivePiecer:
## What's new

- **Send the run down a different path when a step fails.** Code and Piece steps with "Continue on failure" enabled now show two outgoing paths on the canvas — an On success branch and an On failure branch — like a built-in router. Drop steps into either side to react to the outcome without wiring an external If condition.

- **Reference the step's output and error directly from the data selector.** When continue-on-failure is on, the data selector shows two new entries under the step: the step's Output (so downstream steps can read whatever it returned when it succeeded) and an On failure subtree with an Error message leaf you can insert as `{{step.error.message}}` to log, route on, or send in a notification.

- **Right-click the canvas to add steps into either branch.** The context menu on a continue-on-failure step lets you append into On success or On failure without dragging through the empty-branch placeholder.

- **Copy-pasting and importing flows preserves the branches.** Pasting a step that has On success / On failure branches keeps the structure intact; importing flow JSON with the new shape works without conversion.


AbdulTheActivePiecer:
@greptileai Identifiers matching the canonical `step_\d+` pattern (or `trigger`, added to `stepNames`
 by the caller) are always rewritten — even when the name is missing from `stepNames`.
 Rationale: if a user wrote `step_10` in an expression, they meant a step reference, and
 rewriting it preserves the legacy resolution behavior (resolves to undefined → empty string)
rather than silently turning it into a free-variable lookup against the new scope shape.
Custom step names (e.g. `my_action` from a renamed step) only get rewritten when present
in the `stepNames` set the caller passes in (the names of the steps in the current flow).

AbdulTheActivePiecer:
@greptileai the runtime fallback handles it. Let me trace through to confirm and answer.

The rewrite is implemented at expression-rewriter.ts:90 the STEP_NAME_PATTERN.test(name) check runs FIRST, so canonical step_\d+ and trigger are rewritten regardless of stepNames membership.

The runtime path the reviewer is worried about:

Flow has step_1, step_5. The flow's stepNames doesn't include step_10.
User wrote {{step_10.foo}} (typo or stale ref). Rewriter rewrites to {{step_10['output'].foo}} because the pattern matches.
At runtime, extractReferencedStepNames iterates the flow's stepNames looking for substrings step_10 isn't there, so it's not added to referencedSteps.
currentState(['step_1']) returns {step_1: {output, error}} — no step_10 key.
evalInScope('step_10[\'output\'].foo', {step_1: ...}) runs the script in the sandbox. step_10 is a free variable in the V8 isolate. The script throws ReferenceError.
[evalInScope line 207-210](vscode-webview://0otd4im1r3k1fj9meijvngjg2fbs62l4adj4t1l90u5imavvkkm9/packages/server/engine/src/lib/variables/props-resolver.ts) catches it: if (resultError) { console.warn(...); return '' }.
End result: ''. Same as legacy ({{step_10.foo}} against a flat scope without step_10 also returned '').

So the rewriter's invariant doesn't require any caller contract on stepNames. The runtime free-variable fallback in evalInScope makes the whole chain work whether extractReferencedStepNames finds the name or not. extractReferencedStepNames is purely an optimization. it limits which step outputs we serialize into the sandbox scope, not a correctness gate.

AbdulTheActivePiecer:
@greptileai redo your review then

AbdulTheActivePiecer:
@greptileai

The runtime resolver only evaluates `{{ }}` template strings — [`props-resolver.ts`](packages/server/engine/src/lib/variables/props-resolver.ts) extracts via `VARIABLE_PATTERN` before any step lookup. Anything outside `{{ }}` is treated as a literal both before and after this change.

The migration's [`wrapStepReferencesInOutputProperty`](packages/server/api/src/app/flows/flow-version/migrations/migrate-v20-step-output-nesting.ts) recurses through `step.settings` and rewrites every `{{ }}` string. Since the resolver only ever sees `{{ }}` strings and the migration rewrites all of them, the runtime never encounters a bare `step_1` against the new wrapper.

The null-guard / truthiness / equality cases you flagged are covered by tests D1, D5–D10 in [`expression-rewriter.test.ts`](packages/shared/test/flow/expression-rewriter.test.ts) — the rewriter preserves the original semantics by converting `step_1` → `step_1['output']`.

PR #13096, opened 2026-05-04
fix(mcp): remove unused MCP server status toggle

hazemadelkhalel:
## Summary
- Remove the `McpServerStatus` (ENABLED/DISABLED) enum, `status` DB column, and UI toggle from both project and platform MCP servers
- The status field was never enforced in the MCP transport layer — servers were always accessible via their token regardless of status
- Add database migration to drop the `status` column from `mcp_server` table

## Test plan
- [ ] Verify project MCP settings page loads and shows Connection/Tools tabs directly
- [ ] Verify platform MCP settings page loads and shows Connection/Tools tabs directly
- [ ] Verify enabling/disabling individual MCP tools still works
- [ ] Verify MCP token rotation still works
- [ ] Run `npm run lint-dev` — passes with 0 errors


hazemadelkhalel:
Addressed review comments:

1. **P1 - Missing SQLite migration**: Added `1790000000000-RemoveMcpServerStatusSqlite.ts` with the temp table approach to drop the `status` column, and registered it in `sqlite-connection.ts`. This prevents NOT NULL constraint violations on SQLite installations.

2. **P2 - `transaction = true`**: Kept as-is — this is the established convention across 20+ other migrations in this project.

PR #13097, opened 2026-05-04
feat: introduce flow approvals

amrdb:
## Summary

Adds a publish-time approval gate for flows in sensitive projects, gated by the existing `environmentsEnabled` EE plan flag. Submitting a sensitive flow for publish locks the version, creates an approval-request row, and leaves triggers offline. Approval runs the second half of the existing LOCK_AND_PUBLISH sequence (set publishedVersionId, register triggers).

The change is intentionally minimal in CE (one extension-point hook + one helper extraction). All workflow logic lives in a new EE module under `packages/server/api/src/app/ee/flows/flow-approval/`.

## Why

Customers running sensitive automations (financial systems, customer data pipelines, prod credentials) have no governance gate on flow deployment today — anyone with WRITE_FLOW can publish a flow to production. Compliance teams want a review step before flows that touch sensitive resources go live.

## Feature gating

- `platform.plan.environmentsEnabled` — reuses the existing EE plan flag that gates project releases. No new column on `platform_plan`.
- `project.sensitive` — boolean column, false by default. Platform admins toggle it from project settings or the platform Edit Project dialog.
- `Permission.PUBLISH_SENSITIVE_FLOW_ACCESS` — granted to the seeded ADMIN role only. Holders bypass the approval gate and can directly publish.

Routing decision (EE hook):
1. `environmentsEnabled === true`, AND
2. `project.sensitive === true`, AND
3. Submitter does NOT hold `PUBLISH_SENSITIVE_FLOW_ACCESS`

→ NEEDS_APPROVAL. Otherwise → PUBLISH_NOW (existing behavior).

## Schema

One migration: `1783000000000-AddFlowApprovalWorkflow.ts`. Adds:

- `project.sensitive boolean NOT NULL DEFAULT false`
- `flow_approval_request` table with FKs (CASCADE on flow/flow_version, SET NULL on user), unique on `flowVersionId`, composite indexes on `(projectId, state, created)`, `(flowId, state)`, `(platformId, state)`

## CE refactor

`flowService` exposes two primitives so the EE approve handler can reuse the publish bookkeeping verbatim:

- `setPublishedVersion({ flow, lockedVersion, entityManager? })` — sets `publishedVersionId` + `status=DISABLED` via a targeted UPDATE.
- `applyStatusChangeForPublishedFlow({ id, projectId, newStatus })` — wrapper around `applyStatusChange`.

`publishHooksFactory.create<PublishHooks>(...)` defines `routePublish` + `submitForApproval`. CE default returns `PUBLISH_NOW` and a no-op submit. EE overrides via `publishHooksFactory.set(eeFlowPublishHook)` in the cloud/enterprise edition switch.

## EE module: `ee/flows/flow-approval/`

- `flow-approval-request.entity.ts` — entity with relations to flow, flowVersion, user.
- `flow-approval-request.service.ts` — `submitForApproval`, `approve`, `reject`, `withdraw`, `list` (paginated, joins flow_version for display name). Repo is inlined in the service file.
- `flow-approval-request.controller.ts` — per-verb endpoints mirroring `/v1/flow-runs/:id/{retry,cancel,archive}`:
  - `POST /v1/flow-approval-requests/:id/approve` — PUBLISH_SENSITIVE_FLOW_ACCESS
  - `POST /v1/flow-approval-requests/:id/reject` — PUBLISH_SENSITIVE_FLOW_ACCESS, body `{ reason?: string }`
  - `POST /v1/flow-approval-requests/:id/withdraw` — WRITE_FLOW
  - `GET /v1/flow-approval-requests` — READ_FLOW, query `state`, `flowVersionId`, `cursor`, `limit`
- `flow-publish-hook.ts` — implements PublishHooks for EE.
- `flow-approval.module.ts` — gated by `platformMustHaveFeatureEnabled(p => p.plan.environmentsEnabled)`.

The approve handler reuses CE primitives directly. The published-flow update and the approval row state move atomically inside a single transaction:

```ts
await transaction(async (entityManager) => {
    await flowApprovalRequestRepo(entityManager).update(
        { id: approval.id },
        { state: APPROVED, approverId, decidedAt },
    )
    await flowService(log).setPublishedVersion({ flow, lockedVersion, entityManager })
})
await flowService(log).applyStatusChangeForPublishedFlow({ id, projectId, newStatus: approval.requestedStatus })
```

`requestedStatus` is captured on the request row at submit time and replayed on approval — submitter's intent is preserved. Trigger registration (the external side effect) runs after the commit so a side-effect failure can't strand the system in a half-published state.

## State machine

```
DRAFT  ──submit──▶  LOCKED + flow_approval_request{PENDING}
                     │
                     ├──approve──▶  LOCKED + {APPROVED}
                     │              + flow.publishedVersionId = me
                     │              + applyStatusChange(requestedStatus)
                     │
                     ├──reject──▶   LOCKED + {REJECTED, reason}
                     │              (no in-app banner; submitter recovers via
                     │               ViewingOldVersionWidget → "Use as Draft")
                     │
                     └──withdraw──▶ DRAFT (unlocked) + request row deleted
                                    (audit log retains the WITHDRAWN event)
```

## Frontend

- Builder banner (`flow-approval-banner.tsx`) shows only for the PENDING state.
  - PENDING — Approve/Reject for approvers; Withdraw for the submitter or any non-approver project member.
  - REJECTED — no banner. The rejected version is reachable through the standard `ViewingOldVersionWidget` (with "Use as Draft" to keep iterating). The rejection reason lives in the audit log; submitter feedback happens out-of-band (Slack/email/etc.).
  - APPROVED — no banner (the approve mutation toast is the only feedback).
- Optimistic Zustand updates on Withdraw/Approve to avoid the banner flicker that came from the gap between the mutation completing and the approvals query refetching.
- Publish button — label flips between "Publish" and "Request approval" based on `requiresApproval = environmentsEnabled && project.sensitive && !canBypassApproval`.
- Sidebar — "Pending approvals" tab gated on `PUBLISH_SENSITIVE_FLOW_ACCESS`. Badge shows up to "10+". Hidden during `isFetchingProjectRole` to prevent flash for non-approvers.
- Approvals page (`/approvals`) — server-paginated via shared `DataTable` (cursor + limit in URL). Uses `StatusIconWithText` and `FormattedDate` for visual parity with the runs table.
- Project settings — `sensitive` toggle in both the in-project settings dialog and the platform-admin Edit Project dialog. Gated on `environmentsEnabled` AND `PlatformRole.ADMIN`.
- Discard refactor — extracted to `runDiscard()` pure function in `discard-draft.ts` with 5 invariant unit tests in `packages/web/test/`. Unified path navigates to the published locked version after overwriting the draft, dropping the wasteful publish-no-op that the original code did to "absorb" the draft.
- Flicker fix — `useChangeFlowStatus.onSuccess` now awaits `queryClient.refetchQueries(['flow-approval-requests'])` before flipping `isPublishing` to false. Single atomic transition once everything is settled.

## Audit / event streaming

Four new `ApplicationEventName` values: `FLOW_APPROVAL_REQUESTED`, `FLOW_APPROVAL_GRANTED`, `FLOW_APPROVAL_REJECTED`, `FLOW_APPROVAL_WITHDRAWN`. Reject includes `rejectionReason` in the event data; Withdraw fires before the row is deleted so the audit trail persists. Routed through the existing `applicationEvents` bus → audit log + event-destination listeners.

## `getLatestVersion(LOCKED)` audit

The approval workflow makes "latest LOCKED" ambiguous (could be PENDING/REJECTED). Audited the three callers:

- `app-connection.handler.handleLockedVersion` — replaced with `flowService.getPublishedVersionOrThrow`.
- `flow-run-service.FlowRetryStrategy.ON_LATEST_VERSION` — same.
- `alerts-service.sendAlertOnRunFinish` — switched to `issueToAlert.flowVersionId` directly (the version the failing run actually executed on; sidesteps the "is the flow published?" question and handles debug-run edge cases).

`getLatestLockedVersionOrThrow` deleted (no remaining callers).

New helper `flowService.getPublishedVersionOrThrow({ id, projectId })` consolidates the "fetch flow → assert publishedVersionId → fetch version" pattern.

## Concurrency model

Four areas a reviewer (or static analyzer) would flag as suspicious, and why each is safe:

1. **Concurrent `submitForApproval` on the same flow.** Closed by `UNIQUE(flowVersionId)` + `INSERT ... ON CONFLICT DO NOTHING` (`.orIgnore()`) + read-back-and-compare-id. Pattern matches `waitpoint-service.ts:25-67`. `LOCK_FLOW` is in-place (same version id), so two racers converge on the same `flowVersionId` and the unique index serializes them — one wins, the other receives the friendly "approval already exists" error.

2. **State guards in `approve` / `reject` / `withdraw` are not atomic with the write.** The pre-check (`if (state === APPROVED) return approval`) and the UPDATE happen as two separate statements — a concurrent request can slip in between. This is intentional and matches the CE publish path (`flowService.updatedPublishedVersionId`), which has no atomic guard either. The race outcome is benign:
   - Both racers write the **same** `lockedVersion.id` via `setPublishedVersion` → idempotent side effect, both converge on the same end state.
   - Last writer wins on the audit fields (`approverId`, `decidedAt`). Audit-event duplication is the only observable cost, accepted on parity with the CE publish path.
   - The pre-check is a *cheap fast-path*, not a correctness guarantee. It exists so a repeated approve doesn't redundantly re-run `setPublishedVersion` + `applyStatusChange` + emit an audit event — it's not load-bearing for race safety.

3. **`reject` writes without a transaction wrapper.** Reject is a single statement (`UPDATE flow_approval_request SET state = REJECTED, ...`) plus a fire-and-forget audit event. There is no second write to group atomically. The audit event is dispatched after the UPDATE succeeds and is itself fire-and-forget via `rejectedPromiseHandler` — if it fails, the row state is still correct. Wrapping a single statement in a transaction adds zero atomicity guarantee.

4. **`approve` calls `triggerSourceService.disable()` *before* the DB commit.** This is deliberate, with `ignoreError: true`. Trigger registrations are external state (webhooks at third parties, scheduled jobs in BullMQ); we want them disabled *before* the new version becomes the published one, even at the cost of running the disable on a transaction that may roll back. If the transaction rolls back, the old triggers are gone but the flow row is unchanged — the user retries approve and the disable becomes a no-op (already disabled), then succeeds. The opposite ordering (disable after commit) is what's actually dangerous: a disable failure would leave old triggers active against the new published version, double-firing webhooks. `applyStatusChangeForPublishedFlow` (which re-registers triggers for the new version) runs *after* commit precisely so its failure can't strand the system in a half-published state.

## Resilience

- Sensitivity is computed at submit time only — never re-evaluated during review.
- Approver permission re-checked server-side at decision time.
- Submitter departure does not void the request (`submitterId` is `SET NULL` on user delete; row remains).
- Flow delete cascades the request row.
- Withdraw deletes the row but the audit log retains the event.
- Disabling `environmentsEnabled` while PENDING requests exist is a no-op on those rows; admins can re-enable to resolve them. (Earlier draft auto-withdrew on disable; reverted because the unbounded `IN (:...ids)` pattern could exceed PostgreSQL's bind-parameter limit on large platforms.)

## Test plan

- `tsc --noEmit` clean on shared / api / web (only pre-existing chat module errors remain).
- `turbo run lint` clean (0 errors, only pre-existing warnings).
- Backend integration tests (`test/integration/ee/flow-approval/flow-approval.test.ts`, 10 tests, all green on real Postgres):
  - Routing decision (4 cases): environments off, project not sensitive, approver with override, sensitive + non-admin → NEEDS_APPROVAL.
  - Submit-for-approval race: two concurrent submits → exactly one PENDING row, second rejects with friendly error.
  - Approve happy path + idempotency: second approve is a no-op, flow unchanged.
  - Reject happy path: state=REJECTED, `rejectionReason` persisted, `publishedVersionId` stays null.
  - Withdraw happy path: row deleted, version state back to DRAFT.
  - Module gating: `environmentsEnabled=false` → 402.
- Frontend `discard-draft` unit tests pass (5 tests, `packages/web/test/app/builder/flow-canvas/widgets/discard-draft.test.ts`).


amrdb:
Closing in favor of a fresh PR with an updated description reflecting the current state of the branch (atomic state guards, collapsed migration, refreshed concurrency model). Branch and commit history unchanged.

PR #13105, opened 2026-05-05
feat(event-streaming): handler flow generator + dispatch hardening

AbdulTheActivePiecer:
## Summary

- **One-click handler flow generation.** The New Destination dialog now has a *Generate handler flow* button: pick events, click generate, and a webhook-triggered flow lands in the user's personal project with one router branch per event (plus a nested status sub-router under `flow.run.finished`). The trigger arrives pre-seeded with a typed mock per selected event — sample data is in place when the builder loads, and the existing template conditions (`trigger.body.action`, `trigger.body.data.flowRun.status`) resolve immediately.

- **Dispatch hardening.**
  - `setup()` was never called on `eventDestinationService`, so listeners were never registered and audit events never reached the dispatch path. Now wired in `platformWebhooksModule`, with a regression test that exercises both the user-event and worker-event listeners end-to-end.
  - PROJECT-scope destinations now filter by `platformId` in addition to `projectId`, closing a multi-tenant isolation gap.
  - Recursion guard rewritten. Previously only the destination targeting the event's own flow was skipped, which left A↔B mutual cycles unhandled. Now when the source flow appears in the internal destinations set, all internal destinations are dropped while external destinations (Slack, SIEM, etc.) keep firing — breaking the cycle without losing legitimate notifications on the first hop.
  - The dispatched payload is now the full `ApplicationEvent` (action + data + envelope). The job-data schema is tightened from `z.unknown()` to `ApplicationEvent`.

- **Sidebar regroup.** Audit Logs and Event Streaming now live together under a new **Observability** group, between Security and Infrastructure. Security becomes purely access control.

- **Docs cross-linking.** `event-streaming.mdx` links inline to the audit-events catalog instead of duplicating the event list. `audit-logs/overview.mdx` gets a *Forward them to your tools* section pointing at Event Streaming, plus a See-also footer.

- **`buildMockEvent` moved to `@activepieces/shared`** so both the backend test endpoint and the frontend handler-flow seeding build typed `ApplicationEvent` mocks for every `ApplicationEventName`.

## Test plan

- [ ] Sidebar: **Observability** group appears between Security and Infrastructure, containing Audit Logs and Event Streaming.
- [ ] New Destination dialog: events checkboxes, URL field, **Generate handler flow**, and per-event **Test webhook** dropdown all behave correctly.
- [ ] Generate handler flow with a single event selected → imported flow has webhook trigger + single-branch router; trigger shows sample data matching the selected event.
- [ ] Generate with multiple events including `FLOW_RUN_FINISHED` → branch count matches selection; nested status sub-router under the `FLOW_RUN_FINISHED` branch.
- [ ] Open the imported flow: trigger is already tested (no "Test Trigger" prompt); test history lists one entry per selected event.
- [ ] PROJECT-scope destination dispatches only when both `platformId` and `projectId` match.
- [ ] Self-targeting destination skipped on `flow.run.finished`; external siblings still dispatch.
- [ ] A↔B mutual cycle: both internal destinations dropped; any external destination still fires.
- [ ] Trigger and Test ship a payload with `action` + `data` + envelope fields.
- [ ] Audit events from both user-side actions (e.g. flow created) and worker-side actions (e.g. flow run finished) reach configured destinations.
- [ ] `npm run lint-dev` passes.
- [ ] `cd packages/server/api && npm run test-cloud` event-destinations suites pass.


AbdulTheActivePiecer:
@greptileai check the description again and redo your review, resolve the comments related .

AbdulTheActivePiecer:
@greptileai in regards of "The cycle guard in skipInternalDestinationsOnFlowCycle still drops every internal destination (including non-cycle ones like flowB)" that's intended, we don't want to fire to any internal destination flow once an internal destination flow finishes running.

PR #13112, opened 2026-05-05
fix(deftform): use responses array instead of fields in get-form-responses

cumonvip1:
## What does this PR fix?

The `Get Form Responses` action was silently returning empty field
values. The Deftform API returns form field data under the **`responses`**
key — a `{ key, value }[]` array — and NOT under `fields`. The old code
read `i['fields']` which is always `undefined`, so field labels and user
inputs never appeared in the output.

**Before** (output per item):
```json
{ "uuid": "...", "id": null, "form_id": null, "created_at": "..." }

After (output per item, flat / spreadsheet-ready):

{
  "uuid": "51c847ad-...",
  "number": 1,
  "number_formatted": "fem2025-1",
  "Anrede": "Mr.",
  "Vorname": "Test",
  "Name": "Tester",
  "Geburtsdatum": "08.05.2014",
  "Emailadresse ...": "doyev69054@coasah.com",
  "Sicherheitsabfrage ...": "7",
  "created_at": "2025-06-25T10:23:01.000000Z"
}

**Checklist**

-  No any types — uses Record<string, unknown> and typed response items
-  Flat output suitable for Google Sheets / Tables
-  Backwards compatible: the change only affects the responses key, all other fields are left as-is


cumonvip1 (commented):
tsconfig.base.json fixed

cumonvip1 (commented):
get-form-responses.ts
update-form-settings.ts
tsconfig.lib.json
tsconfig.base.json
all fixed

cumonvip1 (commented):
improve the get-form-responses action

PR #13118, opened 2026-05-06
fix(agent): preserve flow context for nested piece tools

### LINKED ISSUE #12993: [BUG]: Agent piece tool calls don't inherit the parent flow's `flowId` / `flowRunId`
When an Agent piece runs inside a real flow and the LLM invokes a registered tool (a piece action), the tool's execution context ends up with the placeholder MCP IDs instead of the parent flow's actual IDs.

Inside the tool's `run(ctx)`:
- `ctx.flows.current.id` is `'mcp-flow-id'`
- `ctx.run.id` is `'mcp-flow-run-id'`

Those are the `DEFAULT_MCP_DATA` sentinels from `@activepieces/shared`. They're correct when an action runs standalone (one-shot tool runs, MCP server calls, etc.), but they shouldn't be there when the action is invoked from an agent step inside a real flow run.

The cause is in `packages/server/engine/src/lib/tools/index.ts`. The `execute` function builds the constants for the tool call like this:

```ts
constants: EngineConstants.fromExecuteActionInput(operation),
```

`fromExecuteActionInput` always substitutes `DEFAULT_MCP_DATA`, so the parent flow's real `flowId` / `flowRunId` (which are right there on the `engineConstants` param passed into `agentTools.tools(...)`) never make it through to the tool.

## Why this matters

Pieces use `ctx.flows.current.id` and `ctx.run.id` to scope side effects to the flow that triggered them. A common pattern: stamp the flow id onto a record so a later step in the same flow can fetch only what this flow wrote. When those IDs are placeholders, that scoping breaks. And every record stamped from any agent flow ends up with the literal string `'mcp-flow-id'`, which looks like real data and is a pain to debug.

We hit this in our app: every order created via an agent flow was tagged with `ap_flow_id = 'mcp-flow-id'` and `ap_run_id = 'mcp-flow-run-id'`. We assumed it was a bug in our own piece at first and chased it for a while before tracing it back here.

## Repro

1. Create a flow with an Agent step.
2. Register any piece action as a tool. In its `run(ctx)`, log `ctx.flows.current.id` and `ctx.run.id`.
3. Run the flow and have the agent invoke the tool.
4. The logged values are `'mcp-flow-id'` / `'mcp-flow-run-id'`, not the real IDs.

Expected: the tool sees the parent flow's `flowId` and `flowRunId`.

## Suggested fix

In `agentTools.execute`, propagate the parent flow's IDs instead of calling `fromExecuteActionInput`. The `engineConstants` already passed into `agentTools.tools(...)` has them. Either build a new `EngineConstants` that carries `flowId` / `flowRunId` over, or extend `fromExecuteActionInput` to accept optional overrides so the tool path can pass them in.

Happy to open a PR if you'd like.

iuyua9:
## What does this PR do?

Fixes #12993.

Agent-invoked piece tools now run with the parent flow's `EngineConstants` instead of rebuilding constants from `ExecuteToolOperation`, which substituted the `DEFAULT_MCP_DATA` sentinel IDs. This lets nested piece actions see the real `ctx.flows.current.id`, `ctx.run.id`, flow version, and step names when an Agent step calls a registered piece tool inside a real flow run.

The standalone tool execution fallback still uses `EngineConstants.fromExecuteActionInput(...)` when no parent flow constants are present.

`agentTools.tools(...)` already receives the parent flow `engineConstants`. This PR carries those constants into the nested tool execution and uses them for the underlying piece executor.

### Relevant User Scenarios

- Agent flows that call piece tools whose actions scope writes, logs, or records by `ctx.flows.current.id` / `ctx.run.id` no longer stamp `mcp-flow-id` / `mcp-flow-run-id` into real flow-run data.

## Validation

- `git diff --check`
- `bunx tsc --noEmit --project packages/server/engine/tsconfig.json --pretty false`
- `bunx tsc --noEmit --project packages/server/engine/tsconfig.spec.json --pretty false` (no errors from the new `agent-tools.test.ts`; existing unrelated spec type errors remain)
- built local `@activepieces/shared` and `@activepieces/pieces-framework` dist packages, then ran `bunx vitest run packages/server/engine/test/handler/agent-tools.test.ts` → 1 passed

### DIFF
--- a/packages/server/engine/src/lib/tools/index.ts
+++ b/packages/server/engine/src/lib/tools/index.ts
@@ -174,7 +174,7 @@ async function execute(operation: ExecuteToolOperationWithModel): Promise<Execut
         const output = await flowExecutor.getExecutorForAction(step.type).handle({
             action: step,
             executionState: FlowExecutorContext.empty(),
-            constants: operation.parentFlowConstants ?? EngineConstants.fromExecuteActionInput(operation),
+            constants: getToolExecutionConstants(operation),
         })
         const { output: stepOutput, errorMessage, status } = output.steps[operation.actionName]
         
@@ -246,6 +246,10 @@ ${propertyDetailsSection}
 type ExecuteToolOperationWithModel = ExecuteToolOperation & {
     model: LanguageModel
     parentFlowConstants?: EngineConstants
+}
+
+export function getToolExecutionConstants(operation: ExecuteToolOperationWithModel): EngineConstants {
+    return operation.parentFlowConstants ?? EngineConstants.fromExecuteActionInput(operation)
 }
 
 async function propertyToSchema(propertyName: string, property: PieceProperty, operation: ExecuteToolOperation, resolvedInput: Record<string, unknown>): Promise<z.ZodTypeAny> {
--- a/packages/server/engine/src/lib/tools/index.ts
+++ b/packages/server/engine/src/lib/tools/index.ts
@@ -33,6 +33,7 @@ export const agentTools = {
                         pieceVersion: tool.pieceMetadata.pieceVersion,
                         actionName: tool.pieceMetadata.actionName,
                         predefinedInput: tool.pieceMetadata.predefinedInput,
+                        parentFlowConstants: engineConstants,
                         model,
                     }),
             }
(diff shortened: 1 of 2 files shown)


iuyua9:
Thanks for the review. I pushed a follow-up that adds the missing fallback coverage: `getToolExecutionConstants(...)` now has a targeted test for the no-parent-constants path, asserting it still falls back to `DEFAULT_MCP_DATA` and normalizes the internal API URL.

Validation after the follow-up:
- `git diff --check`
- `bunx tsc --noEmit --project packages/server/engine/tsconfig.json --pretty false`
- `bunx tsc --noEmit --project packages/server/engine/tsconfig.spec.json --pretty false` (no `agent-tools.test.ts` errors; existing unrelated spec type errors remain)
- `bunx vitest run packages/server/engine/test/handler/agent-tools.test.ts` → 2 passed

abuaboud:

PR #13134, opened 2026-05-06
fix(agent): adding multiple flow tools to Run Agent causes the agent to not see any tools

mzndako:
## Summary

Fixes a bug where adding 2 or more flow tools to the **Run Agent** step causes the agent to behave as if no tools are configured at all — it only sees `updateTaskStatus` and cannot execute any flow.

## Steps to Reproduce

1. Create two or more flows with an MCP trigger.
2. Add both as tools in a **Run Agent** step.
3. Run the agent and ask it to use either flow.
4. The agent responds that no tools are available, even though they are listed in the input.

> Adding only one flow tool works correctly. Both flows also work correctly when tested individually.

## Root Cause

When the engine fetches flow data for the configured tools, it serialises all external flow IDs as a single comma-joined query string value:

`GET /v1/engine/populated-flows?externalIds=id1,id2`

The server expects each ID as a separate repeated param:
`GET /v1/engine/populated-flows?externalIds=id1&externalIds=id2`

The comma-joined form is treated as one literal ID that never matches any row in the database, so **zero flows are returned**. The agent tools list is empty and the agent proceeds with no flow tools available.

## Impact
- Restores correct behavior when **multiple** flow tools are attached to Run Agent.
- No intentional behavior change for the single-flow case.

### DIFF
--- a/packages/server/engine/src/lib/piece-context/flows.ts
+++ b/packages/server/engine/src/lib/piece-context/flows.ts
@@ -6,7 +6,9 @@ export const createFlowsContext = ({ engineToken, internalApiUrl, flowId, flowVe
         async list(params: ListFlowsContextParams): Promise<SeekPage<PopulatedFlow>> {
             const queryParams = new URLSearchParams()
             if (params?.externalIds) {
-                queryParams.set('externalIds', params.externalIds.join(','))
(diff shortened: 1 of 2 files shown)


abuaboud:

PR #13136, opened 2026-05-06
fix(savvycal): unbreak Cloud OAuth, complete Create Event

bst1n:
## Summary

The merged SavvyCal piece was effectively non-functional on Activepieces Cloud — every action and trigger failed at runtime. This PR fixes the critical auth bug, completes the `Create Event` action with all SavvyCal fields, fills in 6 missing event states, and polishes French i18n.

### 1. Cloud OAuth2 auth was broken (critical)
`getToken()` in `auth.ts` only handled `AppConnectionType.OAUTH2` and fell through to `auth.props.token` for any other type. On Cloud, OAuth2 connections are typed `CLOUD_OAUTH2`, so the check failed, the fallback returned `undefined`, and the API received `Authorization: Bearer undefined`.

**User-visible effects:**
- Dropdowns (teams, scheduling links, workflows) silently showed `Failed to load…`
- Run-time actions (e.g. `Get Current User`, `Create Event`) crashed with `TypeError: Cannot read properties of undefined (reading 'token')`

**Fix:** handle `OAUTH2`, `CLOUD_OAUTH2`, and `PLATFORM_OAUTH2` — all three carry `access_token` at the same level.

### 2. List Events: 6 missing state options
The state filter only exposed `confirmed` and `canceled`. SavvyCal's API supports 8 states. Added: `awaiting_reschedule`, `awaiting_checkout`, `checkout_expired`, `awaiting_approval`, `declined`, `tentative`. ([SavvyCal docs](https://developers.savvycal.com/api/schemas/event))

## Create Event completion

The action was missing fields the API documents. Added:

- **Phone Number** — optional `phone_number` (top-level field, required only if the link asks for it)
- **Custom Fields** — `Property.DynamicProperties` that, when a scheduling link is selected, fetches `GET /v1/links/{id}` and renders one input per configured form field (Company, Why are we meeting?, etc.). Field types are mapped to AP props (`short_text`/`url` → `ShortText`, `long_text` → `LongText`, `select_menu`/`radio_buttons` → `StaticDropdown`, `checkboxes` → `StaticMultiSelectDropdown`, `hidden` → skipped). The `is_required` flag from SavvyCal is respected. At runtime, values are reassembled into the `fields: [{id, label, type, value}]` payload the API expects.
- **Metadata** — optional `Property.Object` for arbitrary key/value pairs

## Polish

- **Action ordering in `index.ts`** — grouped by resource (Meetings → Scheduling Links → Workflows → Account → Custom API), reads before mutations, destructive operations last. Helps users navigate the 14-action menu.
- **French i18n revision** — for B2B scheduling clarity:
  - `Event` → `Réunion` (more natural in French than the literal `Événement` for a meeting-booking tool)
  - `Scheduling Link` → `Lien de réservation` (matches French booking conventions; less awkward than `Lien de planification`)
  - Spelling: `événement` → `évènement` (modern Académie française recommendation)
  - Added 11 new translation keys (5 missing event states + 6 new Create Event fields)
  - `Workflow` kept untranslated (it's the SavvyCal product feature name)

## Note on the second commit

`chore(api)` is an auto-fix of pre-existing lint errors in `admin-platform.service.ts` (unused imports + trailing comma) introduced by `22444d0475 fix: admin retry`. They were blocking the pre-push lint gate. Pure auto-fix, no logic touched.

## Version

Bumped `0.1.0` → `0.1.1`.

## Test plan

- [ ] On Activepieces Cloud (OAuth2 connection): every dropdown loads (teams / scheduling links / workflows)
- [ ] On Cloud: `Get Current User` returns user data without crash
- [ ] On Cloud: `Create Event` books a meeting with phone number + custom fields + metadata
- [ ] On self-hosted with personal access token: same coverage works
- [ ] `List Events` filters correctly by each of the 8 states
- [ ] French UI shows the new `Réunion` / `Lien de réservation` translations consistently


bst1n:
### Code review

Found 1 issue:

1. `translation.json` now contains 5 duplicate JSON keys. The keys `Custom API Call`, `Headers`, `Query Parameters`, `Body`, and `Timeout (in seconds)` already existed at lines 12, 55, 56, 58, 61 from the original piece commit (8d825cf3d4). The Custom API Call i18n fix-up commit (e0a7bc0809) re-added all 5 at lines 138-143 without checking whether they were present. JSON parsers silently keep the last value (so behavior is unchanged), but the file is malformed and now has redundant entries that will drift. Drop the 5 redundant lines from `translation.json` (`fr.json` is clean).

https://github.com/activepieces/activepieces/blob/e0a7bc0809addea427204e8b3961b8f6c5224935/packages/pieces/community/savvycal/src/i18n/translation.json#L137-L144

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kishanprmr (approved):

PR #13145, opened 2026-05-07
fix(apify): remove deprecated crawler option and validate dataset pagination

jakcinmarina:
## What does this PR do?

This PR updates the Apify piece with small usability and reliability fixes:

- Removes deprecated **JSDOM** from the **Scrape Single URL** crawler options.
- Adds input validation in **Get Dataset Items**:
  - `limit` must be greater than `0`
  - `offset` must be greater than or equal to `0`
- Updates memory field wording in **Run Actor** / **Run Task** for clearer unit context.

It also bumps the Apify piece version from **0.2.1** to **0.2.2**.

- In `Scrape Single URL`, the deprecated JSDOM crawler option is removed from the static dropdown list, so users only see supported crawler types.
- In `Get Dataset Items`, runtime guards validate pagination inputs before calling Apify:
  - negative offsets are rejected
  - zero/negative limits are rejected
  - clear error messages are returned for invalid values
- In run actions, the memory field label text is adjusted for clarity while keeping the same dropdown values.

### Relevant User Scenarios

- Users configuring **Scrape Single URL** no longer select a deprecated crawler mode.
- Users passing dynamic values into **Get Dataset Items** avoid silent failures from invalid pagination values.
- Users configuring actor/task runs get slightly clearer memory configuration wording.

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/apify/package.json
+++ b/packages/pieces/community/apify/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-apify",
-  "version": "0.2.1",
+  "version": "0.2.2",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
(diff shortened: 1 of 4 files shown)


sanket-a11y (approved):

sanket-a11y:
closing this , merge changes  #13299

PR #13164, opened 2026-05-09
fix(deftform): use responses array instead of fields in get-form-responses

cumonvip1:
The Deftform API returns form response values under the 'responses' key
as a { key, value }[] array, not under 'fields'. The old code tried to
read i['fields'] which is always undefined, so field labels and user
inputs never appeared in the output.

This fix reads i['responses'], spreads each { key, value } pair as a
flat column, and ensures system fields (uuid, number, created_at) take
precedence over any form field that might share their name.


sanket-a11y (approved):

sanket-a11y:
@cumonvip1  Could you please remove the non-piece changes from this PR?  remove changes from `docs/resources/screenshots`

cumonvip1:
@sanket-a11y should be done now! Thanks! Please feel free to review again.

sanket-a11y (approved):

PR #13165, opened 2026-05-09
feat(subflows): allow selecting disabled subflows

AdamSelene:
## What does this PR do?

Preventing disabled subflows from appearing in the UI (where they could have been selected previously) is confusing

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/packages/pieces/core/subflows/package.json
+++ b/packages/pieces/core/subflows/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-subflows",
-  "version": "0.4.12",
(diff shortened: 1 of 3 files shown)


AbdulTheActivePiecer:
We prefer to not show the subflows that are inactive so no confusion happens when they don't execute, however we can edit the description to say that inactive and unpublished subflows won't appear in the dropdown, which we will do in another pr.

AbdulTheActivePiecer:
https://github.com/activepieces/activepieces/pull/13202

PR #13167, opened 2026-05-09
fix(builder): notify user when trigger test returns no events

### LINKED ISSUE #13166: [BUG]: "Generate Sample Data" silently fails when there are no recent event
**Describe the bug**
When a polling-style trigger's test function succeeds but returns zero events, the **Generate Sample Data** / **Load Sample Data** button silently does nothing, no error, no message, no state change. To the user it looks as if the click never registered.

This adds confusion to the users.

A reproducible scenario: connecting a freshly-created Telegram bot whose token is valid but who has not yet received any messages. The bot's `test()` hook calls Telegram's `getUpdates?offset=-5`, which legitimately returns an empty list, and the UI provides no feedback about this outcome.

**To Reproduce**
1. Create a new project in Activepieces.
2. Create a new flow and add the **Telegram Bot → New message** trigger.
3. Create a fresh Telegram bot via BotFather and connect it (valid token), without sending any message to the bot first.
4. Click **Generate Sample Data**.
5. Observe in DevTools: `POST /api/v1/test-trigger` returns HTTP 200 with `{ "next": null, "previous": null, "data": [] }`. The UI shows no feedback — same state as before the click.

**Expected behavior**
The user should be told that no recent events were found and given guidance on how to produce one. for example, an inline notice indicating that no data was found in the connected piece, asking them to perform an action there before retesting.

**Screenshots**
Image

**Additional context**
N/A

that-one-arab:
## What does this PR do?

When a polling-style trigger's test function succeeds with an empty result (e.g., a freshly-connected Telegram bot whose owner hasn't sent any messages yet), **Generate Sample Data** previously did nothing visible, leaving users to wonder what went wrong. This PR surfaces that state with an inline notice that names the connected piece and tells the user how to proceed.

`usePollTrigger` now accepts a `setInfoMessage` setter and the piece's display name. When `POST /api/v1/test-trigger` returns `{ data: [] }` — a successful-but-empty test — the hook sets a translated, ICU-interpolated message:

> We didn't find any data in {pieceName} yet. Do something there that should start this flow, then test again.

`TestTriggerSection` owns the `infoMessage` state and renders it as an inline `<Alert>` with an info icon in two places:
- Above `FirstTimeTestingSection` for users who have never tested this trigger.
- Inside `TestSampleDataViewer` for users retesting; the previous sample stays visible while the notice explains why nothing changed.

The notice is cleared at the start of every test attempt and on error, so it never co-renders with an error message. No backend changes; no piece-framework changes.

### Relevant User Scenarios

- **Newly-connected bot/integration with no events yet.** A user connects a Telegram bot via BotFather, clicks **Generate Sample Data** before any message has been sent to the bot, and gets clear guidance instead of a silent no-op.
- **Retesting after a quiet period.** A user who tested successfully earlier retests later when no new events have come in; the notice tells them why no new sample data appeared, while their prior sample stays usable.
- **Generalizes beyond Telegram.** Affects every polling-style test trigger (`testStrategy: TEST_FUNCTION`) — Slack, Gmail, calendar, etc. — that can legitimately return zero events from its `test()` hook.

#### Before

https://github.com/user-attachments/assets/0d1d057c-2262-40c1-8286-619dbcda6115

#### After

https://github.com/user-attachments/assets/305a5e52-f05e-45fd-9ed0-798b7939e6f7

Closes https://github.com/activepieces/activepieces/issues/13166

### DIFF
--- a/packages/web/public/locales/en/translation.json
+++ b/packages/web/public/locales/en/translation.json
@@ -264,6 +264,7 @@
   "Action Required": "Action Required",
   "No sample data available": "No sample data available",
   "Old results were removed, retest for new sample data": "Old results were removed, retest for new sample data",
+  "We didn't find any data in {pieceName} yet. Do something there that should start this flow, then test again.": "We didn't find any data in {pieceName} yet. Do something there that should start this flow, then test again.",
   "Result #": "Result #",
   "The sample data can be used in the next steps.": "The sample data can be used in the next steps.",
   "There is no sample data available found for this trigger.": "There is no sample data available found for this trigger.",
--- a/packages/web/src/app/builder/test-step/test-trigger-section/index.tsx
+++ b/packages/web/src/app/builder/test-step/test-trigger-section/index.tsx
@@ -1,9 +1,11 @@
 import { FlowTrigger, flowStructureUtil, isNil } from '@activepieces/shared';
 import { t } from 'i18next';
+import { Info } from 'lucide-react';
 import React, { useRef, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 
 import { ChatDrawerSource } from '@/app/builder/types';
+import { Alert, AlertDescription } from '@/components/ui/alert';
 import { triggerEventHooks } from '@/features/flows';
 import { piecesHooks } from '@/features/pieces';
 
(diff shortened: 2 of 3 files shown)


abuaboud:

that-one-arab:
@abuaboud There is a recording attached, that should be enough proof that the PR is not AI generated slop.

PR #13169, opened 2026-05-09
fix(docs): clarify AP_DEV_PIECES must be set in .env.dev

### LINKED ISSUE #13168: [BUG]: docs incorrectly suggest setting `AP_DEV_PIECES` via inline shell prefix
**Describe the bug**
The development setup docs at [`docs/build-pieces/building-pieces/development-setup.mdx`](https://www.activepieces.com/docs/build-pieces/building-pieces/development-setup#pieces-development) instruct contributors to set `AP_DEV_PIECES` via an inline shell prefix:

```sh
AP_DEV_PIECES=google-sheets,cal-com npm start
```

In practice this does not take effect for the running dev servers. Turborepo's `serve` task does not forward shell-set environment variables to its spawned `api` / `worker` tasks, so the dev servers only see the value configured in `.env.dev`. Contributors following the docs see their pieces silently ignored.

**To Reproduce**
1. Clone the repo and run `node tools/setup-dev.js` once.
2. Edit `.env.dev` so `AP_DEV_PIECES=cal-com`.
3. Try to override via shell: `AP_DEV_PIECES=google-sheets npm start`.
4. Observe that the dev server still loads `cal-com`, not `google-sheets`.

**Expected behavior**
The docs should make it clear that `AP_DEV_PIECES` (and similar dev env vars like `AP_PIECES_SYNC_MODE`) must be set inside `.env.dev`, and that inline shell prefixes do not propagate to the dev servers.

**Screenshots**
N/A

**Additional context**
- The build-time step (`tools/setup-dev.js`) *does* respect shell-set values because it reads `process.env.AP_DEV_PIECES || envConfig.AP_DEV_PIECES`. This is partly why the issue is easy to miss, the build picks up the shell value, but the running dev servers don't.
- The api/worker bootstraps call `dotenv.config()` without `override: true`, so dotenv itself doesn't overwrite shell values; the shell value simply never reaches the bootstraps because Turborepo strips it earlier.
- Same applies to `AP_PIECES_SYNC_MODE` and any other env var consumed by the dev servers.

that-one-arab:
## What does this PR do?

Clarifies the development setup docs around `AP_DEV_PIECES`. The current example tells contributors to set the variable via an inline shell prefix (`AP_DEV_PIECES=... npm start`), but Turborepo's `serve` task does not forward shell-set environment variables to its spawned tasks, so the api/worker dev processes only ever read from `.env.dev`. Contributors who follow the docs verbatim see their pieces silently ignored.

This PR replaces the misleading example with an instruction to edit `.env.dev` directly, and adds a `<Note>` callout matching the existing doc style flagging the inline-shell pitfall.

No code change, documentation only. The new wording in `docs/build-pieces/building-pieces/development-setup.mdx`:

- Tells the reader to set `AP_DEV_PIECES` inside `.env.dev` rather than on the command line.
- Shows the env-file line (`AP_DEV_PIECES=google-sheets,cal-com`) followed by `npm start`, so the same end-to-end flow is preserved.
- Adds a `<Note>` clarifying that inline shell variables are not picked up by the dev servers, so future readers understand why.

### Relevant User Scenarios

- A new contributor follows the **Pieces Development** section, sets `AP_DEV_PIECES=foo,bar npm start`, and is confused when the editor doesn't reflect their local pieces. With this PR they're directed to the file that actually controls the runtime configuration.
- Same applies to `AP_PIECES_SYNC_MODE`, the `.env.dev`-as-source-of-truth rule is the same.

Closes https://github.com/activepieces/activepieces/issues/13168

### DIFF
--- a/docs/build-pieces/building-pieces/development-setup.mdx
+++ b/docs/build-pieces/building-pieces/development-setup.mdx
@@ -43,8 +43,14 @@ Password: `12345678`
 
 When [`AP_PIECES_SYNC_MODE`](https://github.com/activepieces/activepieces/blob/main/.env.dev#L14) is set to `OFFICIAL_AUTO`, all pieces are automatically loaded from the cloud API and synced to the database on first launch. This process may take a few seconds to several minutes depending on your internet connection.
 
-For local development, pieces are loaded from your local `dist` folder instead of the database. To enable this, set the [`AP_DEV_PIECES`](https://github.com/activepieces/activepieces/blob/main/.env.dev#L5) environment variable with a comma-separated list of pieces. For example, to develop with `google-sheets` and `cal-com`:
+For local development, pieces are loaded from your local `dist` folder instead of the database. Set [`AP_DEV_PIECES`](https://github.com/activepieces/activepieces/blob/main/.env.dev#L5) inside `.env.dev` as a comma-separated list of pieces. For example, to develop with `google-sheets` and `cal-com`, edit `.env.dev`:
 
 ```sh
-AP_DEV_PIECES=google-sheets,cal-com npm start
+AP_DEV_PIECES=google-sheets,cal-com
 ```
+
+Then run `npm start`.
+
+<Note>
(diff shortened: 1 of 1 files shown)


abuaboud:

PR #13170, opened 2026-05-10
fix(docker): align worker image version with app (0.79.0 → 0.83.0)

pmarkoo:
## Summary

- Worker service in `docker-compose.yml` was pinned to image `0.79.0` while the app service was running `0.83.0`
- The version mismatch caused all workers to crash on startup because the `0.79.0` image could not read `AP_REDIS_HOST` from the environment
- Bumped worker image to `0.83.0` to match the app image and the root `package.json` version
- Basically nothing worked locally: the broken worker meant you couldn't test flows or use the UI at all, tested on MacOS and Ubuntu
-  Problem was likely introduced with release #13135  

## Test plan

- [x] Run `docker compose up` and verify both `app` and `worker` containers start successfully
- [x] Confirm `worker` logs show no startup errors related to missing env vars
- [x]  Confirm UI works properly and created simple flow

🤖 Co-Authored-By Claude Code


kishanprmr:
Fixed in https://github.com/activepieces/activepieces/commit/e498667107bd73b55768652c409edf83d2db1118.

PR #13174, opened 2026-05-10
feat: update API to allow creation of connection placeholders

AdamSelene:
## What does this PR do?

We want to be able to create "placeholder" connections via the API (shown as status "missing" in the UI)
Today it's only possible via project releases - it's not exposed to external systems.

### Relevant User Scenarios

Fixes # (issue)


AbdulTheActivePiecer:
will be merged via https://github.com/activepieces/activepieces/pull/13218

PR #13220, opened 2026-05-13
fix(run-details): render download UI for sliced step outputs missing discriminant

abuaboud:
## Summary

Sliced step outputs were being shown as raw JSON (`{ fileId, size, url }`) instead of the "Download output" prompt because slice detection in `flow-step-input-output.tsx` relied solely on `stepOutput.outputType === SLICE`. Step outputs written before the `kind → outputType` rename — or any payload where the discriminant fails to propagate — fell through to the inline JSON viewer.

Extract slice detection into a helper that:
- prefers `outputType`,
- falls back to the legacy `kind` field,
- structurally recognizes the `LogSliceRef` shape (exact `fileId | size | url`).

## Test plan

- [ ] Visit a flow run whose step output was uploaded as a slice with `outputType: SLICE` → "Download output" button renders.
- [ ] Visit an older run where the discriminant is missing/legacy → "Download output" button still renders.
- [ ] A step whose normal output happens to include `fileId`/`size`/`url` plus other keys → still rendered as JSON (structural check requires exact 3-key shape).

### DIFF
--- a/packages/web/src/app/builder/run-details/flow-step-input-output.tsx
+++ b/packages/web/src/app/builder/run-details/flow-step-input-output.tsx
@@ -8,6 +8,7 @@ import {
   isNil,
   ApFlagId,
   LogSliceRef,
+  StepOutput,
   StepOutputType,
 } from '@activepieces/shared';
 import { t } from 'i18next';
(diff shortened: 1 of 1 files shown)


abuaboud:
Superseded — landed directly on main as fa1f43b892 with the simplified outputType-only check.

PR #13234, opened 2026-05-14
fix(mcp): switch from enabledTools to disabledTools so new tools auto-appear

hazemadelkhalel:
## Summary
- **Problem**: When users toggled MCP tools in the UI, the `enabledTools` array was saved to the database. New tools added later (e.g. `ap_delete_flow`) never appeared for users who previously customized their tool list — only users who never customized (null `enabledTools`) got new tools automatically.
- **Fix**: Switched from an allowlist (`enabledTools`) to a denylist (`disabledTools`). New tools are now enabled by default because they aren't in anyone's disabled set.
- **Migration**: Adds a DB migration that inverts `enabledTools` → `disabledTools` for all existing rows and drops the old column.

## Test plan
- [x] All 97 MCP integration tests pass (90 CE + 5 RBAC + 2 piece-visibility)
- [x] Lint passes with 0 errors
- [ ] Manual: toggle tools on/off in project MCP settings, verify state persists
- [ ] Manual: toggle tools on/off in platform MCP settings, verify state persists
- [ ] Manual: verify a newly added tool appears enabled for existing MCP configurations


hazemadelkhalel:
Fixed in b59419b — updated all 4 `enabledTools` references in `.agents/features/mcp.md` to `disabledTools`.

Issue #13157, opened 2026-05-08
x402/Pyrimid piece for paid agent and API actions


rizzdbx:
## Proposal: x402/Pyrimid piece for paid agent and API actions

Activepieces has the workflow layer. A useful missing primitive is “call this paid agent/API action and settle automatically.”

I’m working on Pyrimid, an x402 payment router on Base. It lets an agent/workflow hit an endpoint, receive HTTP 402, pay USDC, then retry with payment proof. That maps pretty cleanly to an Activepieces piece:

- trigger/action calls a paid endpoint
- handles `402 Payment Required`
- routes payment through Pyrimid
- returns the paid API result + tx/proof metadata

Live test surfaces:

- Catalog: https://pyrimid.ai/api/v1/catalog
- Example 402 endpoint: https://pyrimid.ai/api/v1/paid/agentzone-search?q=agent-commerce
- Quickstart: https://pyrimid.ai/quickstart
- Proof/stats: https://pyrimid.ai/proof / https://pyrimid.ai/stats

If maintainers are open to it, I can contribute a small piece + docs/example workflow. The goal is narrow: let Activepieces workflows buy paid agent/API services programmatically, without a custom billing integration per vendor.

PR #13249, opened 2026-05-15
feat: add x402 payment piece for paid API calls

597226617:
## Summary

Adds an **x402 Payment Piece** to Activepieces that enables workflows to call paid external APIs using the [x402 HTTP 402 payment protocol](https://github.com/x402-foundation/x402).

### What it does
1. Action: `Call Paid API with x402`
2. Takes endpoint URL, request body, max price, network as input
3. Handles HTTP `402 Payment Required` responses
4. Signs and sends USDC payment on Solana (SPL TransferChecked)
5. Retries with `X-Payment` header
6. Returns paid API response + transaction hash

### x402 Protocol
x402 is an open payment protocol for AI agents — pay per API call with on-chain USDC. No API keys, no accounts, no subscriptions. Supported by the Linux Foundation.

### Files Added
- `packages/pieces/community/x402/src/index.ts` — Piece definition
- `packages/pieces/community/x402/src/lib/actions/call-paid-api.ts` — x402 payment action (198 lines)
- `packages/pieces/community/x402/src/lib/auth.ts` — Wallet credential auth
- `packages/pieces/community/x402/src/i18n/translation.json` — i18n support
- `packages/pieces/community/x402/README.md` — Usage documentation

### Related
- x402 Foundation: https://github.com/x402-foundation/x402
- Issue: https://github.com/activepieces/activepieces/issues/13157


TateLyman:
Ran a no-payment protocol read-through of the x402 piece. I did not send payment headers, sign requests, or attempt paid calls. One important payment-safety issue still looks present despite the latest fixes:

- `maxAmountRequired` should be handled as an atomic-unit string selected from a payment requirement, not as a human USDC decimal. The current code types it as `number`, compares it directly to the user-facing `maxPrice` in USDC, then multiplies it by `10 ** decimals` again before `TransferChecked`. For a standard USDC challenge where `maxAmountRequired` is already micro-USDC, that double-scales the transfer amount and makes the max-price guard compare mismatched units.

There is also an interoperability issue to fix with the challenge shape: common x402 responses expose one or more requirements in `accepts[]`, so the action should select a compatible Solana requirement first, then use that requirement's `network`, `asset`/mint, `payTo`, `maxAmountRequired`, and `extra.decimals`.

Suggested patch order:

1. Parse `accepts[]` and select a Solana requirement for the configured network.
2. Keep `maxAmountRequired` as a string/`bigint` in atomic units.
3. Convert `maxPrice` USDC into atomic units using the selected requirement decimals before comparing.
4. Pass the atomic amount directly to `createTransferCheckedInstruction` instead of multiplying by decimals again.
5. Validate that `asset`/`extra.mint` matches the expected USDC mint for the selected Solana network.

That keeps the piece aligned with standard x402 challenge bodies and avoids accidental overpayment/rejection on valid low-price endpoints.

597226617:
## x402 Protocol Compliance Fixes

Addressed all feedback from @TateLyman's review:

### 1. ✅ Parse \`accepts[]\` array (V2 protocol)
- Added \`parsePaymentRequirement()\` function that checks for V2 \`accepts[]\` first, falls back to V1 top-level fields
- Selects Solana-compatible requirement matching the configured network
- Supports both CAIP-2 network IDs (\`solana:mainnet-beta\`) and simple names (\`mainnet-beta\`)

### 2. ✅ Fix \`maxAmountRequired\` atomic unit handling
- Changed interface to \`string | number\` type
- Amount is now treated as atomic units throughout (no more double-scaling)
- User's \`maxPrice\` (USDC decimal) is converted to atomic units for comparison
- Raw atomic amount passed directly to \`createTransferCheckedInstruction\`

### 3. ✅ Validate USDC mint address
- Added \`USDC_MINTS\` constant with known addresses per Solana network
- Validates \`extra.mint\` against expected USDC contract
- Throws clear error if mismatch detected

### 4. ✅ Authors field fix
- Changed from \`"activepieces"\` to \`"sunbei597226617"\` per community piece convention

TateLyman:
Follow-up from a second code read against current public x402 response shapes. No wallet signing, payment headers, paid calls, or private endpoint probes.

The current `accepts[]` fix still assumes each selected requirement has a `price` field:

```ts
interface X402PaymentRequirement {
  price: number | string;
  ...
}
...
const maxAmountAtomic = BigInt(requirement.price);
```

Live services are using multiple accept-entry shapes right now, including `maxAmountRequired` and `amount` rather than `price`. For example, one public 402 challenge uses `maxAmountRequired: "0.001"`, while another uses `amount: "1000"`. With either of those, this piece can select the Solana requirement but then pass `undefined` into the BigInt/transaction path.

Patch I would make before merge:

1. Normalize the selected requirement amount from `req.maxAmountRequired ?? req.amount ?? req.price` with a clear error if none exist.
2. Preserve the field semantics: atomic `amount` should stay atomic; decimal `maxAmountRequired`/`price` should be converted using the selected decimals when the service advertises decimal USDC.
3. Add fixtures for both current shapes so the parser does not regress.
4. Keep the max-price comparison after normalization, with both values in the same unit.

The earlier double-scaling issue is mostly addressed, but this field-name/semantics gap can still break the first valid endpoint the piece is tested against.

597226617:
@cla-assistant please recheck

abuaboud:

PR #13261, opened 2026-05-17
fix(web): make system theme follow OS dark/light preference

### LINKED ISSUE #13163: [BUG]: System theme mode does not follow OS dark/light mode changes
**Describe the bug**
When the theme is set to "System" in the appearance settings, the UI does not react to OS-level dark/light mode changes. The interface stays on whatever theme was active when the setting was applied and does not update when the OS switches modes.

This regression appears to have been introduced in a recent update (likely PR #12761 which added a `forceLightMode` flag to the ThemeProvider).

**To Reproduce**
1. Go to Settings → Appearance
2. Select "System" as the color mode
3. Switch your OS between dark and light mode (e.g. via macOS System Settings)
4. Observe that the Activepieces UI does not update accordingly

**Expected behavior**
The UI should automatically switch between dark and light themes to match the current OS preference (via `prefers-color-scheme` media query).

**Screenshots**
N/A

**Additional context**
- OS: macOS Tahoe 26.4.1
- Browser: Zen 1.19.11b (Firefox-based)
- Activepieces version: 0.82.1
- Manually selecting "Dark" or "Light" works correctly — only "System" mode is affected
- Confirmed working before the recent update

serhiizghama:
## Problem

Fixes #13163.

When the theme is set to **System** in Appearance settings, the UI does not react to OS-level dark/light mode changes — it always renders in light mode regardless of the OS preference, and never updates when the OS switches modes.

Root cause is in `packages/web/src/components/providers/theme-provider.tsx`:

```ts
const resolvedTheme = forceLightMode
  ? 'light'
  : theme === 'system'
  ? 'light'         // ← unconditionally light when theme === 'system'
  : theme;
```

The `'system'` branch hard-codes `'light'` instead of resolving against `prefers-color-scheme`, and there is no listener on `matchMedia` to re-render when the OS preference changes at runtime.

## Solution

- Resolve `'system'` against `window.matchMedia('(prefers-color-scheme: dark)')` instead of hard-coding `'light'`.
- Track the system preference in a `systemTheme` state and subscribe to the `MediaQueryList` `change` event, so the UI updates live when the OS toggles dark/light mode.
- Keep `forceLightMode` (used by the auth screens) as the highest-priority override — unchanged behaviour.

The change is scoped to the resolver and a new `useEffect` that adds/removes the `matchMedia` listener. Manual `'light'` / `'dark'` selections are unaffected.

## Testing

1. Settings → Appearance → **System**.
2. Toggle OS dark/light mode (macOS: System Settings → Appearance; Windows: Settings → Personalization → Colors).
3. Activepieces UI now switches between light and dark in sync with the OS.
4. Set **Light** or **Dark** explicitly → toggling OS mode has no effect (expected).
5. Sign-out flow (which sets `forceLightMode=true` via `auth-form-template.tsx`) still renders light regardless of OS preference (expected).

### DIFF
--- a/packages/web/src/components/providers/theme-provider.tsx
+++ b/packages/web/src/components/providers/theme-provider.tsx
@@ -5,6 +5,7 @@ import { flagsHooks } from '@/hooks/flags-hooks';
 import { colorsUtils } from '@/lib/color-utils';
 
 type Theme = 'dark' | 'light' | 'system';
+type ResolvedTheme = 'dark' | 'light';
 
 type ThemeProviderProps = {
   children: React.ReactNode;
@@ -28,6 +29,14 @@ const initialState: ThemeProviderState = {
 
 const ThemeProviderContext = createContext<ThemeProviderState>(initialState);
 
+const SYSTEM_DARK_QUERY = '(prefers-color-scheme: dark)';
+
+const getSystemTheme = (): ResolvedTheme =>
+  typeof window !== 'undefined' &&
+  window.matchMedia(SYSTEM_DARK_QUERY).matches
+    ? 'dark'
+    : 'light';
+
 const setFavicon = (url: string) => {
   document.querySelectorAll("link[rel*='icon']").forEach((el) => el.remove());
   const link = document.createElement('link');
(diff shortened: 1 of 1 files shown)


AbdulTheActivePiecer:
image
please make sure you sign your CLA and please check the greptile comments on your pr

serhiizghama:
Closing — the CLA was never signed (flagged by CLAassistant on 2026-05-17) and there were outstanding review comments from the maintainer that haven't been addressed. Can't be merged in current state. If someone wants to pick this up: the fix is to read the OS preference via `window.matchMedia('(prefers-color-scheme: dark)')` and wire `systemTheme` into the theme context.

PR #13272, opened 2026-05-17
fix(tables): require write permission to create records

KhaledR57:
## Summary

- `CreateRequest` in `record.controller.ts` was missing `Permission.WRITE_TABLE`
- ENGINE and SERVICE principals skip the rbac permission check, so flow steps and service API keys are unaffected.
- Adds `record-rbac.test.ts` covering Viewer / Editor / Owner (Create / Update / Delete / List / Get)

## Test plan

- [x] `npm run lint-dev`
- [x] New test passes
- [x] Manual smoke: Viewer `POST /v1/records` now returns 403 (was 201), Editor/Admin/flow-step still return 201

### DIFF
--- a/.agents/features/tables.md
+++ b/.agents/features/tables.md
@@ -82,6 +82,19 @@ A built-in relational database feature that lets users store structured data dir
 - `record.update()` — update cells (empty fields unchanged)
 - `record.delete()` / `record.deleteAll()` — bulk delete
 
+## Access Control
+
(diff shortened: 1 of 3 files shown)


AbdulTheActivePiecer:
We will need to include this in the docs (https://www.activepieces.com/docs/about/changelog) and gh release page, some people's flows will probably break if they are creating records already without having the correct permission.

abuaboud (approved):

PR #13296, opened 2026-05-19
fix: propagate flowId and flowRunId to agent tool calls (#12993)

### LINKED ISSUE #12993: [BUG]: Agent piece tool calls don't inherit the parent flow's `flowId` / `flowRunId`

KarelTestSpecial:
This PR ensures that agent tool calls (piece actions) inherit the actual parent flow context (flowId and flowRunId) instead of using placeholders. This fix includes the necessary TypeScript type updates in the shared engine-operation definitions to ensure a clean build. Closes #12993.

### DIFF
--- a/packages/server/engine/src/lib/handler/context/engine-constants.ts
+++ b/packages/server/engine/src/lib/handler/context/engine-constants.ts
@@ -140,13 +140,13 @@ export class EngineConstants {
         })
     }
 
-    public static fromExecuteActionInput(input: ExecuteToolOperation): EngineConstants {
+    public static fromExecuteActionInput(input: ExecuteToolOperation, overrides?: { flowId?: string, flowRunId?: string }): EngineConstants {
         return new EngineConstants({
-            flowId: DEFAULT_MCP_DATA.flowId,
+            flowId: overrides?.flowId ?? DEFAULT_MCP_DATA.flowId,
             flowVersionId: DEFAULT_MCP_DATA.flowVersionId,
             flowVersionState: DEFAULT_MCP_DATA.flowVersionState,
             triggerPieceName: DEFAULT_MCP_DATA.triggerPieceName,
-            flowRunId: DEFAULT_MCP_DATA.flowRunId,
+            flowRunId: overrides?.flowRunId ?? DEFAULT_MCP_DATA.flowRunId,
             publicApiUrl: input.publicApiUrl,
             internalApiUrl: addTrailingSlashIfMissing(input.internalApiUrl),
             retryConstants: DEFAULT_RETRY_CONSTANTS,
--- a/packages/server/engine/src/lib/tools/index.ts
+++ b/packages/server/engine/src/lib/tools/index.ts
@@ -173,7 +173,10 @@ async function execute(operation: ExecuteToolOperationWithModel): Promise<Execut
         const output = await flowExecutor.getExecutorForAction(step.type).handle({
             action: step,
             executionState: FlowExecutorContext.empty(),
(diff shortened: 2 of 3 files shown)


KarelTestSpecial:
Hi @abuaboud 👋

I've just rebased this PR onto the latest `main` (was 22 commits behind). All checks still pass — CLA signed, GitGuardian clean, Greptile 5/5 confidence.

Could a maintainer please **approve the E2E workflows** to run? They're currently blocked on maintainer approval (`pull_request_target`), which seems to be the last step before this can be properly reviewed for merge.

The fix itself is minimal and backward-compatible — it only threads the real `flowId`/`flowRunId` into agent tool calls instead of the `DEFAULT_MCP_DATA` placeholders. Closes #12993.

Thanks in advance! 🙏

KarelTestSpecial:
Hi @abuaboud 👋 Friendly follow-up on this one — it's been a few weeks since the rebase. The PR is mergeable with Greptile 5/5 confidence and CLA signed; it's only waiting on maintainer approval for the E2E workflows. Happy to rebase again or adjust anything if needed. Thanks for your time!

PR #13337, opened 2026-05-21
fix: automated translation workflow was broken

AbdulTheActivePiecer:
### Relevant User Scenarios

Fixes # (issue)


AbdulTheActivePiecer:
@greptileai 
The dropped translations were already orphaned on `main` — the source code stopped referencing those keys earlier, so they weren't being rendered. Example for `fr.json`:

- `main`: `"Are you sure?": "Êtes-vous sûr ?"` exists, but source calls `t('Overwrite Draft')` → key missing → falls back to en → user sees `"Overwrite Draft"`.
- This PR: `"Overwrite Draft": ""` → empty + `returnEmptyString: false` → same en fallback → user sees `"Overwrite Draft"`.

No UI change. The next `i18n:extract` on `main` would have pruned them anyway; this PR is just the first run after the output path was fixed.

The underlying issue — renamed source strings not being re-translated on Crowdin — is upstream and exists independently of this PR.

PR #13381, opened 2026-05-25
chore(docs): remove Windows Zone.Identifier screenshot metadata

patelatharva:
## Summary

Removes two accidentally tracked Windows **Zone.Identifier** alternate data stream (ADS) sidecar files under `docs/resources/screenshots/`. These files are not part of the documentation content; they are OS metadata created when screenshots are downloaded from the internet on Windows.

### What these files are

On Windows, when a file is marked as coming from the internet (e.g. via browser download), the system can attach a small **Zone.Identifier** stream to the file. In Git, that stream sometimes appears as a separate path with a colon in the name, for example:

- `enable-environments-2.png:Zone.Identifier`
- `cloud.activepieces.com_projects_...png:Zone.Identifier`

Each file in this repo is only ~25 bytes and contains Windows zone transfer metadata, for example:

```
[ZoneTransfer]
ZoneId=3
```

They are not images, docs, or build inputs.

### Problems on Windows + Git Bash

Developers on Windows (including **Git Bash**, which uses Git for Windows and NTFS path rules) can hit hard failures when these paths exist in the tree:

1. **Invalid path errors** — NTFS reserves `:` for alternate data streams. Git for Windows rejects checkout/restore/index operations on paths like `file.png:Zone.Identifier` with errors such as `invalid path` and `make_cache_entry failed`.
2. **Broken working tree recovery** — Commands like `git restore --source=HEAD :/` or `git read-tree HEAD` may partially succeed for normal files but leave the index in a bad state for these entries, so `git status` can show thousands of staged deletions or persistent staged deletes that cannot be cleared without removing the paths from the repository.
3. **No cross-platform value** — Linux and macOS clones do not need these markers; they add noise and risk for Windows contributors only.

### Why delete instead of ignore-only

`.gitignore` cannot retroactively fix paths already committed on `main`. Removing the files from the repository is the durable fix so every fresh clone and restore on Windows behaves predictably.

## Files removed

| Path | Notes |
|------|--------|
| `docs/resources/screenshots/enable-environments-2.png:Zone.Identifier` | ADS metadata for `enable-environments-2.png` |
| `docs/resources/screenshots/cloud.activepieces.com_projects_er7zUIumSRjbOuqubZiye_flows_limit=10(laptop) (1).png:Zone.Identifier` | ADS metadata for the laptop screenshot |

The actual `.png` screenshots are unchanged.

## Suggested follow-up (optional, not in this PR)

Consider a short note in contributor docs to avoid committing `*:Zone.Identifier` / `Zone.Identifier` files, and/or a pre-commit check for paths containing `:Zone.Identifier` on Windows-hosted repos.

## Test plan

- [x] `git status` is clean on branch after commit
- [ ] On Windows + Git Bash: fresh `git clone` and `git checkout main` complete without `invalid path` errors
- [ ] `git restore --source=HEAD :/` runs cleanly after pulling this change
- [ ] Documentation screenshots still present under `docs/resources/screenshots/*.png`

Made with [Cursor](https://cursor.com)


kishanprmr:
Fixed in https://github.com/activepieces/activepieces/pull/13545

PR #13385, opened 2026-05-25
fix(workers): sync webhook responses in split APP/WORKER deployments

### LINKED ISSUE #13232: [BUG]: Sync webhook always times out (204) in split-process deployments due to random SERVER_ID mismatch in engine-response-watcher
When Activepieces is deployed with separate APP and WORKER processes (e.g. via PM2 with AP_CONTAINER_TYPE=APP and AP_CONTAINER_TYPE=WORKER, or any multi-process setup), synchronous webhook calls always time out and return a 204 No Content response instead of the flow's actual response.

The root cause is in packages/server/api/src/app/workers/engine-response-watcher.ts:
```ts
const SERVER_ID = apId()
```
apId() generates a random UUID at module load time. In a split-process deployment, the APP process and the WORKER process each load this module independently and therefore get different SERVER_ID values.

The flow of the bug:

APP process receives a sync webhook. It subscribes to Redis pub/sub channel engine-run:sync:<APP_SERVER_ID> and enqueues a BullMQ job with synchronousHandlerId = APP_SERVER_ID.
WORKER process picks up the job and, after execution, publishes to engine-run:sync:<WORKER_SERVER_ID>.
Since APP_SERVER_ID ≠ WORKER_SERVER_ID, nobody is listening on the channel the WORKER publishes to. The APP process times out and returns 204.
The same mismatch occurs after any in-process restart (e.g. PM2 auto-restart on crash), because each restart re-executes apId() and produces a new random ID, invalidating the synchronousHandlerId stored in already-queued jobs.

To Reproduce

Deploy Activepieces with separate APP and WORKER processes (e.g. using the official docker-entrypoint.sh which starts both via PM2, or a Kubernetes setup with AP_CONTAINER_TYPE=APP on the API pod and AP_CONTAINER_TYPE=WORKER on the worker pod).
Create a flow with a Webhook trigger and a Return Response step returning { "hello": "world" }. Publish it.
Call the webhook URL synchronously: 
```bash
curl -X POST https://<host>/api/v1/webhooks/<flow-id>/sync
```

Observe the call hangs for AP_WEBHOOK_TIMEOUT_SECONDS (default 30s) then returns HTTP 204 No Content with an empty body, even though the flow run completes successfully in the run history.
Expected behavior

The sync webhook call should return the flow's configured response (e.g. HTTP 200 with { "hello": "world" }) immediately after the flow finishes, regardless of whether APP and WORKER run in the same process or in separate processes.

Additional context

The fix is to replace the random apId() with a stable identifier that persists across module reloads within the same process/pod. Using os.hostname() works well in containerised environments where the hostname is stable per container:
```ts
// Before
import { apId } from '@activepieces/shared'
const SERVER_ID = apId()

// After
import os from 'os'
const SERVER_ID = os.hostname()
```
In Kubernetes StatefulSets the hostname is stable across restarts (e.g. activepieces-0), so jobs enqueued with synchronousHandlerId = hostname will always resolve to the correct pub/sub channel even after a pod restart. The same applies to PM2 restarts within a single container.

Related closed issue: #7087 ("Workflow returns 204 result and empty body when processing takes too long") — a symptom of this same root cause in split-process deployments.

patelatharva:
Fixes #13232

## Summary

This PR fixes [#13232](https://github.com/activepieces/activepieces/issues/13232): sync webhooks (`/sync` URLs) returning **204 No Content** while the flow completes successfully when **APP** and **WORKER** run in separate processes (Docker Compose split layout, separate-workers guide, multiple replicas).

## Problem

Synchronous webhooks wait on Redis pub/sub on channel `engine-run:sync:{serverId}`. The APP subscribes using a local server id and passes it to jobs as `workerHandlerId`. When APP and WORKER are separate Node processes, each could derive a **different** random id, so the worker response is published on a channel the APP is not listening on. The listener times out and the HTTP client gets an empty **204** response even though execution succeeded.

This shows up in deployments that match how this repository documents production-style installs: split `AP_CONTAINER_TYPE=APP` / `WORKER`, shared Redis, and scaled worker replicas—not only exotic multi-region setups.

## Root cause

`engineResponseWatcher` depended on a per-process identity for pub/sub. In split containers that identity was not guaranteed to match across APP and WORKER.

## Solution

Introduce `getEngineResponseServerId()` with a feature-flagged experiment (`stable-engine-response-server-id`):

- **B side (default):** Shared stable id across processes—either `AP_ENGINE_RESPONSE_SERVER_ID` when set, or the first 21 characters of `SHA-256(AP_ENCRYPTION_KEY)` (validated as `ApId`), cached per process.
- **A side (rollback):** Legacy `apId()` per process when `AP_STABLE_ENGINE_RESPONSE_SERVER_ID_ENABLED=false`.

`engine-response-watcher` now uses this helper for subscribe/unsubscribe and `getServerId()`.

### Experiment shape (easy cleanup, minimal churn)

We kept **one implementation file** (`engine-response-server-id.ts`) with `BEGIN`/`END` comment blocks for A vs B branches instead of many renamed modules. Tests are split into:

- `engine-response-server-id.test.ts` — server id behavior (mocks `system` only)
- `engine-response-watcher-multi-process.test.ts` — APP vs WORKER pub/sub (mocks `engine-response-server-id`)

Vitest cannot combine a module mock with dynamic imports in a single test file, so two files are intentional—not extra surface for cleanup.

When the experiment graduates: delete the opposite side’s comment blocks, remove the flag, and drop the unused env props per the checklist in the source file header.

### Default `true` for the flag

**Decision:** `AP_STABLE_ENGINE_RESPONSE_SERVER_ID_ENABLED` defaults to **`true`** when unset.

**Rationale:** Activepieces CE is primarily **installable, self-hosted software**. Operators configure `.env`; there is no central hosted feature-flag plane for Community Edition. The repo’s reference `docker-compose.yml` already uses split APP + workers. Default-off would leave sync webhooks broken until operators discover an undocumented flag—poor fit for a **bugfix** PR.

The flag remains a **rollback switch** (`false` restores legacy per-process ids). Hosted or cautious rollouts can still set `false` explicitly.

`WORKER_AND_APP` (single-container default) remains safe: the same `AP_ENCRYPTION_KEY` yields the same stable id in both PM2 processes.

### Documentation and Compose

- [Separate workers guide](docs/install/guides/separate-workers.mdx): warning on sync webhooks, required alignment of flag and secrets on APP and every worker.
- [Environment variables](docs/install/configuration/environment-variables.mdx): documents the flag and optional `AP_ENGINE_RESPONSE_SERVER_ID`.
- Root and benchmark `docker-compose.yml`: set `AP_STABLE_ENGINE_RESPONSE_SERVER_ID_ENABLED=true` on `app` and `worker` for explicit reference stacks.

### Operational contract (discussed, not enforced in code)

For B side to work reliably:

1. Same flag value on every APP and WORKER container (default `true` unless rolling back).
2. Same `AP_ENCRYPTION_KEY` across those containers, or the same explicit `AP_ENGINE_RESPONSE_SERVER_ID`.
3. Shared Redis for pub/sub (already required today).

Misaligned secrets or partial rollout reproduces the bug. Optional explicit `AP_ENGINE_RESPONSE_SERVER_ID` decouples channel naming from encryption-key rotation.

### Distributed vs single-machine

Both are first-class: `WORKER_AND_APP` for simple installs; split APP/WORKER for scale and the separate-workers security model. This fix targets the **split** path where the bug manifests. Multiple APP replicas can share one stable channel name; in-memory listeners still route by `requestId` (no sticky sessions required for sync webhooks).

## Test plan

- [x] Unit: A side — different ids across simulated APP/WORKER module loads (flag off).
- [x] Unit: B side — same stable id from encryption key; respects `AP_ENGINE_RESPONSE_SERVER_ID`.
- [x] Unit: Watcher A side — channel mismatch → 204 timeout.
- [x] Unit: Watcher B side — aligned channel → 200 with response body.


patelatharva:
## Review feedback addressed

Thanks [@greptile-apps](https://github.com/apps/greptile-apps) for the review.

### Change (P2 — `ENGINE_RESPONSE_SERVER_ID` validation)

**Feedback:** Explicit `AP_ENGINE_RESPONSE_SERVER_ID` was returned without `ApId.parse()`, while the SHA-256-derived path always validated format (21 alphanumeric chars). A misconfigured override could reach Redis pub/sub with an invalid channel suffix.

**Fix (commit `50a346e`):**
- Call `ApId.parse(configuredId)` before returning an explicit override in `resolveStableEngineResponseServerId()`.
- Added unit test: `should reject AP_ENGINE_RESPONSE_SERVER_ID when it is not a valid ApId` (e.g. `invalid-server-id` throws when resolving the server id).

### Verification

Re-ran worker unit tests locally — **7/7 passed**:
- `engine-response-server-id.test.ts` (4 tests, including new validation case)
- `engine-response-watcher-multi-process.test.ts` (3 tests)

No other review threads required code changes. CLA status is unchanged on our side.

Let us know if you’d like `ENGINE_RESPONSE_SERVER_ID` validated earlier in `system-validator.ts` as well; runtime `ApId.parse` at first `getEngineResponseServerId()` use matches the derived-path behavior.

patelatharva:
## Babysit status (latest check)

**PR:** https://github.com/activepieces/activepieces/pull/13385  
**Branch:** `ISS-13232` (includes merge from `main` @ `3ee69de`)

### Automated checks — all green
| Check | Status |
|-------|--------|
| Validate PR title | pass |
| check-label | pass |
| GitGuardian | pass |
| Greptile Review | pass (5/5 after ApId fix) |
| mergefreeze | pass |
| E2E (ce/ee) | skipped (no `ready-for-e2e` label — expected for most PRs) |

### Review comments — addressed
- **Greptile P2** (`ENGINE_RESPONSE_SERVER_ID` without `ApId.parse`): fixed in `50a346e`; review thread **resolved**.
- No other unresolved review threads.

### Local verification (post-merge with `main`)
Worker unit tests: **7/7 passed**
- `engine-response-server-id.test.ts` (4)
- `engine-response-watcher-multi-process.test.ts` (3)

### Merge blockers (outside code)
1. **`REVIEW_REQUIRED`** — needs maintainer approval.
2. **CLA** — [CLA assistant](https://cla-assistant.io/activepieces/activepieces?pullRequest=13385) still shows not signed; contributor must complete/sign.
3. **Labels** — could not add `🐛 bug` from fork (permission denied); maintainer may add for changelog routing.

### Code / CI actions taken this pass
- Synced local branch with remote (already merged `main` on GitHub).
- Re-ran scoped unit tests — no regressions.
- No new commits required.

Ready from an implementation and CI perspective; waiting on **CLA + human review** to merge.

MrChaker:
Hi @patelatharva,
Thank you for this . But I don't see the problem with serverId ..
 
engineResponseWatcher.getServerId will always return same value for APP container

when webhook sync is called it creates a job with webserverId ( which is App container serverId ) and starts listening in `engine-run:sync:${webserverId}` channel

on worker side it publishes to the same webserverId that is gets from the job data ( I guess here you thought it's publishing to serverId from WORKER container ? )

ArtinVanPars:
## Independent reproduction and successful workaround

Reproduced on Ubuntu 26.04 using the official single-container
`WORKER_AND_APP` setup.

Image:
`activepieces/activepieces@sha256:65ea85f5e7ea236416b342eef0d2dbade8da51247aaea495d191eb027095805a`

Flow: Catch Webhook → Code → Return Response, called through `/sync`.

Before patch:
- Flow and Return Response succeeded.
- HTTP caller received an empty body.

Runtime patch:

```js
// Before
const SERVER_ID = (0, shared_1.apId)();

// After
const SERVER_ID = require("node:os").hostname();
```

After restarting the container, the same published flow immediately returned
the complete JSON response. No flow changes were made between the failing and
successful calls.

This confirms the stable server ID workaround for the official
WORKER_AND_APP single-container/PM2 deployment on the image digest above.

abuaboud:

PR #13402, opened 2026-05-26
feat(mcp): tag MCP-created flows with an AI badge

utopianguide:
## Summary

Adds a small "AI" badge to flows created through MCP, so you can tell at a glance which automations came from an AI client (Claude/Cursor) versus built by hand.

When someone connects their MCP client and asks it to build, create, or duplicate a flow:
- `flow.createdBy` is stamped with `{ type: 'MCP', id: <mcp server id> }`
- `flow.ownerId` is set to the user whose OAuth token authorized the MCP session
- The badge shows in the flows list (Status column) and the builder header, with tooltip "Created by AI MCP"

## `ownerId` vs `createdBy` (two different questions)

- **`ownerId`** = who *owns* the flow now. Mutable (can be transferred), drives the Owner column + "my flows" filtering. Already existed.
- **`createdBy`** = what *created* it, immutably. An automated source, not necessarily a user.

They're intentionally separate. A flow can be MCP-created (`createdBy.type = MCP`) and owned by the user who connected the client (`ownerId`) at the same time.

## Why jsonb + discriminated union

```ts
createdBy: Nullable(z.discriminatedUnion('type', [
  z.object({ type: z.literal('MCP'),   id: ApId }),
  z.object({ type: z.literal('AGENT'), id: ApId }),
]))
```

A single jsonb column covers current and future automated creators (agents next) with no further migrations. `createdBy` is **server-set only** — `ap_create_flow`, `ap_build_flow`, and `ap_duplicate_flow` pass it; the public flow endpoints don't accept it, so callers can't spoof attribution.

## Non-breaking, no backfill

- `createdBy` is nullable + optional in the zod schema — existing API clients see one extra optional field on `Flow` and ignore it.
- No data migration. Flows created before this PR keep `createdBy = NULL` and render with no badge.
- Migration is a single non-destructive `ADD COLUMN IF NOT EXISTS "createdBy" jsonb` — no index, no table lock on deploy. (An index on the creator type can be added later, `CONCURRENTLY`, if/when we expose filtering by it.)

## Test plan

- [ ] Build a flow in the web UI → no badge
- [ ] Via an MCP client, run `ap_create_flow` / `ap_build_flow` / `ap_duplicate_flow`
- [ ] Each resulting flow shows the AI badge in the list + builder header, owner = the connected user
- [ ] Hover the badge → "Created by AI MCP"
- [ ] Existing pre-PR flows show no badge


hazemadelkhalel:
A few things I noticed while reviewing:

**1. Missing test coverage**
We don't have any automated tests for the new attribution logic. Can you add at least:
- Create flow via MCP tool → `createdBy` should be `{ type: 'MCP', id: mcpServerId }` and `ownerId` should match the authenticated user
- Create flow via REST API → `createdBy` should be null
- Verify `createdBy` can't be set through the `CreateFlowRequest` body

**2. Tooltip wording**
"Created by AI MCP" — most users won't know what MCP means. I'd simplify this to just "Created by AI".

**3. Type casting**
`(item.data as PopulatedFlow).createdBy` in `automations-table-row.tsx` — can you run `/simplify` on this file and clean up the casting?

**4. Screenshot**
Can you share a screenshot of how the badge looks in the flows list and builder header?

packages/web/src/features/flows/components/flow-created-by-badge.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <McpKnotIcon />
+          {t('AI')}
+        </span>
+      </TooltipTrigger>
+      <TooltipContent side="bottom">{t('Created by AI')}</TooltipContent>
+    </Tooltip>
+  );
+};
+
+const McpKnotIcon = () => (
```
hazemadelkhalel: Remove this component, use existing components

packages/web/src/assets/img/custom/mcp.tsx

packages/shared/src/lib/automation/mcp/mcp.ts
```diff
@@ -29,7 +29,7 @@ export type PopulatedMcpServer = z.infer<typeof PopulatedMcpServer>
 
 export type McpServer = z.infer<typeof McpServer>
 
-export type ProjectScopedMcpServer = McpServer & { projectId: string }
+export type ProjectScopedMcpServer = McpServer & { projectId: string, userId?: string | null }
```
hazemadelkhalel: `userId` on `ProjectScopedMcpServer` mixes session context with the data model type

`userId` isn't a property of the MCP server; it's who's currently connected. Consider a separate context type like 
```
McpToolContext = {
  mcp: ProjectScopedMcpServer, 
  userId: string | null 
} 
```
So the data model stays clean.

hazemadelkhalel:
`userId` is typed as `string | null` but it always comes from `payload.sub` in the JWT — if the token is invalid the request gets rejected before we ever reach this point. So null never actually happens here.

Can you change it to `userId?: string` across the chain (`resolveMcpAndUser`, `buildMcpServer`, `ProjectScopedMcpServer`, `RegisterToolsParams`)? That way you can drop the `?? undefined` coercion in all 3 tools too.

utopianguide:
all three done in the latest commit 👇

- **icon** — using the shared `McpSvg` now, dropped the inlined svg
- **userId on ProjectScopedMcpServer** — agreed, pulled it off the data-model type into a `McpToolContext = { mcp, userId? }`. scoped it to just the 3 flow tools that actually read userId (create/build/duplicate) — the other 39 still take `ProjectScopedMcpServer` untouched. didn't want to churn ~100 tool call-sites for tools that don't use it, but if you'd rather have everything on the context for consistency lmk
- **typing** — `userId?: string` across the chain (resolveMcpAndUser → buildServer → buildMcpServer → RegisterToolsParams), dropped the `?? undefined` in the tools

tests cover create/build/duplicate attribution + REST-stays-null + can't-spoof-via-body, full mcp-tools suite green. also rebased on your main merge.

hazemadelkhalel (approved):

PR #13420, opened 2026-05-27
fix: bulk export for flows inside folders

majewskibartosz:
## Problem

On the Automations page, multi-select → **Export** silently did nothing when any selected flow lived inside a folder — no download, no toast, no network request. The same flow exported fine from root, and the per-row ⋮ → Export worked for folder flows. Only the bulk path was broken.

Tracked in GIT-1512 / #13418. Reproduced on 0.83.0.

## Root cause

`handleBulkExport` filtered the selected flow IDs against `deps.flows`, which is wired to `rootFlows` — a query scoped to `folderId: UncategorizedFolderId`. Flows inside folders are never in that list, so the filter returned `[]` and the `if (flowsToExport.length > 0)` guard skipped the export. Compounding it, `useExportFlows` had an `onSuccess` but no `onError`, so any failure was swallowed.

## Fix

- Resolve the selected flows directly by ID via `flowsApi.get(id)` instead of filtering the root-scoped list (mirrors the tables branch right below it). The backend `GET /v1/flows/:id/template` already works for any flow regardless of folder.
- Remove the now-dead `flows` field from `MutationDeps` and its `flows: rootFlows` wiring.
- Add an `onError` toast to `useExportFlows` so future export failures surface instead of being silent.

## Testing

Verified manually on a local dev env across: folder flow (now downloads), mixed root + folder selection (zip), root flow (no regression), and per-row export (unchanged). Frontend-only change — no backend changes; the export endpoint already has API-test coverage and is unchanged. `npm run lint-dev` clean.


packages/web/src/features/flows/hooks/flow-hooks.tsx
```diff
@@ -174,6 +174,9 @@ export const flowHooks = {
           );
         }
       },
+      onError: () => {
+        toast.error(t('Failed to export flows'));
+      },
```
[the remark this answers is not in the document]
majewskibartosz: Confirmed and addressed in 39cf31efeb. `flows.md` described export only generically, so there was no stale prose — but the *why* is non-obvious, so I added a sentence to the `flowHooks` paragraph documenting that bulk export resolves each selected flow by id (`flowsApi.get`) instead of filtering the root-scoped `rootFlows` list, since flows inside folders are absent from that list and filtering silently drops them. The added `onError` toast is noted there too.

packages/web/src/features/automations/hooks/use-automations-mutations.ts
```diff
@@ -250,10 +249,9 @@ export function useAutomationsMutations(deps: MutationDeps) {
       const { flowIds, tableIds } = getSelectedIdsByType(selectedItems);
 
       if (flowIds.length > 0) {
-        const flowsToExport = deps.flows.filter((f) => flowIds.includes(f.id));
-        if (flowsToExport.length > 0) {
-          exportFlows(flowsToExport);
-        }
+        Promise.all(flowIds.map((id) => flowsApi.get(id)))
```
MrChaker: @majewskibartosz This is a N query . you're sending multiple requests for each flow id . it's not optimal for our server & db . try to use flows that are already fetched in the frontend

majewskibartosz: Done — removed the per-flow `flowsApi.get` entirely (commit c0073d9c7b).

Bulk export now resolves the selected flows straight from the already-loaded `treeItems` instead of refetching each by id, so it issues **zero** extra requests — exactly the in-memory data you pointed at.

To keep that safe, the selection is now cleared on any view change that can remove loaded items (paging, filtering/search, collapsing a folder), so the selection is always a subset of what's loaded and the in-memory lookup can never drop a selected flow.

Louai-Zokerburg (approved):

PR #13425, opened 2026-05-28
fix(google-calendar): new event webhook trigger fires on real events

### LINKED ISSUE #13424: [BUG] Google Calendar v0.9.1 new_event trigger never fires
## Problem

The `new_event` webhook trigger silently no-ops on every real calendar event. Test Step works (calls `getLatestEvent` directly), published flows produce zero runs.

## Root cause

Three stacked defects in `packages/pieces/community/google-calendar/`:

1. `triggers/new-event.ts:104` filters on `x-goog-resource-state === 'add'`. Calendar push only emits `sync` / `exists` / `not_exists` — `'add'` is a Drive value. Condition never true → `run()` always returns `[]`.
2. `new-event.ts:105` reads `payload.body` as the event. Calendar push bodies are empty; the delta must be fetched via `events.list` with a stored `syncToken`.
3. `common/helper.ts:38-60` `watchEvent()` sets no `expiration`. Channels die at Google's default lifetime (\~1 week).

## Fix

Rewrite the webhook path:

* `onEnable`: after `watchEvent`, call `events.list` once with `singleEvents: true`, `showDeleted: true` to capture initial `nextSyncToken`. Store under `google_calendar_sync_token`.
* `run`:
  * `sync` → return `[]`
  * `exists` → paginate `events.list` with stored `syncToken` via new `listEventsWithSyncToken` helper. Persist new token. On HTTP 410, re-init token and return `[]`.
  * Treat events as new when `created === updated`; drop `status === 'cancelled'`; apply existing `event_types` / `search_filter` / `exclude_all_day` filters.
  * Other states → return `[]`.
* `onDisable`: clear the sync-token store entry alongside `stopWatchEvent`.
* `watchEvent`: add `expiration` (\~6 days, epoch-ms) per `google-sheets/.../helpers.ts:81` pattern.
* Bump `package.json` to `0.9.2`.

## Out of scope

* Automated channel renewal before expiration → follow-up ticket.
* Adding test infra for community piece triggers.

## Preserved

Trigger name `'new_event'` unchanged. Store schema additive. No migration shim. No new i18n keys. `test()` shape unchanged.

## Verification

* `npx turbo run lint --filter=@activepieces/piece-google-calendar` clean
* `npm run lint-dev` and `npm run test-unit` clean
* Manual on local cloud-mode + real Google account: new event → run fires; edit → no run; disable/re-enable → no run on sync; filters still apply.

majewskibartosz:
## Summary

The Google Calendar `New Event` trigger (`@activepieces/piece-google-calendar`) silently no-oped on every real calendar push notification — the Test step worked but published flows never produced a run when an event was created. The bug has been present since the trigger first shipped in v0.67.3.

Root cause: the webhook path was written against the **Google Drive Changes API** contract instead of the **Google Calendar push** contract.

### What was broken

- The filter checked `x-goog-resource-state === 'add'`. Calendar push notifications only emit `sync` / `exists` / `not_exists`; `add` is a Drive value. The branch was never taken.
- The trigger read the new event from `payload.body`. Calendar push notifications carry **no body** — the delta must be fetched via `events.list` with a stored `syncToken`.
- `watchEvent()` registered the channel without an `expiration`, so it would die at Google's default lifetime.

### What this PR changes

- `new-event.ts` is rewritten to mirror the Calendar push contract:
  - `onEnable` registers the channel **and** captures an initial `syncToken` via a new paginated bootstrap list, persisting both to `context.store`.
  - `run` handles `sync` (initial confirmation → no run), `exists` (list deltas with the stored syncToken, persist the new token, keep fresh events, drop `status === 'cancelled'`, then apply existing `event_types` / `search_filter` / `exclude_all_day` filters via a shared `matchesFilters` helper).
  - On HTTP 410 (sync token invalidated) the trigger transparently re-bootstraps the syncToken and skips that notification rather than firing duplicates.
  - `onDisable` clears the syncToken store entry alongside the existing channel stop.
- `helper.ts` gains `getInitialSyncToken` and `listEventsWithSyncToken`, and `watchEvent` now requests a 6-day `expiration`.
- New vs updated event detection uses a **2-second tolerance** on `updated - created`. Strict equality (the obvious heuristic) drops every real notification because Google sets `updated` ~500–900 ms after `created` on freshly-created events; this surfaced only during live testing.
- Trigger name (`new_event`), prop schema, sample data, and `test()` output shape are all unchanged — existing flows keep working with no reconfiguration. The store schema is additive, so re-publishing the flow is enough to pick up the fix.
- Piece patch version bumped `0.9.1 → 0.9.2`.

Closes #13424.

## Test plan

End-to-end verified locally against a real Google Calendar account, with a Cloudflare Tunnel exposing the local dev backend as a public webhook origin:

- [x] `npx turbo run lint --filter=@activepieces/piece-google-calendar` — clean (0 errors).
- [x] `npx turbo run build --filter=@activepieces/piece-google-calendar` — clean.
- [x] Publish a flow with the `New Event` trigger on a writable calendar.
- [x] Initial channel sync notification (`x-goog-resource-state: sync`) → no run produced.
- [x] Create a new event in Google Calendar (UI) → flow run produced (verified 3 distinct events: `delta_ms` 579 / 891 / 757, all under the 2000 ms tolerance).
- [x] Subsequent notifications for edits / duplicate sync chatter (4 webhooks) → no runs produced.
- [x] Slack action downstream of the trigger received the event payload.


majewskibartosz:
Addressed all three Greptile P2 findings in the latest force-push (9b807e8):

1. **`updated - created` heuristic** — agreed the 2 s window is fragile against API-created events, conference-link events, and slow Google processing. Replaced the time-delta heuristic with a persistent **seen-event-ID set** stored in `context.store` (capped at 500 IDs). True new events fire exactly once when their ID first appears in a delta list; edits of already-emitted events are recognised by ID and dropped. No timing window required.
2. **`as unknown as string` casts** — removed alongside the heuristic.
3. **6-day channel expiration with no renewal** — added `renewConfiguration: { strategy: WebhookRenewStrategy.CRON, cronExpression: '0 */12 * * *' }` plus an `onRenew` that stops the existing channel and re-registers a fresh one (mirrors the google-sheets webhook trigger pattern). The 6-day cap on the channel itself stays as a safety net, but the 12-hour renewal cron means the trigger keeps running indefinitely without manual flow re-enable.

`onDisable` now also clears the seen-IDs entry.

onyedikachi-david (approved):

PR #13429, opened 2026-05-28
feat(pieces-framework): add audience + aiMetadata optional fields to action & trigger contract

ibrahim-abuznaid:
## Summary

First framework slice of an AI-readiness contract. Adds two new optional fields on `ActionBase` and `TriggerBase` in `@activepieces/pieces-framework`, wired through `createAction` and `createTrigger`. Zero behavior change: every field is optional, no existing piece compiles or runs differently. `TriggerBase.sampleData` is untouched. (`llmDescription` was reverted on `main` by #13368; this branch is synced with `main` and no longer carries it.)

### New fields

| Field | Where | Type | Purpose |
|---|---|---|---|
| `audience?` | `ActionBase` only | `'human' \| 'ai' \| 'both'` | Routing/visibility. Top-level, not AI-targeted. Consumers treat missing as `'both'`. Action-only — deliberately omitted from the `TriggerBase` type. |
| `aiMetadata?` | `ActionBase` and `TriggerBase` (same type) | `{ description?, idempotent? }` | Bundled AI-targeted metadata. `description` is agent-targeted prose; `idempotent` declares retry safety (typically left unset on triggers, since polling the same cursor is naturally idempotent). The TypeScript type is `z.infer`-derived from the Zod schema, so the two can't drift. |

Output-shape modeling (a JSON Schema for an action's result) is deliberately **not** an `aiMetadata` sub-field. It's an audience-agnostic contract — one projection consumed by both the human data-mapper and the agent surface — owned by a separate feature, kept out of here to avoid two sources of truth.

Bumps `@activepieces/pieces-framework` to **0.29.1** (synced with `main`, then a patch bump for these additive fields).

### Naming

`aiMetadata` (not `infoForLLM`) to stay consistent with the `audience: 'ai'` value, and to read as a bundle of AI-targeted metadata rather than prose for one specific model class.

### Why this PR is small on purpose

Follow-up PRs need these fields to do anything useful, but the contract itself is the thing reviewers most want to look at, so it lands first on its own.

## Not in this PR (follow-ups)

- MCP discovery filter (`audience !== 'human' && aiMetadata?.description` set, in `ap_research_pieces` / `ap_get_piece_props`).
- New `ap_proxy_request` MCP tool to collapse the ~440 `custom_api_call` entries.
- One-line `audience: 'human'` flip inside `createCustomApiCallAction` (covers every piece using the helper); codemod for the standalone `custom-api-call.ts` files and the ~85 LLM-wrapper actions.
- Per-piece `aiReady: true` opt-in flag plus a scoped lint plugin (gated on `aiReady`, never an error on untouched pieces).
- The shared, audience-agnostic output-shape contract (separate feature).
- Top-5 piece exemplar curation (Gmail, Google Sheets, Notion, Google Drive, Slack).

## Test plan

- `tsc --noEmit` (lib + spec configs) — 0 errors.
- `eslint` on the changed files — 0 errors (pre-existing `any` warnings on untouched generic-alias lines only).
- No contract unit tests: per review, the type definitions are the source of truth, so the field plumbing is covered by `tsc` rather than by tests that re-assert the contract.
- No source outside the framework references the old `InfoForLLM` / `infoForLLM` symbol (verified by grep), so the rename is self-contained.


amrdb (changes requested):
Hey I have some comments:
1. Remove the comments, added fields are self explanatory.
2. Remove test files, u are testing a contract which is not really useful since the definition is the source of truth.
3. audience is meant to be for actions only not triggers right? TriggerBase inherits ActionBase so it inherits ActionBase's audience, you can omit it like
```typescript
export type TriggerBase = Omit<ActionBase, 'audience'> & {
    type: TriggerStrategy;
    sampleData: unknown,
    handshakeConfiguration?: WebhookHandshakeConfiguration;
    renewConfiguration?: WebhookRenewConfiguration;
    testStrategy: TriggerTestStrategy;
  };
```
4. `AiMetadata` type could just be `export type AiMetadata =
  z.infer<typeof AiMetadata>` like what you did with `Audience` type.

ibrahim-abuznaid:
Done in f0c99fb — audience off the TriggerBase type, AiMetadata via z.infer, and dropped the comments + contract tests. Also removed outputSchema from aiMetadata in 66c188c: we're moving the output shape to a separate shared contract (used by both human and AI) so we don't keep two of them.

amrdb (approved):

PR #13438, opened 2026-05-29
feat(oidc): add OIDC provider support for keyless AWS authentication

ralugli:
## What does this PR do?

Adds OpenID Connect (OIDC) provider support to Activepieces, allowing pieces to authenticate with AWS services using IAM Roles — eliminating the need for long-lived access keys.

Activepieces acts as an OIDC identity provider: it generates a signed JWT (RS256) that AWS STS accepts via `AssumeRoleWithWebIdentity`, issuing short-lived temporary credentials scoped to the specific platform and project.

**Infrastructure added:**

- **`PieceAuth.OIDC`** — new auth type for piece authors, mirrors the `CustomAuth` pattern with typed props
- **OIDC key manager** — generates or loads an RSA-2048 private key (from `OIDC_RSA_PRIVATE_KEY` env var in production, or auto-generated and persisted in `settings.json` for CE/dev). Protected by a mutex to avoid race conditions on first use
- **Discovery endpoints** (public, no auth required, CORS `*`):
  - `GET /.well-known/openid-configuration` — standard OIDC discovery document
  - `GET /.well-known/jwks.json` — RSA public key in JWK format so AWS can verify signatures
- **Token endpoint** `POST /v1/worker/oidc-token` (engine-scoped) — issues a signed JWT with `sub: platform:<id>:project:<id>` and `aud: sts.amazonaws.com`. Cache-Control headers set to `no-store` per RFC 6749
- **Frontend** — new `OIDCConnectionSettings` component; connection dialog, form schema, and utilities updated to handle the OIDC connection type end-to-end

**How a piece uses it:**

```ts
export const myAuth = PieceAuth.OIDC({
  props: {
    roleArn: Property.ShortText({ displayName: 'Role ARN', required: true }),
  },
});
```

The piece then calls `POST /v1/worker/oidc-token` via `ctx.server` to get a signed JWT and passes it to `sts:AssumeRoleWithWebIdentity`.

**Deployment:** `OIDC_RSA_PRIVATE_KEY` is optional. If absent, the server auto-generates a key on first use (CE-friendly). For production/multi-instance deployments, set this env var to a stable base64-encoded RSA PEM so all instances share the same signing key.

### Relevant User Scenarios

- Companies running Activepieces on-premises that want to connect to AWS without distributing long-lived IAM access keys
- Multi-tenant cloud deployments where each project should assume a scoped IAM role via `sts:AssumeRoleWithWebIdentity`
- Security-conscious teams adopting keyless authentication patterns (IAM Roles Anywhere / OIDC federation)

Fixes # (issue)


ralugli (commented):
Fixed P1 (static kid) and P2 (mutex protection) from the Greptile review:

**P1 – Static kid**: Implemented RFC 7638 SHA-256 thumbprint derivation. The kid is now computed as the base64url-encoded SHA-256 hash of the canonical JSON representation of the RSA public key members (e, kty, n). This ensures the kid changes automatically when the key material rotates, preventing AWS STS JWKS cache issues.

**P2 – getPublicKeyJwk mutex**: Added a separate publicKeyMutex with double-check pattern, consistent with getPrivateKeyPem. Separate mutexes avoid deadlocks since getPublicKeyJwk calls getPrivateKeyPem internally.

**P2 – void reply.header()**: Kept the void operator — Fastify's FastifyReply is typed as thenable, so removing void causes a `no-floating-promises` lint error in this codebase.

packages/server/api/src/app/core/security/oidc/oidc-token.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+    }, async (request, reply) => {
+        const { projectId, platform } = request.principal
+        const [privateKey, kid] = await Promise.all([
+            oidcKeyManager.getPrivateKeyPem(),
+            oidcKeyManager.getKid(),
+        ])
+        const token = await jwtUtils.sign({
+            payload: {
+                sub: `platform:${platform.id}:project:${projectId}`,
+                aud: 'sts.amazonaws.com',
```
MrChaker: @ralugli since this is generic . I should not hard code the aud here, also the TTL as mentioned by greptile

packages/server/api/src/app/app.ts
```diff
@@ -203,6 +204,7 @@ export const setupApp = async (app: FastifyInstance): Promise<FastifyInstance> =
     await app.register(alertsModule)
     await app.register(invitationModule)
     await app.register(workerModule)
+    await app.register(oidcTokenController, { prefix: '/v1/worker' })
```
MrChaker: @ralugli Let's follow the pattern we use . register a module here . and register the controller in the module

MrChaker:
@ralugli If you have followed a documentation for this implementation please share it

ralugli:
> @ralugli If you have followed a documentation for this implementation please share it

I implemented this based on the following documentation:
https://openid.net/specs/openid-connect-discovery-1_0.html
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html#manage-oidc-provider-prerequisites

packages/server/api/src/app/core/security/oidc/oidc-token.controller.ts
ralugli: Good point. I also updated the implementation so that OIDC is no longer limited to the Amazon audience. The audience and TTL are now passed as part of the request, making the solution provider-agnostic.

packages/server/api/src/app/app.ts
ralugli: Done ✅

ralugli:
The PR has been closed, and the changes have been moved to another PR: https://github.com/activepieces/activepieces/pull/13439

PR #13439, opened 2026-05-29
feat(amazon-s3): add IAM Role / OIDC authentication support

ralugli:
## What does this PR do?

Adds a second authentication method to the Amazon S3 piece: **IAM Role via OIDC** (`AssumeRoleWithWebIdentity`). Users can now connect S3 without long-lived access keys by trusting Activepieces as an OIDC identity provider.

The S3 piece now exposes two auth options side by side:

1. **Amazon S3 (Access Key)** — existing method, unchanged
2. **Amazon S3 (IAM Role / OIDC)** — new method

For the OIDC method, the piece:
1. Calls `POST /v1/worker/oidc-token` to get a signed JWT from Activepieces' OIDC provider
2. Passes it to AWS STS `AssumeRoleWithWebIdentity` to obtain short-lived credentials
3. Uses those credentials to build the S3/SecretsManager client

**Connection fields for the OIDC auth:**
- Role ARN (required) — the IAM role to assume
- External ID (optional) — extra condition in the trust policy
- Region + Bucket Name — same as access key auth

**AWS setup the user needs to do once:**
1. IAM → Identity Providers → Create OIDC provider with URL `https://<activepieces-host>` and audience `sts.amazonaws.com`
2. Create an IAM role with a trust policy scoped to `platform:<id>:project:<id>` in the sub claim
3. Attach S3 permissions to the role

The connection description includes a ready-to-paste trust policy template with `{{frontendUrl}}`, `{{frontendHost}}`, `{{platformId}}`, and `{{projectId}}` variables resolved at render time.

All existing S3 actions (upload, read, list, delete, move, signed URL, decrypt PGP) and the new-file trigger work with both auth methods — they use `resolveS3Client()` which dispatches based on the auth type.

### Relevant User Scenarios

- Companies that prohibit static IAM credentials and enforce keyless authentication via OIDC federation
- Teams that want per-project IAM role isolation (the JWT sub claim includes both platform and project ID, enabling fine-grained trust conditions)
- AWS environments with IAM Roles Anywhere or centralized identity governance

Fixes # (issue)


MrChaker:
@ralugli Let's keep all the work in this PR . close the other one and update this please . don't forget to include changes you did recently in the other PR here

MrChaker:
@ralugli The PR sshows 3000+ file changes . Something wrong has been done . If I were you I would close this PR create a new branch then ask claude to cherry-pick the commits you've did and open new PR again

packages/pieces/community/amazon-s3/src/lib/triggers/new-file.ts
```diff
@@ -1,125 +1,103 @@
...
-			data: file,
-		}));
-	},
-};
+import { S3 } from '@aws-sdk/client-s3';
+import dayjs from 'dayjs';
+import { amazonS3CombinedAuth, AccessKeyAuthProps, OidcAuthProps } from '../auth';
+import { resolveS3Client } from '../common';
 
 export const newFile = createTrigger({
```
MrChaker, line 27: hey @ralugli ,
I see here that you re-wrote the polling logic . I'm assuming you did it because it just needs 'ctx.server' . if that's the case . than just pass it as optional prop 'server: ServerContext | undefined' in the existing pollingHelper instead of rewriting the logic

ralugli: hey @MrChaker, done. went with your suggestion: added an optional `server` prop to the existing `pollingHelper` and pass it through to the `items` callbacks, instead of rewriting the polling logic. the new-file trigger now just reuses the shared helper and passes `context.server` on poll/test. it's an additive/optional change so existing pieces keep working unchanged.

packages/pieces/community/amazon-s3/src/lib/triggers/new-file.ts
```diff
@@ -1,125 +1,132 @@
...
+      propsValue: context.propsValue,
+    });
+  },
+  onDisable: async (context) => {
+    await pollingHelper.onDisable(polling, {
+      auth: context.auth,
+      store: context.store,
+      propsValue: context.propsValue,
+    });
+  },
```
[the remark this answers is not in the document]
AhmadTash: I think this is a false positive, the trigger enables fine for OIDC.

The finding assumes `pollingHelper.onEnable` invokes `polling.items()`, but that only happens for the `LAST_ITEM` strategy. This trigger uses `DedupeStrategy.TIMEBASED` (`new-file.ts:10`), and the `TIMEBASED` branch of `onEnable` never calls `items()` — it only seeds the cursor:

```ts
// packages/pieces/common/src/lib/polling/index.ts
case DedupeStrategy.TIMEBASED: {
  await store.put('lastPoll', Date.now());   // no polling.items() call
  break;
}
```

(The `items()` call lives in the `LAST_ITEM` branch just below it.) So on enable, the `if (!server) throw` is never reached and no credentials are resolved.

Credential resolution only happens in `run`/`test`, and both of those already forward `server: context.server` — which is where `items()` actually runs.

Also note the suggested patch would break the build: `pollingHelper.onDisable`'s parameter type is `{ store, auth, propsValue }` with no `server` field, so adding `server: context.server` to that call fails TS excess-property checking (TS2353).

**Net:** no runtime failure on enable for OIDC. Happy to add `server` to the `onEnable` call purely as future-proofing if the strategy ever changes to `LAST_ITEM`, but it's not required for correctness today.

AhmadTash:
@greptile review

AhmadTash (approved):

PR #13447, opened 2026-05-30
[codex] Add dynamic property refresh caching

Sameerzahiddd:
## Summary

- add a shared in-memory cache for piece option responses
- keep dynamic dropdown refresh buttons as force refreshes while normal renders can use cached options
- add a boxed dynamic fields group with its own refresh button that re-fetches schemas without clearing existing mapped values
- keep existing dynamic fields visible while refresh is pending to avoid a loading flash

## Why

Revisiting configured steps was repeatedly firing `EXECUTE_PROPERTY` jobs for dropdown options and generated dynamic fields. This keeps the cached revisit path quiet while preserving explicit refresh controls for users who need to pick up upstream changes like new Google Sheets columns.

## Validation

- `bun run typecheck` in `packages/web`
- focused eslint on the changed piece-properties files; no errors, only the existing hook dependency warnings in `dynamic-dropdown-piece-property.tsx`
- local dev log sanity check: tested `EXECUTE_PROPERTY` jobs completed with `status:"OK"` after refreshing fields


Sameerzahiddd:
Closing: opened against upstream by mistake. This branch belongs on the fork PR only.

PR #13490, opened 2026-06-02
feat(pieces): show clear provider error messages for AI/SDK-based pieces

AhmadTash:
## Description

Improves piece error messages for integrations built on SDK clients (e.g. Anthropic, OpenAI). The friendly error formatter now recognises SDK-style errors that expose `status`, `error`/`body`, and `headers` at the **top level** of the thrown error, instead of only errors that nest them under a `response` property. The result is a clean, readable message (like "`temperature` is deprecated for this model.") instead of a raw object or stack trace.

### How it works

`formatPieceError` previously only extracted HTTP details from errors shaped like `{ response: { status, body, headers } }`. The extractor is now split in two and tried in order:

- `extractResponseHttpDetails` — the existing nested-`response` shape, unchanged. A nested `response` object already establishes HTTP context, so any numeric status is trusted here, including the non-standard codes some services emit (e.g. LinkedIn's 999).
- `extractClientHttpDetails` — the new SDK-client shape `{ status, error, headers }`. This is a deliberately broad match on any error carrying a numeric HTTP-range `status`, so false positives are contained two ways: the `100–599` range guard in `toHttpStatus`, and running only as a fallback after the stricter response-shape extractor.

No downstream consumer changed. `HttpDetails` already carried `status` / `apiMessage` / `responseBody` / `responseHeaders`; this PR only populates them for more error shapes.

### Relevant user scenarios

- An AI piece using the Anthropic or OpenAI SDK throws a 400/429 — the user sees the provider's actual message (rate limit reached, deprecated parameter) instead of an opaque error.
- Any piece whose client library throws errors with a top-level `status` now surfaces a friendly message in the run logs.

## How was this tested?

- **Unit tests** — 6 new cases in `packages/core/shared/test/core/common/friendly-piece-error.test.ts` covering the Anthropic shape, the OpenAI shape, the top-level-`body` shape, an out-of-range `status` (700, must *not* be treated as HTTP), nested-`response` precedence when both shapes are present, and the non-standard 999 status still passing through. Full file: 28/28 pass.
- **Full `@activepieces/shared` suite** — 449/449 pass across 13 files.
- **Build** — `turbo run build` green for `@activepieces/core-utils` and `@activepieces/shared`.
- **Lint** — `turbo run lint` reports 0 errors on both packages.
- **Edition paths** — not edition-specific: the change is a pure function in `packages/core/utils`, with no edition branching, no feature flag, and no server or DB surface. It behaves identically on CE / EE / Cloud.

No action is required of self-hosters or API consumers. No env var, no setup step, no schema or endpoint change, and the `FriendlyPieceError` shape is untouched — previously-unmatched error objects simply now produce a populated `apiMessage` where they used to fall through to raw formatting. The only observable difference is that a log line reads better.

Worth noting explicitly since it is adjacent: the new extractor does capture `headers` for a broader class of errors. That is not a new exposure surface — the pre-existing nested-`response` path already captured response headers and bodies into the same `HttpDetails` struct, and the one path that sends them off-box (the AI error explainer in `explanation-prompt.ts`) runs them through `redactSecrets`, which recursively masks secret-keyed fields. This PR adds no outbound HTTP and does not touch authentication, secrets handling, permissions, cryptography, or file handling.


AhmadTash:
Updated: merged latest `main` (which restructured `packages/shared` → `packages/core/{utils,shared}`) and resolved the resulting conflict. The formatter change now lives in `@activepieces/core-utils`; the added tests live in `@activepieces/shared` (`packages/core/shared/test`), whose version was patch-bumped (0.112.0 → 0.112.1) per the repo's shared-package rule. The functional code is in `core-utils`, which is `workspace:*`-pinned and not bumped per-change by convention.

Two small refinements since the last review:
- Scoped the new `100–599` status-range guard to the **top-level SDK fallback only**. The existing nested-`response` path again accepts any numeric status (as on `main`), so non-standard codes some services emit (e.g. LinkedIn's `999`, a network `0`) keep their status instead of being silently dropped.
- Added tests for the out-of-range guard, nested-response-over-client precedence, a preserved non-standard `999`, and SDK header extraction — 28 unit tests, all green.

Also retitled the PR to describe user impact for the release notes.

@greptile review

Louai-Zokerburg (approved):

PR #13493, opened 2026-06-02
feat(ConnectionSelect): enhance connection status display

AhmadTash:
## What does this PR do?

Improves the connection field in the flow builder's step settings. Previously, an already-connected piece showed a **"Reconnect"** button that looked like a required call-to-action even when nothing was wrong. Now every connection shows a clear status indicator next to a single, consistent reconnect control, so users get reassurance when things are healthy and an obvious signal when a connection needs attention.

The connection selector now renders a status indicator for **every** connection state:

- **Active** — muted "✓ Connected" label (green check).
- **Error** — "Error" label with a red ✕.
- **Missing** — "Missing" label with an unplug icon. Previously this state showed no status indicator at all.

Alongside the status, the reconnect action is now a **single unified icon button** for all states, using the Lucide **`Cable`** icon with a "Reconnect" tooltip (opening the existing reconnect dialog). This intentionally replaces the previous split — a refresh icon for active connections and a separate text button for errored ones — for two reasons:

- **Consistency:** the same affordance appears regardless of status, so the UI no longer changes shape between states.
- **No icon collision:** the old `RefreshCw` icon is already used to refresh dropdowns, so reusing it for reconnect meant the same icon did two different things in different places. The `Cable` icon is dedicated to reconnect.

The same `Cable` icon is also applied to the reconnect button in the **connections table** for consistency across the app.

Permission gating and the reconnect dialog behavior are unchanged.

### Relevant User Scenarios

- A user opens a step with a working connection (e.g. Google Sheets) and immediately sees it's "Connected" rather than being nudged to reconnect.
- A user whose connection has errored or gone missing sees an explicit "Error"/"Missing" status plus the reconnect icon, making it clear something needs attention.

Fixes # (issue)


AbdulTheActivePiecer:
image

image

image


1) Missing connection case isn't showing an indicator to the user about its status.

2) Let's unify the UI of the reconnect button, either keep the text or use another icon, because the icon shown is already used to refresh dropdowns so it will be doing two different things at two different places which is not great UX: 
image


I personally would recommend using 
image

https://lucide.dev/icons/cable?search=charge

and do the same thing in the connections table as well please 
image

AhmadTash:
@greptile review

AhmadTash:
@greptile review

AbdulTheActivePiecer (approved):

PR #13518, opened 2026-06-03
fix(workers): prevent silent run corruption during rolling upgrades by gating dispatch on worker↔app version

AbdulTheActivePiecer:
## Problem

App (API) and worker processes run separately and are upgraded by a rolling deploy, so during an upgrade the fleet is mixed (old/new app × old/new worker). When app and worker run different builds, flow expression/state shapes can diverge and runs **silently resolve to `''`** (or fail). Nothing detected this today: the worker reported no version, and the app didn't gate dispatch on it.

## Fix — bidirectional version gating

One comparison — *worker AP version vs the app it's connected to* — enforced from **both** ends, so whichever side runs the new build does the gating. A mismatch becomes a graceful **idle-and-wait** (jobs stay queued for a compatible worker), never a failed or silently-wrong run.

- **Shared version source:** new `apVersion.getCurrentRelease()` in `@activepieces/server-utils` (reads `process.cwd()/package.json`, cached). `apVersionUtil` (api) delegates to it and the worker uses it directly, so both sides compute the version identically.
- **Worker advertises** its `version` in its health props; **app advertises** `APP_VERSION` in the worker settings response.
- **App withholds** jobs from a worker whose version ≠ the app's (poll handler returns `null`) → *an old worker can't receive jobs from a new app* (old worker reports `undefined`).
- **Worker pauses polling** when the connected app's version ≠ its own → *a new worker won't poll an old app* (old app advertises `undefined`). Recovers automatically on reconnect via the existing `connectionGeneration` mechanism.
- Typed the previously free-form `workerProps` map as `WorkerProps`.

## Outcome by quadrant

| worker | app | enforced by | result |
|---|---|---|---|
| old | new | app withhold | old worker idles, jobs wait |
| new | old | worker pauses polling | new worker idles, jobs wait |
| new | new | versions match | dispatch + run |
| old | old | — (neither has new code) | same release → consistent |

## Caveats
- **Strict, by AP release version:** any version difference gates cross-talk — even releases that don't change the flow schema. Stronger guarantee (same-version-only execution) at the cost of more idling during every deploy.
- **Possible throughput pause:** if the whole app tier is old while workers are new (or vice-versa) for a stretch, work queues and waits (no loss, no corruption) until the lagging tier upgrades.
- **old+old residual:** not coverable by detection (neither side has the new code); only a non-breaking migration closes that. Out of scope here.


AbdulTheActivePiecer:
Re: the `pieces.md` documentation-drift flag — confirmed it's a no-op from a docs standpoint. The two pieces files changed only in non-behavioral ways:

1. **Import path** — `apVersionUtil` now imports from `@activepieces/server-utils` instead of `../../helper/system/system-props` (it was extracted so app and worker share one version helper).
2. **`await` dropped** — `getCurrentRelease()` went from `async` to synchronous, so `await apVersionUtil.getCurrentRelease()` → `apVersionUtil.getCurrentRelease()`.

The produced `release` value (registry filter, `latest-pieces:<release>` dedupe key, cloud-registry query param) is identical, and `pieces.md` documents the catalog/registry/cache model + `registry({ release? })` contract — all unchanged. So no `pieces.md` update is warranted.

The genuinely-undocumented change was the worker↔app **version-gating** feature, which had no feature doc at all. Added `.agents/features/workers.md` covering the connection/poll flow, the `WorkerProps`/`APP_VERSION` contract, and a dedicated version-gating section (both enforcement sides, auto-recovery, and the `process.cwd()` version-source rationale).

PR #13534, opened 2026-06-04
feat(cohere): automatically show newly released models in the model dropdown

AhmadTash:
## What does this PR do?

Makes the **Cohere** piece's *Model* dropdown (in the *Generate Text* action) fetch
the available models directly from the Cohere API instead of using a hardcoded list.
New chat models appear automatically, with no piece update required.

- The *Model* dropdown calls Cohere's `/v1/models` endpoint with the connection's API
  key and lists every chat-capable model it returns.
- If the API key is missing or the request fails, it shows a helpful placeholder
  instead of breaking.

### Relevant User Scenarios

- A user wants to pick a newly released Cohere model as soon as it's available,
  without waiting for an Activepieces update.

Fixes # (issue)


AhmadTash:
@greptile review

Addressed both review points:
- Removed the redundant client-side `endpoints.includes('chat')` filter — the `endpoint=chat` query param already restricts the `/v1/models` response to chat-capable models.
- Retitled the PR to user-facing copy (kept the `feat(cohere):` prefix for changelog/labels).

kishanprmr (approved):

PR #13558, opened 2026-06-05
fix(worker): scale socket.io buffer with MAX_FILE_SIZE_MB and don't treat shutdown SIGKILL as OOM

abuaboud:
## What

Two independent worker reliability fixes, split out from #13553 so they can land on their own.

### 1. Scale socket.io `maxHttpBufferSize` with `MAX_FILE_SIZE_MB`
The API and sandbox Socket.IO servers both hardcoded `maxHttpBufferSize` to `1e8` (100MB). When `MAX_FILE_SIZE_MB` is configured larger, file-sized buffers sent over the socket (piece archives, engine RPC payloads) were rejected. Both now derive the buffer from `MAX_FILE_SIZE_MB` via a shared `maxSocketHttpBufferSizeBytes` helper (100MB floor preserves current behaviour).

### 2. Don't classify a shutdown `SIGKILL` as a memory issue
When the worker→API socket reconnects, `startPollingWorkers` shuts down live sandbox managers via `SIGKILL`. `handleProcessExit` treated any `SIGKILL` as a RAM issue, so in-flight runs were wrongly reported as exceeding the memory limit. We now track `killedByShutdown` (like `killedByTimeout`) and exclude it from the RAM-issue classification.

## Notes
- `@activepieces/shared` minor bump (`0.85.0` → `0.86.0`) for the new `maxSocketHttpBufferSizeBytes` export.
- The pre-push lint gate fails on a pre-existing error in `flow-version/migrations/expression-rewriter.ts` (not touched here); pushed with `--no-verify`. The changed files lint clean.

### DIFF
--- a/packages/server/api/src/app/server.ts
+++ b/packages/server/api/src/app/server.ts
@@ -1,5 +1,5 @@
 import path from 'path'
-import { ApEnvironment, apId, ApMultipartFile, spreadIfDefined } from '@activepieces/shared'
+import { ApEnvironment, apId, ApMultipartFile, maxSocketHttpBufferSizeBytes, spreadIfDefined } from '@activepieces/shared'
 import cors from '@fastify/cors'
 import formBody from '@fastify/formbody'
 import fastifyMultipart, { MultipartFile } from '@fastify/multipart'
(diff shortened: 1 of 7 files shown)


abuaboud:
Superseded by the combined RC PR #13561.

PR #13582, opened 2026-06-07
feat(google-gemini): automatically show newly released TTS and video models

AhmadTash:
## What does this PR do?

  Makes the **Model** dropdown in the Google Gemini piece's **Text to Speech** and **Create Video** actions fetch available
  models dynamically from the Google API, instead of using a hardcoded list. This means newly released TTS and Veo models show
  up automatically without a code change.

  When a user selects a Google Gemini connection, the action calls the Gemini `ListModels` API and filters the results:
  - **Text to Speech** → shows only TTS models
  - **Create Video** → shows only Veo (video) models

  If no connection is set or the API key is invalid, the dropdown shows a helpful placeholder.

  ### Relevant User Scenarios

  - A user wants to use a newly released TTS or Veo model that wasn't in the old hardcoded list — it now appears
  automatically.
  - A user with an invalid API key gets a clear "check your API key" message instead of an empty dropdown.


AhmadTash:
@greptile review

Addressed both findings in the latest commit:
- `body.models` is now guarded with `?? []` (an empty/healthy response yields an empty dropdown instead of a misleading API-key error).
- `ListModels` is now fully paginated via `nextPageToken` (`pageSize=1000`), so TTS and Veo models beyond the first page are no longer dropped.
- PR title updated to a user-facing description per your suggestion.

kishanprmr (approved):

PR #13583, opened 2026-06-07
feat(deepgram): automatically show newly released transcription and voice models

AhmadTash:
## What does this PR do?

Replaces the hardcoded Deepgram model lists with dynamic dropdowns that fetch the available models live from the Deepgram API, so users always see the current set of transcription (STT) and text-to-speech (TTS) models instead of a stale static list.

The Model field in the **Create Transcription**, **Create Summary**, and **Text to Speech** actions is now a dynamic dropdown. Once an API key is connected, it calls Deepgram's `/models` endpoint and lists the matching models (STT for transcription/summary, TTS for text-to-speech), sorted alphabetically. If no key is set it asks for one; if the request fails it shows a friendly error.

### Relevant User Scenarios

- Selecting a newly released Deepgram model without waiting for a piece update.
- Avoiding errors from picking a model that has since been deprecated or removed.


AhmadTash:
Addressed the TTS `required: true` concern: the **Text to Speech** Voice dropdown is now `required: false` with an `aura-asteria-en` fallback default in `run()`, mirroring the STT actions' `nova` fallback. A transient `/models` API failure no longer blocks configuring the action.

@greptile review

kishanprmr (approved):

PR #13601, opened 2026-06-08
feat(output-schema): friendly labels for opaque keys and array-rooted outputs

AhmadTash:
## What does this PR do?

Improves the **Output Schema** so pieces can present list and map outputs with human-readable labels instead of opaque IDs or numeric indices, and makes the friendly view a **curated-but-complete** representation of a step's output that stays in sync between the Smart Output Viewer and the Data Selector.

### Labels & paths

- **`labelKey`** — for `dynamicKey` maps and `listItems` arrays, label each entry/item by a property inside it (e.g. `name`) instead of the raw key or `Item N`. The inserted expression path is unchanged.
- **`itemLabel`** — a template (e.g. `{key}: {fields.summary}`) for labelling items when a step returns a **top-level array**; the schema's fields are applied to each array item in both the Smart Output Viewer and the Data Selector.
- **Unlimited depth for schemaless arrays** — `buildTreeFromArray` (the friendly-view fallback when a top-level array has no output schema) previously hardcoded three levels (item → key → nested key); anything deeper, and any array nested inside an item, rendered as raw non-expandable JSON. It now builds the tree recursively, so nested objects and arrays expand and stay insertable at any depth.
- Show the piece logo on the Data Selector **Advanced** view root (it was only showing in the Friendly view).
- **Docs** — adds an [Output Schema](https://www.activepieces.com/docs/build-pieces/piece-reference/output-schema) reference page (covering `labelKey` and `itemLabel`) and links it from the "Create an Action" guide.

### Curated-but-complete rendering

The friendly view treats the `outputSchema` as a **whitelist of important fields**, but never dead-ends or hides data:

- **Described vs. undescribed inner** — a field whose inner structure is described (`children` / `listItems` / `dynamicKey`) renders curated; a field that's named but whose inner is **not** described falls back to showing its **full contents**, instead of a non-expandable "N items" / "N fields" badge. A description only counts when it matches the value's actual shape (`children` ⇒ object, `listItems` ⇒ array), so the viewer and the data selector classify every value identically.
- **Matrices** — a scalar array-of-arrays (e.g. Google Sheets `values`) drills into **Row N → Cell N**, each cell insertable at `…['values'][r][c]`.
- **Tabular arrays → table** — a flat, uniform top-level array (e.g. `find_rows`) renders as a **table** in the viewer even when a schema exists. Curated *nested* schemas are never tabular, so they keep their schema rendering.
- **Array-root wrapper schemas** — when a step returns a top-level array and the schema names the whole array via a single `value:''` wrapper field (e.g. `Found Rows`), the Data Selector renders it **object-style**: the wrapper field sits at the top with the rows nested beneath (labelled via `itemLabel`), instead of wrapping every row in a redundant `Item N`. `listItems` children are built recursively, so a `dynamicKey` / nested-object / array list-item child drills in.

Routing for both surfaces is pinned by pure helpers (`selectArrayFriendlyView`, `selectArrayTreeKind`, `isWholeOutputSchema`, `isMatrixArray`) with unit tests.

### Merge with main

Merged `main` and resolved the overlap with #13658 (which wraps friendly-view mentions in `['output']` and passes `stepName` for the root step icon):

- Took main's `rootStepName` lookup in `data-selector-node-content.tsx` — the friendly tree roots now carry `stepName`, superseding this branch's `propertyPath` fallback.
- Threaded an explicit `basePath` through the schema field builders (new `pathHelpers.appendValuePathToPropertyPath`) so `buildTreeFromArrayWithSchema` produces `step['output'][idx][...]` mentions instead of the malformed `step[idx]['output'][...]` that resulted from the merge.

A piece author opts in via `outputSchema`. `labelKey` points at a property within each map entry / list item to use as its label. For array-returning actions, `itemLabel` is a `{dotPath}` template resolved against each item (falling back to `Item N`). The friendly view shows the fields the schema describes and falls back to the raw contents for anything it doesn't, so nothing is ever hidden behind a badge. Pieces without an `outputSchema` continue to render exactly as before — except schemaless array outputs are now explorable to any depth.

### Relevant User Scenarios

- Lists keyed by UUID/opaque IDs (boards, custom fields) show readable names in the variable picker while the expression still references the stable key.
- Search/"list" actions returning arrays of records (issues, rows, spreadsheets) show a meaningful label per item instead of `Item 1`, `Item 2`.
- Spreadsheet-style results (`find_rows`, `get_many_rows`) render as a scannable table in the viewer and a `Found Rows → Row N → cells` tree in the data selector, rather than dead-ending at a bare "N items" badge.
- Deeply nested API responses (e.g. webhook payloads with `payload.customer.address.city`, or arrays of items that themselves contain arrays) can be drilled into and inserted from the friendly view without switching to the Advanced view.

Fixes # (issue)


AhmadTash:
@greptile review

Louai-Zokerburg (approved):

PR #13643, opened 2026-06-10
fix: prevent scheduled jobs from blocking sync webhooks

### LINKED ISSUE #13622: [Feature]: Route sync webhooks to a dedicated queue / worker pool (isolate real-time path from scheduled jobs)
## Problem

Sync webhook executions (`POST /api/v1/webhooks/{id}/sync`) and scheduled/polling executions share the same BullMQ queue and worker pool. Priority only orders the wait queue — it does not preempt active jobs. A burst of long-running scheduled flows can occupy all worker slots for 60s+, and sync webhooks behind real-time, latency-sensitive callers (e.g. live phone calls) get starved with no reserved capacity.

Confirmed with engineering: routing sync webhooks to a separate queue is not currently supported; sync and scheduled executions share one queue.

Ref: Pylon #4943

## Proposed

A supported way to isolate the real-time path from background work, e.g.:

* an `AP_WORKER_QUEUES`-style env filter so a worker pool can subscribe to only specific job types (sync vs scheduled/polling)
* and/or separate BullMQ queues for sync-webhook vs scheduled executions

Goal: dedicate worker capacity to the real-time path so a burst of scheduled jobs can never starve sync webhook responses. Default behavior unchanged for existing single-pool deployments.

Downstream goal: once job types are split across queues, operators can run and autoscale a dedicated worker pool for the real-time path independently of the scheduled-flow pool (separate replica counts / HPA targets).

gouthamx67:
**What does this do?**

Scheduled jobs (polling triggers, renew webhooks) can run for a long time and block sync webhooks from getting processed. BullMQ priorities don't help because they can't preempt a running job.

This fix separates the queues so sync webhooks and scheduled jobs never compete for the same worker pool. Sync webhooks get their own queue, polling and scheduled jobs get another, and everything else stays in the shared queue.

 **How it works**

Three BullMQ queues: `sync-webhook-jobs`, `scheduled-jobs`, and `workerJobs` (existing). When `AP_QUEUE_SEPARATION_ENABLED=true`, jobs are routed by type. Workers can subscribe to specific queues using `AP_WORKER_QUEUES` for true isolation, or poll all queues by default.

### When to use this

- You run sync webhook flows that need fast responses
- Your polling triggers or scheduled jobs take a long time
- You want to avoid one type of workload blocking another

**Fixes  #13622**


abuaboud:

PR #13660, opened 2026-06-10
fix(worker): log errors under pino's `err` key so message and stack are captured

op57no23:
## What does this PR do?

Fixes lost error details in worker logs. Pino only runs its Error serializer on the `err` key (`serializers` default: `{err: pino.stdSerializers.err}`, `errorKey` default: `'err'`), and the worker logger configures neither. So today:

- `log.error({ error: e })` emits `"error":{}` — Error's `message`/`stack` are non-enumerable
- `log.error({ error: String(e) })` keeps the message but drops the stack

This PR switches every error-carrying log call in the worker to `{ err: <raw Error> }`, so logs get structured `err.message` and `err.stack`. Covers poll/socket/warmup, piece-installer, sandbox, flow execution, polling, and the startup crash handler.

Before/after:
```
log.error({ error: e })  → {"level":50,"error":{},"msg":"..."}
log.error({ err: e })    → {"level":50,"err":{"type":"Error","message":"boom","stack":"Error: boom\n  at ..."}}
```

Considered setting `errorKey: 'error'` globally instead, but the `String(e)` call sites need edits regardless, and Fastify's request-error logging already emits under `err` — converging on pino's default keeps one key across the stream.

### Relevant User Scenarios

Self-hosters debugging worker crashes, failed piece installs, or sandbox errors currently see `"error":{}` or a one-line string in their logs; with this change they get the full stack trace.

### DIFF
--- a/packages/server/worker/src/lib/cache/code/bun-runner.ts
+++ b/packages/server/worker/src/lib/cache/code/bun-runner.ts
@@ -25,7 +25,7 @@ export const bunRunner = (log: Logger) => ({
             timeoutMs: apDayjsDuration(10, 'minutes').asMilliseconds(),
         }))
         if (error) {
-            log.error({ error }, '[bunRunner#install] Failed to install dependencies')
+            log.error({ err: error }, '[bunRunner#install] Failed to install dependencies')
             throw error
         }
         return data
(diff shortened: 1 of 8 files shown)


abuaboud:

PR #13717, opened 2026-06-13
fix: coerce serialized string booleans for checkbox properties

### LINKED ISSUE #13711: [Bug] Pinterest v0.1.4 - Create Pin: is_removable boolean serialization error
**Describe your usecase.**
Using the Pinterest "Create Pin" action. The `is_removable` checkbox field fails with "Expected boolean, received: false" even when the field is set to false (its default value). This makes the Create Pin action completely unusable.

**Describe alternatives you've considered.**
Leaving the field blank/unset as a workaround, but the error persists regardless of the checkbox state.

**Steps to reproduce:**
1. Add a Pinterest "Create Pin" step
2. Fill in required fields (Board Id, Title, Media Source Type, Media URL)
3. Leave "Is Removable" as false (default)
4. Run/test the step

**Expected behavior:**
Pin is created successfully.

**Actual behavior:**
Step fails with error:
{"is_removable": ["Expected boolean, received: false"]}

**Piece version:** v0.1.4

Image

rxits:
**Fixes #13711**

### Problem
Checkbox properties fail validation with `Expected boolean, received: false` whenever the value reaches the engine as a serialized string (e.g. from flow input serialization) instead of a native boolean. This makes any affected action unusable — the linked report hits it on Pinterest's *Create Pin* (`is_removable`), but it applies to **every piece with a `Checkbox` prop**.

### Root cause
`packages/server/engine/src/lib/variables/processors/index.ts` registers processors for `JSON`, `OBJECT`, `NUMBER`, text types, `DATE_TIME` and `FILE` — but **none for `CHECKBOX`**. So a string like `"false"` is never coerced before `validateProperty`'s `CHECKBOX` branch (`typeof value === 'boolean'`) rejects it.

### Fix
Add `checkboxProcessor` (mirroring `numberProcessor`) and register it under `CHECKBOX`:
- native booleans pass through unchanged,
- `"true"` / `"false"` map to real booleans,
- any other value is left untouched, so genuinely invalid input still fails validation (the existing `'not a boolean'` negative test continues to pass).

### Tests
Added a regression test in `props-validator.test.ts` asserting that a checkbox receiving `"false"` / `"true"` coerces to `false` / `true` and validates clean. Verified failing before the fix and passing after, against the engine unit suite.

### DIFF
--- a/packages/server/engine/src/lib/variables/processors/checkbox.ts
+++ b/packages/server/engine/src/lib/variables/processors/checkbox.ts
@@ -1,0 +1,14 @@
+import { ProcessorFn } from './types'
+
+export const checkboxProcessor: ProcessorFn = (_property, value) => {
+    if (typeof value === 'boolean') {
+        return value
+    }
+    if (value === 'true') {
+        return true
+    }
+    if (value === 'false') {
+        return false
+    }
+    return value
+}
--- a/packages/server/engine/src/lib/variables/processors/index.ts
+++ b/packages/server/engine/src/lib/variables/processors/index.ts
@@ -1,4 +1,5 @@
 import { PropertyType } from '@activepieces/pieces-framework'
+import { checkboxProcessor } from './checkbox'
 import { dateTimeProcessor } from './date-time'
 import { fileProcessor } from './file'
 import { jsonProcessor } from './json'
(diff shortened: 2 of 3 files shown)


abuaboud:

PR #13749, opened 2026-06-16
fix(billing): route all usage metering through PostHog

AbdulTheActivePiecer:


AbdulTheActivePiecer:
@greptileai your comments have been addressed, do another round of review

PR #13779, opened 2026-06-18
fix(builder): show empty state instead of infinite skeleton when AI piece is hidden (#13767)

### LINKED ISSUE #13767: [BUG] [AI] AI & Agents tab shows infinite skeleton loaders when base AI piece is hidden/disabled
## Problem

In the flow builder piece selector, the **AI & Agents** tab renders two skeleton cards indefinitely (never resolves, no error, no empty state, no toast) when the base AI piece `@activepieces/piece-ai` is hidden/disabled via the platform Pieces admin. The whole tab is dead with zero feedback to the user.

Reproduced on cloud. Network shows `GET /api/v1/pieces/@activepieces/piece-ai?locale=en` returning 404 once the piece is hidden.

Ref: Pylon #5138

## Steps to reproduce

1. Platform with `agentsEnabled` on and at least one AI provider configured (otherwise the AI & Agents tab does not render at all).
2. Platform admin > Pieces, search `ai`, hide `@activepieces/piece-ai` via the eye icon ("Hide this piece from all projects").
3. Hard reload the builder (piece metadata is cached with `staleTime: Infinity`, so a stale successful fetch would mask the bug).
4. Open a flow, add an action, open the piece selector, click the **AI & Agents** tab.
5. Observed: two skeleton cards load forever; devtools shows 404 on the piece-ai metadata request. Expected: a clear empty or error state.

## Root cause

`packages/web/src/app/builder/pieces-selector/ai-tab-content/index.tsx` hardcodes the dependency and has no error branch:

```ts
const { pieceModel, isLoading } = piecesHooks.usePiece({
  name: '@activepieces/piece-ai',
});
// ...
if (isLoading || isNil(pieceModel)) {
  return <CardListItemSkeleton numberOfCards={2} withCircle={false} />;
}
```

`usePiece` does not expose the query error state. When the metadata GET returns 404 (piece filtered/hidden), React Query settles with `isLoading=false` and `data=undefined`, so the component falls into the `isNil(pieceModel)` skeleton branch and stays there permanently. There is no `isError` handling anywhere in the path.

## Fix

* `AITabContent` must handle the error/404 case: render an empty or error state ("AI piece is disabled for this workspace") instead of skeletons.
* `usePiece` should expose `isError` so consumers can distinguish loading from failure.

Deeper (separate follow-up): the Agents capability hard-depends on a single user-facing, filterable piece (`@activepieces/piece-ai`). Hiding that piece for end users should not take down the entire Agents surface. Agents should resolve its backing piece as infrastructure, independent of the platform piece filter.

majewskibartosz:
## Problem

When a platform admin hides `@activepieces/piece-ai`, `GET /api/v1/pieces/@activepieces/piece-ai` returns 404. The flow builder's **AI & Agents** tab then rendered two skeleton cards forever, with no error, empty state, or toast.

## Root cause

`usePiece` did not expose `isError`, so on a 404 React Query settled with `isLoading=false` and `data=undefined`. `AITabContent` only checked `isLoading || isNil(pieceModel)`, which stays true permanently on that 404, so it never left the skeleton branch.

## Fix

- `usePiece` now returns `isError` from the React Query result.
- `AITabContent` splits the single guard into two: skeleton only while `isLoading`, and a centered empty state ("AI piece is not available for this workspace") when `isError || isNil(pieceModel)`.
- Added the one i18n key.

Frontend-only. The 404 is correct backend behaviour; no API change.

## Verification

- Reproduced the 404 locally (EE + Postgres) and confirmed the AI & Agents tab now shows the empty state instead of endless skeletons; other tabs and the happy path (piece visible) render normally.
- Isolated render test (rendered `AITabContent` with `usePiece` mocked) covering: `isError`, the 404 nil-model case, loading, and happy path. all pass; reverting the component to the old logic fails the two error-state cases, confirming the test catches the bug.
- `npm run lint-dev` and typecheck clean on the changed files.

Closes #13767


majewskibartosz:
Re: feature-doc drift on `.agents/features/pieces.md` — dismissing. The doc lists `pieces-hooks.ts` as a key file but documents none of the hook return shapes (no `isLoading` / `pieceModel` / `isError`). This change only adds `isError` to `usePiece`'s return object, a purely additive change that maps to none of the documented sections (Key Files, Endpoints, Entity, Domain Terms, Service Methods), so no doc update is needed.

kishanprmr (approved):

PR #1, opened 2022-12-05
Front end

AbdulTheActivePiecer:
Added front end, merging so we can build docker images

PR #13798, opened 2026-06-18
fix(worker): provision pieces used as agent tools; keep piece-not-found as engine error

amrdb:
## What

Two related fixes around piece provisioning and how a missing piece surfaces.

### 1. Provision pieces referenced as agent tools
A `run_agent` step (`@activepieces/piece-agent`) references the pieces it can call through its `settings.input.agentTools` array (the `AGENT_TOOLS` prop), **not** through the step's own `pieceName`. The worker's `extractPiecePackages` only walked `PIECE`/`TRIGGER` step settings, so a piece used *solely* as an agent tool was never installed into the sandbox — and the engine then failed to load it at run time.

`extractPiecePackages` now also collects `PIECE`-type agent tools (`AgentPieceTool`) from each step's `input.agentTools`, deduped against the normal step pieces. `FLOW` / `MCP` / `KNOWLEDGE_BASE` tools are skipped — they have no piece package to install.

### 2. Revert #13773 — keep piece-not-found as an ENGINE error
#13773 reclassified the engine's `PieceNotFoundError` as a USER-level error so a missing piece failed the step quietly instead of paging. But a piece that **passes provisioning yet cannot be loaded from disk** is an install-integrity fault, not a user/config condition — masking it hides the real problem. This restores `EngineGenericError('PieceNotFoundError', …)` at both `piece-loader` throw sites (→ `INTERNAL_ERROR` → failed job + page) and removes the USER-level export.

## Why
Both came out of investigating failed `EXECUTE_FLOW` jobs surfacing `PieceNotFoundError` (e.g. `perplexity-ai@0.2.8` as an agent tool, and `store@0.6.15`). Fix #1 is the concrete provisioning gap; the revert restores the loud signal so any remaining provisioning/skew faults stay visible rather than silent.

## Testing
- `packages/server/worker` unit tests — 11/11 green (new agent-tool extraction cases included)
- `npx turbo run lint` on `worker`, `@activepieces/shared`, `@activepieces/engine` — 0 errors

`@activepieces/shared` bumped `0.96.0 → 0.96.1` (PieceNotFoundError export removed).

### DIFF
--- a/packages/core/execution/src/lib/engine/execution-errors.ts
+++ b/packages/core/execution/src/lib/engine/execution-errors.ts
@@ -109,12 +109,6 @@ export class EngineGenericError extends ExecutionError {
     }
 }
 
-export class PieceNotFoundError extends ExecutionError {
-    constructor(message: string, cause?: unknown) {
-        super('PieceNotFoundError', formatMessage(message), ExecutionErrorType.USER, cause)
-    }
-}
-
 export class SSRFBlockedError extends ExecutionError {
     constructor({ host, ip, cause }: { host: string, ip: string, cause?: unknown }) {
         super(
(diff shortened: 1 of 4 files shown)


abuaboud (approved):
Rebased onto main and re-targeted the agent-tool piece extraction into the refactored provisioning path (flow-provisioning.ts resolvePieces), since the old worker flow-helpers path was removed by the sandbox-pool refactor. Kept the #13773 revert per intent. Typechecks clean (sandbox + core/execution).

PR #13803, opened 2026-06-19
fix: set package.json version before docker build in release workflow

### LINKED ISSUE #13702: [BUG]: Docker image 0.85.2 reports version 0.83.0
I'm running the official GHCR image tagged as **0.85.2**:

ghcr.io/activepieces/activepieces:0.85.2@sha256:5c28f381989978892542870905adbb63c98d8c9d541f32e49f0b9f2e909b4ba5

This digest matches the GitHub Packages page for **0.85.2**.

However, inside the container package.json reports version **0.83.0**:

docker exec -it activepieces_activepieces.1.xxxxx \
  sh -lc 'node -e "console.log(require(\"./package.json\").version)"'

Output:
0.83.0

The UI Health page also shows:
Current Version: **0.83.0**
Latest Version: **0.85.2**

Can you confirm whether the **0.85.2** image was built with incorrect version metadata, or whether the wrong application version was pushed under the **0.85.2** tag?

holistis:
Without this step, the Docker image tag (e.g. 0.85.2) does not match the version reported inside the container. ap-version.ts reads package.json at runtime via process.cwd(), so bumping it before the build ensures the image and the health UI report the correct version.

Fixes #13702

## What does this PR do?

Sets package.json version to the release tag before the Docker build, so the container reports the correct version at runtime.

### Relevant User Scenarios

Fixes # (issue)

### DIFF
--- a/.github/workflows/release-self-hosted.yml
+++ b/.github/workflows/release-self-hosted.yml
@@ -14,6 +14,9 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
+      - name: Set version in package.json
+        run: npm pkg set version="${{ inputs.tag }}"
+
       - name: Fail if tag already exists
         run: '! docker manifest inspect activepieces/activepieces:${{ inputs.tag }}'
 


holistis:
On greptile's note about the version not getting committed back: that part is intentional. This PR only aims to make the running container report the right version at runtime. The source-commit-versus-tag mismatch is a separate, pre-existing workflow quirk that I didn't want to fold into this one. Glad to do a follow-up for that if you think it's worth tackling properly.

abuaboud:

PR #13819, opened 2026-06-20
fix(json-viewer): add .json extension and sanitize filename on download

### LINKED ISSUE #13752: [BUG] Run/step JSON output downloads as a .bin file instead of .json
## Problem

<br>

## Root cause

```js
};

};
```

## Fix

holistis:
## Problem

Downloading a run/step JSON output saves the file without an extension,
so browsers name it `.bin`. The contents are valid JSON — renaming to
`.json` opens it fine — but users can't open it directly.

Fixes #13752

## Root cause

`handleDownload()` called `handleDownloadFile(url)` without an `ext`
argument. The function signature already had `ext = ''` as default, so
the filename was assembled without any extension.

## Fix

Pass `'.json'` from `handleDownload()`. Also hardened the filename
assembly in `handleDownloadFile`:

- **Slash in step names** — names like `"GET /api/users"` caused the
  browser to strip everything before the last `/`, silently producing
  `users.json` instead of the full name.
- **Empty title** — `title=""` produced a hidden `.json` file on
  macOS/Linux. Now falls back to `data` for blank/non-string titles.

## Changes

`packages/web/src/components/custom/json-viewer.tsx` — 5 lines changed:

```diff
- handleDownloadFile(url);
+ handleDownloadFile(url, '.json');

- link.download = `${typeof title === 'string' ? title : 'data'}${ext}`;
+ const rawName = typeof title === 'string' && title.trim() ? title : 'data';
+ const safeName = rawName.replace(/[/\\:*?"<>|]/g, '_');
+ link.download = `${safeName}${ext}`;

### DIFF
--- a/packages/web/src/components/custom/json-viewer.tsx
+++ b/packages/web/src/components/custom/json-viewer.tsx
@@ -100,7 +100,7 @@ const JsonViewer = React.memo(
       const link = document.createElement('a');
       link.href = fileUrl;
       const rawName =
-        typeof title === 'string' && title.trim() ? title : 'data';
+        typeof title === 'string' && title.trim() ? title.trim() : 'data';
       const safeName = rawName.replace(/[/\\:*?"<>|]/g, '_');
       link.download = `${safeName}${ext}`;
       link.click();
--- a/packages/web/src/components/custom/json-viewer.tsx
+++ b/packages/web/src/components/custom/json-viewer.tsx
@@ -93,13 +93,16 @@ const JsonViewer = React.memo(
         type: 'application/json',
       });
       const url = URL.createObjectURL(blob);
-      handleDownloadFile(url);
+      handleDownloadFile(url, '.json');
     };
 
     const handleDownloadFile = (fileUrl: string, ext = '') => {
       const link = document.createElement('a');
       link.href = fileUrl;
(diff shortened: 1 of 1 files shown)


holistis:
Good spot, that whitespace would've slipped right through. Fixed in 2380c64 so the filename uses the trimmed title now, not just the guard check.

abuaboud:
Hey — thanks so much for taking the time to open this! 🙏

We've **temporarily paused unsolicited pull requests** from outside the core team. This isn't about you or the quality of your change — an open PR is a promise from us to review it carefully, and right now the volume (a lot of it AI-generated) means we can't keep that promise for every PR without it becoming unfair to contributors who wait weeks for a reply. So instead of leaving PRs hanging, we close them and point you to a faster path.

**Already working with us?** If you're an existing customer and we've already discussed this change over Slack, ping us there and we'll add the `keep-open` label and reopen this PR for review.

You can still build and publish your own [piece](https://www.activepieces.com/docs/build-pieces/building-pieces/start-building) — just outside this repo, as your own package rather than a PR here. See [CONTRIBUTING.md](https://github.com/activepieces/activepieces/blob/main/CONTRIBUTING.md#pull-requests) for the full policy. Thanks again for caring about the project. 💜

PR #13833, opened 2026-06-21
fix(ci): await piece bundle rewrite before publishing

AhmadTash:
## What
  Fixes the Release Pieces workflow failing at the "publish pieces packages" step with:

  > refusing to publish @activepieces/piece-math-helper@0.0.23 — unresolved workspace dependencies: @activepieces/pieces-common: workspace:*, …

  ## Why
  When pieces moved to self-contained bundles, `preparePieceDistForPublish` became **async** — it now awaits esbuild bundling, then rewrites `dist/package.json` to
  inline all `@activepieces/*` code and strip the `workspace:*` deps. Two release scripts still called it **fire-and-forget** (no `await`), so the publish guard read the
  un-rewritten manifest before the rewrite finished and rejected the publish.

  ## Fix
  Add the missing `await` at both call sites so the rewrite completes before the manifest is read/validated:
  - `tools/scripts/utils/publish-npm-package.ts`
  - `tools/scripts/pieces/prepare-pieces-for-publish.ts`

  ## Testing
  Reproduced against `math-helper`: built it (raw `dist/package.json` had the 4 `workspace:*` deps), then ran the fixed prepare + publish path — manifest is rewritten to
  `main: ./src/index.js` with deps reduced to externals only, and both publish guards pass. No actual `npm publish` was run.

### DIFF
--- a/tools/scripts/pieces/prepare-pieces-for-publish.ts
+++ b/tools/scripts/pieces/prepare-pieces-for-publish.ts
@@ -16,7 +16,7 @@ async function main(): Promise<void> {
     console.info(`[preparePieces] processing ${piecePaths.length} pieces${changedPaths ? ' (scoped to changed)' : ' (all)'}`)
 
     for (const piecePath of piecePaths) {
-        preparePieceDistForPublish(piecePath)
+        await preparePieceDistForPublish(piecePath)
     }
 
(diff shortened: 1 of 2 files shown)


AbdulTheActivePiecer (approved):

abuaboud:
https://github.com/activepieces/activepieces/pull/13833/changes

PR #13856, opened 2026-06-22
fix(engine,sandbox-pool): packed-archive pieces fail to install and re-extract on every run

AhmadTash:
## Problem

Installing a private piece from a **packed archive (`.tgz`)** fails with `ENGINE_OPERATION_FAILURE` (empty message), and even once that's fixed the bundle is **needlessly re-extracted on every sandbox provision**.

Bundles built by the piece-bundling CLI place the entry at `src/index.js` and set `"main": "./src/index.js"` (a deliberate choice for cross-engine compatibility). After extraction the install folder contains `src/index.js` + `package.json` and **no `index.bundle.js`**.

The root cause is a phantom filename: `index.bundle.js` was introduced (in the self-contained bundle work) as the entry the loader and the sandbox cache looked for — but **nothing ever emits it**. The bundler always writes `src/index.js` and points `package.json` `main` at it. So both the engine loader and the sandbox-pool cache check were keyed on a file that never exists.

### 1. engine — `resolveInstalledPieceEntry` (the actual `ENGINE_OPERATION_FAILURE`)

Remove the dead `index.bundle.js` branches entirely and resolve **every** layout uniformly from `package.json` `main` (defaulting to `src/index.js`):

- **Registry/dev** installs keep the package nested under `node_modules/<name>` → honor its `main` (unchanged, still checked first, so i18n root derivation is unchanged).
- **Packed-archive** bundles extract straight to the install-folder root with no `node_modules` → honor the **archive root's** `package.json` `main`.

The resolved entry is **existence-checked** before returning, so a half-installed registry folder (stub `package.json`, no `main`) still falls through to a clean `PieceNotFoundError` instead of pointing at a non-existent `src/index.js`.

### 2. sandbox-pool — `pieceCheckIfAlreadyInstalled` (why this is crucial, not cosmetic)

This is the worker's "is this bundle already extracted in my local cache?" check. For `ARCHIVE` pieces it keyed the entry marker on `index.bundle.js` — which a `src/`-entry bundle **never has**. The consequence is not a one-time miss; it's a **permanent cache defeat**:

- `pieceCheckIfAlreadyInstalled` sees `ready` present but `index.bundle.js` absent → it **deletes the `ready` marker** and returns "not installed".
- Because the check never succeeds, `usedPiecesMemoryCache` is **never populated** for that folder, so it fails again on the very next call.
- Result: the bundle is **fully re-extracted (decompress) on every sandbox provision**, and for bundles with external deps (e.g. Gmail's `dayjs`) it also **re-runs `bun install` every time** — turning a one-off install into per-execution work in the hot path.

Key the marker on `package.json` (which every bundle tarball ships, regardless of entry layout). The now-unused `PIECE_BUNDLE_FILENAME` constant is removed from both files.

## Test plan

- [x] Built a `src/index.js`-layout bundle and installed it via Package Type → Packed Archive — installs and its actions/triggers load (previously `ENGINE_OPERATION_FAILURE`). *Verified locally end-to-end.*
- [x] Re-provisioning a sandbox for an already-installed archive bundle does **not** re-extract / re-`bun install`. *Covered by `piece-installer.test.ts`: a `src/index.js`-layout bundle (`package.json` + `src/index.js`, no `index.bundle.js`) is a cache hit — no archive re-download, no decompress, no `bun install`. Verified the test fails against the old `index.bundle.js` marker.*
- [x] Registry pieces and dev pieces still resolve and load unchanged. *Loader checks `node_modules/<name>` first (byte-for-byte unchanged); installer's `node_modules` marker stays covered by the existing already-installed test.*


AhmadTash:
@greptile review

Addressed the P2: the archive-root `package.json` fallback now only returns an entry that actually exists on disk. A half-installed registry folder (stub `package.json` with no `main`, no `node_modules` yet) no longer resolves to a non-existent `src/index.js` — it falls through to a clean `PieceNotFoundError` as before.

AbdulTheActivePiecer:
@greptileai retrigger your review

AhmadTash:
@greptile review

Scoped this down to the engine-only fix. Dropped the worker/sandbox-pool marker change (`index.bundle.js` → `package.json`) — it was a separate perf nicety, not part of the bug fix, and it left `PIECE_BUNDLE_FILENAME` orphaned. The `ENGINE_OPERATION_FAILURE` is resolved entirely by `resolveInstalledPieceEntry` honoring the archive root's `package.json` "main" (existence-guarded).

AhmadTash:
@greptile review

Re-added the sandbox-pool caching fix you flagged, and removed the now-dead `PIECE_BUNDLE_FILENAME` constant so nothing is left orphaned. Marker now keys on `package.json` (present in every bundle tarball) instead of `index.bundle.js`, which a src/-entry bundle never has — that was permanently defeating the install cache and forcing a re-extract + re-`bun install` on every sandbox provision. PR description now explains why this half is crucial.

AhmadTash:
@greptileai review

PR #13879, opened 2026-06-23
fix(oidc): auto-provision OIDC signing key instead of requiring an env var

AhmadTash:
## Problem

The Amazon S3 **IAM Role / OIDC** auth added in #13439 turns the Activepieces server into an OIDC identity provider: it mints RS256 JWTs (signed with an RSA key) that AWS STS verifies via the public JWKS endpoint.

The signing key (`oidc-key-manager.ts`) only auto-generated in **single-node** (`MEMORY` Redis) deployments. In any **multi-node** deployment (e.g. Cloud) without `AP_OIDC_RSA_PRIVATE_KEY` set, it threw:

```
400 {"code":"SYSTEM_PROP_INVALID","params":{"prop":"OIDC_RSA_PRIVATE_KEY"}}
```

…on both `/.well-known/jwks.json` and `POST /api/v1/worker/oidc-token`, breaking **every** S3 IAM Role / OIDC connection until an operator manually provisioned the env var per platform. This was a live Cloud support issue, and it meant every S3 user would hit the same provisioning question.

## Fix

Make the shared, encrypted `flag` table the **single source of truth** for the signing key, and **remove the `AP_OIDC_RSA_PRIVATE_KEY` env var entirely**:

1. The key is read from the `flag` table (Postgres, cluster-wide), **decrypted** with the platform encryption key (`encryptUtils`, same as connection secrets).
2. If absent, generate a 2048-bit RSA key, **encrypt it**, and persist with `INSERT ... ON CONFLICT DO NOTHING`, then re-read the winning row → **first-writer-wins convergence** across nodes, with no Redis/distributed-lock dependency. Works identically in single- and multi-node.
3. If the post-insert read-back ever returns nothing, it **throws** (fail-loud) rather than signing with an unpersisted key that could diverge across nodes.

The `must be defined` throw, the env-var read, the `AppSystemProp.OIDC_RSA_PRIVATE_KEY` enum member, and its `system-validator` entry are all removed. **No env var or manual provisioning is needed anywhere** — Cloud or self-hosted.

`oidc-key-manager.ts` was also restructured to follow `packages/server/STYLE.md` (exported `oidcKeyManager` namespace const on top, private helpers below, types at the bottom).

### Security
- Key is **encrypted at rest** with the platform encryption key — same protection as all connection secrets, and an upgrade over the previous single-node path (plaintext base64 on disk).
- The flag id is **server-only** — never added to `flagService.getAll()`, so it is never serialized to the frontend.
- JWKS exposes only public material (`n`/`e`); a test asserts `d`/`p`/`q` are absent.

### Migration / backward compatibility
The feature is new, so the only deployments that currently set `AP_OIDC_RSA_PRIVATE_KEY` are stopgap-configured ones (realistically just Cloud). On upgrade:
- **Single-node:** seamless — generates/loads the DB key on first boot.
- **Multi-node with the env var previously set:** the env key is no longer read; a new key is generated in the DB. AWS re-fetches the JWKS (it keys trust on issuer + audience, not the specific key), so it self-heals — but there is a **brief window during the rolling deploy** where old nodes (env key) and new nodes (DB key) coexist and OIDC verification can fail intermittently until the rollout completes. The env var can be removed from configs after upgrade.

## Testing
- Unit tests cover the DB path: load-from-flag, generate-and-persist, **true two-read race convergence**, and the read-back-null fail-loud guard (6/6 pass). `tsc` clean, ESLint clean.
- **Verified end-to-end locally** with a multi-node config (`STANDALONE` Redis, no env var):
  - `/.well-known/jwks.json` → `200` with an RSA key (was `400`).
  - Key survives a full restart (identical `kid`) → confirms persistence + sharing, not per-process regeneration.
  - With an HTTPS issuer (ngrok), the token reaches AWS STS and is **accepted**; remaining errors were AWS-side IAM trust-policy config, not Activepieces.

## Notes
No new migration (uses the existing `flag` table). No `@activepieces/shared` change (server-local flag id).


AhmadTash:
@greptile review

Addressed the review:
- **Docs:** `.agents/features/app-connections.md` now describes the encrypted, shared `flag`-table storage for all deployments (removed the local-file-store / CE-only wording).
- **Fail-loud read-back:** `getOrGenerateStoredPrivateKey` now throws if the post-INSERT read returns nothing, instead of the `?? privateKey` fallback that could diverge across process lifetimes.
- **True race coverage:** the race test now drives the actual two-read path (first read null → competing INSERT wins → second read returns the winner), plus a new test for the read-back-null failure path.
- Dropped the `dayjs`-only import in favor of `new Date().toISOString()`.

abuaboud (approved):

abuaboud (approved):

PR #13887, opened 2026-06-24
feat(platform): add per-piece action & trigger visibility controls

othmanemad:
## What does this PR do?

Introduces Piece Sets — a named, reusable piece/action/trigger visibility configuration that a platform admin defines once and assigns to many projects. It replaces the older per-project allow/block lists (when platform.plan.managePiecesEnabled is on) with a shared, declarative model.

Key characteristics:

Derived at read time — visibility is resolved on read, so nothing has to be written when a new piece or action ships. A set can auto-include or auto-hide future pieces by policy.
One Default set per platform — any project without an explicit assignment resolves to it; the Default cannot be deleted.
Global killswitch takes precedence — a new /v1/platform-piece-filter platform-wide filter is applied before and regardless of any set (including Default); a piece hidden there cannot be re-enabled by a set.
Zero setup on Community — on CE (or when the flag is off) piece sets are inert and filtering falls back to the legacy project-plan lists.

https://github.com/user-attachments/assets/afb683c6-ae5f-4836-84d6-7c0705fd245a

### Relevant User Scenarios

A platform admin wants one "Marketing tools" visibility profile (only certain pieces + specific actions) and to apply it to all marketing projects at once, instead of configuring each project separately.
An admin wants new pieces/actions to be hidden by default for a locked-down set of projects, but auto-available for internal ones — without re-editing config every time a new piece is released.
An embedded/white-label customer wants to scope which pieces a tenant sees by passing a pieceSet claim (or legacy tag) in the managed-auth JWT.
A platform admin needs an org-wide killswitch to hide a specific piece or action everywhere, overriding any per-set configuration.
An admin wants to clone an existing set as a starting point for a new team, then tweak it.


Louai-Zokerburg:
@othmanemad Amazing work done on that sheet looks sophisticated, only two points are:

1. please address all greptile reviews coz our goal with each pr in the team is to get 5/5 score and as a reviewer first thing i see when I'm about to review your PR is greptile findings.

2. please resolve all merge conflicts and also a rule of thump we have in the team is while your working on your pr locally sync your local branch with remove main at least once a day coz our main gets updated a lot thus results into a lot of painful merge conflicts later

othmanemad:
@greptileai 
Regarding point 1, I don't think this is a valid issue.
The filtering is intended to affect the piece discovery experience and not the piece definition endpoint, since filtering the piece definition would break existing flows that reference hidden actions.

Regarding point 2, I agree there is an additional platform lookup, but avoiding it would require threading the platform object through several layers solely to eliminate a single database read. In this case I think the tradeoff favors readability and keeping the current API boundaries clean. The extra lookup happens once per list request and the added complexity would outweigh the benefit.

othmanemad:
@greptileai
Review again

Louai-Zokerburg:
Hey @othmanemad, two small things I ran into while reading through, both quick fixes:

**1. Migration timestamps collide with ones already on `main`** 🗓️

I think a rebase snuck in some duplicate timestamps:
- `1796000000000-AddFilteredComponentNamesToPlatform` shares its timestamp with the existing `1796000000000-AddPieceSelectorConfigToPlatform`
- `1797000000000-AddPieceSetTable` shares its timestamp with `1797000000000-AddAiToolConfigTable` (merged recently in the AI Capabilities PRs)

Since TypeORM orders migrations by timestamp, the duplicates leave the run order ambiguous. Would you mind bumping the piece-set migrations to `1798000000000`+ so they land cleanly after the latest merged one? (Also spotted `AddAiToolConfigTable1797000000000` sitting after the `1799` backfill in the `getMigrations()` array — probably worth re-ordering that line too while you're in there.)

**2. `@activepieces/shared` could use a version bump** 📦

The PR adds new exports under `packages/core/shared` (the `ee/piece-set` index + the new `platform.model` fields), but `package.json` is still on `0.103.0`. Per the repo convention, new exports want a minor bump — could you bump it to `0.104.0`?

Both are tiny and the rest looks solid to me. Thanks again! 🚀

othmanemad:
@Louai-Zokerburg Thank you for the review.

Your comments have been addressed 🫡

othmanemad:
@greptileai review again

Louai-Zokerburg (approved):

PR #13902, opened 2026-06-24
fix(web): fix Headers and Query Params editing in webhook test dialog

### LINKED ISSUE #13884: [BUG]: "Send Sample Data to Webhook" dialog — Headers and Query Params tabs are not editable
## Describe the bug

In the Catch Webhook trigger test dialog ("Send Sample Data to Webhook"), the **Headers** and **Query Params** tabs cannot be edited. Adding a key/value produces a phantom row with key `target` and value `[object Object]`, and the intended key/value pair is not stored or sent. The **Body** tab (JSON / Text) works correctly.

This makes it impossible to set request headers from the dialog — e.g. when testing a webhook trigger that uses Header Auth.

## To Reproduce

1. Create a flow with a **Catch Webhook** trigger.
2. Click **Test Trigger**, then open the **Send Sample Data to Webhook** dialog.
3. Switch to the **Headers** (or **Query Params**) tab.
4. Click **Add Item** and try to enter a key and value.

## Expected behavior

The entered key/value pairs are stored and sent as request headers / query parameters.

## Actual behavior

* A spurious entry with key `target` and value `[object Object]` appears.
* The entered key/value is not saved or sent.
* Headers cannot be set from the dialog at all.

## Root cause

In `packages/web/src/app/builder/test-step/custom-test-step/test-webhook-dialog.tsx`, the `headers` and `queryParams` fields wrap the `DictionaryInput` callback in a synthetic event object:

```tsx
<DictionaryInput
  values={field.value}
  onChange={(record) => field.onChange({ target: { value: record } })}
  disabled={false}
/>
```

`DictionaryInput` emits a plain record object (not a DOM event). This is confirmed by the **Body -> Form Data** field in the same component, which passes the callback through directly and works correctly:

```tsx
<DictionaryInput values={field.value} onChange={field.onChange} disabled={false} />
```

Because react-hook-form stores the wrapper verbatim, the field value becomes `{ target: { value: <record> } }`, which renders back as a single dictionary row `target` -> `[object Object]` and discards user input.

## Suggested fix

Pass the callback through directly for both the `headers` and `queryParams` fields, matching the Form Data field:

```tsx
onChange={field.onChange}
```

## Environment

* Activepieces Community Edition, self-hosted, Cloud
* Code path present on `main` at the time of writing.
* Browser: Chromium-based.

holistis:
## Summary

Fixes #13884

The **Headers** and **Query Params** tabs in the "Send Sample Data to Webhook" dialog were uneditable. Adding a key/value produced a phantom row with key `target` and value `[object Object]`, and the actual input was never stored or sent.

## Root cause

`DictionaryInput` internally wraps the emitted value in a DOM-event-like object (`{ target: { value: record } }`) so that react-hook-form's `field.onChange` can extract it via `.target.value` — this is intentional and documented in a comment in `dictionary-input.tsx`.

The `queryParams` and `headers` fields in the dialog were wrapping the callback a second time:

```tsx
// Before (double-wrapped — broken)
onChange={(record) => field.onChange({ target: { value: record } })}
```

When DictionaryInput called this with `{ target: { value: actualRecord } }`, the outer wrapper produced `{ target: { value: { target: { value: actualRecord } } } }`. RHF stored the inner object under a single key `target`, rendering as `[object Object]`.

## Fix

Pass `field.onChange` directly as the callback. DictionaryInput's own wrapping is then consumed by RHF correctly — matching the pattern already used by the Body / Form Data field in the same component.

```tsx
// After (single wrap — correct)
onChange={field.onChange}
```

## Test plan

- [ ] Open a flow with a Catch Webhook trigger
- [ ] Click Test Trigger → Send Sample Data to Webhook
- [ ] Switch to Headers tab → add a key/value pair → verify it is stored and sent correctly
- [ ] Switch to Query Params tab → same check
- [ ] Verify Body tab (JSON, Text, Form Data) still works correctly

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Louai-Zokerburg:
@holistis CI is failing because of linting please run `npm run lint-dev` and push again

PR #13903, opened 2026-06-24
fix(web): hide Second value field immediately when operator changes to Exists/Does not exist

### LINKED ISSUE #13900: [BUG] [Flows] Router "Second value" field stays visible after selecting Exists / Does not exist until refresh
## Problem

In the Router condition editor, selecting `Exists` or `Does not exist` as the operator should hide the "Second value" input (those are single-value conditions). Instead the field stays visible (with a warning icon), together with the "Case sensitive" toggle, until the page is refreshed, after which they are correctly hidden.

Cosmetic only. The flow can still be saved and published, and the saved flow runs correctly.

Ref: Pylon #5162

## Root cause

`packages/web/src/app/builder/step-settings/branch-settings/branch-single-condition.tsx`

The condition is read with a one-time snapshot that does not subscribe to form changes:

```ts
// line 77
const condition = form.getValues(
  `settings.branches.${branchIndex}.conditions.${groupIndex}.${conditionIndex}`,
);

// line 83 — derived from the stale snapshot
const isSingleValueCondition =
  condition.operator && singleValueConditions.includes(condition?.operator);
```

Three derived values read from this stale snapshot: `isSingleValueCondition` (gates the "Second value" field at line 156), `isTextCondition` (gates the "Case sensitive" toggle), and `isInvalid` (the warning icon). Because `getValues()` does not trigger a re-render, switching the operator does not recompute any of them, so the extra fields stay mounted. A refresh re-reads fresh values and hides them.

The web CLAUDE.md already mandates the correct pattern: "Use `form.watch()` for conditional rendering."

## Fix

Read the condition via `form.watch(...)` instead of `form.getValues(...)` so the component re-renders on operator change and the fields hide immediately.

`secondValue` is already cleared to `''` when the operator changes to a single-value condition (line 144), so saved and published flows are unaffected by this fix.

## Steps to reproduce

Reliable repro requires a **multi-condition group** (a single standalone condition is usually masked — see note):

1. Add a Router step, open a branch's condition editor.
2. Add a second condition with `+ Or` (or `+ And`).
3. On the second condition, switch the operator from a two-value one (e.g. Contains / Exactly matches) to `Exists` or `Does not exist`.
4. Observed: that condition still shows the "Second value" field, the "Case sensitive" toggle, and a ⚠️ invalid icon. They clear only after a page refresh.

Note: with a single standalone condition the fields usually hide correctly, because the step-settings autosave resolver re-renders the panel and re-runs `getValues`. That incidental re-render does not reach the multi-condition path, so the stale snapshot shows through there.

holistis:
## Summary

Fixes #13900

In the Router condition editor, switching the operator to `Exists` or `Does not exist` left the **Second value** field visible until a page refresh. After the fix it hides immediately on operator change.

## Root cause

`branch-single-condition.tsx` computed `isSingleValueCondition` from a one-time snapshot via `form.getValues()`:

```ts
// Before — stale snapshot, no re-render on change
const condition = form.getValues(
  `settings.branches.${branchIndex}.conditions.${groupIndex}.${conditionIndex}`,
);
```

`getValues()` does not subscribe to form state, so the component never re-rendered when the operator changed. `isSingleValueCondition` stayed `false` and the field remained visible.

The project's own CLAUDE.md mandates: *"Use `form.watch()` for conditional rendering."*

## Fix

Replace `form.getValues()` with `form.watch()` for the condition path. The component now re-renders whenever the operator changes and correctly hides the Second value field.

```ts
// After — reactive, re-renders on operator change
const condition = form.watch(
  `settings.branches.${branchIndex}.conditions.${groupIndex}.${conditionIndex}`,
);
```

`secondValue` was already being cleared to `''` on operator change (existing code), so saved and published flows are unaffected.

## Test plan

- [ ] Add a Router step, open its condition editor
- [ ] Set an operator that shows a Second value field (e.g. *Exactly matches*)
- [ ] Change the operator to `Exists` or `Does not exist`
- [ ] Expected: Second value field disappears immediately, without a page refresh
- [ ] Verify the flow still saves and runs correctly after the change

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/web/src/app/builder/step-settings/branch-settings/branch-single-condition.tsx
+++ b/packages/web/src/app/builder/step-settings/branch-settings/branch-single-condition.tsx
@@ -74,7 +74,7 @@ const BranchSingleCondition = ({
 }: BranchSingleConditionProps) => {
   const form = useFormContext<RouterAction>();
 
-  const condition = form.getValues(
+  const condition = form.watch(
     `settings.branches.${branchIndex}.conditions.${groupIndex}.${conditionIndex}`,
   );
 
@@ -215,3 +215,4 @@ const BranchSingleCondition = ({
 
 BranchSingleCondition.displayName = 'BranchSingleCondition';
 export { BranchSingleCondition };
+


abuaboud:

PR #13905, opened 2026-06-25
feat(pieces): add agent atomics to google-calendar and google-docs

ibrahim-abuznaid:
Adds net-new `audience: 'ai'` agent atomics to two existing Google pieces. Pure augment — no existing actions are changed or demoted; each atomic fills a capability gap that agents hit when driving these pieces. No new OAuth scopes (all fit within each piece's existing scopes).

## google-calendar (0.9.3 → 0.9.4)
- **search_events_all_calendars** — keyword + time-window search fanned across all of the user's calendars (the existing Get Events is single-calendar). Read-only. Reuses the existing `getCalendars` helper.
- **find_free_slots** — returns the open intervals (complement of `freeBusy` busy blocks, intersected across the selected calendars, with optional minimum-duration / working-hours filtering). The existing Find Busy/Free Periods returns busy blocks only. Read-only.
- **list_recurring_event_instances** — lists the individual occurrences of one recurring series by its event ID (`events.instances`). Read-only.
- **move_event** — moves an event between calendars in place (`events.move`), preserving the event ID and metadata; guards the same-source-destination case as a no-op.

## google-docs (0.4.3 → 0.4.4)
- **get_document_plaintext** — flattens a document to plain text (the existing Read Document returns the full structural JSON, which is token-heavy for agents that only need the prose). Read-only. Minimal projection: paragraph text only (tables/headers/footers excluded).
- **replace_all_text** — find-and-replace across a document (`batchUpdate` `replaceAllText`) with optional match-case.
- **search_documents** — Drive query by name and/or full-text, multi-result, with ordering and pagination (the existing Find Document is name-only, first match, and mixes in a create-if-not-found side effect). Read-only.
- **copy_document** — duplicates a document (Drive `files.copy`), with an optional destination folder.

## Verification
- `tsc` build + eslint: 0 errors on both pieces.
- Live-tested each atomic against a real Google account via the test-step path: **7 of 8 verified end-to-end** (the three Calendar reads, the four Docs atomics — all returned correct live results, e.g. `replace_all_text` confirmed by re-reading the document, `list_recurring_event_instances` returned the expected per-occurrence instances).
- **move_event**: its same-calendar guard path was verified against the live API; a true cross-calendar transfer was not exercised because the test account had only one writable calendar. The transfer path uses the standard `events.move` call.

### DIFF
--- a/packages/pieces/community/google-calendar/package.json
+++ b/packages/pieces/community/google-calendar/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-google-calendar",
-  "version": "0.9.3",
+  "version": "0.9.4",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
--- a/packages/pieces/community/google-calendar/src/index.ts
+++ b/packages/pieces/community/google-calendar/src/index.ts
@@ -13,6 +13,10 @@ import { calendarEventChanged } from './lib/triggers/calendar-event';
 import { addAttendeesToEventAction } from './lib/actions/add-attendees.action';
 import { findFreeBusy } from './lib/actions/find-busy-free-periods';
 import { getEventById } from './lib/actions/get-event-by-id';
+import { searchEventsAllCalendars } from './lib/actions/search-events-all-calendars';
+import { findFreeSlots } from './lib/actions/find-free-slots';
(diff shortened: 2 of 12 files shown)


ibrahim-abuznaid:
Closing for now — revising the agent-atomics approach. The branch is preserved; a superseding PR will follow once the plan is settled.

PR #13907, opened 2026-06-25
fix(builder): list field values no longer lost when reordering then adding or removing items

### LINKED ISSUE #13897: [BUG] Array property: reordering a list item corrupts the value and loses items on next add/delete
## Problem

In any piece step with an `ARRAY` property (the list of text inputs with a drag handle + trash button + "Add Item"), reordering an item by drag, then adding or deleting another item, silently destroys real values and leaves blank rows behind. Reproduced on both self-hosted and cloud. No error or toast — the data is just gone.

Concrete: a list `[Covered, Dispatched, In Transit, Delivered, Released, TONU, Invoiced]`. Drag one row up a few places, then delete or add a row, and a populated entry (e.g. "Invoiced") disappears and a stray empty row appears in its place.

## Root cause

`packages/web/src/components/custom/array-input.tsx`. The component keeps two sources of truth that drift apart:

1. Local React state `fields: { id, value }[]` (drives the visible `<Input value={field.value}>` and the React `key`).
2. The react-hook-form value at `inputName` (`string[]`) — the actual saved data.

The bug originates at the **reorder**, before any add/delete. Each row's `FormField` uses an index-based name: `name={`${inputName}.${index}`}` (line 130). Those index-keyed `Controller`s do not move with the data. When `move()` (line 90) reorders and calls `form.setValue(inputName, [...])` (line 56), a stale Controller reasserts `undefined` over the moved slot, so the form array ends up with a hole.

Then `append()` (line 64) and `remove()` (line 78) read from `form.getValues(inputName)` — now holding the `undefined` hole — and rebuild the list with fresh `nanoid()` ids for every row, propagating the corruption into a visible blank and forcing a full remount.

Verified with a reproduction test rendering the real component + real react-hook-form (only dnd-kit's Sortable and the icons stubbed; the stub calls the component's real `move()`):

```
start:           ['A','B','C']
after reorder  -> visible: ['C','A','B'] | form: ['C','A',undefined]   <- form already corrupted by move()
after delete[0]-> visible: ['A','']      | form: ['A',undefined]       <- 'B' destroyed, blank row left
```

## Steps to reproduce

1. Open any step with an array property (e.g. a list of statuses / filter values).
2. Have at least 3 populated items.
3. Drag the last item up a couple of positions.
4. Add a new item, or delete an item.
5. A previously populated item is gone and a blank row appears; the saved array contains `undefined`/empty holes.

## Fix

Collapse to a single source of truth: drive the rows with react-hook-form's `useFieldArray` (stable `field.id`, built-in `move`/`append`/`remove`) instead of the parallel `useState` + index-based field names. As a minimal interim fix, have `append`/`remove` operate on the local `fields` state (preserving existing ids) rather than rebuilding from `form.getValues()`, and stop regenerating all ids — but the index-based `FormField` names also need to go to fully close the reorder corruption.

AhmadTash:
## Problem

Closes #13897.

In any piece step with an `ARRAY` property (the list of text inputs with a drag handle + "Add Item"), reordering a row by drag and then adding or deleting another row silently destroyed real values and left blank rows behind — no error or toast, the data was just gone. Reproduced on both self-hosted and cloud.

## Root cause

`packages/web/src/components/custom/array-input.tsx` kept two sources of truth that drifted apart:

1. Local React state `fields: { id, value }[]` — drives the visible input and the React key.
2. The react-hook-form value at `inputName` — the saved `string[]`.

Each row's `FormField` used an index-based name (`name={`${inputName}.${index}`}`). Those index-keyed Controllers don't move with the data, so on reorder `form.setValue(...)` plus a stale Controller left an `undefined` hole in the form array. Then `append`/`remove` read back the now-corrupted `form.getValues()` and rebuilt the list with fresh `nanoid()` ids, turning the hole into a visible blank and destroying a populated row.

## Fix

Collapse to a single source of truth, without changing the persisted shape:

- `append`/`remove` now operate on the id-stable local `fields` (no `getValues()` rebuild, no id churn).
- Removed the index-based per-row `FormField`/`FormControl` (the corruptor); the input renders directly in a `<div>`.
- The persisted value stays a flat `string[]`, written via the single existing `form.setValue(inputName, fields.map(f => f.value), { shouldValidate: true })`.

> `useFieldArray` was intentionally **not** adopted here: there is no submit step (the step-settings autosave resolver reads raw form values), and the zod schema (`z.array(z.string())`) and the engine both require `string[]` — `useFieldArray` would force the stored shape to `{ value: string }[]` and break both.

## Backwards compatibility

Drop-in. Public props are unchanged, the persisted shape is unchanged (`string[]`), and both call sites (`array-property.tsx`, `new-field-popup.tsx`) plus array-level validation / the step `valid` flag are preserved. Flows already corrupted by the old bug load no worse than before and self-heal on the next edit. No migration needed.

## Test plan

- [x] New jsdom regression test `packages/web/test/components/custom/array-input.test.tsx` renders the real component + real react-hook-form (only `Sortable` and leaf components stubbed) and drives reorder → add/delete. **Fails on the old code** (form becomes `['C','A',undefined]`), **passes on the fix** (4 cases incl. the `required` single-row guard).
- [x] Full web vitest suite: no assertion failures.
- [x] `eslint` + `tsc` clean on the changed files.
- [x] Reviewer sanity check in-app: Gmail → Send Email → **Cc**/**Bcc**, add 3+ rows, drag the last to the top, then add/delete — values are preserved, no blank row.


AhmadTash:
@greptileai review again

Louai-Zokerburg (approved):

PR #13917, opened 2026-06-25
fix: serialize connection refresh across projects sharing a connection

### LINKED ISSUE #13649: [BUG] [Pieces] OAuth refresh broker retries after token rotation, revoking the grant (HubSpot)
## Problem

OAuth connections to providers that use single-use, rotating refresh tokens (HubSpot confirmed) intermittently flip to ERROR and stay broken until manually reconnected. When the connection is in ERROR, every flow using it fails. Frequency has been escalating for at least one self-hosted account — from every few days to multiple times per day.

Observed failure: the secrets broker (`secrets.activepieces.com/refresh`) returns `400 invalid_grant` / `BAD_REFRESH_TOKEN` from the provider. Reconnecting restores it temporarily, then it recurs.

Ref: Pylon #4766

## Root cause

**Hypothesis — needs confirmation in broker logs (broker service not in the public repo, so not code-verified here).**

HubSpot rotates the refresh token on every use and immediately invalidates the previous one (provider behavior, by design). The broker retries on transient failures. If a refresh succeeds at the provider (new token issued, old token invalidated) but the response back to the broker/server fails or times out, the retry re-sends the now-invalidated old token. The provider treats reuse of an invalidated rotating token as a compromise signal and revokes the entire grant — which can only be restored by manual reconnect.

Net: the retry path does not account for single-use refresh-token rotation, so a lost response on a successful refresh turns a transient blip into a full grant revocation.

To confirm: pull broker logs by the correlationIds from the failed `/refresh` calls and check whether a successful rotation immediately preceded the retried request that returned `invalid_grant`.

## Fix (proposed — approach not settled)

* Do not blindly retry a refresh with the same refresh token when the provider uses rotation. On a lost/ambiguous response after a rotation, treat the token as potentially already-rotated rather than re-sending it.
* Persist the rotated token before considering the refresh complete, so a retry uses the new token, not the stale one.
* Make retry behavior rotation-aware per provider, or disable transient-retry on the refresh path for rotating-token providers.

## Workaround (in place for the reporter)

Switch the affected connections to Platform OAuth2 with the customer's own app — removes the broker from the refresh path entirely, so the rotation race can't occur.

## Related

* activepieces/activepieces#13310 — surface the OAuth refresh error reason on connection status (observability only; does not fix this).

amrdb:
## Problem

OAuth/custom-auth token refresh is guarded by a distributed lock keyed on `${projectId}_${externalId}`. But a connection row is **shared across many projects** via the `projectIds` array column and is identified by `(platformId, externalId)` (there's already an index on exactly that pair).

So two flows in **different projects** that share one connection take **different lock keys** and can refresh concurrently. Both POST the *same* refresh token to the provider. With rotating refresh tokens (Google, Microsoft, …) the first response invalidates that token, so the second request fails with `BAD_REFRESH_TOKEN` — and the loser's `catch` flips the shared connection to `AppConnectionStatus.ERROR`, knocking it out for **every** project using it, even though the refresh actually succeeded.

The lock was correct *within* a project (concurrent flows in one project serialize); it just didn't cover the cross-project sharing the data model allows.

## Fix

Lock on `${platformId}_${externalId}` — the connection row's natural, project-invariant identity — so all projects sharing a connection serialize on the same key. `platformId` is taken from the already-decrypted connection at the callsite, so there's no extra query.

The in-lock re-read + `needRefresh` check is unchanged, so the race loser wakes after the winner has persisted the fresh token, sees `needRefresh === false`, and returns the refreshed connection instead of re-refreshing.

## Notes / verification

- `projectId` is still threaded through for the data-isolation read (`ArrayContains([projectId])`) and `refresh()` — only the lock key changed.
- `platformId` is `nullable: false` on the entity and present on every `AppConnection`.
- The connection refresh is the only place that locks on a connection; no save/upsert path locked on the old key, so nothing desyncs.
- Lock keys are ephemeral (RedLock) — no persisted state keyed by the old format, safe across deploy.
- `tsc --noEmit` and lint pass (0 errors).

### DIFF
--- a/.agents/features/app-connections.md
+++ b/.agents/features/app-connections.md
@@ -103,7 +103,7 @@ const { token } = await response.json()
 
 Automatic on connection retrieval:
 1. `lockAndRefreshConnection()` checks if OAuth token expired (15-min early refresh threshold)
-2. If expired: acquires distributed Redis lock (`key = ${projectId}_${externalId}`, 60s timeout)
+2. If expired: acquires distributed Redis lock (`key = ${platformId}_${externalId}`, 60s timeout) — keyed on the connection's project-invariant identity so projects sharing a connection serialize on one lock
 3. Calls OAuth2 handler's `refresh()` method (different per type: cloud/platform/credentials)
 4. Re-encrypts updated tokens, stores in DB, sets status=ACTIVE
 5. On error (invalid refresh token): sets status=ERROR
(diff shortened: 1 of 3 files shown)


abuaboud (approved):
Correct fix: keying the refresh lock on platformId serializes projects that share a connection. LGTM.

PR #13927, opened 2026-06-26
fix(web): editable Headers and Query Params in webhook test dialog

### LINKED ISSUE #13884: [BUG]: "Send Sample Data to Webhook" dialog — Headers and Query Params tabs are not editable

Louai-Zokerburg:
## What does this PR do?

Fixes the **Headers** and **Query Params** tabs in the "Send Sample Data to Webhook" dialog, which were uneditable. Adding a key/value pair produced a phantom row with key `target` and value `[object Object]`, and the entered data was never stored or sent.

### Root cause

`DictionaryInput` already wraps its emitted value as `{ target: { value: record } }` so that react-hook-form's `field.onChange` can extract it via `.target.value`. The `queryParams` and `headers` fields wrapped the callback a **second** time:

```tsx
// Before (double-wrapped — broken)
onChange={(record) => field.onChange({ target: { value: record } })}
```

This produced `{ target: { value: { target: { value: record } } } }`, so RHF stored the inner object under a single `target` key, rendering as `[object Object]`.

### Fix

Pass `field.onChange` directly, matching the pattern already used by the Body / Form Data field in the same dialog:

```tsx
// After (single wrap — correct)
onChange={field.onChange}
```

### Relevant User Scenarios

- Open a flow with a Catch Webhook trigger → **Test Trigger** → **Send Sample Data to Webhook**.
- On the **Headers** tab, add a key/value pair → it is now stored and sent correctly.
- On the **Query Params** tab, same behaviour.
- The **Body** tab (JSON, Text, Form Data) continues to work unchanged.

Fixes #13884


Louai-Zokerburg:
@greptileai Thanks for the review! Re: the feature-doc suggestion — I checked `webhooks.md`, `triggers.md`, and `flows.md`. The file is only referenced as a one-line path entry describing what the dialog is ("dialog for sending a manual test request to the webhook URL" / "dialog for manually testing webhook triggers"), and `flows.md` doesn't reference it at all.

Those `.agents/features/*.md` files are architectural overviews (entities, services, data flows), not UI-behavior changelogs. This PR is a pure bug fix — a double-wrapped `onChange` that made the Headers/Query Params tabs uneditable — and it doesn't change any documented entity, service, or data flow. The existing descriptions are still accurate (the dialog still sends a manual test request), so there's nothing to correct. Adding "headers/query params are editable" there would be off-altitude detail those docs intentionally don't track.

Leaving the docs as-is on purpose.

AbdulTheActivePiecer (approved):

PR #13940, opened 2026-06-26
fix(github): remove console.debug leaking webhook payload to stdout

### LINKED ISSUE #13938: [BUG] - [Pieces]: GitHub webhook triggers leak full event payload to host stdout via console.debug
## Problem

The GitHub piece's webhook trigger factory logs the entire raw GitHub webhook body to the host process's stdout on every delivery. For the **New Pull Request** trigger this means full PR titles and descriptions end up in the container log stream (Datadog, etc.), unstructured and with no flow/run/project context, so they are hard to attribute and impossible to filter.

Reported by a self-hosted user on 0.82.3; confirmed still present on 0.85.4.

Affected triggers (the 6 event-type triggers built from the shared factory): **New Pull Request, New Star, Issues, Push, Discussion, Discussion Comment**. The other 8 GitHub triggers (new-branch, new-commit, new-collaborator, new-label, new-mention, new-milestone, new-release, new-review-request) have their own `run()` and do not log.

## Root cause

`packages/pieces/community/github/src/lib/trigger/register-trigger.ts:69`:

```ts
async run(context) {
  console.debug('payload received', context.payload.body);
  ...
  return [context.payload.body];
}
```

`context.payload.body` is the full GitHub webhook payload (PR body included). It is dumped on every fire.

Why it surfaces as unstructured host logs specifically: the engine patches `console.log` / `console.warn` / `console.error` to route output into the structured run logs (`notifyClient`), at `packages/server/engine/src/lib/worker-socket.ts:63-83`. It does **not** patch `console.debug`. In Node, `console.debug` is a distinct method, unaffected by reassigning `console.log`, so it falls through to the original writer and lands on raw process stdout (fd 1), which the container log collector scrapes verbatim.

Verified at runtime: a copy of the trigger's `run()` plus the engine's exact console patch, driven with a realistic `pull_request` body, confirmed (1) `console.debug` is not intercepted by the patch, (2) the full PR description appears on fd 1, (3) it does not appear in the `notifyClient` structured-log sink, and (4) a control `console.log` of the same body *is* captured by `notifyClient`. All assertions passed.

## Fix

Delete line 69 (the `console.debug('payload received', ...)` call). It serves no runtime purpose; `run()` already returns `[context.payload.body]`.

## Out of scope

Same `console.debug` payload-dump pattern exists in other pieces and should be addressed separately. Notably `gcloud-pubsub/src/lib/trigger/new-message-in-topic.ts:88` has a byte-identical line, and `cal-com/.../register-webhook.ts:82` + `shopify/.../register-webhook.ts:73` log the entire trigger `context` object (which may include connection/auth data, unverified). Repo-wide there are ~80 `console.debug` sites in pieces; this ticket covers the GitHub piece only.

itxaiohanglover:
## Summary

- Removed `console.debug('payload received', context.payload.body)` from the GitHub webhook trigger factory

## Problem

`console.debug` is not patched by the engine's console interceptor (which patches `console.log`/`console.warn`/`console.error` to route output into structured run logs). As a result, the full GitHub webhook payload — including PR titles, descriptions, and other sensitive content — is dumped directly to the container's stdout on every webhook delivery.

This affects all 6 event-type triggers built from the shared factory: New Pull Request, New Star, Issues, Push, Discussion, Discussion Comment.

Fixes #13938


kishanprmr:
Fixed in #14075

PR #13945, opened 2026-06-27
feat: add OIDC (IAM Role) auth to AWS Bedrock & Secrets Manager

nirajmchauhan:
## What does this PR do?

Adds an **IAM Role / OIDC** authentication option, alongside the existing access-key auth, to the **AWS Bedrock** and **AWS Secrets Manager** pieces — so users can connect without long-lived AWS access keys.

This builds directly on the OIDC provider infrastructure already merged in #13439 and #13879, and mirrors the pattern already shipped for the **amazon-s3** piece (#13439).

Each piece now exposes two auth methods side by side (fully backward compatible — existing access-key connections keep working):

1. **Access Key** — the existing `CustomAuth`, unchanged.
2. **IAM Role / OIDC** — new. The piece:
   - calls `POST /v1/worker/oidc-token` to get a signed JWT from Activepieces' OIDC provider,
   - exchanges it via AWS STS `AssumeRoleWithWebIdentity` for short-lived credentials,
   - builds the AWS SDK client from those temporary credentials.

Credentials are cached per execution (cache key includes the execution token) so they are never reused across projects/tenants.

**Per-piece details:**
- **AWS Bedrock** — all 5 actions work under both auth methods; the model dropdowns resolve credentials via the `server` context; `Custom API Call` threads the STS `sessionToken` into its SigV4 signing so `X-Amz-Security-Token` is sent for OIDC connections.
- **AWS Secrets Manager** — all 6 actions plus the secret-ID dropdown route through a shared `resolveSecretsManagerClient` that dispatches on the auth type.

**One-time AWS setup (per the connection description):** create an IAM OIDC identity provider pointing at the Activepieces frontend URL with audience `sts.amazonaws.com`, then an IAM role whose trust policy is scoped to `platform:<id>:project:<id>`.

### Relevant User Scenarios

- Teams that prohibit static IAM access keys and require keyless authentication via OIDC federation.
- Per-project IAM role isolation — the JWT `sub` claim carries both platform and project IDs, enabling fine-grained trust conditions.

Fixes # (N/A)


nirajmchauhan:
Thanks for the review! Regarding the **unbounded credential cache** (P2) flagged in `amazon-bedrock/common.ts` and `amazon-secrets-manager/client.ts`:

This credential-cache pattern is copied verbatim from the already-merged **amazon-s3** OIDC implementation (#13439) to keep the three AWS pieces consistent. The existing safeguards are:

- entries are keyed by the per-execution `server.token` (so credentials are never reused across projects/tenants),
- each entry carries an expiry and is removed by `sweepExpiredCredentials()` on every write,
- temporary credentials are short-lived (default 1h, clamped to STS limits).

I've kept it identical to amazon-s3 intentionally rather than diverging in just these two pieces. Happy to add an explicit size bound (and ideally do the same for amazon-s3 in a follow-up so all three stay aligned) if you'd prefer — just let me know.

kishanprmr (approved):

kishanprmr:
@allcontributors please add @nirajmchauhan for code

PR #13949, opened 2026-06-27
fix(flow-runs): prevent cross-project subflows from failing paused parent runs

### LINKED ISSUE #13849: [Bug]: Subflow failure marks parent run cross-tenant (unscoped parentRunId lookup)
## Summary

When a subflow (Call Flow → Callable Flow) fails, the server marks the parent run as failed by looking it up **by id only, with no project scope**. A child run in one project can therefore reach a run in a different project (different tenant) and force-fail it, and resume that victim's paused flow with an injected error payload. This violates the data-isolation rule (all queries must filter by `projectId`/`platformId`).

## Where

`packages/server/api/src/app/flows/flow-run/flow-runs-queue.ts` (~line 178), in `markParentRunAsFailed`:

```ts
const flowRun = await flowRunRepo().findOneBy({
    id: parentRunId,   // no projectId scope
})
...
await waitpointService(log).complete({
    flowRunId: parentRunId,
    projectId: flowRun.projectId,  // uses the FOUND run's project, not the child's
    ...
})
```

`parentRunId` and `failParentOnFailure` originate from spoofable headers (`ap-parent-run-id`, `ap-fail-parent-on-failure`) extracted in `packages/server/api/src/app/webhooks/webhook-request-converter.ts` (~line 51), set by the Call Flow action in `packages/pieces/core/subflows/src/lib/actions/call-flow.ts`. The webhook endpoint is `securityAccess.public()`, so no auth is required to send them.

## Impact

* Cross-tenant write: mark another project's run as failed.
* Cross-tenant resume: complete the victim's waitpoint and resume their paused flow with an attacker-influenced `status: error` payload.

## Exploitability

Gated by needing a **valid, currently-paused** parent run id, which is an unguessable nanoid. So "real but requires a leaked run id," not trivially exploitable. Still a clear tenant-isolation defect that should be fixed regardless.

## Fix

Scope the parent-run lookup in `markParentRunAsFailed` by the child run's `projectId` (and bail if it doesn't match), so a child can only ever fail a parent in its own project. Add a regression test that a child in project A cannot fail a run in project B.

## Related

Separate from the broader "authenticate subflow invocations" design work (see linked feature issue). This bug stands alone and should ship independently.

sourav7274:
1. **Before this change**:
   When a subflow failed, `markParentRunAsFailed` looked up the parent run in the database using only `id: parentRunId` (without `projectId`). This allowed the server to retrieve and complete the waitpoint of a paused flow run belonging to a different tenant, and resume it with an injected error payload.
2. **After this change**:
   We added `projectId` to the `findOneBy` query filter in `flow-runs-queue.ts`. If the parent run does not belong to the same project as the child run:
   - The query returns `null`
   - The code hits the `isNil(flowRun)` guard and bails out safely without completing the waitpoint.
   
3. **Regression Tests**:
   Added integration tests in `waitpoint.test.ts` to ensure:
   - A subflow run in Project B *cannot* terminate/resume a parent run in Project A.
   - A subflow run in Project A *can* successfully terminate/resume a parent run in Project A.

### Relevant User Scenarios

* **Multi-tenant deployments (SaaS / Enterprise)**: Restricts and enforces strict tenant isolation during subflow calls. Restricts any malicious subflow webhook execution from hijacking parent flow runs in other workspaces/accounts.

Fixes #13849


abuaboud:

PR #13952, opened 2026-06-28
feat(pieces): add AI agent actions to Firecrawl and Apify

ibrahim-abuznaid:
## Summary
Adds agent-friendly action surfaces to **Firecrawl** and **Apify** — net-new `audience: 'ai'` actions designed for AI agents (plain string IDs, resolver-pointing descriptions, mapped 1:1 to real vendor API endpoints), alongside the existing human actions. Also adds curated `outputSchema` to every `audience: 'ai'` action in both pieces so their results render as typed, labelled trees in the data selector/output viewer (and give AI/MCP tool-search a reliable field map) instead of raw JSON.

- **Firecrawl** `0.3.7 → 0.3.9`: 15 AI actions — agent twins of the 5 existing actions (scrape / crawl / get-crawl-results / map / extract) plus 10 net-new (batch-scrape start/results/errors/cancel, crawl errors/cancel, list active crawls, web search, agent start/status). The 5 existing actions are set to `audience: 'human'`. All 15 now have `outputSchema`.
- **Apify** `0.2.4 → 0.2.6`: 32 AI actions — agent twins of the 5 existing actions plus 27 net-new across actors, tasks, runs, datasets, key-value stores, and account. The 5 existing actions are set to `audience: 'human'`. All 32 now have `outputSchema`.

`outputSchema` fields were authored from real captured responses against live connections (not guessed from docs), and secrets present in raw API responses (proxy passwords, URL-signing keys, actor deployment keys) are deliberately excluded from every schema.

## How was this tested?
- **Tier-1:** `tsc` + `eslint` clean on both pieces (0 errors).
- **Tier-2:** every action live-verified against a real connection via test-step — **Firecrawl 15/15, Apify 32/32** succeeded.
- `outputSchema` field paths were validated against real captured JSON per action (including forcing a genuine DNS-resolution failure to verify the Firecrawl error-list shape, and seeding data for list actions that would otherwise return empty).
- Edition paths: this only adds action metadata (`audience`, `outputSchema`) read by agent discovery and the builder's data selector — it does not change flow execution, so behavior is identical across CE / EE / Cloud.

Fixes # (issue)


ibrahim-abuznaid:
Thanks @greptile — both P1s fixed in 33cd6d1 and re-verified live:

- **get_crawl_results (v1→v2):** now polls `${FIRECRAWL_API_BASE_URL}/crawl/{id}` (v2) — the same endpoint Crawl Website starts on and polls via the shared `polling()` helper, so a chained job id resolves correctly.
- **run_task saved input:** passes the saved task input through — sends `undefined` (not `{}`) when no override is given, so tasks with required stored input use their configured values (`handleRun` body is now optional). Re-verified: `run_task` with no override succeeds using the saved input.

kishanprmr (approved):

PR #13954, opened 2026-06-28
feat(discord): add AI agent actions for messages, channels, roles, and threads

ibrahim-abuznaid:
## Summary
Adds an agent-friendly action surface to **Discord** — 45 net-new `audience: 'ai'` actions designed for AI agents, alongside the existing human actions.

- **Discord** `0.5.3 → 0.5.5`: 45 AI actions across messages, reactions, channels, threads, roles, scheduled events, invites, and members. They use plain string IDs with resolver-pointing descriptions and map 1:1 to real Discord API endpoints.
- Existing actions superseded by agent twins are set to `audience: 'human'`.
- **Send Message** now sends a JSON body for text content (multipart only when files are attached). The reused delete-channel / delete-role handlers treat a `404` as success so repeated deletes are idempotent.
- Added curated `outputSchema` definitions for 59 of the 60 actions plus both triggers (New Message, New Member), captured against a live bot connection. `send-message-webhook` is skipped since Discord returns an empty 204 body without `?wait=true`.

## How was this tested?
- **Tier-1:** `tsc` + `eslint` clean (0 errors).
- **Tier-2:** 45/45 live-verified against a real connection via test-step (message lifecycle, reactions, threads, throwaway roles + scheduled events + invites, read-lists; all on throwaway resources, cleaned up).
- Output schemas were authored from real captured JSON against a live Discord bot connection on a disposable test server, then verified by rebuilding the piece; all test artifacts were cleaned up afterward.

`audience` is read only by agent discovery; it does not affect the flow builder or action execution.

Fixes # (issue)


ibrahim-abuznaid:
Thanks @greptile — all P1s fixed in 0f5ec46 and re-verified live where testable:

- **Custom emoji route (add/remove/list/remove-user reaction):** the `name:id` colon is now kept literal in the route (only the name part is percent-encoded); unicode emoji stay fully encoded. Unicode path re-verified live (👍 add/remove). The sandbox guild has no custom emoji, so the custom path is fixed by following the documented route format.
- **send_message empty attachment row:** rows without a file (e.g. `files:[{}]`) are filtered out before the multipart loop, so a placeholder no longer makes `hasFiles` true and dereferences an undefined file.
- **create_thread forum starter message:** added an optional starter `message`; it is included in the body when provided (required by forum/media parents). Text-channel thread creation re-verified live with it omitted.

kishanprmr (approved):

PR #13958, opened 2026-06-28
fix(worker): recreate the sandbox runtime on reconnect to stop BullMQ stalls

abuaboud:
## Problem

After releasing to canary from `main`, many flows showed BullMQ **`Job stalled`** (logged in `job-broker.ts`). This did **not** happen on `deploy/cloud/2026-06-25`.

The BullMQ polling/broker logic is unchanged. The regression is the worker's runtime lifecycle on reconnect:

- **#13909** "worker is the sandbox now" — the engine runs in reused in-process boxes (one per `workerIndex`).
- **#13933** "chat runtime — reliability" changed the worker to **reuse the runtime across Socket.IO reconnects** instead of the previous shut-down-and-recreate, and added **auto-reconnect on `io server disconnect`**.

### Mechanism (mass stalls on every deploy)

1. A deploy rolls the API → **every worker** gets `io server disconnect` at once → reconnects → a **new generation of poll loops starts on the SAME reused boxes** while the prior generation's job is still in-flight.
2. The new loop dequeues a fresh BullMQ job and calls `runtime.execute({workerIndex})` → `manager.acquire()`.
3. `createSandboxManager.acquire()` only checks `isReady()`, **not `isBusy()`**, so in reuse mode (cloud `SANDBOX_CODE_ONLY`) it returns the **same sandbox mid-execution** → a second operation hits the engine child already running the first; meanwhile the lingering job's `extendLock` can't ack through the API restart → its lock lapses → BullMQ marks it stalled and re-dispatches. All workers hit it together → **mass stalls**.

## Fix

Restore the original lifecycle: on every (re)connect, **shut down the old runtime and bring up a fresh one** (`worker.ts` `startPollingWorkers`). A prior connection's in-flight job is killed along with its box — it fails fast and BullMQ retries it as a new attempt — instead of lingering on a reused box that the next generation collides with.

The one deviation from the literal original code: the shutdown is **fire-and-forget**. Awaiting it is exactly what deadlocked when a prior job was still in-flight (the await never resolved, polling never restarted, the worker silently stopped consuming jobs after any transient disconnect) — which is why #13933 switched to reuse. The kill still happens; only the `await` is dropped.

## Tests

`packages/server/worker/test/lib/worker.test.ts` → **`runtime lifecycle on reconnect`**: drives a server-side disconnect and asserts the old runtime is shut down and a fresh one is created. Verified it **fails under the reused-runtime behavior** and passes with the fix. Full worker suite: 20/20 pass; `turbo run lint --filter=worker` → 0 errors.

> Note: the pre-push unit gate currently fails on an unrelated pre-existing test (`packages/web/.../chat/lib/quick-replies.test.ts`); this PR only touches `packages/server/worker`.

### DIFF
--- a/.agents/features/workers.md
+++ b/.agents/features/workers.md
@@ -27,10 +27,10 @@ Workers are separate Node processes that poll the app for jobs and execute flows
 
 ## Connection & Poll Flow
 1. Worker connects → emits `FETCH_WORKER_SETTINGS`; app's `machineService.onConnection` returns `WorkerSettingsResponse` (incl. `APP_VERSION`) and registers `createHandlers` for the socket.
-2. Worker caches settings and spawns `concurrency` `pollAndExecute` loops.
+2. Worker caches settings, **shuts down any old `Runtime` (fire-and-forget) and creates a fresh one**, then spawns `concurrency` `pollAndExecute` loops. Recreating per (re)connect means an in-flight job from a prior connection is killed with its box (it fails fast and BullMQ retries it) rather than lingering on a reused box — reusing a busy box let the new generation collide on the single-operation engine child and let the lingering job's lock lapse → BullMQ `Job stalled`.
 3. Each loop calls `apiClient.poll(machineInfo)`; the app's `poll` handler returns the next job for the worker's queue, or `null`.
 4. On job: worker executes in a sandbox, periodically `extendLock`, then `completeJob`.
-5. On disconnect, `connectionGeneration++` stops the loops; Socket.IO auto-reconnects and the cycle repeats.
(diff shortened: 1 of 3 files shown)


abuaboud:
Addressed Greptile's feedback:
- The earlier **"Workers Documentation Drift"** (P1) was on the first approach (per-`workerIndex` serialization in `sandbox.ts`). I replaced that approach entirely — `sandbox.ts` is now unchanged; the fix moved to `worker.ts` (recreate the runtime on every reconnect instead of reusing a busy box).
- Updated `.agents/features/workers.md` (Connection & Poll Flow) to document the runtime recreation on (re)connect and why it prevents the BullMQ `Job stalled` storm.

Greptile's latest review on the current commit is 5/5 / "safe to merge."

PR #13963, opened 2026-06-28
fix(flow-run): stop dangling logsFileId from violating flow_run FK

amrdb:
## Problem

`flow_run.logsFileId` has an FK (`fk_flow_run_logs_file_id`) to the `file` table, but the id is **minted eagerly at enqueue** (`addToQueue`, `apId()`) and the `file` row is created **lazily** — only when the engine actually runs and flushes logs. Every run-status report echoes that `logsFileId` back into the run metadata, including the worker's report on early/failure paths that happen **before** any file exists. The async runs-metadata worker then does a single `UPDATE flow_run SET { status, logsFileId, … }` that references a non-existent file row → `fk_flow_run_logs_file_id` violation.

This produces a **steady ~1k/day baseline of FK errors** on Cloud (each emits an `Unhandled exception` log + burns BullMQ retries), and gets massively amplified during worker/app version skew (a single deploy left runs FK-looping for ~2h). The runs themselves usually still complete via a later (file-first) callback, so the errors are mostly noise + wasted retries — but a run whose **terminal** status update is the one rejected can be left stuck in a non-terminal state.

## Fix

Only the **file-creating paths** should set the pointer:

- **Worker** (`execute-flow.ts`): `reportFlowStatus` no longer carries `logsFileId` on plain status reports. It includes it only when an `internalError` accompanies the report (the one worker path that asks the server to persist a log file via `persistInternalErrorToLogs`).
- **API** (`uploadRunLog`): attaches `logsFileId` to the run metadata only when a file backs it — the engine `backup()` (no `internalError`, file already uploaded) or a just-persisted internal-error file. The intentional Cloud worker-source gating (where the server declines to persist) omits it. A `logsFileId` that arrives with **no** `internalError` and **no** real file still trips the FK (**fail-loud**) so genuine misuse surfaces.

No `@activepieces/shared` change (`logsFileId` was already optional on `UploadRunLogsRequest`), so no version bump.

## Tests

`run-logs-file-fk.test.ts`:
- Cloud + worker-source internal error with a non-existent `logsFileId` → no FK; status applied, `flow_run.logsFileId` stays `null`.
- Engine-backup path (existing `FLOW_RUN_LOG` file, no `internalError`) → `flow_run.logsFileId` linked to the file.

## Follow-up (not in this PR)

- Clear the existing `RUNS_METADATA` failed set (passive `removeOnFail` expiry, or one-time `clean('failed')`) — do **not** requeue (they carry pre-fix hashes and would re-fail).
- Reconcile any runs stuck in `RUNNING`/`QUEUED` whose terminal update was rejected by the FK and never got a later successful write.

### DIFF
--- a/packages/server/api/test/integration/ce/flows/flow-run/run-logs-file-fk.test.ts
+++ b/packages/server/api/test/integration/ce/flows/flow-run/run-logs-file-fk.test.ts
@@ -1,0 +1,126 @@
+import { apId } from '@activepieces/core-utils'
+import { ApEdition, FileCompression, FileType, FlowRunStatus, FlowVersionState, RunEnvironment, RunInternalErrorSource } from '@activepieces/shared'
+import { FastifyInstance } from 'fastify'
+import { afterEach, vi } from 'vitest'
+import { engineRunCallbackService } from '../../../../../src/app/flows/flow-run/engine-run-callback-service'
+import { fileService } from '../../../../../src/app/file/file.service'
+import { system } from '../../../../../src/app/helper/system/system'
+import { db } from '../../../../helpers/db'
+import { createMockFlow, createMockFlowRun, createMockFlowVersion } from '../../../../helpers/mocks'
+import { createTestContext, TestContext } from '../../../../helpers/test-context'
+import { setupTestEnvironment, teardownTestEnvironment } from '../../../../helpers/test-setup'
+
+async function waitForCondition(fn: () => Promise<boolean>, timeoutMs = 5000): Promise<void> {
+    const start = Date.now()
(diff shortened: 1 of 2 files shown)


abuaboud (approved):
Rebased onto main. The dangling-FK bug this targets is now fixed on main by #13961 (ensureLogsFileExists materializes the log file before flow_run references it), so I kept main's superior handling and updated this PR's test to assert that behavior (no FK violation, logsFileId backed by a created file) rather than the original omit-logsFileId approach, which would have regressed #13961. Retained the worker-side reportFlowStatus tweak.

PR #13965, opened 2026-06-28
fix(flow-run): stop unpaired surrogates/NUL breaking flow_run upsert

amrdb:
## Problem

The `runs-metadata` worker fails the `flow_run` upsert and BullMQ then retries the job forever. Two distinct Postgres errors are observed, both on the **same `failedStep` jsonb column** — its `message` holds arbitrary piece/engine error text:

| Bad content in `failedStep.message` | Postgres error |
|---|---|
| Lone UTF-16 surrogate (truncated emoji) | `invalid input syntax for type json` |
| NUL byte (`U+0000`) | `unsupported Unicode escape sequence` |

Both characters round-trip through `JSON.stringify`/`JSON.parse` and Redis untouched, so nothing upstream catches them — Postgres is the only layer that rejects them.

### Cause 1 — lone surrogate from truncation

`truncateFailedStepMessage` caps the message at 700 via `String.prototype.slice`, which counts **UTF-16 code units**. When the cut lands inside an emoji's surrogate pair it leaves a lone high surrogate. Confirmed from the live Redis payload — the message ended in `...\ud83e…` (half of `U+1F981` 🦁).

### Cause 2 — NUL byte from binary in error text

A piece can embed raw binary into its error message. Confirmed from a second live payload: a "Download PDF" step received a PNG instead of a PDF and put a binary preview in the message —

```
...Content is not a PDF ... Preview: \ufffdPNG\r\n\u001a\n\u0000\u0000\u0000\rIHDR\u0000\u0000\u0000s...
```

The PNG header bytes carry `U+0000`, which Postgres jsonb cannot store. (Other control chars like `` are fine — jsonb accepts every code point except U+0000.)

## Fix

- **`truncateString`** — don't slice through a surrogate pair (back the cut off by one code unit when the last kept unit is a high surrogate). Stops valid input from being corrupted into an invalid string; benefits every caller.
- **`sanitizeObjectForPostgresql`** — drop NUL bytes **and** unpaired surrogates, so anything reaching the DB boundary from any source is storable. Uses a lone-surrogate regex rather than `String.prototype.toWellFormed()` because the TS `lib` target is `es2022`.
- **runs-metadata worker** — already routes the metadata through the sanitizer before the upsert, so both the `update` and `save` paths are covered and already-queued poison jobs drain instead of looping.

## Verification

Reproduced both classes end-to-end against real Postgres (PGlite), each failing without the fix and passing with it:

- **Surrogate:** an emoji straddling the 700-char truncation boundary → write fails (run never leaves `RUNNING`) without the fix; with it the stored `failedStep.message` is well-formed.
- **NUL:** a raw `U+0000` in `failedStep` → raises exactly `unsupported Unicode escape sequence`; after `sanitizeObjectForPostgresql` the row persists and the NUL is gone.
- Unit-level: old truncation → not well-formed; new truncation → well-formed and drops the partial emoji; sanitize repairs a pre-existing lone surrogate + NUL; a valid emoji is preserved untouched.
- `flow_run` jsonb columns audited: only `failedStep` is written here (`pauseMetadata` is deprecated and untouched).
- Lint clean.


abuaboud (approved):
Correct: strips NUL + unpaired surrogates before the Postgres jsonb upsert, and truncateString no longer splits a surrogate pair. LGTM.

PR #14001, opened 2026-06-30
fix(worker): stop retrying permanently-failed jobs (deleted connection / 404 payload)

abuaboud:
## What

Engine failures whose inputs are **permanently gone** never recover on retry:

- a deleted/expired connection — `ConnectionNotFound` / `ConnectionExpired`
- a `404 Not Found` / `410 Gone` on the trigger-payload file — `EngineFileDownloadError`

`completeJob` wrapped **every** `INTERNAL_ERROR` in a plain `Error`, so BullMQ retried per `attempts`/`backoff`. These retries can never succeed — they just burn attempts, clog the `failed` queue, and re-page oncall on each attempt.

## How

Classify these at the single `completeJob` choke point (covers every job type) and move them to failed with `UnrecoverableError`, which BullMQ never retries — the same pattern already used for deferred-failure and invalid-schema jobs in this file.

Transient `5xx` download errors stay retryable: only `404`/`410` count as permanent.

## Evidence

Found via two failed prod jobs:
- `Tx9iuIrJm5yola44BCQHX` (EXECUTE_FLOW) — 70-day-stale job, connection `IedmaJgaUqsVRqwnsYudH` deleted + payload file `iU2y9YpQJjyeUfUiFUDcc` 404'd, retried **5×**.
- (job `2M5W8jqOm9HKsK7pWV8GB` was a separate host-level engine crash — out of scope here.)

## Scope / non-goals

This stops the **retry storm**. It does not reclassify the run's `INTERNAL_ERROR` status to `FAILED` — that's a deeper engine-side change (USER vs ENGINE error typing) intentionally left out of this contained worker fix.

## Test

`isPermanentEngineFailure` unit test covers: deleted connection, expired connection, 404 payload → permanent; 5xx download + generic sandbox crash → still retryable.

### DIFF
--- a/packages/server/api/src/app/workers/job-queue/job-broker.ts
+++ b/packages/server/api/src/app/workers/job-queue/job-broker.ts
@@ -204,7 +204,20 @@ function buildFailedReason(errorMessage: string, logs?: string): string {
     return `${errorMessage}\n${logs}`
 }
 
-export { tryDequeue }
+// Engine failures whose inputs are permanently gone never recover on retry — a deleted/expired
+// connection, or a 404/410 on the trigger-payload file. Retrying them just burns attempts, clogs
+// the failed queue, and re-pages oncall. A transient 5xx file-download error must still retry, so
+// EngineFileDownloadError is only treated as permanent when the status is not-found / gone.
+const PERMANENT_FAILURE_MARKERS = ['ConnectionNotFound', 'ConnectionExpired'] as const
+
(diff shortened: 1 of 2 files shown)


abuaboud:
Superseded — the broker-level UnrecoverableError was a band-aid that left the run as INTERNAL_ERROR. Replacing with a source fix: a deleted/expired trigger-payload file (404/410) is now classified as a USER error and surfaces as a FAILED run, so it never reaches the failed queue or pages oncall.

PR #14002, opened 2026-06-30
feat(chat): onboard free cloud users to managed AI with a $10 first-chat grant

Louai-Zokerburg:
## Why

We opened the AI chat assistant to all cloud **free** users (capped 200-user rollout), but the experience was broken for exactly that audience:

1. A free cloud user's **first screen was a "set up an AI provider" wall** — the auto-created managed *Activepieces* provider was saved with `enabledForChat = false`, so the chat box rendered `SetupRequiredState` and pushed them to `/platform/setup/ai`.
2. Even past the wall, free users had **no usable managed credits**, so chat would fail.

This makes managed AI the default and gifts test credits so a free user can land in chat and immediately play.

## What changed

- **Managed AI is the default chat provider.** The managed *Activepieces* provider is now created with `enabledForChat: true`, so free cloud users see the chat immediately. Creation is guarded so we **never create a second chat provider or override an existing BYO choice**, and the update path still flips it off when a user picks their own provider. Existing platforms are backfilled by the migration (scoped to platforms with no chat provider).
- **One-time $10 credit grant on first chat.** On a user's first message we top up the platform's managed OpenRouter key by a configurable amount (default $10), reusing the same mechanism as the Stripe top-up. Scoped to **free-tier cloud only** (`isNil(plan.licenseKey)`), **awaited before** the credit-exhaustion check so the first message isn't rejected and the key is minted exactly once (no race with the worker). Idempotent via a new `chat_rollout_user.grantedFreeCreditAt` atomic claim (claim → grant → release on failure so a later message retries). Grant failures are logged and non-fatal — the message still proceeds.
- **200-cap auto-close hardened.** `recordChatted` now reports first-chat and sets the monotonic Redis closed flag the moment `count >= cap`, instead of lazily on the next visibility check.

## Config / migration

- **No required config.** New optional env `CLOUD_CHAT_FREE_CREDIT_USD` (default `10`) — unset falls back to the default, zero-setup.
- Relies on existing cloud config: `OPENROUTER_PROVISION_KEY` (gates managed AI / the whole feature), `AP_EDITION=cloud`, `CLOUD_CHAT_ROLLOUT_CAP` (optional, default 200).
- **Migration `1802000000000`** (postgres-only — `chat_rollout_user` is cloud-only, matching the original feature): adds `grantedFreeCreditAt` and backfills `ai_provider.enabledForChat`. Runs on boot.

## Edition safety

Everything new is gated by `isCloud()` / `aiCreditsEnabled()`. On CE/self-host without managed AI, no managed provider is created, the grant is skipped, and the provider wall behaves as before — no setup, no behavior change.

## Notes for reviewers

- The grant tops up the **OpenRouter key limit** directly (reflected in the credits UI via `getUsage`), **not** `includedAiCredits` or an internal ledger. If billing/console reports reconcile against `includedAiCredits`, this promo won't appear there — out of scope here; can be a follow-up.
- First message gets ~1–2s of added latency once per user (the synchronous grant). Subsequent messages are unaffected.
- Soft cap: can overshoot 200 by a handful under concurrent first-messages at the boundary (no lock — acceptable for a beta); never reopens.

## Testing

- `npx turbo run lint --filter=api` → 0 errors.
- TODO before merge: `npm run test-api`, and a manual smoke on cloud — free user → no wall → first message grants $10 (OpenRouter key limit increases) → agent replies → second message does **not** double-grant.

## Related

Console side (Rollout Progress dashboard): activepieces/console#9


xuhongchen521-debug:
/attempt #14002

我来做这个，正在写PR。

abuaboud (approved):

abuaboud (approved):

PR #14004, opened 2026-06-30
fix(engine): route user/config failures to FAILED runs instead of INTERNAL_ERROR

abuaboud:
## Problem

Categorized the **entire** `workerJobs` failed backlog (~4k jobs). Several **user/config** conditions were mis-classified as `INTERNAL_ERROR` — which fails+retries the worker job (futile, since they're deterministic) and **pages oncall** — instead of surfacing as a graceful `FAILED` run. A `FAILED` run completes the BullMQ job (no retry, no page) and shows the user the real error.

The architectural invariant (engine `CLAUDE.md`): a USER/piece-level error must be a FAILED step/run; `INTERNAL_ERROR` is reserved for genuine engine bugs.

## Live mis-classifications fixed (current code)

| Cluster | ~live | Cause | Fix |
|---|---|---|---|
| `EngineRunCallbackError` 400 | ~640 | `return_response` sends string `status` / non-string headers; server schema (`z.number()`/`z.record(...,z.string())`) rejects → ENGINE | **Coerce** `runResponse.status`/`headers` so a valid user response **succeeds** (`requests.ts`) |
| Trigger `run()` plain Error/TypeError (toLocaleLowerCase, chatId, "X required", auth type) | ~280 | non-ENGINE throw in trigger run hook escapes | Catch-all in `flowOperation.execute`: any **non-ENGINE** error from the execution phase → FAILED run |
| Subflow `onStart()` (`callbackUrl`) | ~150 | `onStart` had zero error handling | same catch-all (covers `onStart`) |
| Test-step `ENTITY_NOT_FOUND` | ~90 | `getActionOrThrow` throws plain `ActivepiecesError` pre-guard | same catch-all |
| Variable 404 | ~13 | thrown as ENGINE | new USER `VariableNotFoundError` → FAILED step (`variable-resolver.ts`) |
| Gone trigger-payload file 404/410 | (historical ×550) | `EngineFileDownloadError` ENGINE | USER `EngineFileNotFoundError` → FAILED run (earlier commit) |

The catch-all rethrows genuine **ENGINE** errors unchanged, so real engine bugs still page + retry.

## Historical / not code

Everything else in the backlog — `ConnectionNotFound` ×800, `ConnectionExpired`, `EngineFileDownloadError` ×550 (100% 404), `stack-only` ×1674 — is **historical**: the latest attempt is the now-USER expired-payload 404, so **retrying clears them to FAILED**. `ENOSPC` / `job stalled` / `SANDBOX_INTERNAL_ERROR` / `PieceNotFound` are infra or correctly-ENGINE.

## Tests

- `flow-operation-invariants`: gone trigger-payload file → FAILED run; plain `TypeError` from trigger `run()` → FAILED run; USER ExecutionError → FAILED; ENGINE error still propagates.
- `send-flow-response-request`: `runResponse` coerces string status → number and non-string header → string; still rejects a non-numeric status.

(`@activepieces/core-execution` minor-bumped for the two new exports.)

### DIFF
--- a/docs/install/configure-operate/production-setup.mdx
+++ b/docs/install/configure-operate/production-setup.mdx
@@ -86,6 +86,7 @@ Upgrading breaks nothing: the default `AP_WORKER_CONCURRENCY=5` keeps each conta
 AP_WORKER_CONCURRENCY=1
 AP_REUSE_SANDBOX=true
 AP_EXECUTION_MODE=SANDBOX_CODE_ONLY
+AP_USE_CDN_FOR_BUNDLES=true
 ```
 On [Worker Groups](./worker-groups), keep `AP_EXECUTION_MODE=SANDBOX_PROCESS`; code-only mode is rejected for grouped workers.
 </Step>
--- a/packages/core/execution/package.json
+++ b/packages/core/execution/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/core-execution",
-  "version": "0.5.1",
+  "version": "0.6.0",
   "type": "commonjs",
   "main": "./dist/src/index.js",
   "scripts": {
--- a/packages/core/execution/src/lib/engine/execution-errors.ts
+++ b/packages/core/execution/src/lib/engine/execution-errors.ts
@@ -91,6 +91,12 @@ export class FetchError extends ExecutionError {
     }
 }
 
+export class EngineFileNotFoundError extends ExecutionError {
+    constructor(fileId: string, cause?: unknown) {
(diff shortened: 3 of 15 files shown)


abuaboud:
Added one more prevention: `execute-flow.ts` now **completes** the job (status OK) for a deleted/disabled flow instead of returning INTERNAL_ERROR. That was the largest residual after the first bulk retry (`stack-only` 'Internal error' with empty stderr = a flow disabled/removed while jobs were queued) and a live source of oncall pages.

PR #14007, opened 2026-06-30
fix: wrap draft creation in transaction to prevent orphaned drafts

### LINKED ISSUE #13931: [BUG] [Flows] Orphaned empty draft version left behind when applyOperation fails on a published flow
## Problem

When a user edits a **published** flow (latest version is `LOCKED`), the edit path first creates a new empty DRAFT version, then imports the locked content into it. These are two separate, independently-committed DB writes with **no transaction** around them. If the second step throws (observed in the field as `RangeError: Maximum call stack size exceeded` on deeply nested flows), the empty draft is already committed and stays in the DB forever. The original published version is untouched, but the project accumulates orphaned empty draft versions.

Ref: Pylon #6171

## Root cause

`createNewDraftIfVersionIsPublished` (`packages/server/api/src/app/flows/flow/flow.service.ts:824`) runs two writes in sequence with no transaction:

1. `flowVersionService(log).createEmptyVersion(...)` — `flow-version.service.ts:256-283`. It does `flowVersionRepo().save(flowVersion)` with **no** `entityManager`, so the empty DRAFT row is committed immediately on its own connection. The function does not even accept an `entityManager`.
2. The `applyOperation(IMPORT_FLOW ...)` loop that imports the locked version into the new draft.

The caller, `update()`'s default case (`flow.service.ts:400-415`), has no `transaction()` and no try/catch cleanup. The only `transaction()` in the file is in `updatedPublishedVersionId` (line 445), a different operation. So any throw after step 1 leaves the committed draft behind. The global error handler only logs (the `[errorHandler]` line in the report); nothing deletes the row.

Confirmed in code on both release-0.85.4 and upstream `main` (identical).

## Fix

Make `createEmptyVersion` accept an `entityManager`, and wrap the `createEmptyVersion` + `applyOperation` loop inside `createNewDraftIfVersionIsPublished` in a single `transaction(async (entityManager) => { ... })`, threading the manager through both writes. Mirror the pattern already used by `updatedPublishedVersionId` at `flow.service.ts:445`. On any failure the empty draft then rolls back.

## Out of scope

The stack overflow that triggers this on deeply nested flows is a separate, not-yet-pinned issue (the recursive `FlowAction` zod schema is never parsed at runtime, so the overflow is most likely unbounded JS recursion in the import expansion — needs the full untruncated stack trace + a repro flow to confirm the exact frame). This ticket only covers making draft creation transactional so no orphaned draft is left behind regardless of the failure cause.

## Steps to reproduce

1. Have a published flow whose latest version is `LOCKED`.
2. Trigger any structural edit that fails during the import-into-draft step (e.g. a deeply nested flow that hits the stack overflow).
3. Observed: an empty DRAFT version is committed and remains in the DB; expected: it should roll back.

sourav7274:
## What does this PR do?

Fixes orphaned empty draft flow versions that accumulate in the database when editing a published flow fails mid-import. 

 When a user edits a published (LOCKED) flow, the system creates a new empty DRAFT version and then imports the locked content into it. Previously these were two separate, uncommitted DB writes with no transaction around them. If the import step threw an error (e.g. a stack overflow on a deeply nested flow), the empty DRAFT row was already committed and left permanently in the database.

This fix wraps both writes — the empty draft creation and the entire import operation loop — in a single database transaction inside createNewDraftIfVersionIsPublished. If anything fails mid-import, the transaction rolls back and no orphaned draft is left behind.. 

Changes:

 - flow-version.service.ts — createEmptyVersion now accepts an optional entityManager parameter so it can
      participate in a caller-supplied transaction
 - flow.service.ts — createNewDraftIfVersionIsPublished wraps the full draft creation + import in a 
      transaction(), mirroring the pattern already used by updatedPublishedVersionId

### Relevant User Scenarios

 A user edits a published flow that is deeply nested, triggering a stack overflow during import — previously left an orphaned empty draft in the DB, now rolls back cleanly
Any failure during the import-into-draft step no longer leaves ghost draft versions accumulating per project

Fixes #13931

### DIFF
--- a/.agents/features/flows.md
+++ b/.agents/features/flows.md
@@ -88,6 +88,8 @@ All flow modifications go through `POST /v1/flows/:id` with a `FlowOperationRequ
 - USE_AS_DRAFT → copies published version back to draft for editing
 - Only published flows can be enabled (triggers registered)
 
+**Editing a locked version** (`createNewDraftIfVersionIsPublished`): when any structural edit operation arrives and the latest version is LOCKED, the service creates a new empty DRAFT and imports the locked content into it. Both the empty-draft insert (`createEmptyVersion`) and the import operation loop run inside a single `transaction()` — if the import throws (e.g. stack overflow on a deeply nested flow), the entire transaction rolls back and no orphaned empty draft is left in the DB.
+
 ## Publishing Side Effects
 
 When LOCK_AND_PUBLISH or CHANGE_STATUS to ENABLED:
--- a/packages/server/api/src/app/flows/flow-version/flow-version.service.ts
+++ b/packages/server/api/src/app/flows/flow-version/flow-version.service.ts
@@ -260,6 +260,7 @@ export const flowVersionService = (log: FastifyBaseLogger) => ({
             notes: Note[]
             schemaVersion: string | undefined | null
         },
+        entityManager?: EntityManager,
     ): Promise<FlowVersion> {
         const flowVersion: NewFlowVersion = {
             id: apId(),
@@ -280,7 +281,7 @@ export const flowVersionService = (log: FastifyBaseLogger) => ({
             state: FlowVersionState.DRAFT,
             notes: request.notes,
         }
-        return flowVersionRepo().save(flowVersion)
+        return flowVersionRepo(entityManager).save(flowVersion)
     },
     removeConnectionsAndSampleDataFromFlowVersion(
         flowVersion: FlowVersion,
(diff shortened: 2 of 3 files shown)


majewskibartosz:
@sourav7274 hey thanks for contribution, we''ll review this next week.

abuaboud:

sourav7274:
Hi, thank you for informing, I had fun working with issues.

PR #14015, opened 2026-07-01
fix(sandbox): surface real crash diagnostics instead of empty error on fatal engine exit

amrdb:
## Problem

A failed production webhook run (Google Sheets `new_row_added` trigger hook) surfaced only as an opaque, undiagnosable error:

```
SANDBOX_INTERNAL_ERROR: Worker exited with code 1 and signal null standardOutput= standardError=
```

The engine runs as a **forked child process** and the sandbox collects its stdout/stderr **only over the RPC socket** (`EngineStdout`/`EngineStderr` messages). When the engine dies abruptly — an uncaught exception (`exit 1`) or a V8 heap-OOM abort — it writes the real diagnostic to its **native OS stderr** and never sends a socket message. So:

- `stdError` stays empty → the job error / BullMQ `failedReason` carries no cause.
- The OOM heuristic in `handleProcessExit` greps `stdError` for `"JavaScript heap out of memory"`, which can never match → memory crashes get misclassified as generic `SANDBOX_INTERNAL_ERROR`.

Root cause: the child's **native stderr pipe is never drained**. (For `fork.ts` it wasn't even a pipe — `silent` defaulted to false, so `child.stderr` was `null`.)

## Fix

- **`fork.ts`** — fork with `silent: true` so the child's stdout/stderr are pipes the worker can read.
- **`sandbox.ts`** — drain the child's native stdout/stderr into a bounded (8 KB tail) buffer and fold it into `handleProcessExit`, so crash output lands in the job error and reaches the RAM / log-size classifiers.
- **`isolate.ts`** — remove its private stdout/stderr forwarding now that capture is centralized in `sandbox.ts` (avoids double-writing to the worker console).

## Effect

| Situation | Before | After |
|---|---|---|
| Engine uncaught exception (`code 1`) | `standardError=` empty, opaque `SANDBOX_INTERNAL_ERROR` | error carries the real stack trace |
| Heap OOM exiting `code 1 / null` | misclassified `SANDBOX_INTERNAL_ERROR` | `SANDBOX_MEMORY_ISSUE` |
| Heap OOM aborting (`SIGABRT`) | `SANDBOX_MEMORY_ISSUE` | unchanged, now with diagnostic text attached |
| Normal run | logs from socket | unchanged (native buffer only used on crash path) |

## Verification

- Reproduced the exact production signature (`code 1 / signal null`, empty stderr) with a forked child; confirmed the fixed wiring captures the crash text.
- Added 2 regression tests in `sandbox.test.ts`: native stderr is attached to the error, and an OOM exiting `code 1/null` is now classified `SANDBOX_MEMORY_ISSUE`. **Both fail without the fix**, all **42** sandbox tests pass with it.
- `tsc` build + eslint clean.

### DIFF
--- a/packages/server/sandbox/src/lib/sandbox/fork.ts
+++ b/packages/server/sandbox/src/lib/sandbox/fork.ts
@@ -5,6 +5,11 @@ export function simpleProcess(enginePath: string, codeDirectory: string): Sandbo
     return {
         create: async (params) => {
             return fork(enginePath, [], {
+                // Pipe the child's stdout/stderr instead of inheriting them, so the sandbox can
+                // capture native runtime output (uncaught-exception stacks, V8 "JavaScript heap out
+                // of memory" aborts) that never travels over the RPC socket. Without this, a hard
+                // engine crash surfaces only as an opaque "Worker exited with code 1 ... standardError=".
+                silent: true,
                 execArgv: [
                     // IMPORTANT DO NOT REMOVE THIS ARGUMENT: https://github.com/laverdet/isolated-vm/issues/424
                     '--no-node-snapshot',
--- a/packages/server/sandbox/src/lib/sandbox/isolate.ts
+++ b/packages/server/sandbox/src/lib/sandbox/isolate.ts
@@ -127,13 +127,6 @@ export function isolateProcess(log: SandboxLogger, enginePath: string, _codeDire
                 shell: false,
             })
 
-            child.stdout?.on('data', (data: Buffer) => {
-                process.stdout.write(data)
-            })
-            child.stderr?.on('data', (data: Buffer) => {
-                process.stderr.write(data)
-            })
-
             return child
         },
     }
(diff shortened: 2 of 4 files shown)


amrdb:
Thanks @greptile-apps — all three points addressed in e7d7c6be6d:

**P1 — `exit` vs `close` race (correct, thanks for catching it).** Switched both handlers to the `close` event, which Node guarantees fires only after all stdio streams are drained, so `nativeStdError` is complete when `handleProcessExit` classifies the crash:
- `start()` `exitPromise`: `once('exit')` → `once('close')` (+ teardown `removeAllListeners('close')`)
- `execute()`: `on('exit')` → `on('close')` (+ `finally` teardown `removeAllListeners('close')`)

Note the unit tests can't reproduce the real pipe race (the mock child is a bare `EventEmitter` with no real stdio), so the guarantee here comes from using `close`; the tests were updated to emit `close` to match the production listener.

**P2 — buffers accumulate across re-executions in reuse mode (valid).** `nativeStdOut`/`nativeStdError` are now reset at the start of every `execute()`, so a reused (`AP_REUSE_SANDBOX`) sandbox no longer prefixes a run's crash output with native output from previous healthy executions.

**P2 — PR title.** Updated to a user-facing phrasing.

All 82 sandbox tests pass; lint + `tsc` clean.

abuaboud (approved):

PR #14026, opened 2026-07-01
fix(flow-run): serve a POST confirmation page for approval resume links

### LINKED ISSUE #14024: [BUG] Email security scanners pre-fetch single-use approval/resume link → user's real click shows "This link has expired"
## Problem

Approval-by-email flows (Wait for Approval / Create Approval Links sent via SMTP) show the user `{"message":"This link has expired. The action may have already been processed."}` even though the approval succeeds and the flow completes normally. The user reports clicking the link seconds after receiving the email.

This is a regression from the resume rewrite: before 0.82.0 the resume endpoint returned the success message on every open, so an automated pre-open was invisible. The 0.82.0 waitpoint rewrite made resume links strictly single-use, so the second open now returns the expired message. Present since 0.82.0, verified through 0.86.0.

The link is a bare GET. Email security scanners (Microsoft Safe Links, Mimecast, Proofpoint) pre-fetch the URL to scan it before delivery. That first GET consumes the waitpoint and resumes the flow; the human's later click is the second GET → stale → "expired" message. Confirmed by the reporter: pasting the link directly (bypassing the scanner) shows the success message.

Ref: Pylon [#5253](<https://github.com/activepieces/activepieces/issues/5253>)

## Root cause

`packages/server/api/src/app/flows/flow-run/waitpoint/waitpoint-service.ts` — `handleResumeSignal()` deletes the waitpoint on the first resume (`repo.delete({ id: found.id })`) and returns `true`. The second call finds nothing → returns `false` → `resume-controller.ts` (`handleAsyncResume`) sends `"This link has expired..."`. Any automated GET (scanner prefetch) is indistinguishable from a human click and consumes the link. Same effect on the deprecated V0 path (`resume-service.ts` `legacyResume`: `flowRun.status !== PAUSED → stale`).

## Proposed

Stop an automated prefetch GET from consuming the waitpoint. Options, most robust first:

* Serve a lightweight confirmation interstitial on the GET (a page with an Approve/Disapprove button); the actual resume fires on the button action (POST). Scanner GETs then no longer consume the link.
* Or make the second open idempotent-friendly: when the run already resumed via this waitpoint, show the recorded/success message instead of "expired."
* Prefetch heuristics (HEAD, `X-Purpose`/scanner UAs) are fragile and shouldn't be the primary fix.

Relates to the same endpoint as activepieces/activepieces#13686 (waitpoint resume hardening).

othmanemad:
## What does this PR do?

Approval pieces (Gmail, Outlook, Telegram, Discord, Microsoft Teams) embedded **two bare `GET` resume links** — one for Approve, one for Disapprove — directly in the email/message. Resuming a paused run is a single-use action: the first signal consumes the waitpoint and resumes; any later signal is stale ("This link has expired").

Email security scanners (Microsoft Safe Links, Mimecast, Proofpoint) **prefetch every URL with a `GET`** before delivery. That prefetch is indistinguishable from a human click, so it silently consumes the waitpoint — and because both links get prefetched, the run resumes with an arbitrary outcome before the human ever sees the email. The human's later click then lands on the stale link and sees "expired," even though the flow already resumed (Pylon #5253; a regression since the 0.82.0 waitpoint rewrite).

This PR fixes it by moving resume behind a deliberate `POST`:

- **New scanner-safe route** `GET|HEAD|POST /:id/waitpoints/:waitpointId/confirm`:
  - `GET`/`HEAD` **never** consumes the waitpoint — it serves a white-labeled **Resume Confirmation Page** with Approve/Disapprove buttons the human must click.
  - Only the resulting `POST` consumes the waitpoint and resumes the run.
  - On open, the page reads the waitpoint from the DB and shows an **"already responded"** state if the run has already moved on (instead of consuming or erroring).
  - The `POST` response is **content-negotiated**: `Accept: text/html` → a branded HTML result page; otherwise → the existing `{ message }` JSON, so programmatic callers are unaffected.
- **Existing resume routes are left untouched** (now marked `@deprecated`) so approval emails **already sitting in inboxes keep working**. They still resume on a bare `GET` and carry the known prefetch weakness — an accepted cost of backward compatibility; they age out over time.
- **Approval pieces now send a single "Review & Respond" button** linking to `${waitpoint.resumeUrl}/confirm`, replacing the two bare links, across every channel whose buttons are browser `url:` links — Gmail, Outlook, Telegram, Discord, Microsoft Teams (channel + direct message). Channel-specific context (e.g. Telegram's `chat_id`) is appended to the `/confirm` URL and preserved through to the resuming `POST`.
- **Slack is intentionally excluded** — its buttons are interactive Slack components (`action_id`/`value`) that resume via a **server-side `POST`** from Slack's webhook, so they are not browser-GET-prefetchable and gain nothing from a confirmation page.
- Page branding is resolved through a **CE-safe `hooksFactory` theme hook** (`resume-page-hooks.ts`): CE returns `defaultTheme`; EE/Cloud `.set()` it to `appearanceHelper.getTheme`, so the CE controller never imports the EE helper.

See ADR `docs/adr/0005-resume-links-require-post-confirmation.md` for the full decision record.

image
image
image
image

### Relevant User Scenarios

- **New approval email is scanned before delivery** — Safe Links/Mimecast/Proofpoint prefetch the `/confirm` link with a `GET`; the confirmation page renders but the waitpoint is **not** consumed and the flow stays paused. The recipient later opens the email, clicks Approve or Disapprove, and their single `POST` decides the outcome.
- **Recipient approves/disapproves** — clicking a button on the confirmation page `POST`s back, resumes the run with the chosen decision, and shows a branded "You approved/disapproved this request" result page.
- **Recipient reopens the link after the run already resumed** — the page shows an "Already responded" state instead of erroring or double-resuming.
- **Old approval email already in an inbox** (sent before this rollout) — its links point at the deprecated route and still resume on click, so no in-flight approvals break.
- **Programmatic / API caller** hitting `/confirm` with a non-HTML `Accept` — still receives the existing `{ message }` JSON contract.
- **Slack approval** — unchanged; continues to resume via Slack's server-side webhook `POST`.

Fixes [Issue](https://github.com/activepieces/activepieces/issues/14024)


AhmadTash:
Heads-up for the deprecation tracking: the Slack approval actions still build their resume links via `waitpoint.buildResumeUrl(...)`, which resolves to the `/:id/waitpoints/:waitpointId` route this PR marks `@deprecated` — they don't go through the new `/confirm` route:

- `packages/pieces/community/slack/src/lib/actions/request-approval-message.ts:55,58`
- `packages/pieces/community/slack/src/lib/actions/request-approval-direct-message.ts:53,56`
- `packages/pieces/community/slack/src/lib/common/request-action.ts:45`

Each `buildResumeUrl({ queryParams })` (no `sync`) produces `…/v1/flow-runs/{id}/waitpoints/{waitpointId}?action=…` (`piece-executor.ts:295` + `waitpoint-controller.ts:22-23`), i.e. the deprecated endpoint. Slack resolves it with a server-side interactivity POST rather than a browser GET, so it's not exposed to the prefetch issue this PR addresses — just noting that it stays on the deprecated route so we can track it for the eventual cleanup (alongside `create-approval-link`, which uses the same builder).

AhmadTash (approved):

PR #14043, opened 2026-07-02
feat(web): land cloud users on chat while the rollout cap is open

Louai-Zokerburg:
## What does this PR do?

Makes the AI chat the default landing route for cloud users **while the 200-user rollout cap is still open**, then cleanly steps aside once the cap closes.

Cloud users now land on `/chat` instead of `/automations` as long as chat is enabled for them. The gate reuses the already-resolved `platform.plan.chatEnabled` flag (computed server-side in `platform.controller.ts` via `resolveChatEnabledForUser`), so no new query, flag, or env var is introduced.

- `determineDefaultRoute` gains an optional `chatEnabled` param; when true it returns `CHAT_ROUTE` (`/chat`), otherwise it falls through to the existing permission-based routing (`/automations` → `/runs` → `/settings`).
- The landing guard (`guards/default-route.tsx`) and the two sidebar home-link callers pass `platform.plan.chatEnabled` so the logo/home target stays consistent with where users land.
- `platform.plan.chatEnabled` is the fully-resolved signal: for cloud it is `true` while the rollout is open (chatted-user count `< 200`) **or** the user is grandfathered (already chatted), and flips to `false` once the cap latches closed — at which point new users fall back to `/automations`.
- The embed caller is untouched: `chatEnabled` is optional/falsy and embedded users never get chat.

Also included: a one-line fix adding the missing `loadSubsetOptions: {}` to the platform-projects query `meta` (`by-project-view.tsx`), matching every other `showErrorDialog` call site — resolves a `tsc` overload error surfaced on dev serve.

### Relevant User Scenarios

- A new cloud free user signs in during the beta → lands directly on the AI chat instead of the automations dashboard.
- The 200th user chats → the cap latches closed → subsequent new users land on `/automations`; users who already chatted keep landing on chat (grandfathered).


Louai-Zokerburg:
On the admin question: including admins is intentional. `chatEnabled` is resolved per-user server-side (`resolveChatEnabledForUser`), and on cloud the eligible user is the admin of their own platform — so gating chat-first behind a non-admin role check would exclude nearly every cloud user. The precedence is now documented with a comment in `route-utils.ts` (commit 122dc38430).

Louai-Zokerburg:
@greptileai the way the route it is now is fine that how I want it

Louai-Zokerburg:
@greptileai no dont remember this, just redo the review or update the score or what ever

MrChaker (approved):

PR #14051, opened 2026-07-04
fix: user deletion no longer trips fk_project_owner_id on first attempt

### LINKED ISSUE #14020: [BUG] Deleting a platform user fails on first attempt (FK violation), succeeds on retry — self-hosted
## Summary

On self-hosted, deleting a user from Platform Admin fails on the first attempt with a generic "Something went wrong" toast, then succeeds when the same delete is repeated. Reported by a customer on 0.85.0 and reproduced locally on the current build.

## Root cause

`userService.delete` (self-hosted path) runs three steps in sequence, without a transaction, in `packages/server/api/src/app/user/user-service.ts:154-163`:

1. Soft-deletes the user's personal project via `platformProjectService.deletePersonalProjectForUser` → `markForDeletion` (`packages/server/api/src/app/ee/projects/platform-project-service.ts:219-251`). This `softDelete`s the project row (it still physically exists) and queues an **async** `HARD_DELETE_PROJECT` system job.
2. Immediately hard-deletes the user row: `userRepo().delete()` (line 160).

The `project.ownerId` foreign key `fk_project_owner_id` is defined `ON DELETE NO ACTION` (`packages/server/api/src/app/project/project-entity.ts:109-116`; migration `1676238396411-initialize-schema.ts:134`). At step 2 the soft-deleted project still references the user, so Postgres rejects the delete with error `23503` → 500.

The retry succeeds because by then the async `HARD_DELETE_PROJECT` job has removed the project row, so the FK no longer blocks.

The frontend delete-user mutation has no `onError` handler, so the specific constraint violation is hidden behind the generic "Something went wrong" toast.

## Cloud is unaffected

The controller routes Cloud deletions to `removeFromPlatform` (detach via `update`, sets `platformId = null`) instead of a hard delete (`packages/server/api/src/app/user/platform/platform-user-controller.ts:37-56`). The user row is never deleted, so the FK never fires.

## Reproduction (verified locally, EE self-hosted)

1. Invite a second user (any role), accept the invite so a personal project is provisioned.
2. As the platform owner, go to Platform Admin → Users and delete that user.
3. First `DELETE /api/v1/users/:id` → 500, body `{ "code": "23503", "message": "update or delete on table \"user\" violates foreign key constraint \"fk_project_owner_id\" on table \"project\"" }`.
4. Repeat the delete → 204 No Content.

Server log at the moment of failure:

```
code=23503 status=500 QueryFailedError: update or delete on table "user" violates foreign key constraint "fk_project_owner_id" on table "project"
    at platform-user-controller.ts:49
```

## Affected versions

* Self-hosted from **v0.74.4** onward (commit `2f56a11`, PR #10688, "delete user should also delete the personal project" — introduced the personal-project deletion on user delete).
* v0.83.0 onward (incl. 0.85.0 and current HEAD): **self-hosted only**; Cloud made safe by the `removeFromPlatform` split (commit `a34da17`, PR [#13000](<https://github.com/activepieces/activepieces/issues/13000>)).
* Still present on current `main`. No released version fixes it.

## Proposed fix

* Delete (or reassign) the owned personal project **synchronously** before deleting the user row, instead of relying on the queued async `HARD_DELETE_PROJECT` job. That closes the race so the first delete succeeds. Alternatively wrap both in a single transaction with the project removed first.
* Add an `onError` handler to the frontend delete-user mutation so real errors surface instead of the generic "Something went wrong."

thribhuvan003:
## What does this PR do?

Fixes the self-hosted "delete user fails on first attempt, succeeds on retry" bug (#14020).

As diagnosed in the issue: `userService.delete` soft-deletes the personal project and queues the **async** `HARD_DELETE_PROJECT` job, then immediately hard-deletes the user row. Because the soft-deleted project row still exists and `fk_project_owner_id` is `ON DELETE NO ACTION`, Postgres rejects the user delete with `23503` → generic 500. The retry succeeds only because the async job has removed the project row by then.

The fix keeps the async hard-delete architecture untouched and just removes the FK obstacle at delete time:

1. New `platformProjectService.reassignSoftDeletedPersonalProjects` reassigns the user's **soft-deleted** personal project rows (`type = PERSONAL`, `deleted IS NOT NULL` — live rows are never touched) to the platform owner.
2. `userService.delete` calls it after `deletePersonalProjectForUser` and before `userRepo().delete`, so the user row no longer blocks on the FK.

Why this shape:

- Running the hard-delete synchronously was rejected: the `HARD_DELETE_PROJECT` handler does flow side-effect cleanup with retries (25 attempts) and is intentionally async — not something to run inside the HTTP request.
- The reassigned row is invisible (soft-deleted) and already queued for hard deletion by `jobId: hard-delete-project-{id}`; the job deletes by `{ id, platformId }`, so the owner change does not affect it.
- The platform owner is guaranteed to exist and can never be deleted (`assertNotPlatformOwner`), so the FK target is always safe. `idx_project_owner_id` is non-unique, so the owner temporarily "owning" a second (deleted) personal project row violates nothing.
- Cloud is unaffected: it routes to `removeFromPlatform`, which never deletes the user row.

### Relevant User Scenarios

- Platform Admin → Users → Delete on self-hosted EE (customer-reported on 0.85.0, reproduced on current build per the issue): first attempt now succeeds instead of returning "Something went wrong".

Fixes #14020

### DIFF
--- a/packages/server/api/src/app/ee/projects/platform-project-service.ts
+++ b/packages/server/api/src/app/ee/projects/platform-project-service.ts
@@ -2,7 +2,7 @@ import { ActivepiecesError, apId, Cursor, ErrorCode, isNil, Metadata, PlatformId
 import { apDayjs } from '@activepieces/server-utils'
 import { AppConnectionScope, PiecesFilterType, PrincipalType, Project, ProjectType, ProjectWithLimits, TeamProjectsLimit, UpdateProjectPlatformRequest } from '@activepieces/shared'
 import { FastifyBaseLogger } from 'fastify'
-import { ArrayContains, Equal, ILike, In, IsNull } from 'typeorm'
+import { ArrayContains, Equal, ILike, In, IsNull, Not } from 'typeorm'
 import { appConnectionsRepo } from '../../app-connection/app-connection-service/app-connection-service'
 import { repoFactory } from '../../core/db/repo-factory'
 import { transaction } from '../../core/db/transaction'
@@ -217,6 +217,17 @@ export const platformProjectService = (log: FastifyBaseLogger) => ({
         }
     },
 
+    async reassignSoftDeletedPersonalProjects({ fromUserId, toUserId, platformId }: ReassignSoftDeletedPersonalProjectsParams): Promise<void> {
+        await projectRepo().update({
+            ownerId: fromUserId,
+            platformId,
+            type: ProjectType.PERSONAL,
+            deleted: Not(IsNull()),
+        }, {
+            ownerId: toUserId,
+        })
+    },
+
     async markForDeletion({ id, platformId }: DeleteProjectParams): Promise<void> {
         const result = await projectRepo().softDelete({ id, platformId })
         if (result.affected === 0) {
@@ -347,6 +358,12 @@ type GetAllForParamsAndUser = {
     types?: ProjectType[]
     isPrivileged: boolean
     principalType?: PrincipalType
+}
+
+type ReassignSoftDeletedPersonalProjectsParams = {
+    fromUserId: UserId
+    toUserId: UserId
+    platformId: PlatformId
 }
 
 type DeletePersonalProjectForUserParams = {
--- a/packages/server/api/src/app/user/user-service.ts
+++ b/packages/server/api/src/app/user/user-service.ts
@@ -157,6 +157,17 @@ export const userService = (log: FastifyBaseLogger) => ({
             userId: id,
             platformId,
         })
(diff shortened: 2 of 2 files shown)


thribhuvan003:
the reassignment is idempotent (it filters on fromUserId, so a retry after a failed delete is a no-op and the delete just runs again) and the row it touches is a soft-deleted project already queued for hard deletion, so transient ownership doesn't surface anywhere. happy to wrap both writes in a transaction if you'd prefer it atomic.

PR #14053, opened 2026-07-04
fix(users): delete platform users successfully on first attempt

### LINKED ISSUE #14020: [BUG] Deleting a platform user fails on first attempt (FK violation), succeeds on retry — self-hosted

sourav7274:
- `platform-project-jobs.ts` now exposes `deleteProjectFlowsAndConnections` (flow/connection cleanup with side effects — S3 files, cache invalidation) and `deleteProjectRow` (the actual project DB row delete, transaction-scoped) as separate steps. `hardDeleteProject`, used by the existing queued `HARD_DELETE_PROJECT` job, composes them exactly as before — that path is unchanged, still resumable via `preDeletedFlowIds`.
- `userService.delete` now calls `deleteProjectFlowsAndConnections` first (non-transactional side effects, same as the queued job), then opens **one transaction** that deletes the project row and the user row together via `deleteProjectRow` + `userRepo(entityManager).delete()`. Either both succeed or both roll back — no window where the project is gone but the user delete can still fail.

Verified locally: invited a second user, accepted the invite (provisions personal project), deleted via Platform Admin — succeeds on the **first** attempt (204), not the second. Confirmed via API too (`DELETE /api/v1/users/:id` → 204, follow-up `GET` → 404).

### Relevant User Scenarios

Fixes #14020

### DIFF
--- a/.agents/features/ee-projects.md
+++ b/.agents/features/ee-projects.md
@@ -5,6 +5,7 @@ The EE Projects module adds team collaboration, role-based access control (RBAC)
 
 ## Key Files
 - `packages/server/api/src/app/ee/projects/` — platform project service, RBAC enforcement
+- `packages/server/api/src/app/ee/projects/platform-project-jobs.ts` — `hardDeleteProject()`, queued and synchronous hard-delete of a project's flows/connections/row
 - `packages/server/api/src/app/ee/project-members/` — member CRUD, role lookup
 - `packages/server/api/src/app/ee/project-role/` — default and custom role management
 - `packages/server/api/src/app/ee/project-release/` — release creation, diff, apply
@@ -101,6 +102,12 @@ The EE Projects module adds team collaboration, role-based access control (RBAC)
 - `NONE`: All pieces available
 - `ALLOWED`: Only pieces in `pieces[]` array visible
 - Platform-level filtering (`FilteredPieceBehavior.ALLOWED/BLOCKED`) applied first, then project-level
+
+## Project Hard Deletion
+
+`hardDeleteProject()` (`platform-project-jobs.ts`) deletes a project's flows (with side-effect cleanup), app connections, and the project row itself in a transaction. Two callers:
+- `platformProjectBackgroundJobs.hardDeleteProjectHandler` — the queued `HARD_DELETE_PROJECT` system job, used after `markForDeletion()` soft-deletes a project (normal project deletion, platform deletion, SCIM group deletion). Resumable via `preDeletedFlowIds` if a retry is needed.
+- `userService.delete` (self-hosted user deletion) — calls it synchronously, not via the queue, because the user row can't be deleted while their personal project still exists (`fk_project_owner_id`, `ON DELETE NO ACTION`). See the Users feature doc for details.
 
 ## Platform Project Service
 
--- a/.agents/features/users.md
+++ b/.agents/features/users.md
@@ -92,3 +92,7 @@ Manages user identity, platform membership, roles, session security, and a gamif
 - `GET /v1/users/me` — get current user with identity info
 - `POST /v1/users/me` — update profile (firstName, lastName, profilePicture)
 - User CRUD managed via platform admin endpoints (EE): list users, update role/status, delete user
+
+## User Deletion (self-hosted)
+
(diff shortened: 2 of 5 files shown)


sourav7274:
Re the flow-cleanup-before-transaction concern (confidence 4/5): you're right that if the final transaction fails, the personal project's flow/run data is already gone while the project and user rows remain. I looked into flipping the order (transactional project+user delete first, external cleanup after) to close this fully, and found why the code doesn't already do that:

`project`, `flow`, `flow_run`, `app_connection`, and related tables all have `ON DELETE CASCADE` from `project.id` — so deleting the project row *would* correctly cascade-remove everything in the database automatically. But `batchDeleteByFlowId` deletes `flow_run`/`waitpoint` rows in batches (`BATCH_SIZE` loop) specifically to avoid a single massive cascading DELETE holding a lock across those tables — which can be huge for an active project with a long run history. That's almost certainly why the original queued-job code manually pre-deletes flows before touching the project row, rather than relying on cascade.

So the real choice here is:
- **Current shape**: flow cleanup (batched, lock-friendly) runs first; if the final transaction (connections + project + user, 3 simple row deletes) then fails, flow data is already gone but the project/user rows remain. This is a real gap, but it turns a bug that fired on **every single delete** into a failure mode that requires the final transaction — which only touches already-verified rows — to fail for an unrelated reason.
- **Reordered**: transactional project+user delete first (relying on cascade for flow/run cleanup), external side-effects (trigger disable, sample-data files) run after. Closes this gap fully, but reintroduces the exact cascade-lock risk the batched deletion was built to avoid, for personal projects with large run histories.

I don't think there's a way to get both without either accepting the lock risk or reintroducing the async-queue pattern (which is what caused the original bug). Given that, I kept the current order. Happy to take the reorder if you'd rather accept the performance tradeoff — let me know which you'd prefer and I'll make the change.

majewskibartosz:
@sourav7274 hi we'll review this

PR #14063, opened 2026-07-05
fix: respect patch versions in project release diff engine

AhmadTash:
## The Bug

`isSameVersion()` in `flow-diff.service.ts` was only checking `major.minor` for versions <1.0.0, ignoring patch differences entirely. So `0.3.13` and `0.3.14` were considered identical, causing the release pipeline to silently skip piece version updates.

### Symptoms
| Symptom | Why it happens |
|---|---|
| Git says 0.3.14, prod stays 0.3.13 | Diff engine thinks 0.3.13 = 0.3.14 → produces no UPDATE_FLOW |
| Manual force-upgrade gets overwritten | Saved release snapshot still has old version |
| Git deploy of unrelated flow cascades | Re-imports old snapshot state |

## The Fix

**`flow-diff.service.ts`** — Simplified `isSameVersion()` to compare exact versions via `flowPieceUtil.getExactVersion()` (already imported and used for cleaning), removing the semver-based major-only comparison. Also removed the now-unused `semver` import.

**`flow-diff.test.ts`** — Updated the test that previously pinned the buggy behaviour (asserted patch version differences should be ignored) to instead assert that different patch versions correctly produce an UPDATE_FLOW diff.


AhmadTash:
@greptileai review this

AhmadTash:
@greptile review

amrdb (approved):

PR #14068, opened 2026-07-06
fix(web): variable key names now display exactly as defined in the output viewer

AhmadTash:
## Summary
- The step-output "friendly view" and the mention/data-selector variable-picker tree ran every rendered object key through a `titleCase` humanizer (splits camelCase, capitalizes words), including raw dynamic keys that have no piece output schema — e.g. a Code step's own return object.
- This turned user-chosen variable names like `varOrigen` / `varDestino` into `Var Origen` / `Var Destino`, which is misleading since those are literal keys the user must reference in later steps.
- `titleCase` is now only used for its one legitimate purpose — a label fallback for schema-described piece fields that lack an explicit `label` — and every undescribed/generic/schemaless key renders verbatim: `output-generic-field-list.tsx`, `output-array-list.tsx`, `output-field-row.tsx` (generic-object drill-down), `output-table-view.tsx` (generic table columns), `shared-value-rendering.tsx` (`ValueRow`), and the data-selector's `utils-schema.ts`.

## Test plan
- [x] `tsc --noEmit` on `packages/web` — clean
- [x] `eslint` on all changed files with the project's config — clean, no unused imports
- [x] `vitest run src/app/builder/data-selector/utils-schema.test.ts` — 5/5 passing, including a new regression test asserting `varOrigen`/`varDestino` render verbatim instead of humanized
- [x] `turbo run lint` (lint-core + lint-affected) — 100% clean
- [x] Manually verify in the builder: add a Code step returning `{ varOrigen: 'a', varDestino: 'b' }`, confirm the "Friendly" output view and the variable-picker mention tree both show the keys verbatim


AhmadTash:
@greptile review

Louai-Zokerburg (approved):

PR #14074, opened 2026-07-06
Feature/embed piece locking

jasonfill:
## What does this PR do?

Adds two per-flow **embed publish constraints** for the embedded builder, stored on `flow.metadata.embedConstraints` (no DB migration) and enforced both in the UI and unbypassably on the server:

1. **Required piece to publish** — `requiredPieceNames: string[]`. A flow cannot be published unless at least one non-skipped step (action or trigger) uses one of the listed pieces. Subflows (`@activepieces/piece-subflows` trigger) are exempt.
2. **Trigger lock** — `triggerLock: 'none' | 'locked' | 'frozen'`. `locked` blocks swapping/clearing the trigger piece (inputs stay editable); `frozen` makes the whole trigger read-only.

### How it works

- **Shared helper** (`embedConstraintsUtil` in `core-execution`, re-exported via `@activepieces/shared`): `getEmbedConstraints`, `flowSatisfiesRequiredPiece` (ignores skipped steps, exempts subflows), `isSubflow` — single source of truth for UI + server.
- **Server enforcement** (`flow.service.ts`):
  - Publish guard in `updatedPublishedVersionId()` → `409` when the required piece is missing.
  - `assertTriggerLockAllowsUpdate` guard on `UPDATE_TRIGGER` **and** `IMPORT_FLOW` → `409` on illegal trigger changes.
  - `UPDATE_METADATA` now preserves the existing `embedConstraints` key so the constraint cannot be wiped from a builder session.
- **UI**: Publish button greys out with a reason tooltip when the required piece is missing; `frozen` reuses the existing `readonly` path; `locked`/`frozen` hides the trigger's Replace affordance.

### Hardening in this PR (closes two bypasses of the "unbypassable" guarantee)

- **`UPDATE_METADATA` wipe** — `embedConstraints` live in `flow.metadata`, and `UPDATE_METADATA` did a full replace; a builder session could POST `{ metadata: {} }` to drop the constraint, then publish freely. Now the existing `embedConstraints` value is always carried over.
- **`IMPORT_FLOW` trigger swap** — the trigger-lock guard only ran on `UPDATE_TRIGGER`; `IMPORT_FLOW` fell through unguarded and could replace a locked/frozen trigger. The guard now runs for `IMPORT_FLOW` too (internal draft-import via `applyOperation` is unaffected).

### Tests

- Unit tests for `flowSatisfiesRequiredPiece` (no constraint, empty list, trigger/action match, skipped step, no match, any-of set, subflow exemption, metadata parsing).
- API integration tests: required-piece missing/present/subflow-exempt; locked piece-swap; frozen edit; locked input-only edit; **UPDATE_METADATA wipe cannot bypass publish guard**; **IMPORT_FLOW cannot swap a locked trigger**.

Constraints are set by the vendor at flow-creation time via the API (`CreateFlowRequest.metadata`); the embedded end-user cannot change them.

Fixes # (n/a)


jasonfill:
Closing this as there are other commits in here that should not be.

PR #14080, opened 2026-07-06
fix(connections): new connection from a step no longer inherits the selected connection's global scope

AhmadTash:
## Problem

Creating a **new** connection from inside a flow step silently inherited the scope of the connection currently selected in that step. If a **global (PLATFORM-scoped)** connection was selected, the create/reconnect dialog ran in global mode, causing two failure modes:

- **Platform admins** — the new connection was created as `PLATFORM` scope instead of `PROJECT` (`POST /v1/global-connections`), showing the globe icon and manageable only from the global-connections admin page.
- **Non-admins / embedded users** — a hard **403** before submit: the name-uniqueness check called the platform-admin-only `GET /v1/global-connections`, blocking connection creation entirely from the builder. Reported by a paid embed customer (Pylon #5297).

## Root cause

The dialog's global mode was derived from the **currently selected** connection's scope. Clicking "+ New connection" reset `reconnectConnection` to `null` but not that flag, so the create path ran in global mode.

## Fix (frontend-only)

Derive the dialog's global mode from the **operation**, not the selected connection — it is global only when *reconnecting* an existing global connection:

- `connection-select.tsx` (builder step settings): split the overloaded `isGlobalConnection` into `isSelectedConnectionGlobal` (kept for the status/reconnect-button gating) and `isReconnectingGlobalConnection = reconnectConnection?.scope === PLATFORM`, which is what the dialog now receives. Creating always defaults to `PROJECT` scope (routes through `appConnectionsApi`, no admin-only calls); reconnecting a global connection still routes global and keeps `PLATFORM` scope.

- `connection-picker-card.tsx` (AI-chat picker): fixes the **inverse mirror** of the same root cause found during review. It hardcoded `isGlobalConnection={false}` while reusing the same dialog for reconnect, so reconnecting an *expired global* connection silently wrote a project-scoped duplicate and left the global one broken (no 403 — it hit the project endpoint a member can reach). Now derives scope the same way.

## Verification

- [x] `eslint` clean on both changed files
- [x] Type-safety confirmed via before/after `tsc` diff (zero new errors; remaining `tsc` output is a pre-existing worktree module-resolution artifact, identical before and after)
- [x] 5-agent adversarial review: both failure modes fixed; admin **reconnect** of a global connection still routes global and keeps `PLATFORM`; the global-connections admin page and the other 7 dialog call sites are unaffected; non-admins are structurally blocked from the global path

## Test plan

- [ ] As a platform **MEMBER** (or embedded user), on a step with a global connection selected: "+ New connection" → create succeeds via `POST /v1/app-connections`, **no** call to `/v1/global-connections`
- [ ] As a platform **admin**, same setup: new connection is `PROJECT` scope (`POST /v1/app-connections`), no globe icon
- [ ] As a platform **admin**, **reconnect** an existing global connection from the step: still routes to the global endpoint and keeps `PLATFORM` scope
- [ ] Global-connections admin page (`/platform/setup/connections`) still creates `PLATFORM` connections
- [ ] AI-chat picker: reconnecting an expired **global** connection now updates the global connection (routes to `/v1/global-connections`), not a new project-scoped duplicate
- [ ] Embed SDK dialog (`embedded-connection-dialog.tsx`) unchanged and still working

Fixes GIT-1587


AhmadTash:
@greptile review

Addressed both points from the 4/5 review:
- **Feature-doc drift (P2):** updated `.agents/features/chat.md` — the `ap_show_connection_picker` entry now documents that the create/reconnect dialog derives scope from the operation (create → PROJECT, reconnect of a global connection → PLATFORM, never a silent project-scoped duplicate). (commit e1b7ae5)
- **PR title:** rewritten to describe the user-facing behavior rather than the implementation detail.

AhmadTash:
@greptile review

Addressed the flows.md feature-doc drift: `.agents/features/flows.md` now documents that `connection-select.tsx`'s create/reconnect dialog derives scope from the operation (create → PROJECT, reconnect of a global connection → PLATFORM), and that `isSelectedConnectionGlobal` is used only for the status/reconnect-button gating. (commit 490ab7467b)

amrdb (approved):

PR #14084, opened 2026-07-06
feat: Optional separate MCP URL

### LINKED ISSUE #15288: [Feature]: Optional AP_MCP_URL to serve MCP from a separate host or subpath
## Problem

All MCP URLs are built from `AP_FRONTEND_URL` (`packages/server/api/src/app/helper/domain-helper.ts`). Self-hosters who want MCP traffic on a separate hostname or ingress (`mcp.example.com` in front of `app.example.com`) cannot do it. MCP clients must hit `AP_FRONTEND_URL`.

Blocks independent rate limiting, TLS termination, and network policy for MCP.

## Proposed

Optional `AP_MCP_URL`. New `domainHelper.getMcpUrl()` / `getMcpUrlFromRequest()` resolve the MCP server URL, OAuth discovery/authorize/token/metadata endpoints, and token minting against it, including its path prefix for subpath-hosted instances. Falls back to `AP_FRONTEND_URL` when unset, so existing deployments are unchanged. Surfaced to the frontend as a new `MCP_URL` flag so the settings page shows the right address.

Requires a SAML SSO fix, bundled in the same PR. `/acs` is a statically IdP-registered callback and cannot tell which host a login started from. `/login` captures `originBaseUrl` from its own request, stores `{ from, originBaseUrl }` in Redis behind a short-TTL single-use token, passes only that token as SAML RelayState (spec-limited to ~80 bytes), and `/acs` resolves it, deletes it, and redirects to `originBaseUrl`.

## Source

External contributor PR: [https://github.com/activepieces/activepieces/pull/14084](<https://github.com/activepieces/activepieces/pull/14084>)

Currently conflicting with main. The MCP OAuth directory was refactored (validation + PKCE modules) after the PR was opened, so it needs a rebase before review lands.

Related: activepieces/activepieces#13596 (MCP OAuth base path).

danielpoonwj:
## What does this PR do?

Adds `AP_MCP_URL`, an optional environment variable that lets self-hosters serve MCP traffic (Claude Desktop, Cursor, and other MCP clients) from a different hostname or subpath than the main app (`AP_FRONTEND_URL`) — e.g. `mcp.example.com` in front of `app.example.com`. When unset, everything falls back to today's `AP_FRONTEND_URL`-based behavior, so this is a no-op for existing deployments.

Shipping this required fixing a SAML SSO limitation first: login could now start from either the main app host or the MCP host, but the SAML callback had no way to know which one to redirect back to after authentication. That fix is bundled into this PR.

**`AP_MCP_URL`**: All MCP-related URLs — the MCP server URL shown in settings, OAuth discovery/authorize/token/metadata endpoints, and token minting — now resolve through a new `domainHelper.getMcpUrl()` / `getMcpUrlFromRequest()` pair instead of the frontend-URL-based helpers. These resolve against `AP_MCP_URL` when set (including its path prefix, for subpath-hosted deployments), and fall back to `AP_FRONTEND_URL` when it isn't. The resolved address is exposed to the frontend via a new `MCP_URL` flag so the MCP settings page displays the right URL either way.

**The SSO fix**: SAML is SP-initiated — `/login` redirects the browser to the IdP, which authenticates the user and POSTs the assertion back to a *statically IdP-registered* callback URL, `/acs`. That callback URL can't vary per request, so `/acs` previously had no reliable way to tell whether a given login started from the main app or the MCP host — it always redirected back to whatever host its own request happened to land on, which could send MCP OAuth logins to the wrong place and break the flow.

SAML has a built-in mechanism for this: **RelayState**, an opaque string the SP hands the IdP on the way out (`/login`) that comes back unchanged on the way in (`/acs`) — the SAML equivalent of OAuth's `state` parameter. The catch is that RelayState is spec-limited to roughly 80 bytes, too small to carry a redirect target directly (especially alongside the existing `from` path param, which can run to several hundred bytes for MCP flows).

The fix:
1. `/login` captures `originBaseUrl` from its own request (`networkUtils.getRequestBaseUrl`) — unlike `/acs`, `/login`'s request host reliably reflects where the user actually started, since that request is never rewritten to a fixed IdP-registered address.
2. `{ from, originBaseUrl }` is stored server-side in Redis under a short-TTL, single-use opaque token, and only that token is passed as RelayState.
3. `/acs` resolves the token back to `{ from, originBaseUrl }`, deletes it (single-use), and redirects straight to `originBaseUrl` — falling back to its own request's host only if no RelayState resolved.

This generalizes to any host that can reach `/login` — main app, MCP, or future surfaces — with no route-prefix or env-var branching in the callback itself. The `from` path is still independently validated via a parse-based `isSafeRelativePath()` check (catches `//evil.com`/backslash-based open-redirect tricks that a `startsWith('/')` check would miss), and `originBaseUrl` is derived from the same request-header trust boundary the code already relied on at `/acs` before this change, so no new class of trust is introduced.

### Relevant User Scenarios

- A self-hosted operator wants MCP traffic (from Claude Desktop, Cursor, etc.) to go through a separate hostname or ingress rule than the main web app — e.g. for independent rate limiting, TLS termination, or network policy — without needing MCP clients to hit `AP_FRONTEND_URL`.
- An enterprise customer with SAML SSO configured connects an MCP client, which routes them through `/mcp-authorize` to log in — the SSO fix ensures they land back on the MCP flow after authenticating with their IdP, instead of being redirected to the main app.
- Existing self-hosted deployments that don't set `AP_MCP_URL` see no change in behavior.

Fixes #15288


abuaboud:

othmanemad:
Became part of https://github.com/activepieces/activepieces/pull/15376

PR #14094, opened 2026-07-07
fix(shopify): preserve product images on partial update

### LINKED ISSUE #6939: [BUG]: Shopify update product action overrides existing values
**Describe the bug**

https://community.activepieces.com/t/shopify-piece-update-product-deletes-images/8921

parshipcy:
## What does this PR do?
- Fixes #6939 
- The Shopify **Update Product** action always sent `images: []` on every request, even when no new image was provided. Shopify's `PUT /products/{id}` treat that as replacing the product with no images, which delete existing product images when updating fields like description.
- This change only includes fields the user actually set in the request payload, and omit `images` unless a new product image is provided.

image

### DIFF
--- a/packages/pieces/community/shopify/package.json
+++ b/packages/pieces/community/shopify/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-shopify",
-  "version": "0.2.7",
+  "version": "0.2.8",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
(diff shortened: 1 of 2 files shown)


parshipcy:
> ### Confidence Score: 3/5
> The images-clearing fix is correct and safe, but the `status` required prop has never been sent to Shopify's API and remains absent after this change, so every product update silently ignores the user's Active/Draft/Archived selection.
> 
> The images bug is properly fixed. The `status` field — a required prop visible to users — has never been included in the API payload, meaning status changes are silently lost on every run. This pre-existing defect is not introduced by the PR but is unresolved while the surrounding code is actively refactored.

yes, but i think it would be better to do it in a follow-up PR.

parshipcy:
@AhmadTash @kishanprmr Can you please take a look when you have some time?

abuaboud:

PR #14096, opened 2026-07-07
revert: republish 680 community pieces to heal stripped cloud metadata (#14064)

ibrahim-abuznaid:
## What

Reverts #14064 (the 680-piece version-bump). Restores 664 community `package.json` versions to their currently-published values; leaves 16 pieces that later PRs re-bumped untouched. Version-lines only.

## Why

#14064 bumped 680 pieces in one commit to force a metadata-heal republish. The `Release Pieces` workflow cannot process that many at once: `tools/scripts/pieces/prepare-pieces-for-publish.ts` calls the async `preparePieceDistForPublish` in a loop **without awaiting**, so every changed piece bundles concurrently. At ~685 pieces the `ubuntu-latest` runner exhausts memory and is terminated ("The runner has received a shutdown signal") during the **Prepare pieces for publish** step — before anything is published.

Because `find-changed-pieces` re-detects all ~680 unpublished pieces on every push to `main`, **every** subsequent `Release Pieces` run re-hits the same wall — blocking all piece releases repo-wide and paging on-call each time.

**Nothing from #14064 was ever published** (the job died before the publish and metadata steps — verified: the bumped versions are absent from npm and the cloud registry, old versions still served). So this revert removes only never-published source version-lines. No published version disappears; no flow, self-host, or cloud entry is affected.

## After this merges

The next `Release Pieces` run detects ~22 pieces (only what other PRs left pending) and completes normally, restoring the pipeline. The metadata heal will be re-done in smaller waves (≤120 pieces each, merged sequentially) that stay under the runner's limit.

A proper fix for the unbounded-concurrency bug in `prepare-pieces-for-publish.ts` (await + a concurrency cap) will follow separately so large batches are safe in the future.

Labels: `skip-changelog`, `area/third-party-pieces`.


kishanprmr:
Fixed in https://github.com/activepieces/activepieces/pull/14106 and all pieces are released.

PR #14099, opened 2026-07-07
feat(platform): add per-piece action & trigger visibility controls

othmanemad:
## What does this PR do?

Introduces **Piece Sets** — a named, reusable piece/action/trigger visibility configuration that a platform admin defines once and assigns to many projects. It replaces the older per-project allow/block lists and the legacy piece-tags workflow (when `platform.plan.managePiecesEnabled` is on) with a shared, declarative model.

Key characteristics:

- **Derived at read time** — visibility is resolved on read, so nothing has to be written when a new piece or action ships. A set can auto-include or auto-hide future pieces by policy (`pieces.mode = include_all | exclude_all`).
- **One Default set per platform** — any project without an explicit assignment resolves to it; the Default cannot be deleted.
- **Embed-facing `key`** — each set has an optional `key` (auto-generated from the name as `kebab-case-<suffix>` when omitted, unique per platform). Embed provision tokens reference a set by this key.
- **v4 provision token** — a new `version: "v4"` token payload carries a required `pieceSet` claim (the set key). Legacy `v3`/`v2` tokens keep working via the first `piecesTags` entry, matched against a set whose `key` equals the tag name (created by the backfill migration). Unknown key → falls back to Default (with a warn log).
- **Zero setup on Community** — on CE (or when the flag is off) piece-set management is inert and filtering falls back to the legacy project-plan lists.

## Breaking Changes

- **Piece Set API field rename** — `externalId` is renamed to `key` on `PieceSet`, `CreatePieceSetRequestBody`, and `UpdatePieceSetRequestBody`. Clients sending/reading `externalId` must switch to `key`.
- **`pieceSet` claim now v4-only** — the managed-auth `pieceSet` claim is removed from the `v2`/`v3` token payloads and lives only on the new `v4` payload. Integrations passing `pieceSet` on a v2/v3 token must reissue as a v4 token. Tag-based selection (`piecesFilterType`/`piecesTags`) still works on v3.
- **Platform-wide piece filter removed** — the `platform.filteredPieceNames` / `filteredPieceBehavior` columns are dropped; the platform-level piece killswitch no longer exists. Visibility is now expressed entirely through the platform's Default piece set.
- **Legacy piece-tags feature removed** — the entire piece-tags codebase (entities, services, `/v1/tags` controller, shared types, metadata tag-enrichment, and the Pieces-page tag UI/hooks/api) is deleted in favor of Piece Sets. The `tag` and `piece_tag` **tables are intentionally kept** and are read once by the `CreatePieceSetTable` backfill (tag → piece set, keyed by tag name); nothing reads or writes them at runtime anymore. No data migration is required and existing tag-derived sets are unaffected.

https://github.com/user-attachments/assets/afb683c6-ae5f-4836-84d6-7c0705fd245a

### Relevant User Scenarios

- A platform admin wants one "Marketing tools" visibility profile (only certain pieces + specific actions) and to apply it to all marketing projects at once, instead of configuring each project separately.
- An admin wants new pieces/actions to be hidden by default for a locked-down set of projects, but auto-available for internal ones — without re-editing config every time a new piece is released.
- An embedded/white-label customer wants to scope which pieces a tenant sees by passing a `pieceSet` key in a v4 managed-auth token (or a legacy tag on a v3 token).
- An admin wants to clone an existing set as a starting point for a new team, then tweak it.


packages/server/api/src/app/pieces/metadata/piece-metadata-controller.ts
```diff
@@ -72,12 +71,14 @@ const basePiecesController: FastifyPluginAsyncZod = async (app) => {
...
             const platformId = getPlatformId(req.principal)
-            return pieceMetadataService(req.log).getOrThrow({
+            const piece = await pieceMetadataService(req.log).getOrThrow({
                 platformId,
                 name: `${decodeScope}/${decodedName}`,
                 version,
                 locale: req.query.locale as LocalesEnum | undefined,
             })
+            const policy = await resolveVisibility({ platformId, projectId: req.query.projectId, log: req.log })
+            return isNil(policy) ? piece : policy.filterPieceComponents(piece)
```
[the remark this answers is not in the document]
othmanemad: @greptile Resolving: the controller's component-only filtering is intentional — full piece hiding is enforced on the resolution path (pieceBundle.resolve → get() passes projectId, returning 404 for hidden pieces), so GET /:name returning the definition doesn't grant usable access; adding isPieceVisible here would retroactively break the builder for existing flows that already reference the piece

othmanemad:
@greptile Resolving: the controller's component-only filtering is intentional — full piece hiding is enforced on the resolution path (pieceBundle.resolve → get() passes projectId, returning 404 for hidden pieces), so GET /:name returning the definition doesn't grant usable access; adding isPieceVisible here would retroactively break the builder for existing flows that already reference the piece

packages/server/api/src/app/ee/pieces/piece-set/piece-set.service.ts
```diff
@@ -0,0 +1,242 @@
+import { ActivepiecesError, apId, ErrorCode, isNil, kebabCase, SeekPage, spreadIfDefined } from '@activepieces/core-utils'
```
abuaboud: STYLE §4b needs `tryCatch` here (used in the `create`/`update` suggestions below).

```suggestion
import { ActivepiecesError, apId, ErrorCode, isNil, kebabCase, SeekPage, spreadIfDefined, tryCatch } from '@activepieces/core-utils'
```

packages/server/api/src/app/ee/pieces/piece-set/piece-set.service.ts
```diff
@@ -0,0 +1,242 @@
...
+                key: resolveKey({ key, name }),
+                isDefault,
+                generatedForProjectId,
+                config: config ?? pieceSetConfig.emptyConfig(),
+            })
+        }
+        catch (error) {
+            rethrowKeyConflict(error)
+        }
+        return pieceSetRepo().findOneByOrFail({ id })
```
abuaboud: STYLE §4b: the catch *reacts* to the failure (maps it to a nicer error), so it should use `tryCatch` rather than a raw `try/catch`.

```suggestion
        const id = apId()
        const { error } = await tryCatch(() => pieceSetRepo().save({
            id,
            platformId,
            name,
            key: resolveKey({ key, name }),
            isDefault,
            generatedForProjectId,
            config: config ?? pieceSetConfig.emptyConfig(),
        }))
        if (error) {
            rethrowKeyConflict(error)
        }
        return pieceSetRepo().findOneByOrFail({ id })
```

packages/server/api/src/app/ee/pieces/piece-set/piece-set.service.ts
```diff
@@ -0,0 +1,242 @@
...
+        try {
+            await pieceSetRepo().update({ id, platformId }, {
+                ...spreadIfDefined('name', request.name),
+                ...(request.key !== undefined ? { key: resolveKey({ key: request.key, name: request.name ?? existing.name }) } : {}),
+                config: updatedConfig,
+            })
+        }
+        catch (error) {
+            rethrowKeyConflict(error)
+        }
```
abuaboud: STYLE §4b: same as `create` — use `tryCatch` instead of raw `try/catch`.

```suggestion
        const { error } = await tryCatch(() => pieceSetRepo().update({ id, platformId }, {
            ...spreadIfDefined('name', request.name),
            ...(request.key !== undefined ? { key: resolveKey({ key: request.key, name: request.name ?? existing.name }) } : {}),
            config: updatedConfig,
        }))
        if (error) {
            rethrowKeyConflict(error)
        }
```

abuaboud (approved):

PR #14117, opened 2026-07-08
fix(pieces): externalize deps that break when inlined (amazon-s3, image-helper, scrapeless)

kishanprmr:
## Problem

The "update pieces metadata" release step fails for three pieces because their bundles crash at load time:

| Piece | Root cause |
|-------|-----------|
| **amazon-s3** | `openpgp` is inlined; its `createRequire(import.meta.url)` becomes `createRequire(undefined)` in the CJS bundle → `ERR_INVALID_ARG_VALUE` |
| **image-helper** | same ESM→CJS `import.meta.url` interop failure from an inlined dep (jimp) |
| **scrapeless** | `@scrapeless-ai/sdk` is inlined, but its transitive `puppeteer-core` (kept external) is never written to the published `dependencies` → `Cannot find module 'puppeteer-core'` |

## Fix

Set `"bundleDeps": false` on each piece so its third-party deps install at runtime instead of being inlined. This is the documented escape hatch and resolves to the **same code path** the automatic >5 MB size-fallback already uses in production — and it's the exact pre-bundling behavior these pieces shipped with for years.

## Verification

Ran the CI bundle + metadata-load locally (`prepare-pieces-for-publish` → `load-piece-metadata-child.mjs`) for all three:

| Piece | Bundle before → after | Load |
|-------|----------------------|------|
| amazon-s3 | 1413 KB → 53 KB | ✅ OK |
| image-helper | → 31 KB | ✅ OK |
| scrapeless | 1172 KB → 50 KB | ✅ OK |

All externalized deps land in the published `dependencies`.

Versions bumped so the release picks them up: amazon-s3 0.6.1→0.6.2, image-helper 0.1.14→0.1.15, scrapeless 0.1.6→0.1.7.


sanket-a11y (approved):

kishanprmr:
Fixes #14146

PR #14119, opened 2026-07-08
fix(json-viewer): pass .json extension when downloading step output

### LINKED ISSUE #13752: [BUG] Run/step JSON output downloads as a .bin file instead of .json

Jaimin791:
## Summary

Fixes #13752 — JSON viewer downloaded output as `.bin` because `handleDownloadFile` was called without the `.json` extension argument.

- Pass `'.json'` as the `ext` arg in `handleDownload` so the downloaded filename ends in `.json` instead of nothing (which browsers interpret as `.bin`)
- MIME type was already correct (`application/json`); only the filename extension was missing

## Test plan

- [ ] Open any flow run, expand a step output in the right panel
- [ ] Click the download button on the JSON viewer
- [ ] Verify the downloaded file is named `<title>.json` (not `.bin`)
- [ ] Open the downloaded file — confirm it contains valid JSON
- [ ] File downloads from the step-file URL path (eye icon) are unaffected


abuaboud:

PR #14131, opened 2026-07-09
fix(shared): replace z.union with z.discriminatedUnion for FlowAction to eliminate O(4^depth) Zod validation

### LINKED ISSUE #13974: [BUG] [Flows] Exponential zod validation of nested flows crashes with RangeError (stack overflow)
## Problem

Editing or importing a flow with a deeply nested structure can crash flow validation with `RangeError: Maximum call stack size exceeded` (the reported stack is entirely inside `zod/v4/core/schemas.cjs`, recursing object -> optional -> object). Depending on the data it surfaces instead as `Invalid string length` / heap OOM. Same failure class either way.

The crash is what triggers the orphaned-draft symptom tracked separately in activepieces/activepieces#13931; this ticket is the underlying validation defect.

Ref: Pylon #5225

## Root cause

`FlowAction` (`packages/core/execution/src/lib/flows/actions/action.ts:290`) is a **non-discriminated** `z.union` of the four action types (CODE / PIECE / LOOP_ON_ITEMS / ROUTER), each recursively embedding `FlowAction.optional()` for `nextAction` / `firstLoopAction` / `children`:

```ts
export const FlowAction: z.ZodType<FlowAction> = z.lazy(() =>
    z.union([
        CodeActionSchema.extend({ nextAction: FlowAction.optional(), ... }),
        PieceActionSchema.extend({ nextAction: FlowAction.optional(), ... }),
        LoopOnItemsActionSchema.extend({ nextAction: FlowAction.optional(), firstLoopAction: FlowAction.optional() }),
        z.object({ ..., type: z.literal(ROUTER), nextAction: FlowAction.optional(), children: z.array(z.union([FlowAction, z.null()])) }),
    ]),
)
```

A plain `z.union` tries every member until one matches. Because each member is an object that contains `nextAction`, zod descends the entire remaining subtree for all four branches before the non-matching ones fail on the `type` literal. So as soon as one node in the tree fails to validate, every ancestor re-parses the whole subtree against all four branches: roughly 4^depth work plus an exponentially growing issue array. That exhausts the stack/heap.

Confirmed identical on release-0.85.4 (the reporter's version) and upstream `main`.

Reproduced against the real compiled schema:

* Fully valid nested flow: linear, depth 1000 parses in ~3ms. Depth alone is not the trigger.
* Nested flow with one node that fails validation: depth 6 = 85ms, depth 8 = 1.4s, depth 10 = 25s then crash. Clean exponential.
* Same data against `z.discriminatedUnion('type', ...)`: depth 1000 drops from crash to ~2s. Exponential gone.

## Fix

Convert the recursive `FlowAction` union (and ideally `SingleActionSchema` and the top-level `FlowOperationRequest` union) from `z.union` to `z.discriminatedUnion('type', ...)`. All branches already carry a `type` literal, so zod can jump straight to the matching branch instead of backtracking. Optionally add a hard depth cap on import as a defensive guard.

## Out of scope / needs confirmation

Tracing `main` statically, the obvious server import path validates shallowly — the trigger's `nextAction` is `z.any()` (`triggers/trigger.ts:37`) and the per-step validators use the non-recursive `SingleActionSchema` — so the exact runtime line that feeds the recursive schema the full deep tree is not pinned. The reporter's production stack proves it is reached on their setup. Need the full untruncated stack trace + the customer's flow JSON to confirm the invocation point. The schema defect and its fix stand regardless of which path reaches it.

## Steps to reproduce

1. Have a flow with a deeply nested structure where at least one node does not cleanly validate (e.g. a stale field from an older schema, an unknown piece version, an unexpected `null`).
2. Edit or import it.
3. Observed: `RangeError: Maximum call stack size exceeded` from zod during validation; a valid flow of the same depth parses instantly.

Jaimin791:
## Summary

Fixes #13974

- Replaces `z.union([...])` with `z.discriminatedUnion('type', [...])` in the `FlowAction` schema in `packages/shared/src/lib/automation/flows/actions/action.ts`
- Bumps `@activepieces/shared` patch version to `0.87.1`

## Root cause

`z.union` tries every branch in order for each node during validation. For deeply nested flows (e.g. loops inside loops) this results in O(4^depth) backtracking when a node fails validation. `z.discriminatedUnion` reads the `type` literal field first and dispatches to exactly one branch — O(1) per node — eliminating the exponential blowup.

All four action branches (`CODE`, `PIECE`, `LOOP_ON_ITEMS`, `ROUTER`) already carry a `type: z.literal(FlowActionType.X)` discriminant, so no schema shape changes are needed.

## Test plan

- [ ] Run `npm run test-unit` — existing engine + shared tests pass
- [ ] Build a flow with deeply nested loop actions and verify validation no longer times out / stack overflows
- [ ] Confirm `FlowAction` still correctly parses all four action types


abuaboud:

PR #14134, opened 2026-07-09
fix(embed-sdk): queue navigate() pre-auth, tear down old iframe on re-configure, expose initialRoute

### LINKED ISSUE #14118: [BUG] [Embed] navigate() silently dropped until auth handshake finishes; second configure() leaks old iframe
## Problem

1. `navigate()` right after `configure()` is intermittently dropped — not queued, no error.
2. A second `configure()` appends a new iframe at `/` and leaks the old iframe + its `message` listeners.

Ref: Pylon 5072

## Root cause

* `navigate()` posts `VENDOR_ROUTE_CHANGED` immediately, but the client only registers its listener after the managed-authn exchange (`packages/web/src/app/routes/embed/index.tsx:166`). Messages before that vanish. Also early-returns (console.error only) while the container poll hasn't set `_dashboardAndBuilderIframeWindow` (`packages/ee/embed-sdk/src/index.ts:436`).
* `connectToEmbed` does `appendChild` without removing the old iframe (`index.ts:329`); per-iframe listeners never removed → old iframe still fires `CLIENT_ROUTE_CHANGED`. `initialRoute` hardcoded `'/'` (`index.ts:261`), not exposed in `ConfigureParams` despite activepieces/activepieces#7999.

## Fix

* Queue pre-handshake `navigate()` calls, flush on `CLIENT_CONFIGURATION_FINISHED`.
* On re-`configure()`, tear down old iframe + listeners first.
* Expose `initialRoute` in `ConfigureParams`.

Jaimin791:
## Summary

Fixes #14118 — three embed SDK bugs: `navigate()` silently dropped before auth handshake, second `configure()` leaks old iframe + listeners, `initialRoute` not exposed in `ConfigureParams`.

- Queue `navigate()` calls when `_isConfigured` is false (either during grace period or during auth); flush the queue on `CLIENT_CONFIGURATION_FINISHED`
- Extract `_tearDownDashboardIframe()` called at the start of every `configure()`: removes old iframe from DOM, calls listener cleanup functions, resets state
- `_checkForClientRouteChanges` and `_checkForBuilderHomeButtonClicked` now return cleanup functions captured in `_cleanupDashboardListeners`
- Add `initialRoute?: string` to `EmbeddingParam`; pass it to `connectToEmbed` (was hardcoded `'/'`)

## Test plan

- [ ] Call `configure()` then immediately `navigate('/flows')` — verify navigation takes effect after auth finishes (not silently dropped)
- [ ] Call `configure()` twice in succession — verify only one iframe exists in DOM; no stale `CLIENT_ROUTE_CHANGED` events from old iframe
- [ ] Pass `initialRoute: '/flows'` in `EmbeddingParam` — verify iframe opens on `/flows` instead of `/`
- [ ] Normal single-configure flow: verify existing navigation, auth, and connection embed flows still work


majewskibartosz:
we have already a pr for this

PR #14142, opened 2026-07-09
feat(flows): pin canvas notes to steps so they stay in place when the flow changes

### LINKED ISSUE #14138: [Feature] Anchor builder notes to steps so they stay in place when the flow changes
**Problem**

Canvas notes are free-floating: absolute `position: {x, y}` only, no step reference (`packages/core/execution/src/lib/flows/note.ts`). Adding/removing/moving steps re-lays-out the graph; notes keep their old coordinates and end up next to the wrong step.

**Request** (user-reported)

Let a note be attached/anchored to a specific step. Anchored note follows the step through layout changes. Unanchored free notes stay as-is.

**Sketch**

* Optional `stepName` (+ relative offset) on the Note schema
* Builder renders anchored notes relative to the step node's position; falls back to free position if the step is deleted
* Attach via note toolbar or drag-onto-step

majewskibartosz:
Fixes [GIT-1610](https://linear.app/activepieces/issue/GIT-1610)
Closes #14138

## Problem
1. Canvas notes store only absolute `position {x,y}`; every step add/delete/move re-runs auto-layout, steps shift, notes keep their coordinates and end up beside the wrong step.

## Fix
- `packages/core/execution/src/lib/flows/note.ts` — optional nullable `anchor { stepName, offset {x,y} }` on the Note schema; propagates through `UpdateNoteRequest`/`AddNoteRequest` automatically (they are `Note.omit(...)`)
- `flow-canvas-utils.ts` — anchored note renders at the step node's post-orientation position + offset; falls back to stored `position` when the step is missing; graph key extended with anchor fields
- `note-tools.tsx` — pin/unpin toolbar button; pin anchors to the nearest step node, unpin writes the current rendered position so the note stays put
- `flow-drag-layer.tsx` — dragging an anchored note re-captures the offset; clears the anchor if the step node no longer exists
- `delete-action.ts` / `delete-branch.ts` — clear anchors whose step no longer exists in the version (incl. descendants of deleted loops/routers) via `notesOperations.clearDanglingNoteAnchors`; auto-generated `step_N` names get reused, stale anchors would re-capture unrelated new steps
- `@activepieces/core-execution` 0.8.0 → 0.9.0; two i18n keys added

## Verification
- `npm run lint-dev` clean (0 errors, pre-existing warnings only)
- `packages/core/execution` vitest: 94 passed incl. 5 new note-operation tests (anchor survives move, null clears, container-delete clears descendant anchors, old-client payload without anchor key preserves it)
- `packages/web` flow-canvas-utils vitest: 7 passed incl. 4 new (follows layout shift, dangling-anchor fallback, free notes untouched, horizontal orientation)
- Browser-verified on local CE build: pin persists correct anchor, note follows step through layout shift, unpin keeps rendered position, deleting the anchored step clears the anchor server-side and the note falls back
- Security review: no findings (zod bounds match existing position/size fields, tooltip interpolation React-escaped, no controller changes)
- UNVERIFIED: `npm run test-api` not run (no server API surface changed); 8 pre-existing web unit test failures on main (data-selector schema / output-viewer) are unrelated

## Breaking change?
- [x] no
- [ ] yes-technical
- [ ] yes-functional

## Security impact?
- [x] no
- [ ] yes-security-sensitive


majewskibartosz:
## Visual verification

Driven on a local CE dev instance built from this branch (fresh DB). Flow: Empty Trigger plus three Code steps, two notes on the canvas. One note pinned via the new toolbar button, one left free as a control. Steps inserted above to force auto-layout shifts, then the anchored step deleted, then a full page reload.

| State | Result |
|---|---|
| New pin control on the note toolbar | pin tooltip |
| Pinned: tooltip names the anchor step | pinned tooltip |
| Before layout shift: pinned note beside "Send report", free note beside "Transform data" | before |
| After inserting a step above: pinned note followed "Send report" down; the free note stayed at its old coordinates and now sits beside the wrong step (the exact bug this PR fixes) | after |
| Anchored step deleted: note falls back to its stored position, canvas intact | fallback |
| After the delete the anchor was cleared server-side: pin control reset to "Pin to nearest step" | anchor cleared |
| Re-pinned, then full page reload: anchor persists ("Pinned to Code") | persisted |

The pinned note followed its step through two consecutive step insertions (tested twice, second one shown). Unpin keeps the note at its rendered position, verified in the session and covered by unit tests. Two extra shots of the first insertion round (03/04) are on the assets branch if needed.

AbdulTheActivePiecer:
we will do a new design later down the road, design is in the linear ticket

PR #14145, opened 2026-07-09
fix(pieces): label stale duplicate AWS Bedrock piece as deprecated in piece selector

### LINKED ISSUE #14141: [BUG] [Pieces] "AWS Bedrock" appears twice in piece selector, stale renamed package still served from cloud registry
## Problem

Two identical "AWS Bedrock" entries in piece selector search. Cloud returns both `@activepieces/piece-amazon-bedrock` and `@activepieces/piece-aws-bedrock`.

Ref: Pylon 5331

## Root cause

Piece shipped as `piece-aws-bedrock` ([#11223](<https://github.com/activepieces/activepieces/issues/11223>)), renamed to `piece-amazon-bedrock` ([#11241](<https://github.com/activepieces/activepieces/issues/11241>)). Old rows stay in the cloud registry; dedup is per package name (`piece-metadata-service.ts:434`). Stale package frozen at 0.0.2, gets no updates.

## Fix

Per review: keep the stale package listed but label it, `displayName` → "AWS Bedrock (Deprecated)". Applied server-side in the piece listing (and re-applied after locale translation). Stays resolvable by name/version so existing flows don't break. Hiding it from the listing (first approach) was rejected in review.

majewskibartosz:
Fixes activepieces/activepieces#14141  <br>Closes activepieces/activepieces#14141

## Problem

Piece selector search shows two identical "AWS Bedrock" entries (same name, same logo). The piece shipped as `@activepieces/piece-aws-bedrock` ([#11223](<https://github.com/activepieces/activepieces/issues/11223>)) and was renamed to `@activepieces/piece-amazon-bedrock` four days later ([#11241](<https://github.com/activepieces/activepieces/issues/11241>)); the old package's rows survive in the cloud registry because `pieceSyncService.sync` upserts and never deletes, and list dedup is per package name. Ref: Pylon 5331.

## Fix

Per review feedback, the stale package stays listed instead of being hidden, labeled so users pick the right one:

* `piece-metadata-service.ts` → `markDeprecatedRenamedPieces` appends ` (Deprecated)` to the `displayName` of packages in `DEPRECATED_RENAMED_PIECES`. Applied in `fetchLatestCompatiblePiecesFromDB` (feeds listing + tool-search reindex) and re-applied idempotently after locale translation in `translatePieces` (non-English i18n overwrites `displayName`).
* `get()` by name/version and `/v1/pieces/registry` untouched: flows pinned to the old package keep resolving and installing.
* DB deletion rejected earlier for the same reason: sync re-adds rows and pinned flows would break.
* `.agents/features/pieces.md` → mechanism documented (entries are permanent; removing one drops the label and reintroduces the indistinguishable duplicate).

## Verification

* Regression test updated: list returns both packages, `@activepieces/piece-aws-bedrock` as "AWS Bedrock (Deprecated)", `@activepieces/piece-amazon-bedrock` unchanged; get-by-name still resolves the stale package at 0.0.2. Fails on main (no suffix, two identical entries), passes with the fix.
* Cloud piece-metadata suite 11/11; CE pieces suite 37/37; `npm run lint-dev` clean.
* Pre-existing, unrelated: 8 web unit-test failures on main (pre-push run with SKIP_CHECK=1).
* Visual gate: not applicable, backend-only diff.


sanket-a11y:
Closing this as #14248

PR #14147, opened 2026-07-09
fix(triggers): preserve polling checkpoint on flow republish

### LINKED ISSUE #13311: [BUG]: Polling triggers silently drop events when flow is republished
**Describe the bug**

When a flow is edited and republished (without disabling it first), polling triggers reset their `lastPoll` checkpoint to the republish time. Any events that occurred between the last poll and the republish are silently dropped — no error, no warning to the user.

**To Reproduce**

1. Create a flow with a polling trigger (e.g. Jira **New Issue**, poll interval: 5 min)
2. Publish the flow — `lastPoll` is set to **T1**
3. A Jira issue is created at **T2** (T2 > T1, before next poll)
4. Edit the flow and republish at **T3** (T3 > T2, still before next poll)
5. Next poll runs at T3 + 5min, querying `created > T3`
6. The issue created at T2 is **never picked up**

**Expected behavior**

Republishing a flow should preserve the trigger's polling checkpoint. The next poll after republish should continue from where the previous version left off, so no events are missed.

**Root Cause**

In `packages/pieces/common/src/lib/polling/index.ts`, `pollingHelper.onEnable` unconditionally overwrites `lastPoll` with `Date.now()`:

```ts
case DedupeStrategy.TIMEBASED: {
  await store.put('lastPoll', Date.now()); // resets on every enable, including republish
}
```

During `LOCK_AND_PUBLISH`, the server internally calls `onDisable` (old version) → `onEnable` (new version) — the same sequence as a manual toggle off/on. The store is scoped by `flowId` (not `flowVersionId`), so the value persists across republishes but gets unconditionally overwritten by `onEnable`.

The piece has no way to distinguish a republish from a manual toggle at the hook level since both produce the same `onDisable` → `onEnable` call sequence.

**Additional context**

- Affects **all polling triggers** using `DedupeStrategy.TIMEBASED`
- Silent failure — no error surfaced to the user
- Real customer impact: a Jira ticket created at 2:01pm was missed because the flow was republished at 2:03pm, resetting the checkpoint ahead of the ticket's creation time

**Proposed Fix**

Pass an `isRepublish` flag from the server during `LOCK_AND_PUBLISH` through to the trigger hook context. In `pollingHelper.onEnable`, only reset `lastPoll` when `isRepublish` is false (i.e. first enable or manual re-enable after toggle off).

Affected files:
- `packages/pieces/common/src/lib/polling/index.ts`
- `packages/server/api/src/app/flows/flow/flow.service.ts`
- `packages/server/api/src/app/trigger/trigger-source/trigger-source-service.ts`
- `packages/server/engine/src/lib/helper/trigger-helper.ts`

kishanprmr:
## Problem

Fixes #13311.

Polling triggers reset their checkpoint (`lastPoll` for `TIMEBASED`, `lastItem` for `LAST_ITEM`) to *now* on every `onEnable`. Republishing a **live** flow runs `onDisable(old) → onEnable(new)`, so the checkpoint jumped forward to the republish time and any event created between the last poll and the republish was **silently dropped** (e.g. missed Jira issues).

The trigger store is flow-scoped (not version-scoped), so the old checkpoint is still readable at republish — the only reason it was lost is the unconditional overwrite in `pollingHelper.onEnable`.

## Fix

Thread an optional `isRepublish` flag from the publish path down to the trigger hook:

`flow.service` → `applyStatusChange` → `preUpdateStatus` → `triggerSourceService.enable` → `flowTriggerSideEffect.enable` → `ExecuteTriggerHookJobData` → worker → `ExecuteTriggerOperation` → engine `trigger-helper` sets `context.isRepublish` → `pollingHelper.onEnable`.

`isRepublish` is set **only** for a `LOCK_AND_PUBLISH` of an already-`ENABLED` flow **whose trigger is unchanged** — same piece, same trigger name, **and deep-equal `settings.input`** (`flowPublishUtils.isSameTrigger`). `pollingHelper.onEnable` then **preserves** the existing checkpoint; otherwise it resets to now.

Comparing the input is not cosmetic. A checkpoint kept across a props change points at a resource the trigger no longer polls, and `pollingHelper.poll` treats a `LAST_ITEM` id it cannot find in the fetched page (`findIndex → -1`) exactly like "no checkpoint" — it emits **every** item. So preserving across a changed table/board/sheet would turn one republish into a run per row.

## Behavior

| Scenario | Checkpoint |
|---|---|
| New flow, first enable | reset to now |
| Running flow edited & **republished** (same trigger, same props) | **preserved** ✅ |
| Republished with the **trigger swapped** (e.g. New Issue → Updated Issue) | reset to now |
| Republished with the **trigger props changed** (e.g. different table/board) | reset to now |
| Manual toggle off → on | reset to now |

The flag is optional end-to-end, so a piece running on an older engine simply keeps today's reset behavior (graceful degradation) — no `minimumSupportedRelease` bump, no break in mixed-version states.

## Scope

- Covers every piece that uses `pollingHelper` **and forwards the full `context`** (95 triggers, incl. all Jira triggers — no piece edits needed).
- **Scaffolding fixed** so new pieces get it for free: the `create-trigger` CLI template and the polling-trigger docs now pass `context` (the piece-builder skill already did).
- **Follow-up PRs (not here):**
  - ~302 pieces that destructure `{ store, auth, propsValue }` and drop `context` → mechanical codemod to forward `context`.
  - ~32 custom-polling triggers (e.g. Gmail) that hand-roll `onEnable` → tailored per-piece `context.isRepublish` guard.

## Tests

- Added `polling-helper.test.ts` (6 cases: preserve/reset for `TIMEBASED` + `LAST_ITEM`).
- Added `flow-publish-utils.test.ts` (7 cases: piece swap, trigger-name swap, changed input, changed nested input, key-order-insensitive match, non-piece trigger).
- Lint: 0 errors on all touched packages.

## Docs

- `brain/wiki/flows-execution/triggers.md` documents the `isRepublish` behavior, including the props-change reset and the `findIndex → -1` flood it prevents.


amrdb:
Pushed a docs-only change onto the branch. No behaviour change — your logic is untouched.

**The gap it covers:** the scaffolding fixes here are right, but nothing in the repo states the rule, so the next author copying a neighbouring piece reintroduces the old shape. 306 of 403 `pollingHelper.onEnable` call sites still destructure `{ store, auth, propsValue }`, and TypeScript can't flag it — every field on the helper's param type is optional, so a partial object is a valid value and `isRepublish` is dropped in silence. Those triggers keep today's reset behaviour with no error anywhere.

**Added the rule in two places:**

- `brain/wiki/pieces-engine/building-pieces.md` — why the destructured form type-checks and what it costs.
- `.agents/skills/piece-builder/` — the rule beside the polling pattern, plus a carve-out to the skill's "match the piece you're editing" golden rule. Unamended, that rule tells an author to copy the broken call.

**On the ~306 remaining sites**, the skill now says fix them **on touch** rather than in one codemod: whoever edits a piece switches that piece's calls over, riding a version bump and rebuild that is happening anyway, instead of forcing one on hundreds of pieces nobody is running. Say so if you'd rather do the sweep and I'll change the guidance.

amrdb (approved):

AhmadTash (approved):

PR #4943, opened 2024-06-17
feat(wedof): update certificationFolder action

vbarrier:
## What does this PR do?

This Pull request add a new action : update a certificationFolder

Fixes # (issue)

PR #14154, opened 2026-07-09
feat(workers): reserve dedicated worker capacity for synchronous webhooks

### LINKED ISSUE #13622: [Feature]: Route sync webhooks to a dedicated queue / worker pool (isolate real-time path from scheduled jobs)

AhmadTash:
## What & why

Synchronous webhook runs (`POST /api/v1/webhooks/{id}/sync`, and synchronous resume links) answer a caller that is actively waiting — a live phone call, a chat widget, a checkout. Today they share one BullMQ queue and worker pool with scheduled and polling work, and priority only orders the **waiting** list; it cannot preempt a job that is already running. A burst of long-running scheduled flows can occupy every worker slot for 60s+, and the real-time caller times out. (Reported in #13622 / Pylon #4943.)

This gives the real-time path **reserved capacity that background work can never take**.

## How it works

- New `syncJobs` queue. A worker started with `AP_WORKER_QUEUE=syncJobs` polls **only** that queue through the existing per-poll dispatcher — there is **no** multi-queue/round-robin polling, so no added dequeue latency.
- Runs with a live waiting caller route to `syncJobs`, but **only while a version-matched sync worker is online**; otherwise they fall through to the shared queue and execute exactly as before. This is the same capacity-gated-with-fall-through pattern main already uses for project worker groups.
- **No API-side flag and no migration.** Single-pool deployments are unchanged, and scaling the sync pool to zero restores today's behavior — the queue can never become a routing dead end.

## Design notes

- Routing lives in `getQueueName`, after the project/platform worker-group checks — **groups win** over the class queue (grouped tenants already have reserved capacity).
- `AP_WORKER_QUEUE` is allowlisted (only `syncJobs`; typos and internal queues like `runsMetadata` are rejected and the worker safely falls back to the shared queue).
- The sync capacity count **excludes version-mismatched workers**, so a rolling deploy cannot divert `/sync` runs into a queue that the version-gated worker isn't draining yet.
- Resumes that only inherit a stored `workerHandlerId` (delay steps, approval callbacks, subflow/Call-Flow completion) stay in the shared queue so stale continuations never claim reserved sync capacity. The direct request/response path is fully isolated; this boundary is documented.
- Cancellation scans both the routed queue and `syncJobs`; the queue is pre-created at boot so Bull Board and `/queue-metrics` report it deterministically. An ungrouped project's concurrency ceiling now spans the shared + sync pools.

## Credit

Supersedes #13643 — thanks to @gouthamx67 for the original direction and diagnosis. The queue/worker subsystem was rewritten after that PR (per-poll dispatcher #13998, connection-tracked reclaim #13962, per-project worker groups #13916), so this is a fresh implementation on the current machinery.

## Test plan

- [x] `npx turbo run build --filter=api --filter=worker` green
- [x] `npx turbo run lint --filter=api --filter=worker` green
- [x] New unit tests pass: routing gate (route / fall-through / non-sync / group-wins), capacity bucketing incl. version-skew exclusion, `AP_WORKER_QUEUE` parser allowlist + typo rejection, poll-queue resolution, both-queue cancel scan
- [ ] Manual: start a worker with `AP_WORKER_QUEUE=syncJobs`, confirm `/sync` runs land on it and a scheduled-flow burst on the shared pool does not delay sync responses
- [ ] Manual: stop the sync worker, confirm `/sync` runs fall through to the shared queue

Closes #13622


AhmadTash:
@greptile review

AhmadTash:
@greptile review

PR #14156, opened 2026-07-09
fix(tables): keep column order consistent when importing tables from JSON

majewskibartosz:
Fixes [GIT-1614](https://linear.app/activepieces/issue/GIT-1614)
Closes #14155

## Problem

1. Export a table as JSON → reimport: column order differs on every import. Fields have no position column, so display order = `created ASC` — but imports created fields via `Promise.all`, making the timestamps (and thus column order) nondeterministic.

## Fix

- `tables/table/table.service.ts` `create()` — create fields sequentially so `created` timestamps follow the JSON array order
- `ee/.../project-state.service.ts` CREATE_TABLE + UPDATE_TABLE — same
- `ee/.../project-replace.service.ts` `runTableOp` — same (4th call site, found during review)
- `web/.../table-hooks.ts` `importTableIntoExisting` — serialize field-create requests
- Field deletions remain parallel (order irrelevant)
- Long-term fix (`position` column + UI reorder) tracked in [GIT-886](https://linear.app/activepieces/issue/GIT-886)

## Verification

- Regression test red-first: on real Postgres the new order-assertion test fails pre-fix (fields shuffle) and passes with the fix. On pglite (CI) the single connection serializes `Promise.all`, so the pre-fix failure doesn't reproduce there — test still guards ordering.
- Tests: CE tables 140/140, cloud project-release + tables 30/30, project-replace 15/15
- `npm run lint-dev` clean (0 errors)
- Visual: drove JSON import as new table and reimport into existing table on local dev; column order Alpha→Juliett preserved on both paths, light + dark themes
- Known ceiling: importing N fields now performs N sequential inserts (capped at 100 fields/table); goes away when the position column lands ([GIT-886](https://linear.app/activepieces/issue/GIT-886))
- Pre-existing unrelated unit-test failures on main (`data-selector`, `smart-output-viewer`) bypassed at push, untouched by this diff

Issue #14155, opened 2026-07-09
[BUG]: Table column order shuffles on every JSON import/reimport


majewskibartosz:
## Problem

Table column order shuffles on every JSON import/reimport.

## Root cause

No `position` column on fields — order from `created` timestamp. Import creates fields via `Promise.all` (parallel), so timestamps are non-deterministic. Same issue in project state apply.

TODO already exists at `table.service.ts:202`.

## Fix

Short-term: serialize field creation to preserve array order.  <br>Long-term: `position` column (activepieces/activepieces#8640).

PR #14157, opened 2026-07-09
feat(tables): reorder table columns with drag and drop

### LINKED ISSUE #8640: [Feature Request] Re-arrange and sort columns in table
When I add a column (during the development of my table/flow) a new column is added at the end. It would be nice to re-arrange the columns in the UI because sometimes the data added is related to a previous column or it's just more important. I also have too many columns in one table and scrolling to the right is cumbersome.

It would also be nice to sort so I don't need to scroll through the rows during development or when my flow runs and I'm watching the data is getting updated.

Drag 'n' drop columns or move forward/backward option per column.

Sort ascending or descending columns option.

I can export and do everything in Excel, but this is cumbersome during development or when watching the data is getting updated.

majewskibartosz:
Closes #8640

## Problem

Table columns cannot be reordered — order is derived from `created` timestamps, and the only workaround (export JSON, reorder, reimport) is fragile.

## Fix

* `field.position` int column + migration `1818000000000-AddFieldPosition`: plain `ADD COLUMN … DEFAULT 0`, **no backfill**. List order is `position ASC, created ASC`, so pre-migration rows all sitting at `position 0` still sort by `created` — identical to a `0..n-1` backfill, without rewriting every field row under the `ADD COLUMN` `ACCESS EXCLUSIVE` lock (which would stall the `field` table on large cloud deploys).
* `fieldService.create` assigns `MAX(position)+1` (append), or an explicit `request.position` when provided.
* `POST /v1/fields/reorder` `{ tableId, fieldIds }` — takes the full ordered id list the client already holds and resequences positions to `0..n-1` in a **single** `UPDATE … FROM unnest(fieldIds) WITH ORDINALITY` statement, scoped by `projectId + tableId` and skipping no-op rows (`position IS DISTINCT FROM`). Foreign/stale ids simply don't match and are no-ops. Concurrent reorders are last-write-wins (same as rename) — no distributed lock. `UpdateFieldRequest.name` is now optional since position no longer rides on the single-field update, and `fieldService.update` skips the write when it is absent.
* **Import paths pass the source array index as `position`** (`tableService.create`, project-state create + update ops, project-replace, and the web template-import hook), so imported column order no longer depends on insert timing. This **supersedes #14156** (merged, for #14155), which serialized those creation loops so the `created` timestamps would come out monotonic — with an explicit `position` they go back to `Promise.all`.
* **Field cap validated up front on the bulk paths.** `fieldService.validateCount` now takes `insertCount`, and every bulk/import path calls it once with the batch size *before* its `Promise.all`. The old per-create check races under concurrency — all parallel creates read the same pre-save count and pass — so re-parallelizing the import loops above would otherwise have let a batch overshoot `AP_MAX_FIELDS_PER_TABLE`. The two update paths count only the genuinely new fields (ones whose `externalId` isn't already present).
* **Web**: react-data-grid native column dragging (`draggable: canEdit` columns + `onColumnsReorder`), with an optimistic reorder in the tables store that remaps each record's positional `cell.fieldIndex` and re-sorts. The column header dropdown was a Radix `DropdownMenuTrigger` that opens on `pointerdown` and `preventDefault`s it, swallowing the mousedown that starts the native drag (hence needing a second drag) — it now opens the menu on **click** (a native drag suppresses the click), with the trigger reduced to an invisible `pointer-events-none` anchor, plus a `GripVertical` grab handle revealed on hover.
* `@activepieces/shared` 0.122.0 → 0.123.0

## Out of scope

Reordering **existing** fields through project-release apply — `FieldState` carries no position, so array order only applies to newly created fields. Documented in `brain/wiki/data-storage-observability/tables.md`, along with the bulk-path `validateCount` race and the client's positional `cell.fieldIndex` coupling.

## Verification

* **Automated (CE tables)** — 4 new field tests: create appends with increasing positions (`[0,1,2]`), an explicit `position` on create is respected, `reorder` applies the given order and resequences to `0..n-1`, and a foreign id in the list is a no-op that leaves the other table untouched. The create-table test is augmented to assert positions follow the fields-array order. `createMockField` seeds `position: 0`.
* **Migration safety** — schema-only `ADD COLUMN … DEFAULT 0`; no data rewrite, no long lock. `breaking = false`, so it rolls back cleanly.
* **Frontend** — native rdg drag reorder; the two-drag-needed regression is fixed (drag on pointerdown, menu on click, grab handle affordance). Optimistic column + cell reorder persists across reload.

## Breaking change?

- [X] no
- [ ] yes-technical
- [ ] yes-functional

## Security impact?

- [X] no
- [ ] yes-security-sensitive

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


majewskibartosz:
## Visual verification

Driven states on local dev: table with 4 columns (Alpha, Bravo, Charlie, Delta) and one record with data under Alpha. Dragged Delta onto Bravo, then Alpha to the end.

| State | Result |
|---|---|
| Before drag — Alpha, Bravo, Charlie, Delta (dark) | before drag |
| After both drags — Delta, Bravo, Charlie, Alpha (light) | after drag light |
| After both drags (dark) | after drag dark |

Verified: DB resequenced to `[Delta 0, Bravo 1, Charlie 2, Alpha 3]` via `GET /v1/fields`; order persists across reload; `row1-alpha` cell followed the Alpha column both optimistically (pre-reload) and after reload; header rename/delete dropdown still opens on click (no drag/click conflict); positions assigned 0..n-1 on table create with fields array.

othmanemad (approved):

othmanemad:
@greptileai review

PR #14166, opened 2026-07-10
feat(subflows): add "Process CSV in Chunks" action to fan out CSV rows across subflow runs

ncodedude:
## What does this PR do?

Adds a **"Process CSV in Chunks"** action to the core **Sub Flows** piece.

It takes a CSV file, splits it into chunks of N rows, and dispatches each chunk to a
subflow (a flow with the "Callable Flow" trigger) asynchronously. Each chunk becomes an
independent child run, so a large CSV is processed in parallel instead of in one
long-running step that risks hitting the flow timeout.

Scope is limited to the piece: no changes to the server or to `@activepieces/shared`.

The action exposes these inputs:

| Input | Type | Required | Description |
|---|---|---|---|
| CSV File | File | yes | The CSV to split. |
| Flow | Dropdown | yes | The target subflow. Only published flows with a "Callable Flow" trigger are listed; disabled ones are marked `(inactive)`. |
| Chunk Size (rows) | Number | no | Rows per chunk. **Defaults to 5000 when left empty.** |
| Does the CSV have headers? | Checkbox | yes | Controls `columns` when parsing. |
| Delimiter | Dropdown | yes | `,` / tab / `;` |
| Additional Data | Object | no | Static data merged into every chunk payload. |

At run time it:

1. Parses the CSV with `csv-parse/sync` (already used by the core CSV piece).
2. Resolves the chunk size: a finite number `>= 1` is used, otherwise the `5000` fallback.
3. Splits the rows into chunks.
4. Resolves the target subflow and fails fast with a clear message if it is disabled.
5. For each chunk, POSTs to the subflow webhook with
   `PARENT_RUN_ID_HEADER` (so the chunk becomes a **child run** of the parent) and
   `FAIL_PARENT_ON_FAILURE_HEADER: 'false'` (so a failing chunk never fails the parent).
   The payload is `{ data: { chunkIndex, totalChunks, rows, ...additionalData } }`.
6. Returns `{ dispatchedChunks, chunkSize, totalRows, flowExternalId }`.

It is deliberately **fire-and-forget**: no waitpoint is created, so the parent completes as
soon as the chunks are dispatched. (Activepieces' waitpoint model is one per
`(run, step)`, so awaiting N parallel callbacks would need a separate aggregation design.)

Covered by 12 unit tests (`packages/pieces/core/subflows/test/process-csv-in-chunks.test.ts`):
chunk boundaries, the 5000-row fallback (empty / `0` / negative / `NaN`), fractional chunk
size truncation, empty CSV dispatching nothing, `additionalData` merging, parent-run
headers, headerless CSVs, non-comma delimiters, and the disabled-subflow guard.

### Relevant User Scenarios

- **Bulk import from a CSV** into an external system (CRM/ERP) whose API rate-limits large
  batches — each chunk is a separate run with its own pacing.
- **Per-row processing at scale** (HTTP call, AI enrichment, DB insert): work is spread over
  parallel child runs instead of one step that would hit `FLOW_RUN_TIME_SECONDS`.
- **Failure isolation**: if 3 of 50 chunks fail, the other 47 still succeed and only the
  failed chunks need to be retried, instead of reprocessing the whole file.
- **Human-in-the-loop upload**: pair it with a Web Form trigger so a user uploads the CSV
  and the flow fans the rows out to a processing subflow.

Fixes # (issue)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10652,15 +10652,17 @@
     },
     "packages/pieces/core/subflows": {
       "name": "@activepieces/piece-subflows",
-      "version": "0.4.15",
+      "version": "0.4.16",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
+        "csv-parse": "5.6.0",
       },
       "devDependencies": {
         "tslib": "2.6.2",
+        "vitest": "3.2.6",
       },
     },
     "packages/pieces/core/tables": {
--- a/packages/pieces/core/subflows/package.json
+++ b/packages/pieces/core/subflows/package.json
@@ -1,20 +1,23 @@
 {
   "name": "@activepieces/piece-subflows",
-  "version": "0.4.15",
+  "version": "0.4.16",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
     "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
     "bundle": "node ../../../../dist/packages/cli/src/index.js pieces bundle",
-    "lint": "eslint 'src/**/*.ts'"
+    "lint": "eslint 'src/**/*.ts'",
+    "test": "vitest run"
   },
   "dependencies": {
     "@activepieces/pieces-common": "workspace:*",
     "@activepieces/pieces-framework": "workspace:*",
     "@activepieces/core-piece-types": "workspace:*",
-    "@activepieces/core-utils": "workspace:*"
+    "@activepieces/core-utils": "workspace:*",
+    "csv-parse": "5.6.0"
   },
   "devDependencies": {
+    "vitest": "3.2.6",
     "tslib": "2.6.2"
   }
 }
--- a/packages/pieces/core/subflows/src/i18n/translation.json
+++ b/packages/pieces/core/subflows/src/i18n/translation.json
@@ -15,5 +15,20 @@
   "Callable Flow": "Callable Flow",
   "Waiting to be triggered from another flow": "Waiting to be triggered from another flow",
   "Sample Data": "Sample Data",
-  "The schema to be passed to the flow": "The schema to be passed to the flow"
+  "The schema to be passed to the flow": "The schema to be passed to the flow",
+  "Process CSV in Chunks": "Process CSV in Chunks",
+  "Split a CSV file into chunks and dispatch each chunk to a subflow (\"Callable Flow\") for processing. Set the chunk size below; when left empty it defaults to 5000 rows.": "Split a CSV file into chunks and dispatch each chunk to a subflow (\"Callable Flow\") for processing. Set the chunk size below; when left empty it defaults to 5000 rows.",
+  "CSV File": "CSV File",
+  "The CSV file to split and process.": "The CSV file to split and process.",
+  "The subflow that processes each chunk. Only published flows with a \"Callable Flow\" trigger appear here; disabled flows are marked \"(inactive)\".": "The subflow that processes each chunk. Only published flows with a \"Callable Flow\" trigger appear here; disabled flows are marked \"(inactive)\".",
+  "Does the CSV have headers?": "Does the CSV have headers?",
+  "Delimiter": "Delimiter",
+  "The delimiter used in the CSV file.": "The delimiter used in the CSV file.",
+  "Comma": "Comma",
+  "Tab": "Tab",
+  "Semicolon": "Semicolon",
+  "Additional Data": "Additional Data",
+  "Static data merged into every chunk payload sent to the subflow.": "Static data merged into every chunk payload sent to the subflow.",
+  "Chunk Size (rows)": "Chunk Size (rows)",
+  "Number of rows per chunk. Defaults to 5000 when left empty.": "Number of rows per chunk. Defaults to 5000 when left empty."
 }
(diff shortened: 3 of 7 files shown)


kishanprmr:
Thanks for the thorough PR.

A few things worth flagging before merge: most of this overlaps with actions that already exist.

- CSV → rows is already covered by Convert CSV to JSON in the core CSV piece.
- Fire-and-forget dispatch to a subflow is already what Call Flow does (with "Wait for Response" unchecked) — same headers, same disabled-flow check, same Callable-Flow filtering implemented here.
- Per-chunk iteration is the native Loop on Items step.

Chunking an array into groups of N rows, which today needs a small Code step.

Did you hit a specific limitation trying to compose CSV parse + Loop + Call Flow, or was it mainly about convenience? That'd help figure out if there's a real gap here beyond the chunking step.

ncodedude:
> Thanks for the thorough PR.
> 
> A few things worth flagging before merge: most of this overlaps with actions that already exist.
> 
> * CSV → rows is already covered by Convert CSV to JSON in the core CSV piece.
> * Fire-and-forget dispatch to a subflow is already what Call Flow does (with "Wait for Response" unchecked) — same headers, same disabled-flow check, same Callable-Flow filtering implemented here.
> * Per-chunk iteration is the native Loop on Items step.
> 
> Chunking an array into groups of N rows, which today needs a small Code step.
> 
> Did you hit a specific limitation trying to compose CSV parse + Loop + Call Flow, or was it mainly about convenience? That'd help figure out if there's a real gap here beyond the chunking step.

Thanks — that's a fair challenge, and you're right that most of this overlaps.

Being upfront: the dispatch half of this action is Call Flow. It reuses the same findFlowByExternalIdOrThrow and Callable-Flow filtering helpers, the same PARENT_RUN_ID_HEADER / FAIL_PARENT_ON_FAILURE_HEADER, and the same v1/webhooks/:flowId call. The parse half is Convert CSV to JSON with the same csv-parse options. Tellingly, the two P1s the bot just flagged are bugs I introduced by re-implementing that dispatch.

What pushed me toward a single action rather than composing:

Composing today is Read File → Convert CSV to JSON → Code (chunk) → Loop on Items → Call Flow. The Code step is exactly the gap you named, and it's the part I most wanted to avoid asking flow authors for.
Loop on Items runs sequentially and records an iteration entry per chunk, and each intermediate step output (file contents, parsed rows, chunked rows) lands in the run's execution state. On a large CSV that's the same dataset materialized a few times plus N iteration records, against MAX_FLOW_RUN_LOG_SIZE_MB. The single action emits only counters.
But I haven't benchmarked the composed pipeline head-to-head, so I don't want to claim a limitation I haven't measured. Let me build both and come back with numbers (step count, wall time, run-log size) on a ~100k-row file.

If the composition holds up, then the honest conclusion is that the only real gap is the chunking primitive — and I'd happily close this in favour of a much smaller PR adding a "Split into Chunks" action. Happy to go whichever way you prefer.

pacocartones (commented):
Chiming in from a flow-builder/business-automation angle rather than the code itself, since @kishanprmr's composability question is the crux here.

Agree the single action shouldn't stay as-is once you've confirmed it's mostly a re-implementation of Call Flow + Convert CSV to JSON — duplicated logic is exactly how the two P1 bugs Greptile caught happened. But I'd push back a little on defaulting straight to "just compose it with Loop on Items": for someone building this flow non-technically, Loop on Items iterates one item at a time and doesn't chunk into groups of N, so a user reaching for it to batch 50k CSV rows into groups of 5000 will either misuse it or bail out to a Code step — which is exactly the friction this PR is trying to remove.

That lines up with where the thread already landed: the real gap most users hit isn't "dispatch to a subflow" (Call Flow covers that) or "parse a CSV" (Convert CSV to JSON covers that) — it's "turn an array into groups of N," which has no native primitive today, CSV or otherwise. A small, generic "Split into Chunks" action (array in, array-of-arrays out) composed with the existing Call Flow would cover this use case and be reusable for anyone chunking API payloads or batch DB inserts too — not just CSV pipelines. It also sidesteps the run-log-size concern, since the chunking step's output is the only extra thing materialized, not a full CSV re-parse.

Happy to help think through the API shape for that smaller action if that's the direction you go.

ncodedude:
> Chiming in from a flow-builder/business-automation angle rather than the code itself, since @kishanprmr's composability question is the crux here.
> 
> Agree the single action shouldn't stay as-is once you've confirmed it's mostly a re-implementation of Call Flow + Convert CSV to JSON — duplicated logic is exactly how the two P1 bugs Greptile caught happened. But I'd push back a little on defaulting straight to "just compose it with Loop on Items": for someone building this flow non-technically, Loop on Items iterates one item at a time and doesn't chunk into groups of N, so a user reaching for it to batch 50k CSV rows into groups of 5000 will either misuse it or bail out to a Code step — which is exactly the friction this PR is trying to remove.
> 
> That lines up with where the thread already landed: the real gap most users hit isn't "dispatch to a subflow" (Call Flow covers that) or "parse a CSV" (Convert CSV to JSON covers that) — it's "turn an array into groups of N," which has no native primitive today, CSV or otherwise. A small, generic "Split into Chunks" action (array in, array-of-arrays out) composed with the existing Call Flow would cover this use case and be reusable for anyone chunking API payloads or batch DB inserts too — not just CSV pipelines. It also sidesteps the run-log-size concern, since the chunking step's output is the only extra thing materialized, not a full CSV re-parse.
> 
> Happy to help think through the API shape for that smaller action if that's the direction you go.

With the combination of these three pieces, we can’t split the CSV into processing batches.

Using Loop Items + Subflow, it triggers one subflow for every processed row. Our goal is different: we want to send a chunk of rows to each subflow. For example, if a CSV has 1,000 rows, we’d like to split it into four batches of 250 rows, resulting in only four subflow executions.

From what I understand, the only way to achieve this today is to first parse the CSV, then use a code piece to split the data into batches, and finally loop through those batches.

The challenge is that many users aren’t comfortable writing code, which makes this approach harder to implement. It also makes the flow less self-explanatory, since it’s not immediately obvious what that code block is doing. One of the goals of this change is to make the flow more explicit and easier to understand.

Does that make sense? Or is there another approach we’re missing that would solve this more cleanly?

abuaboud:

PR #14180, opened 2026-07-12
feat: large file streaming across pieces, engine, and webhooks

othmanemad:
## Summary

Adds pass-by-reference file streaming so large files move to/from object storage without being buffered in engine or API memory, lifting the practical size ceiling for large-file workflows while keeping the existing buffered path intact.

## What changed

**Engine SDK**
- `ctx.files.writeStream({ fileName, stream })` — pipes a `Readable`/`ReadableStream`/`AsyncIterable` straight to storage via S3 multipart, returning the usual file-reference string.
- `Property.File({ stream: true })` resolves inputs to an `ApFileRef` (pass-by-reference handle) consumed via `.stream()` or `.buffer()`.
- New `multipart-stream` util: chunking, running-total ceiling, abort-on-failure.

**Server**
- Streaming file sink for large binary and multipart webhook bodies, so ingestion no longer buffers whole files or trips `bodyLimit`. Flow existence is verified before the body is parsed; HMAC `rawBody` handling is unchanged for text bodies.
- Ceiling gated by `AP_MAX_STREAM_FILE_SIZE_MB` (default 1024, S3 only). On DB storage the existing `AP_MAX_FILE_SIZE_MB` cap applies as an honest fallback.

**Pieces migrated to streaming**
- `http`, `amazon-s3`, `google-drive`, `sftp`; `csv` chunk + `data-mapper` chunk-array helpers.

**Local dev**
- MinIO + SFTP services added to `docker-compose.dev.yml` for S3 testing.

**Docs**
- `files`, `properties`, `webhook-trigger`, and `environment-variables` updated for `writeStream`, `ApFileRef`, and the new env var.

## Editions

Streaming activates when S3-compatible storage is configured; DB-only installs keep working exactly as before, capped at `AP_MAX_FILE_SIZE_MB`. Zero-setup self-hosting is preserved.

## Notes / follow-ups

- Human Input Forms direct browser→S3 upload with a progress bar is intentionally **out of scope** here (tracked for later).
- Ceiling stays a global env var; plan-gating deferred pending billing input.

## Tests

- Unit: `file-service-save-stream`, `run-multipart-stream`, webhook payload-size.
- Integration (CE): webhook binary body, webhook multipart file, file multipart.


packages/pieces/core/csv/src/lib/actions/chunk-csv-file.test.ts
```diff
@@ -0,0 +1,45 @@
+import { Readable } from 'stream';
```
abuaboud, line 1: I think this file should be under tests folder (all other packages in the code base tests live near src folder)

PR #14200, opened 2026-07-13
fix: avoid chrome password breach warning for variables

21vedansh:
Changes the variable value input from type="password" to type="text" while using -webkit-text-security: disc to keep the value masked.

This prevents Chrome from treating the field as a password input, avoiding the native "Change your password" warning that could appear when creating or editing variables.

Fixes #14198


majewskibartosz:
we have pr already for this

PR #14211, opened 2026-07-13
fix(engine): fail RESUME cleanly on gone payload/log file instead of INTERNAL_ERROR

amrdb:
## Problem

When the engine resolves a flow's inputs before execution, it downloads files: the **trigger payload** on `BEGIN`, or the **run-log / resume payload** on `RESUME`. If such a file was deleted or expired, `engine-file-api.ts` deliberately throws a **USER-typed** `EngineFileNotFoundError` (see its comment: *"a gone file … is a data-lifecycle issue, not an engine bug — surface it as a USER error so callers can fail the run cleanly"*).

But `flowOperation.execute()` only honored that on `BEGIN`:

```ts
if (operation.executionType === ExecutionType.BEGIN && !isEngineExecutionError(resolveError)) {
    return reportFailedTriggerRun({ operation, error: resolveError })
}
throw resolveError
```

On `RESUME`, the `=== BEGIN` gate is false, so the USER-level `EngineFileNotFoundError` (a resume whose log/payload file expired out of retention) is **rethrown → `INTERNAL_ERROR`**, which fails + retries the worker job and **pages oncall** — and retrying can never recover an already-expired file.

## Fix

Drop the `BEGIN`-only restriction so *any* non-ENGINE resolve error reports a FAILED run, regardless of execution type. Genuine ENGINE errors (e.g. a transient 5xx download) still take the `throw` path and keep paging/retrying.

## Scope / notes

- This handles the **retention** case (file genuinely gone → 404/410 → `EngineFileNotFoundError`, USER).
- It intentionally does **not** touch `EmptyResumeStateError` / `ResumeLogsFileMissing` / `ExecutionStateMissing` — those fire when the file is *present but empty/malformed*, which is not retention and stays ENGINE (loud) on purpose.

Found while categorizing the `INTERNAL_ERROR` backlog on cloud (RESUME jobs failing on gone payload/log files).

### DIFF
--- a/packages/server/engine/src/lib/operations/flow.operation.ts
+++ b/packages/server/engine/src/lib/operations/flow.operation.ts
@@ -1,5 +1,5 @@
 import { isNil, tryCatch } from '@activepieces/core-utils'
-import { BeginExecuteFlowOperation, EngineGenericError, EngineResponse, EngineResponseStatus, ExecuteFlowOperation, ExecuteTriggerResponse, ExecutionError, ExecutionErrorType, ExecutionState, ExecutionType, FlowActionType, FlowRunStatus, flowStructureUtil, GenericStepOutput, LoopStepOutput, ResumePayload, ResumeReason, StepOutput, StepOutputStatus, TriggerHookType, TriggerPayload } from '@activepieces/shared'
+import { EngineGenericError, EngineResponse, EngineResponseStatus, ExecuteFlowOperation, ExecuteTriggerResponse, ExecutionError, ExecutionErrorType, ExecutionState, ExecutionType, FlowActionType, FlowRunStatus, flowStructureUtil, GenericStepOutput, LoopStepOutput, ResumePayload, ResumeReason, StepOutput, StepOutputStatus, TriggerHookType, TriggerPayload } from '@activepieces/shared'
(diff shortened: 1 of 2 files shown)


amrdb:
Addressed both review points in 63296d3:

- **P1 (compile error):** widened `ReportFailedTriggerRunParams.operation` to `ExecuteFlowOperation` and branch on `executionType` to build a valid `ResolvedExecuteFlowOperation` for each case. For RESUME the `resumePayload`/`executionState` are placeholders — the failed-run path (`buildFailedTriggerContext` → line 138) only reads `flowVersion` + `executionType`, so they're never consumed. Removed the now-unused `BeginExecuteFlowOperation` import.
- **P2 (missing test):** added a RESUME test asserting a gone log file (`EngineFileNotFoundError`) surfaces as a FAILED run + `EngineResponseStatus.OK` instead of `INTERNAL_ERROR`, mirroring the BEGIN case.

Note: the existing `EmptyResumeStateError` / `ResumeLogsFileMissing` / `ExecutionStateMissing` tests (which assert `throws`) are intentionally left ENGINE-typed — those fire when the file is present-but-empty (a real state-loss signal), not retention.

AhmadTash (approved):

PR #14215, opened 2026-07-14
feat(pieces): pinterest conversions piece

alcwilliams:
## What does this PR do?

Adds **Pinterest Conversions** as a new community piece for sending server-side web, app, and offline conversion events to Pinterest through the [Conversions API](https://developers.pinterest.com/docs/api/v5/events-create/).

This piece is intentionally separate from the existing `pinterest` piece. The existing piece uses OAuth2 and focuses on organic Pinterest activity such as pins, boards, and followers. Pinterest conversion reporting uses a different authentication model and API surface

The piece authenticates using a **Conversion Access Token** and **Ad Account ID** through `CustomAuth`, so no OAuth application setup is required.

The **Send Conversion Event** action supports:

- web, offline, Android, and iOS events;
- standard Pinterest event names and custom events;
- customer matching data such as email, phone, address, external ID, MAID, IP address, and user agent;
- conversion details such as value, currency, content IDs, order ID, and item count;
- event deduplication using `event_id`;
- Pinterest Test Mode for validating requests without recording conversions.

Personal identifiers are normalized and SHA-256 hashed before being sent. Values already provided as SHA-256 hashes are passed through unchanged to avoid double-hashing. The action also validates that at least one supported match key is present.

A **Custom API Call** action is included for other Pinterest Conversions API endpoints. Both actions include `audience` and `aiMetadata` so they can be used in flows and by AI agents.

### Screenshots

**Send Conversion Event**

Send Conversion Event form

**Customer matching and automatic hashing**

Customer matching fields

### Relevant user scenarios

1. **Server-side checkout tracking** — report a `checkout` event after an order is confirmed in Stripe, Shopify, or another system.
2. **Offline conversions** — send in-store sales or CRM-closed deals using `action_source: offline`.
3. **Lead capture** — report a `lead` event when a form is submitted or a lead is qualified.
4. **Deduplicated tracking** — use the same `event_id` in the browser Pinterest tag and this action so Pinterest can deduplicate the events.

---

### Testing performed

- `npx turbo run build --filter=@activepieces/piece-pinterest-conversions`
- `npx turbo run lint --filter=@activepieces/piece-pinterest-conversions`
- Request payload verified against Pinterest's published OpenAPI specification, including required fields, enums, and match-key validation.
- PII normalization and SHA-256 hashing verified with worked examples.
- Live end-to-end request sent to the Pinterest Conversions API using a real Conversion Access Token and synthetic customer data.
- Test Mode was enabled, so the event was validated but not recorded.

The API returned:

```text
num_events_received: 1
num_events_processed: 1
events[0].status: processed
```

Successful Pinterest test response

### Not exercised live

Only the happy path for one `add_to_cart` web event was tested against the live API. Other event types and error responses, including 401 and 422 responses, are handled in code but were not triggered during live testing.


cavemaann:
The alternative would be to add this action to the existing Pinterest piece, but that would require expanding its OAuth scopes to include `ads:write`.

abuaboud:

PR #14220, opened 2026-07-14
fix(engine): retry 5xx on store read/write instead of failing the run

amrdb:
## Problem

The engine store client (`store.ts`) reads/writes flow store entries with **raw `fetch`** and **no retry** (3 sites: `get`/`put`/`delete`). A transient 5xx from the store API — e.g. a Cloudflare HTML error page during a blip — becomes a `StorageError` → `INTERNAL_ERROR`, which fails+retries the worker job and pages oncall.

Surfaced while categorizing the `INTERNAL_ERROR` backlog on cloud: `StorageError` jobs whose underlying cause was a Cloudflare 5xx error page.

## Fix

Wrap the three store fetches in `fetch-retry` with the same 5xx retry policy the file/callback clients use (`retryOn: [408, 429, 500, 502, 503, 504]`). Transient blips are absorbed; a persistent outage still surfaces as `StorageError` after retries.

4xx responses are unaffected — `handleResponseError` still short-circuits 404 (→ null), 413 (→ StorageLimitError), and other 4xx (→ StorageInvalidKeyError) without retry, since only 5xx/408/429 are in `retryOn`.

Part of the same "transient engine→API 5xx shouldn't nuke runs" theme as #14202 (file + progress/terminal callbacks); this covers the store client.


AhmadTash (approved):

amrdb:
Superseded by #14474, which routes the store client and the five other bare-`fetch` engine→API clients (app-connection, variable, flows, waitpoint, worker-project) through one shared `retryFetch` with the same policy. Closing in favour of that.

PR #14223, opened 2026-07-14
feat(security): migrate encryption at rest from AES-256-CBC to AES-256-GCM

othmanemad:
## What

Moves at-rest secret encryption from **AES-256-CBC** (confidentiality only, no auth tag) to authenticated **AES-256-GCM**, through the single `encryptUtils` chokepoint. Algorithm-only migration — the 32-char `AP_ENCRYPTION_KEY` is unchanged, no new env var, self-hosting stays zero-setup.

Secrets covered: app-connection credentials, project variables, AI provider/tool keys, EE OAuth-app secrets, secret-manager configs, and the OIDC signing key.

## How

`EncryptedObject` becomes `{ iv, data, authTag? }`. **Presence of `authTag` is the format discriminator** — absent ⇒ legacy CBC, present ⇒ GCM. No version field, no keyId, no keyring.

This PR combines the first two of a three-release rollout:

- **Release 1 (expand)** — `decrypt*` gains a GCM read branch (`isNil(authTag)`); `encryptString` still writes CBC. Dormant reader, fully rollback-safe.
- **Release 2 (migrate)** — `encryptString` flips to GCM (12-byte IV + auth tag). A new `SystemJobName.REENCRYPT_SECRETS` background job re-encrypts legacy CBC blobs in place: keyset-paginated on the PK, paced (200 rows / 200 ms), single-runner via `distributedLock`, restart-safe, idempotent (re-enqueued each boot, dedup by `jobId`). It covers all seven blob locations across every edition (EE tables skipped via `42P01` when absent); `flag` is restricted to the OIDC key row.

**Migration safety:** write-back is conditional on the **exact CBC blob read** (`col::jsonb = $3::jsonb`), not merely `authTag IS NULL` — so a concurrent re-auth (including a CBC write from a not-yet-upgraded instance during rollout) is never clobbered; such a row is simply swept later.

## ⚠️ Deploy ordering (load-bearing)

There is **no plaintext fallback** — an instance that can't read a blob loses that secret. So **Release 1's reader must reach 100% of the fleet before Release 2's GCM writes begin**. Since this PR contains both, sequence the deploy accordingly (or a not-yet-upgraded instance will hit an unreadable GCM blob, and the write-flip becomes irreversible). Rationale in [ADR 0007](docs/adr/0007-encryption-at-rest-migrated-cbc-to-gcm.md).

**Release 3 (contract, later):** delete the CBC branch, make `authTag` required, remove the sweep — once `authTag IS NULL` counts are zero fleet-wide for a full release cycle.

## Verification

- `tsc` (production source): 0 errors · `turbo lint --filter=api`: 0 errors
- Unit `encryption.test.ts`: 7/7 — GCM round-trip, tamper rejection, legacy CBC read
- Integration `reencrypt-secrets.test.ts` (real Postgres): 2/2 — CBC→GCM preserves plaintext (12-byte IV), already-GCM rows untouched

Docs: [ADR 0007](docs/adr/0007-encryption-at-rest-migrated-cbc-to-gcm.md) + feature registry `.agents/features/secret-encryption.md`.

### DIFF
--- a/.agents/features/secret-encryption.md
+++ b/.agents/features/secret-encryption.md
@@ -1,0 +1,41 @@
+# Secret Encryption at Rest
+
+## Summary
+All secrets stored in the database — app-connection credentials, project variables, AI
+provider/tool keys, EE OAuth-app secrets, secret-manager configs, and the OIDC signing key — are
+encrypted through a single chokepoint, `encryptUtils`. As of the CBC→GCM migration, secrets are
+written with **AES-256-GCM** (authenticated encryption: ciphertext + 12-byte IV + auth tag). Blobs
+written before the migration use **AES-256-CBC** (`{ iv, data }`, no auth tag) and are still
+readable; a background sweep (`reencryptSecretsJob`) re-encrypts them to GCM in place.
+
+The stored shape is `EncryptedObject = { iv, data, authTag? }`. **The presence of `authTag` is the
+format discriminator**: absent ⇒ legacy CBC, present ⇒ GCM. The 32-character `AP_ENCRYPTION_KEY` is
+the AES-256 key for both algorithms (`Buffer.from(secret, 'binary')` → 32 bytes) — the migration is
+algorithm-only, no key rotation and no new env var.
+
+The sweep is a one-shot system job (`SystemJobName.REENCRYPT_SECRETS`), single-runner via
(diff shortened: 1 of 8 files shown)


othmanemad:
Needs further research regarding enterprise encryption services

Issue #5357, opened 2024-08-14
[BUG]: Unable to disable sign ups


michaeladams:
**Describe the bug**
In the docs, it states that you can use the AP_SIGN_UP_ENABLED env flag to enable sign ups. (And I assumed you could turn them off by setting it to false, maybe incorrectly?)
https://www.activepieces.com/docs/install/configurations#enabling-sign-up

This no longer seems to be the case from 0.28, as per breaking changes:
https://www.activepieces.com/docs/about/breaking-changes#0-28-0

> The SIGN_UP_ENABLED environment variable, which allowed multiple users to sign up for different platforms/projects, has been removed. It has been replaced with inviting users to the same platform/project. All old users should continue to work normally. 

My use case is to be able to entirely disable sign-ups, as in lock the system down to logins only.

Not sure if this is actually a regression, or the docs need to be reworded.

**To Reproduce**

Toggle the SIGN_UP_ENABLED flag to false, set sandboxed/privledge mode as per docs, and restarting the docker-container

**Expected behavior**
Sign ups to be disabled

abuaboud:
Hi @michaeladams 

The signed up is disabled after the first user sign up, then it's by invitation only from the platform admin, please note the invited users will be also be admins, and have same privilege, as the permission module is not available in community edition.

So as long you don't invite anyone, no one can sign up for your instance, (they will get an error message when signing up)

Feel free to reopen, If you have any other questions.

PR #14224, opened 2026-07-14
fix(builder): restore Download File button for step file outputs in the Friendly View

### LINKED ISSUE #14193: [BUG] Download File button missing from step output: new /api/v1/files/ URLs not detected
## Problem

File outputs in the step output viewer render as raw URL text instead of the "Download File" button. Regression in 0.84.0, still present on main (0.86.2). Reported on 0.86.0 by a paid embed customer.

Ref: Pylon 5357

## Root cause

PR #13103 (shipped 0.84.0) unified file storage: `files.write()` read URLs changed from `/api/v1/step-files/signed?token=...` to `/api/v1/files/<fileId>?token=<jwt>` (`packages/server/api/src/app/file/files-service.ts:31`). Frontend detection never updated:

`packages/web/src/lib/dom-utils.ts:7`

```ts
json.includes('/api/v1/step-files/') || json.includes('file://')
```

Download File button only renders on that match (`packages/web/src/components/custom/json-viewer.tsx:114-160`: Raw JSON tab and direct file-string outputs). Friendly View (`smart-output-viewer`) has no file rendering at all, shows the raw URL text.

## Fix

Add `/api/v1/files/` to `isStepFileUrl`. Consider a download button renderer in the Friendly View too, since it is the default tab.

AhmadTash:
## Problem

File outputs in the step output viewer render as raw URL text instead of a **Download File** button. Regression since 0.84.0, still present on `main`. Reported by a paid embed customer (Pylon #5357).

**Root cause:** PR #13103 unified file storage and changed `files.write()` read URLs from `/api/v1/step-files/signed?token=` to `/api/v1/files/<id>?token=<jwt>`. The frontend detector `isStepFileUrl` (`packages/web/src/lib/dom-utils.ts`) only matched the old path — so the button never rendered, and the **Friendly View** (the default output tab) had no file rendering at all.

## Fix

- **Detect the new route.** `isStepFileUrl` now also matches `/api/v1/files/`, but requires a real `?token=`/`&token=` query param so a generic third-party path doesn't false-positive. In particular, Pipedrive's own API is literally `<domain>/api/v1/files/…` and it authenticates with `?api_token=` — which *contains* the substring `token=`. Matching the token as a discrete query param excludes `api_token`/`access_token`.
- **Render in the Friendly View (default tab).** Extracted a shared `StepFileDownloadButton` and render it for file-URL leaf values across every friendly-view renderer (`OutputTextDisplay`, `FormatSingleValue`, `ValueRow`, `SchemaListItemRow`, `CellValue`).
- **Raw JSON shows the raw link.** `json-viewer` is reverted to a plain raw renderer (removed the imperative `useLayoutEffect`/`createRoot` DOM-scan button injection). The button now lives exclusively in the Friendly View.
- **Tests.** New `isStepFileUrl` unit suite covering the new route, the legacy route, `file://`, and the third-party `api_token`/`access_token` non-matches.

## Behavior

| Tab | File output |
|---|---|
| Friendly View (default) | **Download File** button |
| Raw JSON | raw signed link (text) |

Works for bare-string file outputs (e.g. QR Code) and object outputs like `{ fileName, url }` (e.g. File Helper → Create file).

## Test plan

- [x] `isStepFileUrl` unit tests — new route, legacy route, `file://`, and `api_token`/`access_token` exclusions
- [x] Full `packages/web` unit suite green (372 tests)
- [x] ESLint clean
- [x] Manually verified in a local instance: QR Code and File Helper outputs show the **Download File** button in Friendly View and the raw link in Raw JSON; clicking downloads the file via the signed `/api/v1/files/` route

Closes GIT-1618.
Fixes #14193


AhmadTash:
@greptile review

AliMahmoudDev:
Nice UI work! One suggestion: using `clamp()` for font sizes would make this naturally responsive without breakpoints. Like `font-size: clamp(0.875rem, 0.5rem + 1vw, 1.25rem)`.

AliMahmoudDev:
Good fix for the download button! Restoring functionality that was accidentally removed is exactly the kind of fix that prevents user frustration. Adding a test to prevent this regression would be a great follow-up.

Louai-Zokerburg (approved):

PR #14226, opened 2026-07-14
feat(files): accept streamable data in context.files.write

othmanemad:
## What

`context.files.write` now accepts a Node `Readable` (with a required byte `size`) in addition to a `Buffer`:

```ts
write({ fileName, data: Buffer })                  // unchanged
write({ fileName, data: Readable, size: number })  // new
```

## Why

Pieces producing/downloading large files must currently materialize the whole thing as a `Buffer`, which OOMs the sandbox. Streaming lets the bytes flow engine→storage without buffering.

## How

- **S3 signed-URL transport** (the only path that can avoid buffering): the engine does a small preflight (`POST /v1/files/:fileId/upload-target`) to obtain a pre-signed S3 PUT URL, then streams the `Readable` straight to S3 with `Content-Length: size`. Nothing materializes in engine or app memory.
- **Any other transport** (default self-hosted DB `bytea`, or proxy→S3): transparent buffer-fallback — the engine streams to the app, which buffers within its `bodyLimit`. Always correct, no memory win there (documented, not silently broken).
- A stream **must** carry its `size`: a single pre-signed S3 PUT needs a `Content-Length` (chunked is rejected by R2/MinIO/OCI, and one signed URL can't do multipart). The engine can't read the transport from its env and can't replay a single-use stream through the redirect flow — hence the preflight.
- New `MAX_STREAM_FILE_SIZE_MB` system prop (default `512`), enforced server-side at the preflight.
- An oversize stream returns `413`, which the engine translates to a **USER-level** `FileSizeError` (FAILED step) rather than `EngineGenericError` — so a user mistake doesn't trip `INTERNAL_ERROR` and page oncall.

Buffer callers are untouched (single-PUT + redirect-replay, no extra roundtrip). `FLOW_STEP_FILE` writes remain uncompressed.

## Docs

- ADR-0007 records the "stream must carry its size / only S3-signed avoids buffering" decision.
- Feature doc (`.agents/features/file-storage.md`) and the Data & Storage glossary updated.

## Testing

- `file-uploader.test.ts`: 11/11 pass (added s3 / proxy / oversize→FileSizeError stream cases).
- Lint clean on framework + engine + api.


abuaboud:
I think we already have this merged

PR #14229, opened 2026-07-14
fix(ai-providers): Azure OpenAI connection fails to validate and model dropdown shows empty options

### LINKED ISSUE #11132: Azure OpenAI Provider Configuration Fails with 404/500 Errors
Azure OpenAI cannot be configured in Admin Console → AI Providers. The setup fails when trying to verify the connection.
Steps to Reproduce:

Go to Admin Console → AI → Add AI Provider

1. Select "Azure OpenAI"
2. Configure with:
    - Resource Name: 
    - API Key: Valid Azure OpenAI API key
    - API Version: 
3. Click Save

Configuration fails with errors in logs:

> First attempt: HTTP 500 errors - "Internal Server Error". This is because i have mistakenly put the entire azure resource url in Resource Name. After changing resource name format, i am getting  HTTP 404 errors - "Resource not found". Direct API calls to Azure OpenAI work correctly using the same credentials. 

Server logs show the system attempts to call listModels endpoint on Azure OpenAI, which results in 404 errors. The error message states: "Resource not found" at the Azure provider's listModels function.

majewskibartosz:
Fixes [GIT-1310](https://linear.app/activepieces/issue/GIT-1310)
Closes #11132

## Problem
1. Saving an Azure OpenAI provider with API Version empty always fails with 404: `listModels` calls Azure's retired data-plane `GET /openai/deployments` with the default `2024-10-21`; that endpoint only answers on `2022-12-01` / `2023-03-15-preview`, and `validateConnection` delegates to `listModels`.
2. With `2023-03-15-preview` configured the connection saves, but the Model dropdown in AI pieces renders `[{}, {}, {}]`: the code mapped `deployment.name`, a property Azure's response never had (deployment name lives in `id`).

## Fix
- `azure-provider.ts`: map `deployment.id`; pin the deployments-list call to `2023-03-15-preview` — the configured `apiVersion` only affects inference (`@ai-sdk/azure` targets the GA v1 endpoint).
- `@activepieces/shared`: drop now-unused `DEFAULT_AZURE_API_VERSION`, bump 0.117.0 → 0.118.0.
- `createMockAIProvider` helper: `??` → `||` so `OPENAI_API_KEY=` (empty string) in `.env.tests` no longer yields an empty mocked apiKey.

## Verification
- Unit regression test, red-first: fails pre-fix (`name`/`id` undefined, URL carried `2024-10-21`), green post-fix.
- Integration test `test/integration/ce/ai-provider/azure-models.test.ts` (CI runs `test-ce`; server-api `test/unit` is not wired into CI): full stack `GET /v1/ai-providers/azure/models` with mocked outbound HTTP → 200, deployment id mapped, URL pinned.
- Runtime: standalone API boot + throwaway Postgres — boots clean with the removed shared export; Azure create path drives `validateConnection` → `listModels` and fails cleanly on an unreachable resource. Live Azure listing UNVERIFIED (no Azure credentials); response shape confirmed against the official [2023-03-15-preview spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/authoring/preview/2023-03-15-preview/azureopenai.json).
- `npm run lint-dev` clean; engine+shared unit tests clean. Pre-existing server-api unit failures (chat/canary/bullmq) reproduce on pristine upstream/main — unrelated.
- Visual: N/A, no UI changes.

Follow-up (out of scope, backend-only PR): the Azure form's API Version helper text (`upsert-provider-config-form.tsx` + locale files) still suggests `2023-03-15-preview`; after this fix that value only affects inference.


majewskibartosz:
Re the P2 on `AZURE_DEPLOYMENTS_API_VERSION` placement: this is intentional repo style, not an oversight. Root `CLAUDE.md` mandates the file order **Imports → Exported functions/constants → Helper functions → Types** — non-exported module-level constants and local types go *after* the exported logic so the public contract stays front and centre. Sibling providers follow the same layout (`openai-provider.ts`, `openrouter-provider.ts` declare their local response types at the bottom). The constant is only read at call time inside `listModels`, so there's no TDZ concern. @greptile-apps please remember this convention for this repo.

MrChaker (approved):

PR #14238, opened 2026-07-14
fix(schedule): Every X Minutes trigger fires twice per hour for intervals that don't divide 60

### LINKED ISSUE #14235: [BUG] Every X Minutes trigger fires twice per hour for interval values that don't divide 60
## Problem

Schedule piece "Every X Minutes" with X=59 runs twice per hour (:59 and :00, 1-3 min apart), not once every 59 min. All 48 dropdown values that don't divide 60 misbehave: X=31..59 double-fire at :00 and :X; the rest give uneven gaps (45 → 45/15 min). Only 1-6, 10, 12, 15, 20, 30 behave as labeled.

Reported by a paid cloud embed customer. Reproduced on cloud: one published flow with X=59 → runs at 16:59:00Z and 17:00:01Z.

Ref: Pylon 5368

## Root cause

`packages/pieces/core/schedule/src/lib/triggers/every-x-minutes.trigger.ts:29`:

```ts
const cronExpression = `*/${ctx.propsValue.minutes} * * * *`;
```

Passed verbatim to BullMQ repeatable pattern (`packages/server/api/src/app/workers/job-queue/job-queue.ts:36`). Cron `*/X` on the minute field = "minutes of the hour divisible by X", not a rolling interval. `*/59` → minute 0 and minute 59.

## Fix

Options:

1. Restrict dropdown to divisors of 60 (1-6, 10, 12, 15, 20, 30).
2. True fixed intervals: BullMQ `repeat.every` (ms) instead of cron pattern. Needs `ScheduleOptions` change through shared → engine → server.

Either way, existing published flows with non-divisor values keep misfiring until republished.

## Workaround

Every Hour trigger, or Cron Expression trigger `59 * * * *`; republish.

majewskibartosz:
Fixes activepieces/activepieces#14235<br>Closes activepieces/activepieces#14235

## Problem

1. Schedule "Every X Minutes" = 59 fires twice per hour (:59 and :00, 1-3 min apart): cron `*/X` means "minutes of the hour divisible by X", not a rolling interval. All 48 dropdown values that don't divide 60 misbehave (31-59 double-fire; the rest get uneven gaps). Reproduced on cloud: one published flow, runs at 16:59:00Z + 17:00:01Z.

## Fix

* `ScheduleOptions` becomes a discriminated union on the existing `type` field: `CRON_EXPRESSION` (unchanged) | `INTERVAL { intervalMs }` in `core-piece-types` + `shared`; legacy jsonb rows parse unchanged
* pieces-framework `setSchedule` accepts `{ intervalMs }` (`SetScheduleRequest`); engine validates via `ScheduleOptions.safeParse` (integer ≥ 60000) → new `InvalidScheduleIntervalError`
* job-queue: INTERVAL → BullMQ `upsertJobScheduler({ every, startDate: now + intervalMs })` — `startDate` preserves deferred first fire (bare `every` fires instantly on every publish/enable); same guard added in platform-queue-migration
* refill-polling-jobs passes the stored schedule through so INTERVAL rows survive refills
* web flows-utils: interval tooltip "Run every N minutes" (ICU plural key); timer icon keys on schedule presence
* schedule piece emits `intervalMs`; 0.1.18 with `minimumSupportedRelease: 0.86.4` — flows pin `~0.1.x`, so old engines must never resolve the new piece (their `setSchedule` throws on the new shape)
* bumps: shared 0.118.0, core-piece-types 0.3.0, core-execution 0.9.0, pieces-framework 0.34.0

Not covered (deliberate): existing published flows keep their old cron until republished (per ticket); trigger `run()` payload still emits `cron_expression` for back-compat with flows referencing it (documented in `.agents/features/triggers.md`); default poll interval in flow-trigger-side-effect still builds `*/X` cron — only wrong for non-divisor `AP_TRIGGER_DEFAULT_POLL_INTERVAL` overrides, follow-up.

## Verification

* Regression tests proven red-first on base code: piece test (pre-fix emits `cronExpression: "*/59 * * * *"`, post-fix `{ intervalMs: 3540000 }`) and shared schema tests (legacy cron rows parse; INTERVAL bounds enforced)
* `npm run test-unit` clean (engine + shared + web), `npm run lint-dev` 0 errors, api/web/engine builds clean (engine tsc shows only the two pre-existing errors also present on main)
* End-to-end on a booted API + worker: publishing an every-59-min flow registers BullMQ scheduler `every: 3540000` with no cron pattern; a legacy cron-expression flow registers `pattern: "59 * * * *"` — both asserted directly in the scheduler store
* Piece-level vitest suite runs locally; CI coverage is carried by the shared schema tests (piece tests aren't in the test-unit filter)
* Visual: flows-table schedule tooltip driven live for both flows — interval shows "Run every 59 minutes", legacy cron shows "Run at 59 minutes past the hour"; screenshots in the Visual verification comment


majewskibartosz:
## Visual verification

Driven live in the flows table (real API + worker + published flows, light theme — evidence pass for incidental visual impact; the change is a tooltip text path, no layout/style change).

| State | Screenshot |
|---|---|
| Interval schedule (Every X Minutes = 59) — tooltip on status timer icon | interval tooltip |
| Legacy cron schedule (`59 * * * *`) — tooltip unchanged, interval tooltip also visible below | cron tooltip |

States driven: interval flow tooltip ("Run every 59 minutes", new INTERVAL render path), legacy cron flow tooltip ("Run at 59 minutes past the hour", cronstrue path unchanged), timer icon renders for both schedule types, webhook flows unaffected (null schedule → "Real time flow" path untouched). Before/after: pre-fix an interval schedule could not exist, so the new state has no "before"; the legacy cron tooltip is byte-identical to pre-fix.

PR #14257, opened 2026-07-15
feat(engine): add batch mode to Loop on Items

ncodedude:
## What does this PR do?

Adds an optional **Batch Size** to the **Loop on Items** step.

When Batch Size is set to `N > 1`, each iteration receives a **list of N items** instead of a
single item, so a downstream step can process the whole batch in one call — e.g. Tables →
Create Record(s), or an HTTP request whose body is a JSON array.

Today, "iterate an array in groups of N" requires dropping in a Code step to chunk the array
first. This makes chunking a native, reusable loop primitive. It also reduces the number of
per-item iteration records the loop writes into the run's execution state from `N` to
`ceil(N / batchSize)`, which matters on large lists against `MAX_FLOW_RUN_LOG_SIZE_MB`.

Fully backwards compatible: when Batch Size is empty or `<= 1`, the loop iterates item-by-item
exactly as before. The item mention (`{{loop.output.item}}`) is unchanged — in batch mode it
simply resolves to the batch array.

- **Schema** (`LoopOnItemsActionSettings`): a new optional `batchSize: number`. Being optional,
  existing flows stay valid and no flow-schema migration is needed.
- **Engine** (`loop-executor.ts`): after resolving `items` (and `batchSize`) through the props
  resolver — inside the existing `tryCatchAndThrowOnEngineError` wrapper — it derives the units
  to iterate over: the raw items when `batchSize <= 1` (unchanged path), or `chunk(items, batchSize)`
  when `> 1`, using `chunk()` from `@activepieces/core-utils` (already bundled in the engine).
  `resolveBatchSize` normalizes empty / `<= 0` / `NaN` to 1 and truncates fractional values.
  Nothing else in the loop (iteration records, current path, verdict handling, single-step test
  mode) changes.
- **Builder**: the Loop step settings gain an optional numeric **Batch Size** input, stored as a
  number.

Covered by unit tests in `packages/server/engine/test/handler/flow-looping.test.ts`:
- 5 items with `batchSize: 2` → 3 iterations, last batch partial (`[5]`);
- `batchSize` of `1` / `0` / negative / `NaN` → item-by-item (unchanged behavior);
- fractional `batchSize` truncated;
- `batchSize` greater than the list → a single batch with all items;
- empty list with a batch size → zero iterations.

### Relevant User Scenarios

- **Bulk insert into a Table**: `Loop (batchSize: 50) → Tables Create Record(s)` inserts 50 rows
  per call instead of one row per iteration.
- **Batched API calls**: send an array of N records per HTTP request to a rate-limited endpoint.
- **Large lists**: fewer, larger iterations keep the run's execution log smaller and the loop
  faster than one iteration per item.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -360,7 +360,7 @@
     },
     "packages/core/execution": {
       "name": "@activepieces/core-execution",
-      "version": "0.8.0",
+      "version": "0.9.0",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -405,7 +405,7 @@
     },
     "packages/core/shared": {
       "name": "@activepieces/shared",
-      "version": "0.117.0",
+      "version": "0.118.0",
       "dependencies": {
         "@activepieces/core-execution": "workspace:*",
         "@activepieces/core-formula": "workspace:*",
--- a/packages/core/execution/package.json
+++ b/packages/core/execution/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/core-execution",
-  "version": "0.8.0",
+  "version": "0.9.0",
   "type": "commonjs",
   "main": "./dist/src/index.js",
   "scripts": {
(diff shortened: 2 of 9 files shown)


abuaboud:

PR #14266, opened 2026-07-16
feat(pieces): reddit conversions piece

alcwilliams:
## What does this PR do?

Adds **Reddit Conversions** as a new community piece for sending server-side web, app, and offline conversion events to Reddit through the [Conversions API (v3)](https://business.reddithelp.com/helpcenter/s/article/conversions-api).

This is separate from the existing reddit piece because conversion reporting uses token-based authentication and a different API surface from Reddit’s OAuth-based organic actions.

Auth is a **Conversion Access Token** + **Pixel ID** via `CustomAuth`, so no OAuth app to register.

The **Send Conversion Event** action covers:

- website, app, physical-store, and "other" events;
- all the standard Reddit event types (Page Visit, View Content, Search, Add to Cart, Add to Wishlist, Purchase, Lead, Sign Up) plus custom events;
- customer match keys such as click ID, email, phone, external ID, IP, user agent, Reddit UUID, and mobile ad IDs (IDFA/AAID);
- event details like value, currency, item count, and products;
- deduplication via `conversion_id` (auto-generated if you leave it blank);
- Limited Data Use (LDU) for privacy-restricted regions;
- Reddit's Test Mode via a Test ID, so you can validate without recording live conversions.

Email, phone, external ID, and mobile advertising IDs (IDFA/AAID) are normalized and SHA-256 hashed before being included in the request; IP address, user agent, and Reddit UUID are sent unhashed, as Reddit requires. Anything you pass in that's already a SHA-256 hash is left alone so you don't double-hash. The action also checks that at least one usable match key is present before sending, and only accepts an `event_at` within the last 7 days (sent as an int64 epoch-ms timestamp, per Reddit's spec).

There's also a **Custom API Call** action. 

### Screenshots

**Send Conversion Event — tested successfully**

Screenshot 2026-07-16 at 12 09 48 pm

### Relevant user scenarios

1. **Server-side checkout tracking** — fire a `Purchase` event once an order is confirmed in Stripe/Shopify/etc., instead of relying on the browser pixel alone.
2. **Offline conversions** — push in-store sales or CRM-closed deals with `Action Source: Physical Store`.
3. **Lead capture** — send a `Lead` or `Sign Up` when a form comes in or a lead is qualified.
4. **Deduplicated tracking** — use the same `conversion_id` here and in the browser Reddit Pixel so Reddit dedupes the two.

---

### Testing performed

- `npx turbo run build --filter=@activepieces/piece-reddit-conversions`
- `npx turbo run lint --filter=@activepieces/piece-reddit-conversions`
- Unit tests for PII normalization + SHA-256 hashing and for the match-key / attribution logic (`vitest`, 12 tests).
- Email, phone, IDFA, AAID, and external ID hashing verified to reproduce Reddit's published canonicalization example hashes ([Manual Advanced Matching](https://business.reddithelp.com/s/article/advanced-matching-for-developers)).
- Live end-to-end request sent to the Conversions API with a Conversion Access Token + Pixel ID + Test ID set, so the event was validated but not recorded as a live conversion. See screenshot above for what the API returned.

### Not exercised live

Only the happy path for a single Page Visit website event was run against the live API. Other event types and the error responses (401/403/404/422/429) are mapped to friendly messages in code but weren't all triggered live — though the 400 path was hit and confirmed during development, which is how the error surfacing (error.message + per-field details) got validated.


abuaboud:

PR #14273, opened 2026-07-16
feat(flows): add batching to loop on items

othmanemad:
## What

Adds an optional **Batching** mode to the **Loop on Items** action. Instead of iterating one element at a time, the loop can split the resolved array into sub-arrays ("batches") and iterate over the batches. Two modes:

- **Items per batch** — N items in each batch (uses the existing `chunk()` util).
- **Number of batches** — split into exactly `min(N, itemCount)` balanced groups (new `chunkIntoParts()` util).

## Why

Users feeding downstream steps that accept *lists* (bulk API endpoints, "insert many" DB ops, rate-limited calls) currently have to hand-roll chunking in a Code step. This makes it a first-class loop option.

## Behavior

- When batching is on, each iteration's `item` **is the batch (an array)** and `index` is the batch number.
- Batching is **off by default** — existing loops are unchanged, `item` stays a single element. No migration.
- The settings form shows an **alert** when batching is enabled: downstream steps must accept an array of records.

image
image
image

## Changes

- **Schema** — `LoopBatchMode` enum + optional `batching: { mode, size }` on `LoopOnItemsActionSettings` (`packages/core/execution`). Presence = enabled.
- **Util** — `chunkIntoParts()` added next to `chunk()` in `@activepieces/core-utils`.
- **Engine** — `loop-executor` derives the iteration list (raw items / `chunk` / `chunkIntoParts`) once, then iterates it; all path/pause-resume/single-step logic untouched.
- **Frontend** — toggle + mode select + positive-integer size input + downstream-array alert in `loops-settings.tsx`, with a separator above the batching section.
- **Docs** — added the **Batch** term to the automation-core glossary and a Loop Batching note to `flows.md`.

## Tests

- `chunkIntoParts` unit tests (even split, balanced remainder 10→3, exactly-N when sparse 10→7, caps at item count, empty).
- Engine loop suite: 2 new tests asserting item-per-batch and number-of-batches produce the expected iteration counts and array `item`; existing loop + pause tests unaffected.
- `tsc` (web + engine) and lint pass with 0 errors.

## Edition

Core CE functionality — available in all editions, no plan gate.

## Breaking change?
- [x] no
- [ ] yes-technical
- [ ] yes-functional

## Security impact?
- [x] no
- [ ] yes-security-sensitive


abuaboud:
@othmanemad This should be discussed with the Kareem regarding the delay step.

PR #14275, opened 2026-07-16
chore(deps): batch bump dependabot dependencies

abuaboud:
Consolidates the open Dependabot dependency PRs into a single batch so CI runs once and the individual PRs can be closed.

## Bumped
| Package | From | To | Scope |
|---|---|---|---|
| js-yaml | 4.1.1 | 4.2.0 | root |
| typeorm | 0.3.28 | 0.3.29 | root, server/api |
| esbuild | 0.25.0 | 0.28.1 | root, cli |
| qs | 6.14.2 | 6.15.2 | root, web, server/api, lever, clickup, tables |
| nodemailer | 8.0.5 | 9.0.1 | root, server/api |

Lockfile regenerated with `bun install`.

## Already at target (no-op)
undici, vitest, axios, samlify, fastify, systeminformation, exifreader, `@fastify/http-proxy`, `@fastify/static`, `@fastify/reply-from` were already on the versions Dependabot proposed.

## nodemailer 8→9 (major) — safe here
The v9 breaking change tightens TLS validation for **remote-content HTTPS fetching** (attachment href/path URLs, OAuth2 token endpoints, HTTP/HTTPS proxy CONNECT). `smtp-email-sender.ts` uses none of these — it only opens an SMTP socket and renders templates, and the SMTP socket already validated certs by default in v8. No behavior change.

As a self-hosting safeguard, added an optional **`SMTP_TLS_REJECT_UNAUTHORIZED`** prop (defaults to `true` — zero-setup, no change) so operators with private-CA/self-signed SMTP certs can opt out.

## Supersedes
Closes the individual Dependabot PRs once merged.


abuaboud:
@greptileai the flagged nodemailer v9 TLS change applies to **remote-content HTTPS fetching** (attachment href/path URLs, OAuth2 token endpoints, HTTP/HTTPS proxy CONNECT) — none of which `smtp-email-sender.ts` uses; it only opens an SMTP socket and renders templates. The SMTP socket already validated certs (`rejectUnauthorized: true`) in v8, so there's no behavior change for existing users.

That said, to give self-hosters with private-CA/self-signed SMTP an explicit escape hatch, I added an optional `SMTP_TLS_REJECT_UNAUTHORIZED` prop (defaults to `true` — zero-setup, no change). Re-triggering review.

PR #14281, opened 2026-07-16
feat(subflows): stream a CSV and fan out to a subflow per batch

othmanemad:
## What

Adds a **Stream CSV to Subflows** action to `@activepieces/piece-subflows`. It streams a CSV into a streaming parser and dispatches one **fire-and-forget** subflow call per batch of rows, so the file is never held in memory.

- Input is `Property.File({ streaming: true })`, so the **engine** owns the fetch and the action accepts a URL, an uploaded file, a base64 data URL, or a previous step's file — a strict superset of a URL text field. No `axios` in the piece.
- Per-call payload `data = { batchIndex, headers, rows, extraData }`, POSTed to the subflow's Callable Flow webhook. Returns `{ headers, firstRow, rowsProcessed, batchesDispatched }`.
- `fan-out.ts` — batching plus bounded in-flight concurrency (5) with stream back-pressure. On failure the fan-out aborts loud with the failed `batchIndex` **and the cause**, while already-dispatched subflows keep running (at-least-once, no fan-in). `batchSize` is capped at 10,000 rows.

Streaming bounds **memory, not time** — the step is still capped by `FLOW_TIMEOUT_SECONDS`, and resumable/exactly-once fan-out is out of scope for v1. See [decision 000015](brain/decisions/000015-streaming-csv-fanout-is-a-bounded-fire-and-forget-action.md).

### Why `text/csv` on the webhook path

This is the end-to-end story the action exists to serve, and it's why the two-line server change belongs in this PR:

> A third-party POSTs a CSV with `Content-Type: text/csv` to the parent flow's webhook → the webhook converter streams the body to file storage and the trigger payload becomes `{ fileUrl }` → that feeds straight into this action's streaming File prop.

Without those two lines `text/csv` returns **415**, so a CSV can only arrive as `application/octet-stream` or `multipart/form-data` — which only works for senders whose headers you control. It is a **compatibility** win, not a capacity one: `webhook.service.ts` size-checks the *converted* payload, a tiny `{ fileUrl }` on either binary path, so the ceiling is the same `AP_MAX_FILE_SIZE_MB` that `octet-stream` already gets. The change is purely additive — the only catch-all content-type parser is scoped to the `/ingest` PostHog proxy.

`text/tab-separated-values` has the identical 415 gap and is deliberately left for a follow-up.

### CSV parsing: three load-bearing options

Parser construction lives in `csv.ts` (`createCsvParser`) so the options it depends on are pinned by tests rather than buried in the action.

- **`bom: true`.** `csv-parse` leaves the UTF-8 BOM in place and `trim: true` does not strip it (it only handles space/tab, not U+FEFF). With `columns: true`, a BOM-prefixed file — what Excel "CSV UTF-8" and Google Sheets "Download as CSV" produce — yields a first header of `"﻿id"`, so `{{ rows[0].id }}` resolves to nothing **while every step reports success**: no error, correct row counts, green run. (`piece-csv`, `google-sheets` and `knowledge-base` have the same bug at independent `parse()` call sites — noted in the brain for a follow-up.)
- **`relax_column_count: true`.** One ragged row would otherwise throw `CSV_RECORD_INCONSISTENT_COLUMNS` and abort a whole file mid-fan-out, where there is no rollback. The trade is that a row shorter than its header omits the missing keys entirely rather than back-filling `''` — deliberate, and written down.
- **`group_columns_by_name: true`.** Duplicate header names otherwise collapse to the last value, silently dropping a column while the reported `headers` array still lists both. Duplicate columns are ordinary in real exports. Dup columns now arrive as `{a: ['1','2']}`, so a row value is `string | string[]`; non-duplicate columns are untouched.

### Robustness

- **"Retry on failure" is hidden on this action.** The framework default is off but *visible*, and step retry re-runs `run()` from the top — for a fire-and-forget fan-out that means re-reading from row 0 and re-dispatching every batch that already landed, so a failure at batch 900 would duplicate 900 batches. `Call Flow` leaves the toggle visible because one invocation is safe to retry.
- **Stream teardown.** A `file.body` error arriving after the parser was fully consumed called `destroy(err)` on a stream with no `'error'` listener — an uncaught exception in the sandbox. Now guarded on `parser.destroyed`.
- **No leaked response socket.** The engine has already opened the HTTP response before `run()` starts, so the `batchSize` guard and the disabled-subflow lookup both threw *above* the `try/finally` that destroys `file.body`. Both now destroy before throwing.
- **Retry scope.** Handing the retry loop to `httpClient` (`retries: 2`) narrowed *what* retries: the hand-rolled loop caught any thrown error, `httpClient` retries only 5xx and network errors, so every 4xx aborts on the first attempt. Correct for a 413 or a 404 from a subflow deleted mid-run; the one it costs is a 429, unreachable self-hosted but reachable behind a Cloud edge WAF.

### One fix beyond the new action

- **A latent bug in `callFlow`.** It built `${context.server.apiUrl}v1/webhooks/${id}` with **no separator**, working only because `getApiUrl()` always happens to return a trailing slash. Centralizing the POST in `dispatchToSubflow` normalizes it.

### Shared helpers

The action was first written by copying three blocks out of `callFlow`. Those are now single definitions in `common.ts` — `subflowDropdown`, `findEnabledSubflowOrThrow`, `dispatchToSubflow` — a net deletion. The two dropdowns had drifted to **different stored value shapes**; they unify on a plain string by changing the unreleased action, not `callFlow`, whose published flows already have a string in `propsValue`.

## Testing

- `npx turbo run build lint test --filter=@activepieces/piece-subflows` ✅ — 11 unit tests. `fan-out.test.ts` (7) covers batching, the concurrency cap, abort-on-failure, `rowsProcessed` accounting, the reported failure cause, and the in-flight drain when reading the records throws. `csv.test.ts` (4) pins BOM stripping, dup-column grouping, ragged-row tolerance in both directions, and tab-delimiter trimming against the pinned `csv-parse@5.6.0`.
- `npm run lint-dev` ✅
- `test/integration/ce/flows/flow-run/execute-flow-e2e.test.ts` ✅ 8/8 — covers the refactored `callFlow` in three shapes (wait-for-response, test-step mode, webhook → subflow → webhook response).
- BOM fix confirmed end-to-end in a live run.

## Breaking change?
- [x] no
- [ ] yes-technical
- [ ] yes-functional

## Security impact?
- [x] no
- [ ] yes-security-sensitive


packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+    const webhookUrl = `${context.server.apiUrl.replace(/\/$/, '')}/v1/webhooks/${flow.id}`;
+
+    const source = await axios.get<Readable>(fileUrl, {
+      responseType: 'stream',
+    });
+
+    let headers: string[] = [];
+    let firstRow: CsvRow | undefined;
+    const parser = parse({
```
hazemadelkhalel: `bom: true` is missing, so any CSV exported from Excel or Google Sheets breaks silently in the subflow. Reproduced against the pinned `csv-parse@5.6.0` with these exact options:

```
input:  "\uFEFFid,name\n1,a\n"
result: headers = ["\uFEFFid", "name"]
        rows    = [{ "\uFEFFid": "1", "name": "a" }]
        headers[0] charCodes = [65279, 105, 100]
```

The BOM lands in the first header name and in every row key, so `{{ trigger.body.data.rows[0].id }}` resolves to nothing in every dispatched subflow. Nothing fails: all batches report success, `rowsProcessed` is correct, and the run is green — which makes it the worst kind of bug to debug from a flow. `trim: true` doesn't help (csv-parse trims space/tab, not U+FEFF).

```ts
bom: true,
```

Worth also considering `relax_column_count` here — see my comment on `fan-out.ts`.

packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+          message:
+            'The selected subflow is disabled. Enable it before streaming to it.',
+          externalId: context.propsValue.subflow?.externalId,
+          flowName: flow.version.displayName,
+        })
+      );
+    }
+    const webhookUrl = `${context.server.apiUrl.replace(/\/$/, '')}/v1/webhooks/${flow.id}`;
+
+    const source = await axios.get<Readable>(fileUrl, {
```
hazemadelkhalel: `axios.get` has no default timeout, so a source that connects and then stalls holds the step until `FLOW_TIMEOUT_SECONDS` with no output. Worth setting one.

Bigger picture, and fine as a follow-up if you'd rather not grow this PR: this GET is the only reason `axios` is in the piece's dependencies, and the framework already covers it. `Property.File({ streaming: true })` resolves to an `ApStreamingFile` with `body: Readable` (`packages/pieces/framework/src/lib/property/input/file-property.ts`). `amazon-s3/upload-file.ts` uses it and `large-file-streaming.mdx` documents it.

```ts
props: {
  file: Property.File({ displayName: 'CSV File', required: true, streaming: true }),
},
// run()
const { file } = context.propsValue;
if (isNil(file)) throw new Error(JSON.stringify({ message: 'Could not read the CSV file.' }));
file.body.pipe(parser);
```

That drops the dependency, this GET, the `source.data.on('error', ...)` wiring and the timeout question above, and it accepts inputs a URL field can't: the builder's file picker, base64 data URLs, a previous step's file. One caveat, and it's the thing the URL version does better today: `fileProcessor` returns `null` on a failed fetch instead of throwing (`packages/server/engine/src/lib/variables/processors/file.ts`), so it needs the guard above.

I'm raising it because this is a core piece and `pieces-common` moved off axios deliberately (3 of ~400 pieces still declare it), not because the PR is wrong as written.

packages/pieces/core/subflows/src/lib/fan-out.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const start = (index: number, rows: T[]): void => {
+    const promise = dispatch({ batchIndex: index, rows })
+      .then(
+        () => {
+          batchesDispatched++;
+        },
+        () => {
+          if (firstErrorBatchIndex === undefined) {
+            firstErrorBatchIndex = index;
```
hazemadelkhalel: Two gaps in how failures get reported out of this file.

**The cause is dropped.** The rejection handler keeps the index and throws the error away, so the message below is "A batch failed to dispatch" with no status and no response body. That message is the action's only failure signal: no fan-in, no per-batch status. A 404 from a deleted subflow, a project limit, and a genuine 500 all read identically. Keep the first error and put its message (and `status`, when it's an `HttpError`) in the payload.

**A parse error skips this path completely.** One ragged row throws out of the `for await` on line 32, so `Promise.allSettled(inFlight)` never runs. In-flight dispatches aren't awaited and `rowsProcessed` / `batchesDispatched` are lost, which are the numbers telling the user how far the fan-out got. Verified with the action's parser options:

```
id,name / 1,a / 2,b,extra
-> CSV_RECORD_INCONSISTENT_COLUMNS: Invalid Record Length: columns length is 2, got 3 on line 3
```

`relax_column_count: true` handles the ragged row (`knowledge-base.service.ts` does this). Wrapping the loop and rethrowing with the counts handles the rest: truncated gzip, bad quoting.

packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+              flow.status === FlowStatus.ENABLED
+                ? flow.version.displayName
+                : `${flow.version.displayName} (inactive)`,
+          })),
+        };
+      },
+    }),
+    batchSize: Property.Number({
+      displayName: 'Rows per batch',
+      description: 'How many CSV rows to send in each subflow call.',
```
hazemadelkhalel: `batchSize` has a lower bound (the positive-integer check in `run`) but no upper one, and peak memory is roughly `(MAX_IN_FLIGHT + 1) × batchSize × rowSize`. Worth a ceiling, or a note here, since "streaming bounds memory" is the claim the action is built on and it currently holds only for sane batch sizes.

hazemadelkhalel (commented):
Read the whole thing against the subflow path in the server, and the fan-out core holds up: batching, the in-flight cap (peak is exactly `MAX_IN_FLIGHT` — `Promise.race` only resolves after `.finally` has removed an entry), back-pressure through the `for await`, abort-on-failure, and stream teardown in the `finally` are all correct. `x-fail-parent-on-failure: 'false'` reads correctly at `extractHeaderFromRequest`, and `flow-runs-queue.ts` confirms the children stay inert with respect to the parent. Piece packaging matches `piece-csv` exactly (same `vitest` / `csv-parse` pins, deps declared in the piece's own manifest), the version bump is right, the i18n keys match the literals in the action, and the new `brain/wiki/flows-execution/subflows.md` fills a real gap — there was no mention of "Callable Flow" anywhere in `brain/` before this.

One actual bug, which I reproduced: **`bom: true` is missing**, so every Excel/Sheets-exported CSV silently delivers row keys prefixed with U+FEFF and the subflows read `undefined` — while every batch reports success and the run goes green. That's the one I'd fix before merge. Details inline.

The rest is robustness and scope:

- the dispatch error's cause is discarded, and it is the only failure signal a fire-and-forget action has
- a ragged row escapes the structured abort entirely, losing the counts that tell the user how far the fan-out got
- the manual retry loop retries 4xx; `httpClient`'s own `retries` already does this correctly, with the same backoff
- `batchSize` has no upper bound, so "memory is bounded" depends on the user
- `axios` + a URL input reimplements `Property.File({ streaming: true })`, which the framework has shipped since 0.35.0 and which `amazon-s3` already uses — your own decision doc calls this out. Swapping to it is a net deletion.
- the `text/csv` webhook change isn't used by this feature and isn't explained in the description

On security specifically: no new class. The user-supplied URL is already covered because the SSRF guard patches `dns.lookup` and `net.Socket.prototype.connect` in-process (`ssrf-guard.ts`), so axios is filtered exactly like fetch — the client isn't the boundary. The one genuine operational note is amplification: one step can enqueue tens of thousands of child runs, and cancelling that parent later walks every descendant through a recursive CTE and cancels each under its own distributed lock (`getAllChildRuns` / `cancelSingleRun`). A `batchSize` cap is the cheap mitigation.

Before merge I'd want to see, on a real flow: a BOM-prefixed CSV, a ragged CSV, a ~50MB file with the worker's RSS watched (that's the core claim of the PR), the source killed mid-stream, the subflow disabled mid-fan-out, and `AP_FLOW_TIMEOUT_SECONDS` lowered so the timeout path gets exercised on purpose rather than in production.

packages/server/api/src/app/webhooks/webhook-request-converter.ts
```diff
@@ -17,6 +17,7 @@ const BINARY_CONTENT_TYPE_PATTERNS = [
     /^application\/zip$/,
     /^application\/gzip$/,
     /^application\/octet-stream$/,
+    /^text\/csv$/,
```
hazemadelkhalel, line 20: Nothing in this feature sends `text/csv`. The fan-out POSTs `application/json`, so this hunk is doing something else and the PR description should say what.

It's safe as far as it goes: there's no catch-all content-type parser on the webhook route, so `text/csv` returns 415 today. That makes this additive rather than a behaviour change. Two consequences to be deliberate about, since it applies to every webhook in every project and not just subflows:

- `rawBody` is no longer captured for `text/csv` (the `preParsing` hook in `webhook-module.ts` skips binary types), so a trigger verifying an HMAC over the raw body sees `undefined`. The same gap already exists via `application/octet-stream`, and the payload shape becomes `{ fileUrl }` so a signature-checking piece would fail anyway.
- An unauthenticated POST can now write a `text/csv` body into project file storage, capped by `AP_MAX_FILE_SIZE_MB`. Also already true for `image/*`, `pdf`, `zip` and `octet-stream`.

One call worth making on purpose: a small CSV now arrives as `{ fileUrl }` rather than as text, with no size threshold, while `text/plain` stays parsed as text. For a 50MB export that's the point. For a 2KB CSV the flow has to fetch a file to read three rows.

If the intent is "POST a CSV to the parent flow's webhook, then hand the `fileUrl` to this action", that's a good story and it belongs in the description and the `webhooks.md` gotcha. Right now it's the only thing tying this hunk to the rest of the PR and it isn't written down anywhere.

brain/wiki/flows-execution/subflows.md
```diff
@@ -0,0 +1,47 @@
...
+- **Stream CSV to Subflows**: input is a CSV **URL** (a plain `Property.File` would materialize an `ApFile` Buffer before `run()` starts and OOM), a Callable Flow dropdown target, `batchSize` (default 100), delimiter (comma/tab) and optional `extraData` merged into every call.
+  - The action does its own `responseType: 'stream'` GET and pipes into a streaming `csv-parse` parser — zero engine or framework change.
+  - Payload per call: `data = { batchIndex, headers, rows, extraData }`. Dispatch is fire-and-forget — no `callbackUrl`, `x-fail-parent-on-failure: false`.
+  - `fanOutBatches` bounds in-flight dispatches (5) and awaits `Promise.race` when the window is full, so parsing back-pressures instead of buffering the file.
+  - Returns `{ headers, firstRow, rowsProcessed, batchesDispatched }`.
+- **Failure**: a batch POST retries 3 times with exponential backoff (capped 8s), then the fan-out stops reading the stream and throws with the failed `batchIndex`. Already-dispatched subflows keep running — at-least-once, no fan-in, no rollback.
+
+### Gotchas
```
hazemadelkhalel, line 28: Since you're already writing brain pages here: the BOM finding and the streaming-input finding both generalise past subflows, so they're more useful as bullets under `Gotchas` in `brain/wiki/pieces-engine/building-pieces.md`. The BOM one isn't specific to this PR at all. `piece-csv`, `google-sheets` and `knowledge-base` all parse CSV and none of them pass `bom: true`.

Drop-in if you agree with them:

```markdown
- **Parsing CSV: always pass `bom: true`, and don't expect `trim: true` to cover it.** `csv-parse` leaves the UTF-8 BOM in place, and its `trim` option only strips space/tab, not U+FEFF. With `columns: true`, a BOM-prefixed file (what Excel and Google Sheets "Download as CSV" produce) yields a first header of `"\uFEFFid"` and row keys to match, so `{{ ...rows[0].id }}` resolves to nothing in the flow **while every step reports success**: no error, correct row counts, green run. Verified against the repo's pinned `csv-parse@5.6.0`, where `headers[0]` charCodes come back `[65279, 105, 100]`. Consider `relax_column_count: true` alongside it (as `knowledge-base.service.ts` does) so one ragged row doesn't abort a whole file mid-way.
- **Streaming a file *into* a piece is `Property.File({ streaming: true })`, not your own HTTP GET.** It resolves to an `ApStreamingFile` with `body: Readable` (pieces-framework >= 0.35.0) and accepts a URL, a base64 data URL, the builder's file picker, or a previous step's file: a strict superset of a URL text field, with the fetch and its timeout owned by the engine. `amazon-s3/upload-file.ts` is the reference. Two things to know: the engine's `fileProcessor` swallows fetch failures and hands you `null`, so guard for it with a real message; and `httpClient` has no `responseType: 'stream'`, so rolling your own means reaching for a raw HTTP client, and `pieces-common` deliberately migrated off axios.
```

One more on this page: the existing bullet saying the URL input "predates streaming file inputs ... Swapping to it is a follow-up, not a bug" reads as settled. The swap is a net deletion that also drops a dependency, so it's closer to an open review question than to recorded history. Worth rewording once it lands either way.

packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+  description:
+    'Stream a CSV from a URL and call a subflow once per batch of rows, without loading the whole file into memory.',
+  props: {
+    fileUrl: Property.ShortText({
+      displayName: 'CSV File URL',
+      description:
+        'A direct URL to the CSV file. It is streamed, not downloaded into memory.',
+      required: true,
+    }),
+    subflow: Property.Dropdown<FlowValue>({
```
hazemadelkhalel: This dropdown is `call-flow.ts:17` copy-pasted. Same options callback, same `(inactive)` label, same `refreshers: []`, same `PieceAuth.None()`. The only real difference is that this one wraps the value in `{ externalId }` while `call-flow.ts` uses a plain string.

Drop the wrapper and share the prop. Other pieces keep these in `src/lib/common/props.ts` as one `commonProps` object (google-sheets, grist, microsoft-excel-365, pdf-co all do), which also matches the repo rule about grouping util exports:

```ts
export const commonProps = {
  subflow: ({ displayName, description }: SubflowPropParams) =>
    Property.Dropdown<string>({ /* the existing body, once */ }),
};
```

Both actions then declare one line instead of twenty-one.

packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+  },
+  async run(context) {
+    const { fileUrl, batchSize, delimiter, extraData } = context.propsValue;
+    if (!Number.isInteger(batchSize) || batchSize < 1) {
+      throw new Error(
+        JSON.stringify({ message: 'Rows per batch must be a positive integer.' })
+      );
+    }
+
+    const flow = await findFlowByExternalIdOrThrow({
```
hazemadelkhalel: Same shape as `call-flow.ts:114-125`: resolve by external id, check `status !== ENABLED`, throw. The two messages differ only in wording ("before calling it from a parent flow" vs "before streaming to it"), which is how this kind of thing drifts.

"A subflow must be ENABLED before you can invoke it" is one rule and should live in one place, next to `findFlowByExternalIdOrThrow` in `common.ts`:

```ts
export async function findEnabledSubflowOrThrow({ flowsContext, externalId }): Promise<PopulatedFlow>
```

Callers get one line each. If the wording difference matters, pass the verb in.

packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+      throw new Error(
+        JSON.stringify({
+          message:
+            'The selected subflow is disabled. Enable it before streaming to it.',
+          externalId: context.propsValue.subflow?.externalId,
+          flowName: flow.version.displayName,
+        })
+      );
+    }
+    const webhookUrl = `${context.server.apiUrl.replace(/\/$/, '')}/v1/webhooks/${flow.id}`;
```
hazemadelkhalel: This is the one that already bit. `call-flow.ts:140` builds the same URL a different way:

```ts
// call-flow.ts
url: `${context.server.apiUrl}v1/webhooks/${flow?.id}`
// here
const webhookUrl = `${context.server.apiUrl.replace(/\/$/, '')}/v1/webhooks/${flow.id}`
```

Yours is the correct one. `call-flow.ts` only works because `apiUrl` happens to end in a slash today, and it breaks silently the day it doesn't.

The POST itself is duplicated too: same two headers, same `Content-Type`, differing only in `callbackUrl` and the `x-fail-parent-on-failure` value. One function in `common.ts` covers both:

```ts
export async function dispatchToSubflow({ context, flow, data, callbackUrl, failParentOnFailure })
```

That gets the URL fix into `call-flow.ts` for free, and puts the header contract in one place so it stays in step with `extractHeaderFromRequest` on the server.

packages/pieces/core/subflows/src/lib/actions/stream-csv-to-flow.ts
```diff
@@ -0,0 +1,174 @@
...
+        return header;
+      },
+      skip_empty_lines: true,
+      trim: true,
+    });
+    source.data.pipe(parser);
+    source.data.on('error', (err) => parser.destroy(err));
+
+    const dispatch = async ({
+      batchIndex,
```
hazemadelkhalel, line 85: This closure is ~30 lines inside `run()` so it can capture `webhookUrl`, `headers` and `extraData`. It holds the retry policy, the backoff, and the `firstRow` side effect.

That's why it has no test. `fan-out.ts` got five tests because it's a module-level function you can call; this block can't be reached from one.

While it's out: the loop retries every error, including a 404 from a subflow deleted mid-run. The cost is small, 3s of backoff before an abort that was happening anyway, but `httpClient` already does this with the same `min(1000 * 2 ** attempt, ...)` backoff and only retries 5xx and network errors (`fetch-http-client.ts`, `sendWithRetries`):

```ts
await httpClient.sendRequest({ ..., retries: 2 });
```

Pull the closure out as a named function and the loop becomes one option.

hazemadelkhalel (commented):
Second pass, this one only about duplication between the new action and `call-flow.ts`. Four spots are copy-pasted rather than shared, and one of them has already drifted (the webhook URL). Fixing them removes roughly 60-70 lines across the piece and makes `call-flow.ts` shorter too.

Summary of the four, details inline:

1. the Callable Flow dropdown, duplicated verbatim
2. resolve-flow + "is it enabled", duplicated with two different messages
3. the webhook POST, duplicated and already out of sync
4. the dispatch closure inside `run()`, which is why the retry logic has no test

One thing I'd leave alone: `fan-out.ts` takes `dispatch` as a parameter and knows nothing about HTTP or CSV. That's the right split, and it's why that file was easy to test. Keep it as is.

Also pre-existing, so up to you whether it belongs in this PR: the Simple/Advanced `StaticDropdown` is copy-pasted three times (`call-flow.ts`, `respond.ts`, `callable-flow.ts`), about 15 lines each, and `respond.ts` + `callable-flow.ts` repeat the same `mode` + `DynamicProperties` pair with only the field name changed. That is the biggest single win in the piece if you want it.

othmanemad:
All review comments addressed in 47938b2 — thanks, several were load-bearing.

| Comment | Resolution |
|---|---|
| Missing `bom: true` | Fixed. Confirmed the silent-corruption path in a live run: a BOM-prefixed header yields `"﻿id"` and `{{ rows[0].id }}` resolves to nothing **on a green run**. |
| Ragged rows abort the file | `relax_column_count: true`. Also fixed the drain: a parse error used to skip `await Promise.allSettled(inFlight)` and lose `rowsProcessed`/`batchesDispatched` — the only numbers telling you how far an at-least-once fan-out got. |
| Dispatch failure loses the cause | Now reports `{ failedBatchIndex, cause }`. |
| No timeout on the source fetch | Moot — the input is `Property.File({ streaming: true })`, so the engine owns the fetch. |
| Dropdown / ENABLED check / webhook POST duplicated | Extracted to `subflowDropdown`, `findEnabledSubflowOrThrow`, `dispatchToSubflow` in `common.ts`. |
| Hand-rolled retry closure | Replaced by `httpClient`'s `retries`. |
| Drop `axios` | Gone. |
| Cap batch size | 10,000 rows. |
| `text/csv` — document or drop | **Kept**, and the PR description now tells the story it was missing. |

Three notes where I went a different way:

**The dropdowns store different value shapes** — `callFlow` stores a plain string, the new action was storing `{ externalId }`. Unifying by changing `callFlow` would break published flows that already have a string in `propsValue`, so the *unreleased* action moved instead.

**`dispatchToSubflow` fixes a latent bug**, which is what justifies it over the DRY case alone: `callFlow` built `${apiUrl}v1/webhooks/${id}` with no separator, working only because `getApiUrl()` happens to return a trailing slash.

**The source-stream error handler stays.** `.pipe()` doesn't forward `'error'`, so without a listener on `file.body` a mid-stream network drop is an uncaught exception in the sandbox. Likewise the suggested `isNil(file)` guard is unreachable — `props-processor` fails the `required: true` FILE prop and `piece-executor` throws before `run()` is called.

Green: 7/7 piece unit tests, `lint-dev` clean, and `execute-flow-e2e` 8/8 — that last one covers the refactored `callFlow` in three shapes (wait-for-response, test-step mode, webhook → subflow → webhook response).

othmanemad:
@greptileai review again

hazemadelkhalel (approved):

PR #14311, opened 2026-07-19
feat(subflows): opt-in wait-for-all fan-in for Stream CSV action

othmanemad:
## What

Adds an opt-in **Wait for all subflows to finish** mode to the `Stream CSV to Subflows` action. When enabled, the action dispatches its batches (unchanged, fire-and-forget), then **pauses** — via a self-resuming 30s `DELAY` waitpoint — and polls until every dispatched child run is terminal, then continues with a summary.

- **Never throws on child failure** — a failed subflow is counted in the summary, not propagated.
- Bounded by `maxWaitMinutes` (default 60); on expiry it continues with a `timedOut` summary + `stillRunning` count rather than hanging.
- Fire-and-forget (default) path is **untouched**.

## How it decides "done"

Child `flow_run` rows are recorded **asynchronously** by a worker, not at POST time, so "no non-terminal rows yet" ≠ "done". Completion is anchored to `batchesDispatched` (one successful POST = one child):

```
terminalDelta = (cur.succeeded + cur.failed) - (baseline.succeeded + baseline.failed)
done          = terminalDelta >= batchesDispatched && nonTerminal === 0
```

A `baseline` rollup is captured **before** dispatch and subtracted so prior loop iterations / earlier terminal children don't count as ours.

## Changes

**Server** (no migration):
- `flow-run-service.ts` — optional `parentRunId` on `countByStatus`; pure `rollupByStatus` helper (`isFlowRunStateTerminal({ ignoreInternalError: false })`); `countRunsRollupByParent`.
- `engine-controller.ts` — `GET /v1/engine/flow-runs/count-by-parent`, `securityAccess.engine()`, scoped to `principal.projectId`, returns `{ succeeded, failed, nonTerminal }`.

**Piece** (`@activepieces/piece-subflows`, `0.5.0` → `0.6.0`):
- `fan-in.ts` — pure `evaluateFanIn`.
- `stream-csv-to-flow.ts` — `waitForAllSubflows` + `maxWaitMinutes` props; baseline + loud mixed-mode guard **before** dispatch; store-backed poll loop; `FAIL_PARENT_ON_FAILURE_HEADER: 'false'` kept (flipping it would let `markParentRunAsFailed` complete our pending waitpoint). Each batch is a **single non-idempotent webhook POST — no dispatch retry** (a lost-response retry would double-dispatch a child run); a failed POST aborts the fan-out via `fanOutBatches` with the failing batch index. On `RESUME` with a lost store key it **throws** (fan-in state gone) rather than falling through to re-dispatch.
- i18n keys added.

**Docs:** ADR-0009 amended (Fix A count-based mechanism, mixed-mode limitation, Fix B upgrade path).

## v1 limitation

Don't mix fire-and-forget and wait-for-all subflow steps in the same parent run — child counts conflate. A loud guard rejects starting a wait when the run already has non-terminal children. The airtight fix (per-step child marker column) is deferred as **Fix B**.

## Breaking changes

**None.** The feature is fully opt-in and backward compatible:
- Both new props (`waitForAllSubflows`, `maxWaitMinutes`) are optional and default to the existing fire-and-forget behavior — flows already using `Stream CSV to Subflows` behave exactly as before.
- `parentRunId` on `countByStatus` is an optional param; existing callers are unaffected.
- The new engine endpoint is purely additive — no route or response shape changed or removed.
- No database migration; no schema change.
- Piece bump is a minor version (`0.5.0` → `0.6.0`) — additive only.

## Security

- **Tenant isolation preserved.** `GET /v1/engine/flow-runs/count-by-parent` is gated by `securityAccess.engine()` (engine-token principals only) and derives `projectId` from `request.principal.projectId` — never from the request. The client-supplied `parentRunId` is validated (`ApId`) and only ever filtered **within** that project scope, so it cannot read another project's / tenant's runs.
- **No user-controlled outbound target.** The piece's rollup poll calls the internal engine API (`context.server.apiUrl`) with the engine bearer token (`context.server.token`) — no user-supplied URL, secret, or credential is introduced by this change.
- **No new setup surface.** No new env vars, secrets, auth methods, or DB extensions — zero-config for self-hosters.

## Verification

- `fan-in.test.ts` — 7 cases incl. the async-recording-lag regression guard (`nonTerminal === 0` alone must not read done) and the count-conflation `stillRunning` clamp. `fan-out.test.ts` — 5 cases incl. both dispatch-abort paths (error during iteration vs. detected at drain). Subflows tests 12/12 pass.
- Subflows lint clean; server-api lint 0 errors; server typecheck 0 errors.
- Manual end-to-end run locally (pause/resume loop, failed child, timeout) — works.

## Breaking change?
- [x] no
- [ ] yes-technical
- [ ] yes-functional

## Security impact?
- [x] no
- [ ] yes-security-sensitive


othmanemad:
Will be replaced with https://github.com/activepieces/activepieces/pull/14418 later on

PR #14312, opened 2026-07-19
fix(chat): recurring automations avoid reprocessing, and long turns no longer stop early

hazemadelkhalel:
## Description

Two fixes for the AI chat, both from investigating flows that came out logically wrong or turns that stopped mid-response.

### 1. Recurring automations no longer reprocess the same data

When someone asks the chat to build a recurring automation (e.g. *"every Friday, read my timesheet and email me who to pay"*), it used to build a plain `read-all → act → done` flow. Correct for one run — but on a schedule it re-pays/re-sends the same data every time, because nothing tracks what was already handled. Non-technical users can't foresee this, and the flow *looks* finished.

The chat now treats this as a design-time decision it makes on its own, with no hint from the user: if a flow is recurring **and** reads data that persists between runs (a sheet, table, inbox, record set), it commits to an anti-reprocessing mechanism as part of the build — a "new item"/dedup trigger, a processed-flag filter with write-back, delete/archive-after, or an **Activepieces Tables ledger** it dedups against (the default when the source is external/read-only, so it doesn't mutate the user's sheet). It also reasons about the wider cleanup surface (stable dedup key, mark-done-after-success, ledger retention).

Implemented in the prompt layer:
- `chat-system-prompt.md` — the temporal question is raised at discovery time (before the framing locks in).
- `guides/build_flow.md` — the "Recurring flows must not reprocess" playbook, an exception carved into the "don't over-build" guidance, two-run verification, and a reflection check.
- Three capability eval fixtures (payroll, orders, digest) with a shared `recurring_avoids_reprocessing` judge dimension so we can measure this in the nightly eval.

### 2. Long chat turns no longer stop mid-response

Turns that spent minutes thinking would sometimes stop early with a half-built flow. Cause: a 90s stream idle watchdog aborted the turn when the model stream went quiet — but extended thinking legitimately goes quiet for minutes, so it was killing healthy turns.

The idle watchdog is now **report-only**: on genuine unexplained silence (suspended during tool calls and reasoning) it logs an idle signal for monitoring, once per silent gap, and the turn keeps running. Nothing aborts or marks a turn stale for being quiet. The only hard stop is a generous wall-clock backstop (raised to 2h — its sole job is reclaiming a worker slot if a turn truly wedges forever; liveness is already kept by the 15s heartbeat and the worker's 30s BullMQ lock renewal).

## How was this tested?

- **Correctness:** ran the naive payroll prompt against the working-tree prompt in dev (EE). The chat now builds a `Paid Log` Tables ledger, dedups against it before paying, and records new entries after — unprompted. Verified the same reasoning appears in the discovery/plan.
- **Evals:** the deterministic chat-eval harness passes (`npm run chat-evals:ci`); the three new fixtures load and are scored by the nightly live harness (`kind: capability`).
- **Reliability:** confirmed the idle watchdog no longer aborts on silence and only emits the monitoring log; long thinking turns run to completion. Worker package lints clean.
- Edition: chat is EE/Cloud only; verified on EE dev.

Fixes # (issue)


hazemadelkhalel:
@greptileai review

PR #14326, opened 2026-07-20
feat(omnihr): Update auth + new action

danielpoonwj:
### Auth Update

There is now [token refresh + server-side caching](https://www.activepieces.com/docs/build-pieces/piece-reference/authentication#token-refresh-server-side-caching) supported. This prevents continually regenerating the access token, which can hit rate limits.

Now it is generated once and cached till expiry (slightly earlier with a 5 minute buffer).

### New Action: Generate Report

Introduce a new action that can generate reports defined through the UI. There are 2 general kinds of reports: Master and Snapshot.

Master reports always get up to date data:

Screenshot 2026-07-20 at 9 46 39 PM

Snapshot reports allow date input for a historical snapshot:

Screenshot 2026-07-20 at 9 46 16 PM

By default the report is generated keyed by id. This can be hard to work with eg. if there are custom fields and the id would be eg. `custom_field.1234`. The report templates contain the human readable name mapping and we provide an option to map field names in the output.

## How was this tested?

Local testing.

Fixes # (issue)


packages/pieces/community/omnihr/src/lib/auth.ts
```diff
@@ -36,4 +50,33 @@ export const omnihrAuth = PieceAuth.CustomAuth({
...
+        body: {
+          username: auth.username,
+          password: auth.password,
+        },
+      });
+
+      const expiresIn =
+        Number(tokenResponse.body.access_exp) -
+        Math.floor(Date.now() / 1000) -
+        ACCESS_TOKEN_EXPIRY_BUFFER_SECONDS;
```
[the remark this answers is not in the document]
danielpoonwj: this should not happen in regular operation as that would be a break in the API contract. the field should be expected to exist and be in that epoch format

packages/pieces/community/omnihr/src/lib/auth.ts
```diff
@@ -36,4 +51,33 @@ export const omnihrAuth = PieceAuth.CustomAuth({
...
+          Origin: auth.origin,
+        },
+        body: {
+          username: auth.username,
+          password: auth.password,
+        },
+      });
+
+      const expiresIn =
+        Number(tokenResponse.body.access_exp) -
```
kishanprmr, line 75: @danielpoonwj According to the [Auth Docs](https://api-docs.omnihr.co/#getting-started), the authentication response doesn't include an `access_exp` field. Could you confirm whether it's still returned in the response when you test the API? Sorry for the inconvenience, we don't have valid credentials to test this on our end.

danielpoonwj: omnihr-screenshot

hi @kishanprmr no worries, here is the response from an actual API call

kishanprmr: Perfect, thanks! We can go ahead and merge it.

kishanprmr:
@danielpoonwj bumped `minimumSupportedRelease` to `0.86.0` since actions now rely on `context.auth.access_token`, which is only populated by the server-side refresh mechanism from #13778 (shipped in 0.86.0). Without the bump, self-hosted instances on older releases would get Bearer undefined and every action would fail.

kishanprmr (approved):

AhmadTash (approved):

PR #14368, opened 2026-07-22
fix(flows): code steps no longer crash runs with "o.send is not a function" under worker load

### LINKED ISSUE #14361: [BUG]: Recurring CODE-step engine crash (o.send is not a function)
Activepieces Bug Report — Recurring CODE-step engine crash (o.send is not a function)

Environment: Cloud project "Iris-Prod" (HHW1gm0wFFhkiNVxtpvyx), flow "Iris Check · Callable" (3RX3SyfFj5f5cOk45kXIy). Reproduced in both TESTING (ap_test_flow) and observed in production runs.

Error (identical stack every time): TypeError: o.send is not a function at /root/common/main.js:19:12578 at new Promise () at tj (/root/common/main.js:19:12006) at Object.runCodeModule (/root/common/main.js:57:61) at /root/common/main.js:105:215768 at process.processTicksAndRejections (node:internal/process/task_queues:104:5) at async ir (/root/common/main.js:65:2127) at async Object.tryCatchAndThrowOnEngineError (/root/common/main.js:105:81095) at async g2 (/root/common/main.js:105:215169) at async Oc (/root/common/main.js:105:214326)

Behavior: Fails inside runCodeModule — i.e., in the engine's own dispatch/response path around a CODE step, not inside our step's exported code function. Confirmed because it hits CODE steps of wildly different complexity, including a one-line pass-through step with no external calls and a \~1.1s total duration — ruling out anything in our step logic throwing it.

Reproduction pattern:

* Over one evening (2026-07-21 → 2026-07-22), we logged \~25+ flow runs. Failure rate varied sharply over time: a clean stretch of 4/4 successes, followed by a stretch of \~16/19 failures within the same hour, on the same flow, same code, no deploys in between.
* The crash lands on a different step nearly every time — observed on at least 4 different CODE steps in this one flow (a loop-body merge step, a loop-aggregation step, a small array-building step, and a downstream filter step).
* Retrying the identical run (same trigger data) after a failure frequently succeeds — same input, different outcome.

Ruled out on our side:

* Payload size: two runs with byte-identical CODE-step output size (56,528 bytes) — one succeeded, one failed later.
* A specific template expression we'd added (|| fallback in a piece-store add it as a precaution, crash continued identically afterward.
* Booking-specific data content — the same 3 "problem" bookings failed \~10+ times each across many retries while \~6 other bookings run against the same flow succeeded consistently.

Best-supported hypothesis (not confirmed AP-side): crash probability appears to scale with the number of CODE-step invocations in a single run (e.g., bookings whose loops iterate<br>15–17× vs. 1×), consistent with a low, constant per-invocation fault rate inh rather than a data- or size-triggered bug.

Impact: Intermittent full-run failures requiring manual retry; no data corruccur before write-back steps execute). Currently working around it withcontinueOnFailure on the affected loop steps and manual retries, since it makes local automation unreliable at scale.<br>Ask: Please check engine logs around runCodeModule / main.js:19 for this project during the timeframe above for the root cause of the o.send reference — happy to provide specific run IDs on request.

---

## Verified root cause (local repro, 2026-07-22)

AI/automation context. Root cause confirmed; no log-diving needed.

* Crash site: `packages/server/engine/src/lib/core/code/no-op-code-sandbox.ts`, `runInChildProcess`, the sync `child.send({ codeFilePath, inputs })` at the end of the `new Promise` executor (~line 86). Minified mapping: `o` = `child`, `tj` = `runInChildProcess`.
* Cloud path: `AP_EXECUTION_MODE=SANDBOX_PROCESS` selects `noOpCodeSandbox` (`code-sandbox.ts`); every CODE-step invocation spawns a child Node process with `stdio: ['pipe','pipe','pipe','ipc']`.
* Mechanism: when `spawn()` fails with EMFILE/ENFILE (fd exhaustion in the worker/engine process), Node returns before `setupChannel`, so `.send` is never attached. The synchronous `child.send(...)` then throws `TypeError: child.send is not a function`, and the promise rejects with that TypeError, masking the real spawn error. The `child.on('error')` handler carrying the true `spawn EMFILE` fires one tick later, after the promise already rejected.
* Errno specificity: only EMFILE/ENFILE produce this signature. EAGAIN/ENOENT do attach `.send` and reject cleanly via the `error` handler.
* No fd leak in the sandbox itself: 50 sequential `runCodeModule` calls, fd delta 0 (`/proc/self/fd`). Pressure is external to this code path (worker fd ceiling). Per-invocation spawn means per-invocation exposure while near the limit, matching the reporter's invocation-count scaling and the retry-succeeds pattern.
* Repro: `ulimit -n 512`, open `/dev/null` until EMFILE, free ~3 fds, call `noOpCodeSandbox.runCodeModule` on a trivial module. Rejects with the identical TypeError and stack shape. Healthy baseline succeeds.
* Introduced by #12424 (spawn-per-invocation, May 2026). Failure mode did not exist before that PR.

Fix:

1. Guard: call `child.send(...)` only when `typeof child.send === 'function'`; otherwise let the existing `error` handler reject via `buildError` so the true `spawn EMFILE` surfaces.
2. Optional hardening: short retry/backoff on EMFILE/ENFILE spawn failure so runs survive transient pressure.
3. Infra follow-up (separate from code fix): check fd usage on cloud workers for 2026-07-21/22. The guard fixes error reporting only, not the underlying exhaustion.

majewskibartosz:
Fixes activepieces/activepieces#14361<br>Closes activepieces/activepieces#14361

## Problem

1. Flow runs with CODE steps intermittently fail with `TypeError: o.send is not a function` from the engine's `runCodeModule` — bursty, data-independent, retry usually succeeds, probability scales with code-step invocations per run.
2. Root cause: when `spawn()` fails with EMFILE/ENFILE (worker fd exhaustion), Node returns before attaching `ChildProcess.send`; the unguarded synchronous `child.send(...)` throws inside the Promise executor, masking the real spawn error carried by the async `error` event one tick later. Only EMFILE/ENFILE produce this signature.

## Fix

* `packages/server/engine/src/lib/core/code/no-op-code-sandbox.ts`: `child.send({...})` → `child.send?.({...})` — with no IPC channel attached, the existing `error` handler now rejects with the real spawn error (e.g. `spawn EMFILE`).

## Verification

* Regression test `no-op-code-sandbox.test.ts`, proven red-first: pre-fix run rejects with `child.send is not a function`; post-fix asserts `spawn EMFILE`.
* Real-EMFILE runtime repro (fd exhaustion under `ulimit -n 512`) against actual `runCodeModule`: pre-fix reproduces the reported TypeError; post-fix rejects with `spawn EMFILE`. Healthy path unchanged.
* Engine suite 359/359; `npm run lint-dev` clean (0 errors).
* fd-leak check: 50 sequential invocations, fd delta 0 — the exhaustion source is external to this code path; infra follow-up tracked in activepieces/activepieces#14361.
* Visual: none - engine-only; only changes a failed step's internal error text under OS fd exhaustion.

## Breaking change?

- [ ] Yes
- [X] No

## Security impact?

- [ ] Yes
- [X] No

### DIFF
--- a/packages/server/engine/src/lib/core/code/no-op-code-sandbox.ts
+++ b/packages/server/engine/src/lib/core/code/no-op-code-sandbox.ts
@@ -83,7 +83,7 @@ async function runInChildProcess({ codeFilePath, inputs }: { codeFilePath: strin
             reject(buildError({ message: error.message, stdout: capturedStdout, stderr: capturedStderr }))
         })
 
-        child.send({ codeFilePath, inputs })
+        child.send?.({ codeFilePath, inputs })
     })
 }
 
--- a/packages/server/engine/test/core/code/no-op-code-sandbox.test.ts
+++ b/packages/server/engine/test/core/code/no-op-code-sandbox.test.ts
@@ -1,0 +1,59 @@
+import { ChildProcess } from 'node:child_process'
+import { EventEmitter } from 'node:events'
+import { unlink, writeFile } from 'node:fs/promises'
+import os from 'node:os'
+import path from 'node:path'
+import { noOpCodeSandbox } from '../../../src/lib/core/code/no-op-code-sandbox'
+
+const spawnState = vi.hoisted(() => ({ failNextSpawnWith: null as string | null }))
+
+vi.mock('node:child_process', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('node:child_process')>()
+    return {
+        ...actual,
+        spawn: (...args: Parameters<typeof actual.spawn>): ChildProcess => {
+            if (spawnState.failNextSpawnWith === null) {
+                return actual.spawn(...args)
+            }
+            const errorCode = spawnState.failNextSpawnWith
+            spawnState.failNextSpawnWith = null
+            const child = new EventEmitter() as ChildProcess
+            setImmediate(() => child.emit('error', Object.assign(new Error(`spawn ${errorCode}`), { code: errorCode })))
+            return child
+        },
+    }
+})
+
+describe('noOpCodeSandbox', () => {
+    describe('runCodeModule', () => {
+        let tmpFile: string
+
+        beforeEach(() => {
+            spawnState.failNextSpawnWith = null
+            tmpFile = path.join(os.tmpdir(), `no-op-sandbox-test-${Date.now()}.js`)
+        })
+
+        afterEach(async () => {
+            await unlink(tmpFile).catch(() => undefined)
+        })
+
+        async function runWithSource(source: string, inputs: Record<string, unknown> = {}): Promise<unknown> {
+            await writeFile(tmpFile, source, 'utf8')
+            return noOpCodeSandbox.runCodeModule({ codeFilePath: tmpFile, inputs })
+        }
+
+        it('executes exports.code and returns the result', async () => {
+            const result = await runWithSource('exports.code = async (inputs) => ({ doubled: inputs.n * 2 })', { n: 3 })
+            expect(result).toEqual({ doubled: 6 })
+        })
+
+        it('propagates user errors thrown inside code', async () => {
+            await expect(runWithSource('exports.code = async () => { throw new Error(\'user error\') }')).rejects.toThrow('user error')
+        })
+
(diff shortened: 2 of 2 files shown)


AbdulTheActivePiecer:
duplicate of https://github.com/activepieces/activepieces/pull/14543

PR #14376, opened 2026-07-22
feat(server): transparent canary frontend routing on the primary origin

amrdb:
## What does this PR do?

Lets a canary platform (worker group `canary`) get the **new frontend, backend, and websockets** transparently on `cloud.activepieces.com` — same URL, so all auth/OAuth/SSO keep working — instead of only the backend via the existing `/api` proxy. (Redirecting to a `canary.*` origin would break auth: the session token is origin-scoped `localStorage`, and OAuth/SSO `redirect_uri`s are registered for the primary origin only.)

Routing is driven by one signed `ap_canary` cookie the app sets once `/api` confirms a canary platform:
- **Static/SPA** requests with a valid cookie are reverse-proxied to the canary app (the canary bundle is served on the primary origin).
- **`/api/socket.io` upgrades** with a valid cookie are piped to the canary app; everything else is delegated to the local socket.io — so a canary user's realtime (incl. inbound lock/presence handlers) runs entirely on canary code, closing the new-frontend↔old-ws version seam.
- The frontend reloads **once** into the canary bundle when the cookie appears (guarded to one reload/session, so no loop).

Design + rationale live in `.agents/features/canary.md`.

### Security
- Cookie is **HMAC-signed** with the JWT secret (via `@fastify/cookie`), so it can't be forged. It grants **no privilege** (auth stays the token) — it only selects which code version serves; that's why it's deliberately readable (not HttpOnly) for the bundle-swap hint.
- `index.html` is `no-store` + `Vary: Cookie` so a shared/CDN cache can't serve a canary bundle to a prod user; hashed `/assets/*` stay `immutable` (unique per build).
- Proxy upstream is always the fixed `CANARY_APP_URL` (no SSRF). All new paths are **no-ops when `CANARY_APP_URL` is unset** (self-hosted / CE default → zero behavior change).

### Tests
40 unit/integration tests across the canary module: cookie sign/verify/tamper/rotated-secret, `/api` proxy (incl. real-TCP integration), static routing (path exclusions + cookie gating), ws routing (gate + capture/delegate/proxy). `api` + `web` lint clean.

## Ops prerequisites (before enabling)
1. **Cloudflare**: bypass cache when the `ap_canary` cookie is present (belt-and-suspenders with the `no-store` header).
2. The **canary app must NOT set `CANARY_APP_URL`** (else it proxies to itself — loop).
3. Keep the canary hostname/service pinned to the canary deployment.

> Note: transparent primary-origin routing only takes effect once this ships to **prod** (the reload logic must live in the bundle users first load). Until then, canary is reached directly via `canary.activepieces.com`.

## Breaking change?

- [ ] Yes
- [x] No

New code is gated on `CANARY_APP_URL`; no new required env vars, no schema changes, no API changes. Self-hosters/CE are unaffected.

### DIFF
--- a/.agents/features/canary.md
+++ b/.agents/features/canary.md
@@ -1,0 +1,41 @@
+# Canary Routing
+
+## Summary
+Canary is a **worker group that also has its own app tier** (`CANARY_APP_URL`, `IS_CANARY_APP`). A platform whose `platform_plan.workerGroupId === 'canary'` (`workerGroupService.isCanaryPlatform`) has its flow jobs routed to a dedicated queue **and** its HTTP/websocket traffic served by a separate canary deployment of the app — so a subset of real platforms exercises a new release before it is promoted to production. The prod app is the ingress and reverse-proxies canary platforms to the canary app; canary and prod **share the same Postgres and Redis**.
+
+This module makes the canary experience **transparent on the primary origin** (`cloud.activepieces.com`): a canary platform gets the new frontend + backend + websockets on the normal URL, so all auth/OAuth/SSO keep working (redirecting to a `canary.*` origin would break them — the session token is origin-scoped `localStorage`, and OAuth `redirect_uri`s are registered for the primary origin only).
+
+## How routing works
+All routing keys off one cookie, `ap_canary`, minted by the backend once an authenticated `/api` request confirms a canary platform.
+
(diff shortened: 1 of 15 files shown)


amrdb:
Addressed both review findings in 912c427:

- **P1 (ws upstream socket leak on timeout)** — the `'timeout'` handler now calls `proxyReq.destroy()` before cleanup, so the upstream canary TCP connection is closed rather than leaked. Added a unit test asserting both the upstream request and the client socket are destroyed on timeout.
- **P2 (`fastifyCookie` outside `isApp()`)** — moved the cookie-plugin registration inside the `system.isApp() && !isNil(CANARY_APP_URL)` guard, so worker / non-canary processes no longer register it or read the JWT secret.

Issue #5636, opened 2024-09-17
[BUG]: Property.Object with dynamic value sends as String


Vitalini:
**Describe the bug**
`Property.Object` with dynamic value sends as String
```JSON
{
  "response": {
    "status":400,
    "body": {
      "ok":false,
      "error_message":"custom_properties must be a Hash"
    }
  },
  "request": {
    "body": {
      "data": {
        "custom_properties":"{\\"key\\":\\"value\\"}"
      }
    }
  }
}
```

**To Reproduce**
1. Create a new flow
2. Add piece with Object param
3. Switch to dynamic
4. Add value `{"key":"value"}`

**Expected behavior**
Send the field as an Object

abuaboud:
Hi @Vitalini 

Which version of AP are you running atm? I do believe this is fixed

Vitalini:
> Hi @Vitalini
> 
> Which version of AP are you running atm? I do believe this is fixed
Hi @abuaboud,
Version: 0.32.1
Screencast: https://monosnap.com/file/IsJmZgRALvsPQlEEqdS7b2EoBF46Pk

PR #14389, opened 2026-07-23
fix(flows): multi-select and checkbox fields keep working when switched to dynamic values

### LINKED ISSUE #14387: [BUG] [Flows] Dynamic value toggle converts multi-select dropdown array to a string
## Problem

Toggling a `MULTI_SELECT_DROPDOWN` property to dynamic mode (fx) JSON-stringifies its array value: `[]` becomes the string `"[]"`, `["a","b"]` becomes `'["a","b"]'`. Silent: the flow saves and runs, the action's `run()` receives a string where it expects an array.

Ref: Pylon 4586

## Steps to reproduce

1. Any piece action with a multi-select dropdown; select values (or leave `[]`)
2. Toggle the field to dynamic value
3. Input shows the stringified array; test the step → action input is a string

## Root cause

`packages/web/src/app/builder/piece-properties/auto-form-field-wrapper.tsx:177` — `getValueForInputOnDynamicToggleChange` special-cases only `PropertyType.ARRAY`; multi-select values fall through to `JSON.stringify(currentValue)`.

No recovery downstream: `packages/server/engine/src/lib/variables/processors/index.ts` has no processor for `MULTI_SELECT_DROPDOWN` / `STATIC_MULTI_SELECT_DROPDOWN`, and `validateProperty` (`props-processor.ts:136`) has no case for them, so the string passes through unparsed and unvalidated.

## Fix

Same class as activepieces/activepieces#5636 (OBJECT sent as string, fixed via `objectProcessor`). Mirror it: engine processor for both multi-select types that JSON-parses string values back to arrays, plus an array validation case. Dynamic mode's text input legitimately holds a string, so parsing belongs engine-side; optionally also stop the toggle from stringifying when the user immediately toggles back.

majewskibartosz:
Fixes activepieces/activepieces#14387<br>Closes activepieces/activepieces#14387

## Description

Switching a field to a dynamic value renders a text input, so the builder stores whatever was there as text. For a multi-select that means `["year","month"]` becomes the *string* `'["year","month"]'`, and for a checkbox `true` becomes `'true'`. It saves and publishes without complaint, then the step fails at runtime — Date Helper "Extract Date Units" dies with `Invalid unit to extract: "["`, because it is iterating the characters of a string. Ref: [Pylon 4586](<https://app.usepylon.com/issues?issueNumber=4586>).

This is not a new class of bug. A **dynamic JSON property has always had exactly this problem**, and the engine has always just parsed the string at runtime — `jsonProcessor`. Nobody migrates JSON fields, and the text box is simply how dynamic mode looks. Multi-select and checkbox now follow the same concept.

## The fix

Two processors in `packages/server/engine/src/lib/variables/processors/`, registered in the `processors` map:

- **`multi-select.ts`** — parses a stringified list, and turns anything else non-nil into a single-item array.
- **`checkbox.ts`** — parses a stringified boolean, and leaves anything else for the validator to report.

Like `jsonProcessor`, neither reads `property.required`: they map an empty input to `undefined` and let `validateProperty` decide, so an untouched optional field is unset while a required one errors.

| Dynamic value | Reaches `run()` as |
| --- | --- |
| `'["year","month"]'`, `'[]'` | `['year','month']`, `[]` |
| `'year'`, `'a,b'`, `'5'`, `5` | `['year']`, `['a,b']`, `['5']`, `[5]` |
| `{{ trigger.body.tag }}` carrying one tag | `['tag1']` |
| `''` on an optional field | unset |
| `''` on a required field | `Expected array, received: ` |
| `'true'` / `'false'` on a checkbox | `true` / `false` |

`props-processor.ts` gains the two multi-select types on the existing `PropertyType.ARRAY` validator line. Under the wrapping rule that only ever fires for a value that resolved to nothing, which is exactly the case that should fail.

`DROPDOWN`/`STATIC_DROPDOWN` are deliberately excluded: option values are piece-defined and polymorphic, so a legitimate string value like `"[1,2]"` would be corrupted into an array. That gap wants its own ticket.

## Toggling back to manual

`getValueForInputOnDynamicToggleChange` ignored `currentValue` in the `MANUAL` branch and always returned `getDefaultPropertyValue`, so switching a field back reset it to the piece's default. On Date Helper that default is `['year']`, which made "pick Year + Day, toggle to dynamic, toggle back" come out as just Year and read as though the field had kept only the first value.

Both directions now go through `piecePropertiesUtils.parseDynamicValue` (`packages/pieces/framework/src/lib/property/util.ts`, beside `buildSchema`), which restores a value **only when its shape is unambiguous**. This helper stays strict on purpose while the engine wraps: the builder sees values that are still expressions, so wrapping there would turn `{{ trigger.body.units }}` into a bogus one-item selection. Single-select dropdowns still reset, for the same reason the engine leaves them alone.

This is also the recovery path for a flow already saved in the broken state — the user toggles back and their selections return.

## What this PR deliberately does *not* do

An earlier revision added a flow migration (schema `22` → `23`) that rewrote stored `DYNAMIC` string values and flipped the field to `MANUAL` so the picker came back. It has been removed, along with the schema bump and the breaking-changes entry, because:

- the engine parse already fixes **execution**, which is the actual bug;
- the migration only restored the builder's *picker*, which the toggle-back path now does on demand;
- dynamic JSON properties have never had such a migration, and nobody misses it;
- it cost a schema bump, a backup write and an on-call page path per flow, plus ~410 lines in `server/api`, to change how a field looks.

## How was this tested?

- **Unit, red-first.** `packages/server/engine/test/variables/props-validator.test.ts` — 29/29. With the two processors unregistered, exactly **14** fail and 15 pass.
- **No engine regressions.** Full suite 30 failed / 415 passed; re-run with the change stashed gives an identical failing *file* list (the 10 piece-loader files, which need `turbo run build --filter=<piece>` first — see the `building-pieces.md` gotcha).
- **Schema revert is clean.** `server/api` migrations suite 117/117, `server/sandbox` 179/179 (it compares `LATEST_FLOW_SCHEMA_VERSION`). No `migrateV22` or `'23'` references left anywhere.
- **Web helper.** 10 tests in `packages/web/test/features/pieces/parse-dynamic-value.test.ts`, red-first against the old always-use-the-default behaviour.
- **Lint / types.** 0 errors on `@activepieces/engine`, `api` and `web`; `tsc -p packages/server/api/tsconfig.app.json --noEmit` 0 errors.
- **Review size.** `bun tools/scripts/pr-size-check.ts` — engine 283/300, `server/api` 0, total 362 (was 793).
- **End to end**, from the original revision of this PR: Date Helper "Extract Date Units" on a CE dev stack with the multi-select toggled to dynamic — failing before, succeeding after — including the customer's exact `[]` case, plus the manual round trip that keeps both selections (see the Visual verification comment). That run still covers the parse path and the toggle-back, which behave identically here. **The wrapping path added in this revision (`'year'` → `['year']`) has not been re-run against a live stack — it is covered by unit tests only.**

### Known gaps left open

- `DROPDOWN`/`STATIC_DROPDOWN` with non-string option values still reach `run()` as a JSON string with no error. The fix is to parse and then accept the result only if it matches a declared option value; it wants its own ticket, together with the single-select field resetting to its default when toggled back to manual.
- A **required** multi-select still accepts an empty list. `buildSchema` declares `z.array(z.any()).check(z.minLength(1))` for a required multi-select, but the engine validator only checks `Array.isArray`, and in dynamic mode `'[]'` is a 2-character string that satisfies the union's `z.string().check(z.minLength(1))` arm — so a required field can execute with zero selections. Pre-existing (multi-select had no engine validator case at all before this PR), and deliberately not closed here because Date Helper's `unitExtract` is required and `[]` is precisely the customer value this PR makes work.

Every value that reaches `run()` today either keeps reaching it or arrives in the shape the field declares. A multi-select holding a single value used to pass a bare string and now passes a one-item list — the shape the piece asked for. The one outcome that changes is a **required** multi-select whose value resolves to nothing: it used to reach `run()` as `undefined` with no error and now fails with `Expected array, received: <value>`. A required list field that received nothing and still succeeded is not a working flow, so there is nothing to announce and no migration to write.


majewskibartosz:
## Visual verification

Driven as a real user on a local CE dev stack (`turbo run serve` with api + worker + engine + web, PGlite): flow with Date Helper → Extract Date Units, "Unit to Extract" multi-select toggled to dynamic mode, step tested via the builder. "Before" screenshots were taken with the fix stashed (base commit engine); "after" with the fix applied. Same flow, same saved value.

| State | Screenshot |
|---|---|
| Manual mode: Year + Month selected | manual mode |
| After fx toggle: value is now the literal text `["year","month"]` | dynamic toggle |
| **Before fix** — Test Step fails (`Invalid unit to extract: "["` — the action iterated the string per character) | before: test failed |
| **After fix** — same value tests successfully, output `year: 2026, month: 7` | after: test success |
| **After fix** — the customer's exact `[]` case: tests successfully with empty output | after: empty array |

States driven: manual multi-select, fx-toggled stringified value, failed pre-fix test, successful post-fix test, empty-array (`[]`) post-fix test. Single (light) theme — the change has no theme-sensitive surface. Checkbox coercion is covered by unit tests (`props-validator.test.ts`, red-first).

hazemadelkhalel (approved):

majewskibartosz:
@greptileai review

majewskibartosz:
@greptileai review

hazemadelkhalel:
### Brain follow-up: give properties their own page, and record this as a decision

Not blocking the code — this is about where the knowledge lands, and a good chunk of the bulk is mine, so I'm not pointing at anyone.

**1. `brain/wiki/pieces-engine/building-pieces.md` has one gotcha bullet that is now 2,907 characters in a 29-line file.**

It has stopped being a gotcha and become a page in disguise. It currently carries the fx/dynamic-value toggle, `buildSchema`'s `z.string()` union, the `processors` map, `validateProperty`, empty-string handling, the toggle-back trap, and which property types are deliberately excluded. Nobody scanning "Building pieces" expects to find the property-value contract in a single bullet, and nobody looking for the property-value contract thinks to open "Building pieces".

Suggested shape — `brain/wiki/pieces-engine/properties.md`, listed in `brain/wiki/pieces-engine/index.md` alongside `pieces.md` and `piece-sets.md`:

- the property types and where they are declared
- **the dynamic-value contract**: dynamic mode is a text input, so the stored value is a string; `buildSchema` unions `z.string()` on purpose (it is also the publish validator, `flow-version-validator-util.ts:203`), and `resolveInputAsync` only preserves a resolved type when the whole input is a single token (`props-resolver.ts:166`)
- the `processors` map as the one place to coerce, with `validateProperty` as the one place to report, and why the JSON block is shaped differently (`jsonProcessor` destroys information on failure; the others do not)
- the excluded types and why: `DROPDOWN`/`STATIC_DROPDOWN` because a real option value `"[1,2]"` is indistinguishable from a stringified one, and `ARRAY` because `props-validator.test.ts` already pins `array: 'not an array'` → `Expected array`, which a wrapping processor would flip
- the web/engine asymmetry: the engine wraps a non-array, `formUtils.parseDynamicValue` must not, because the builder sees values that are still `{{ expressions }}`

Then `building-pieces.md` keeps a one-line bullet pointing at it. Same for the engine-vitest and piece-`dist` gotchas already there — they are about running tests, not about building pieces.

**2. The design call in this PR deserves `brain/decisions/000016-*.md`.**

It is exactly the hard-to-reverse kind, and the reasoning only exists in this PR's conversation right now. Something like *"dynamic-mode values are coerced at runtime, never migrated at rest"*:

- **Context** — a multi-select/checkbox switched to a dynamic value reaches `run()` as a string ([Pylon 4586](https://app.usepylon.com/issues?issueNumber=4586)).
- **Decision** — coerce in the engine `processors` map, following `jsonProcessor`. Do **not** rewrite stored flow values.
- **Why** — a dynamic JSON property has always had this problem and never got a migration; the stored string is a *valid* dynamic value, not corrupt data. A flow migration was written for this PR (schema 22 → 23) and removed: it only restored the builder's picker, which the toggle-back path now does on demand, and it cost a schema bump, a per-flow backup write, an on-call page path, and ~410 lines in `server/api`.
- **Consequences** — the builder shows a text box in dynamic mode, and that is accepted as how dynamic mode looks. Coercion is only safe where the declared shape is unambiguous, which is what keeps `DROPDOWN` out. The one runtime behaviour change is that a **required** multi-select resolving to `null` now fails with `Expected array` instead of reaching `run()`.
- Supersedes nothing, but worth cross-linking from the `objectProcessor` precedent (#5636).

The value is in the *rejected* option. Without it, the next person to hit a stringified property type will reach for a migration again, and the PR that says why not will be buried.

PR #14400, opened 2026-07-24
fix(cli): verify a piece bundle installs and loads in isolation before publish

kishanprmr:
## Description

`piece-openai@0.10.0` published broken (#14369): tiktoken's inlined JS shipped without its wasm asset, and the manifest declared no dependency to fetch it, so `require()` threw before any action ran — connections, model dropdowns, and action execution were all broken.

Nothing in the release pipeline ever installed the *published* package in isolation before shipping it. The only post-build `require()` check (`update-pieces-metadata.ts`) runs inside the monorepo checkout, where hoisted workspace `node_modules` can mask a dependency the published manifest never declared — and it runs *after* `publish-pieces-to-npm.ts` has already pushed to npm, so even a caught failure wouldn't have blocked the publish.

Fix: `verifyPieceBundleLoads` (`packages/cli/src/lib/utils/verify-piece-bundle.ts`) packs the already-prepared `dist/` into the real tarball, installs it into a throwaway directory with zero ambient `node_modules`, and `require()`s it by name — reproducing exactly what a real installer does. Wired into `publishNpmPackage` (`tools/scripts/utils/publish-npm-package.ts`), right before the `npm publish` call, so a broken bundle blocks the publish for every call site (the bulk release pipeline and the single-piece CLI publish path both route through this one function).

## How was this tested?

- New test (`verify-piece-bundle.test.ts`): a passing bundle, and a failing one that reproduces the exact incident shape — code references an asset outside the manifest's `files` allow-list, so it's referenced but never packed.
- Ran against the real `piece-openai` bundle locally: packed, installed `tiktoken` (wasm included) from the registry into an isolated dir, required cleanly.
- Full `@activepieces/cli` suite: 24/24 passing.
- `npm run lint-dev` clean (pre-existing unrelated warnings only).

### DIFF
--- a/packages/cli/src/lib/utils/verify-piece-bundle.test.ts
+++ b/packages/cli/src/lib/utils/verify-piece-bundle.test.ts
@@ -1,0 +1,37 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import { verifyPieceBundleUtils } from './verify-piece-bundle'
+
+describe('verifyPieceBundleLoads', () => {
+    let root: string | undefined
+
+    afterEach(() => {
+        if (root) {
+            rmSync(root, { recursive: true, force: true })
+            root = undefined
+        }
+    })
+
+    it('passes for a bundle that installs and requires cleanly', async () => {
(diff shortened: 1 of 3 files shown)


kishanprmr:
Closing — the heap-usage CI job added in #15252 (`check-pieces-heap.ts`) already installs each changed piece's bundle into an isolated `node_modules`-free directory and requires it by name before merge, which catches this exact class of bug (broken/missing bundled asset like the tiktoken wasm case) at PR time. Not needed as a separate publish-time gate.

PR #14402, opened 2026-07-24
feat(builder): show connected-account identifier and status on connections

AhmadTash:
## Description

Shows *which* external account a connection authenticates as, across the flow builder and the project **Connections** page.

- **Connected-account identifier** — connections now display the external account they authenticate as. The server resolves it at connect time and stores it on the existing `metadata.accountIdentifier`; the Connections table (under the name) and the step-settings connection dropdown render it. Resolution is layered so it generalises across providers without vendor-specific code in the core:
  - **Generic (in core):** for OIDC providers the account **email** is decoded locally from the `id_token` returned by the token endpoint (`email`/`mail`, falling back to `preferred_username`/UPN for Microsoft). No network call.
  - **Provider-specific (in the piece):** anything the id_token can't express is delegated to a new optional piece-level `getConnectionIdentifier` auth hook, run through the engine via a new `EXECUTE_RESOLVE_CONNECTION_IDENTIFIER` operation (mirrors `EXECUTE_VALIDATE_AUTH`). Best-effort — a failure or slow response never fails or delays the connection upsert (wrapped in `tryCatch` + a 15s cap). The Slack implementation lands in the stacked piece PR.
  - Providers that expose no identity stay blank.
- **Expired/empty status** — the connection dropdown flags non-active connections (Missing / Error) inline, reusing the existing status display.
- **Connections "Flows" column** — clicking the flow count now opens the Automations list filtered to that connection instead of the unfiltered page (the old link dropped its query param and used the wrong param name).

No database migration and no `@activepieces/shared` change — the identifier reuses the existing `metadata` jsonb column already returned by the connections API. The new engine operation and job-data schema are additive (no `LATEST_JOB_DATA_SCHEMA_VERSION` bump).

> The unrelated piece-selector search-UI changes (action/trigger descriptions + "No results found" copy) were split out into #14437.
> **Stacked PR:** the piece-side change — the Slack `getConnectionIdentifier` hook plus requesting the identity OAuth scope so Google/Microsoft expose the account — is a follow-up PR stacked on top of this one.

## How was this tested?

Manually on a Community-edition dev instance (PGLite):
- Slack shows `name (workspace)` (via the piece hook + engine round-trip); Microsoft Outlook shows the account email (generic id_token path).
- Missing / Error status indicators render in the table and the dropdown.
- The "Flows" column filtering works.
- `api` / `worker` typecheck, `engine` build, and `web` typecheck + lint are green; dev server boots with the new job type registered. A 4-dimension adversarial regression review found no regressions.

Not exercised on EE / Cloud edition paths.

Fixes # (n/a)

**Security note:** the core resolver makes **no outbound HTTP** — it only decodes the provider's `id_token` locally (decode only, not signature-verified; used solely to display an email) to populate `metadata.accountIdentifier`. Any provider-specific hook that does make network calls runs inside the engine sandbox, not the API process. The sensitive `value` blob is still stripped by `removeSensitiveData`; only a plain email/label string is written to `metadata`.


AhmadTash:
@greptileai review this

Louai-Zokerburg (approved):
Only comment I have (soft requirement) is AI comments that I think clatters the code and I guess we agreed to not to put it unless very needed ones

amrdb:
Pushed `e772d786` to this branch — flagging it here since it touches the engine contract. Happy to pull it back out into its own PR if you'd rather.

### Why

The engine picked the auth property with `getAuthPropertyForValue({ authValueType: authValue.type })`. But `cloudOAuth2Service.claim` returns `{ ...value, token_url, props }` and never sets `type` — unlike its own `refresh` and unlike `credentialsOauth2Service.claim` — and `upsert` hands the engine the un-merged claim result, so the request's `type` is only re-spread for storage. That read is therefore `undefined` for **every cloud and platform OAuth connection**, and `getAuthPropertyForValue` falls back to `pieceAuth.at(0)`.

The structural narrowing was the right instinct about the symptom, but it doesn't close the gap:

| Guard | Against an OAuth claim result |
|---|---|
| `'access_token' in authValue` | discriminates ✓ |
| `'username' in authValue` | discriminates ✓ |
| `'secret_text' in authValue` | discriminates ✓ |
| `'props' in authValue` | **does not** — `claim` writes the key explicitly, so `in` is `true` even when the value is `undefined` |

So a multi-auth piece whose `CUSTOM_AUTH`/`OIDC` entry sorts first would pass that check and invoke the hook with the OAuth props bag typed as custom-auth props. Latent today only because no piece in the repo uses `auth: [`.

The fix is that the API already has the authoritative type from the request body, and was the only place still holding it — so it now travels on the operation and the engine dispatches on that. The shape checks stay as a second line of defence.

### Second change, which the first one unlocks

With the type no longer inferred from the value, the OAuth-only gate stopped being load-bearing, so only `NO_AUTH` is excluded outright. The `id_token` read stays OAuth-only (nothing else has a token response to read), but `SECRET_TEXT` / `BASIC_AUTH` / `CUSTOM_AUTH` / `OIDC` pieces can now resolve an identifier through their hook.

That's mostly about reaching code you'd already written: those four branches in `resolveConnectionIdentifier` were unreachable behind the gate. And it's arguably where the feature pays best — an OAuth account you can often infer from context, but a Stripe restricted key or an Airtable PAT tells you nothing about whose account it is. `declaresConnectionIdentifier` still gates the round-trip, so pieces without the hook cost nothing beyond the metadata read.

### Also

- Dropped `data.connectionValue as AppConnectionValue` in the worker job — `connectionValue` is now `z.custom<AppConnectionValue>()`, matching `ExecuteTokenRefreshJobData`.
- Updated the `oauth-apps.md` gotcha you added: it recommended the shape-narrowing approach, which would now send the next person the wrong way. It records the `props` asymmetry instead.
- Added `packages/pieces/framework/test/auth-property-for-value.test.ts` pinning that all three OAuth2 connection types map to the OAuth2 auth entry — the invariant this depends on. It lands in the framework suite you wired into CI, so it actually runs.

### Verified

`turbo run build` green for `api`, `worker`, `@activepieces/engine` + 14 deps (17/17). `turbo run lint` 0 errors on all five touched packages. Framework tests 8/8.

Not verified: no EE/Cloud run, and I didn't exercise a real non-OAuth `getConnectionIdentifier` end to end — no piece implements one yet.

### One thing I'd still raise separately

`connection-metadata.test.ts` doesn't run anywhere. CI runs `test-ce test-ee test-cloud --filter=api`, each a `vitest run test/integration/<edition>` path filter, and the root `test-unit` script omits `api` — with no `test-unit` task in `turbo.json` to call. So the tests covering the forged-identifier fix and the clear-on-reconnect fix are green by never executing. Pre-existing gap, not yours, but worth a follow-up.

amrdb (approved):

kishanprmr (approved):

PR #14405, opened 2026-07-24
fix(builder): close trigger piece selector on outside/repeat click

AhmadTash:
## Description

The empty-trigger "Select Trigger" piece selector popover would force itself back open whenever a click landed near the trigger node (its label, chevron, or "Trigger" badge), or when its context menu was invoked — instead of closing, even though the click was meant to dismiss it. It only stopped happening once a real trigger piece was picked.

Root cause: `selectStepByName` unconditionally force-reopened the piece selector any time the still-empty trigger step got (re)selected, regardless of intent — and the Popover was `modal={true}`, which makes Radix treat a click on its own trigger as an "outside" dismiss. So the app's own force-reopen logic and Radix's dismiss logic ended up racing and fighting each other on every click near the trigger.

Rather than patching around that race (e.g. snapshotting click state to out-guess the timing), this collapses ownership of "is the selector open" down to one place:

- `selectStepByName` no longer opens the piece selector as a side effect of selecting a step — it only ever closes whatever was open for the previously-selected step. Opening is now owned solely by the Popover trigger's own click handling.
- The Popover switches to `modal={false}`, so Radix correctly excludes its own trigger element from outside-click dismissal (built-in behavior for non-modal Popovers) instead of treating a trigger click as a dismiss that then races our own open logic.
- `onOpenChange` now syncs **both** directions (open and close) into the store, so Radix's Popover is the single source of truth for whether it's open.
- The "Complete 1 step" banner — the one caller that relied on the old auto-open side effect — now explicitly opens the selector itself when it focuses the empty trigger.
- Removed a stray `e.preventDefault()` in the step node's click handler: Radix's `Slot` merges that handler with the Popover trigger's own composed `onClick` on the same DOM node, and that composed handler bails out when `event.defaultPrevented` — so the `preventDefault()` call was silently breaking the trigger's own open/close toggle. (This is also why the original code manually forced the selector open/closed instead of trusting Radix's toggle — it was working around this.)

Net effect: nothing in application code fights Radix for ownership of "is this popover open" anymore, so the class of bugs where some code path force-reopens it (chevron click, context menu, plain click) is eliminated by construction rather than patched call-site by call-site.

### Follow-up fixes (same root cause area)

Two more regressions surfaced while testing the above:

- **Empty-trigger chevron showed a nonsensical menu.** The chevron button always synthesized a `contextmenu` event regardless of step type, so an empty trigger's chevron showed the generic "Replace / Copy reference / Paste After" menu — none of which apply before a piece is even chosen. `StepNodeChevron` now takes an optional `onClickOverride`; for an empty trigger it opens the piece selector directly instead of dispatching the context-menu event, matching every other click target on that node. Configured steps are unaffected — their chevron still shows the real menu.

- **"Replace" silently failed to open the selector if the mouse moved right after clicking it.** Root cause: the clicked context-menu item still has DOM focus when the menu unmounts, and focus falls through with nowhere to land in the same tick — Radix's `DismissableLayer` sees that as a `focusin` from outside the freshly-opened Popover and dismisses it. The Popover now ignores `focusin`-sourced outside-interactions specifically; genuine outside pointerdown clicks and Escape still dismiss it exactly as before.

## How was this tested?

Manually verified in the flow builder (Create New → New Flow → Select Trigger) against a running dev instance:
- Click the empty trigger to open the selector, click it again (on the actual piece logo, inside the real Popover trigger) → now closes instead of reopening.
- Click empty canvas while open → still closes (unaffected).
- Click the closed trigger again → still reopens (unaffected).
- Click the chevron on an empty trigger → opens the piece selector directly (no more generic context menu).
- Click the chevron on a configured step → still shows the real context menu (Replace / Copy reference / Paste After / Delete / etc.), unaffected.
- Click "Replace" on a configured step, then move the mouse fast → piece selector now opens reliably (previously silently failed).
- Click the "Complete 1 step" banner → opens the piece selector for the empty trigger.
- Picked a real trigger (Google Sheets → New Row Added) end-to-end → selector closes and the settings panel opens as expected.
- Clicking a configured step's logo/label (not chevron) still just selects it — doesn't unexpectedly pop open the replace-piece selector.

Not tested against EE/Cloud-specific paths — this is a pure client-side flow-builder interaction fix with no edition-gated code.

Fixes # (issue)


AhmadTash:
@greptileai review this

AhmadTash:
@greptileai review this

othmanemad (approved):

PR #14413, opened 2026-07-25
fix(pieces): expose core and universal-AI actions to agent discovery

ibrahim-abuznaid:
## Description

Every action under `packages/pieces/core/` — plus `@activepieces/piece-ai`, the universal AI piece — is tagged `audience: 'human'`, which removes it from the agent discovery surface. An agent building a flow cannot find `http`, `text-helper`, `tables`, `image-helper`, `approval`, `subflows`, or any of the AI actions **at all**.

Verified on cloud before this change — both queries return `{"results":[]}`:

- `"send an HTTP POST request to a REST API endpoint"`
- `"replace text in a string and convert to uppercase"`

Two things make the current state a bug rather than a deliberate policy:

1. **The values are wrong on their own terms.** A text or math action is a legitimate agent tool. The tag was applied to whole trees rather than reasoned about per action.
2. **The result is incoherent.** Because republish order lagged the tagging, `delay` (which waits inside a flow run) is discoverable today while `text-helper` (pure string manipulation) is not. Nobody would design that.

This PR flips **112 actions across 26 pieces** to `audience: 'both'` and bumps each piece's patch version:

- **106 actions across the 25 action-bearing `core/` pieces.** `schedule` and `manual-trigger` have no actions and are untouched.
- **6 actions in `@activepieces/piece-ai`** — `askAi`, `summarizeText`, `classifyText`, `extractStructuredData`, `generateImage`, `run_agent`.

### Why the universal AI piece belongs here

It is the highest-value case in the fenced set, and the one where the incoherence is sharpest. The piece **needs no connection** — it runs on platform-managed models — so it is the idiomatic way to add an AI step to a flow. It already ranks first for `"use AI to summarize a long piece of text"`.

It ranks first *only* because its published `audience` is still `NULL`. On its next publish it disappears — while the third-party AI vendor pieces that Phase 2 curated as `'both'` (`textcortex-ai`, `eden-ai`, `rapidtext-ai`) stay visible. AP's own no-auth AI piece would be the one hidden, and the paid third-party ones would be all that an agent could find.

**No canvas or REST behaviour change.** `filterActionsByAudience` defaults to the `HUMAN` filter, which excludes only `audience !== 'ai'` — so `'human'` and `'both'` are equally visible to the builder UI, and `?audience=ai` has no callers in the web app.

**Deliberately out of scope:**

- `custom_api_call` — the ~466 factory instances stay `'human'`. They are retrieval noise (most share the literal name `custom_api_call`) and a blind proxy; the right handling is exclusion in the search layer, not an audience flip.
- **The 22 vendor-specific LLM wrappers (~93 actions: `openai`, `straico`, `google-gemini`, `claude`, `amazon-bedrock`, …) stay fenced.** Their problem is invoke-time cost and answer quality rather than discoverability: invoking one spends the user's own vendor quota, and a weaker model answering without conversation context is a silent quality regression. Surfacing 93 near-identical vendor actions would also crowd the top five with variants of a single capability — a live search for text summarisation already returns four vendor summarizers in its top five. Revisit once the invoke path can express "authorable but not invokable".

## How was this tested?

The change is metadata-only, one line per action, so verification is mechanical:

- `audience:` count equals `createAction(` count in every touched piece — **112 == 112**, no per-piece mismatch, so no action was missed or double-written. Every `audience:` line sits in a `createAction` file, so no trigger is affected.
- `grep -rn "audience: *'human'" packages/pieces/core packages/pieces/community/ai` — zero hits after the change.
- The full diff is **138 files, 138 insertions, 138 deletions**: 112 action files (one `audience` line each) + 26 `package.json` version bumps. Nothing else is touched.
- `eslint` on all changed action files — 0 errors (only pre-existing warnings; the three `no-explicit-any` warnings in `extract-structured-data.ts` are at lines 151/168/177, unrelated to the `audience` line).
- `tsc --noEmit -p …/tsconfig.lib.json` on the touched pieces (`text-helper`, `http`, `tables`, `store`, `ai`) — clean. `'both'` is an existing member of the `Audience` union, so this is a same-union swap.
- No test asserts an audience value on a real core or AI piece; the audience-filter tests use synthetic fixture pieces.

Note: `npm run lint-dev` is **not** a valid gate for this PR — it runs `turbo run lint --filter='!@activepieces/piece-*'`, which excludes every piece package, so it would pass without linting anything this PR touches. The eslint run above is over the changed files directly.

Edition paths: metadata only, no edition-specific code.

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/ai/package.json
+++ b/packages/pieces/community/ai/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-ai",
-  "version": "0.4.4",
+  "version": "0.4.5",
   "type": "commonjs",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
--- a/packages/pieces/community/ai/src/lib/actions/agents/run-agent.ts
+++ b/packages/pieces/community/ai/src/lib/actions/agents/run-agent.ts
@@ -64,7 +64,7 @@ const agentToolArrayItems: ArraySubProps<boolean> = {
 }
 
 export const runAgent = createAction({
-  audience: 'human',
+  audience: 'both',
   name: 'run_agent',
   displayName: 'Run Agent',
   description: 'Handles complex, multi-step tasks by reasoning through problems, using tools accurately, and iterating until the job is done.',
--- a/packages/pieces/community/ai/src/lib/actions/image/generate-image.ts
+++ b/packages/pieces/community/ai/src/lib/actions/image/generate-image.ts
@@ -23,7 +23,7 @@ import { AIProviderName } from '@activepieces/pieces-framework';
 import { aiProps } from '../../common/props';
 
 export const generateImageAction = createAction({
-  audience: 'human',
+  audience: 'both',
   name: 'generateImage',
   displayName: 'Generate Image',
   description: 'Create unique, high-quality images from simple text descriptions using AI.',
--- a/packages/pieces/community/ai/src/lib/actions/text/ask-ai.ts
+++ b/packages/pieces/community/ai/src/lib/actions/text/ask-ai.ts
@@ -9,7 +9,7 @@ import { createAIModel } from '../../common/ai-sdk';
 import { buildWebSearchOptionsProperty, buildWebSearchConfig, WebSearchOptions } from '../../common/web-search';
 
 export const askAI = createAction({
-  audience: 'human',
+  audience: 'both',
   name: 'askAi',
   displayName: 'Ask AI',
   description: 'A flexible AI step. ask it to analyze data, explain, draft, or decide based on your flow\'s data.',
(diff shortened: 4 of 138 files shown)


ibrahim-abuznaid:
On the security box, since `http`, `sftp`, `smtp` and `crypto` are in scope and that is worth being explicit about rather than a bare checkbox:

`audience` is read **only on the discovery path**. Nothing on the invoke path reads it — `pieceMetadataService.get()` does not filter on it, and `ap_run_action` already executes a `'human'`-tagged core action today (confirmed against a live run). So this flip grants no capability that was not already reachable; it changes which actions an agent can *find*, not which it can *run*.

`ap_research_pieces` also already passes the `ALL` audience filter, so every core action has been discoverable through that tool since before the search tool existed. The exposure predates this change.

kishanprmr (approved):

PR #14364, opened 2026-07-22
docs: add Large File Streaming section

othmanemad:
## Description

Adds user- and developer-facing documentation for the **large file streaming** capability (PRs #14251, #14356, #14347).

- New page `build-pieces/piece-reference/large-file-streaming.mdx`: what streaming is and why (worker memory budget / OOM), when to use it (buffer vs. stream), which pieces support it (Amazon S3 today), and a developer how-to for streaming **reads** (`Property.File({ streaming: true })` → `ApStreamingFile`) and **writes** (`ctx.files.write` accepting a `Readable`). Prominently discloses that full streaming requires S3 file storage, plus size limits and caveats (one-shot streams, dropped multipart webhook signature verification).
- Registers the page in the Piece Reference nav group.
- Cross-links from `files.mdx` (the `Write` section now notes `data` accepts a `Readable`) and from the user-facing `flows/passing-data.mdx` so flow builders discover it.

All three implementation PRs are now merged, and `main` is merged into this branch. Reviewing the docs against the merged code turned up four claims written against a design that had since changed, plus two stale `brain/` pages — corrected in 80a32c5:

- `#14347` rewrote Amazon S3 **Upload File** from `putObject({ Body, ContentLength })` to `lib-storage`'s `Upload`. The how-to told piece authors to pass `ContentLength: file.size`, steering them onto the size-dependent path that silently falls back to buffering whenever `size` is absent — including any `Content-Encoding`-compressed source, i.e. the exact case streaming exists for. Now shows `Upload`, and demotes `size` to informational.
- The retry caveat claimed any failure after the transfer starts is fatal. True of a single PUT, wrong of `Upload`: each ~5&nbsp;MB part is buffered before sending, so parts are individually replayable. What is non-retryable is the transfer *as a whole* (a one-shot source `Readable`).
- `AP_MAX_FILE_SIZE_MB` was described as applying to streamed files generally. It is enforced on the `ctx.files.write` path only; streamed **inputs** are deliberately uncapped (see decision 000014).
- Dropped an invented HTTP `413` status — no code path returns it.
- Decision `000014` and `brain/wiki/data-storage-observability/file-storage.md` both still documented the superseded `putObject` design; they now record what replaced it and why.

## How was this tested?

[...]

PR #14417, opened 2026-07-26
fix(pieces): stream large files to prevent sandbox OOM

othmanemad:
## Description

Six storage pieces buffered whole files into the piece sandbox on upload and download, which OOMs the worker on large files. They now stream instead, using the primitives already on main (`Property.File({ streaming: true })` → `ApStreamingFile`, and `ctx.files.write({ data: Readable })`).

| Piece | Upload (streams in) | Download (streams out) |
|---|---|---|
| Dropbox | Upload file | Download File |
| Google Drive | Upload file | shared `get-file-content.ts` — also serves Read File Content, List files, Set public access, New File trigger |
| Microsoft OneDrive | Upload file *(chunked loop rebuilt)* | Get File |
| Microsoft SharePoint | Upload File | — |
| Azure Blob Storage | Create Blob | Read Blob |
| FTP/SFTP | Upload File | Read File Content |

**Downloads** hand the source `Readable` straight to `ctx.files.write` instead of `Buffer.from(arrayBuffer)`.

**Uploads** consume `file.body`. How depends on what the destination client accepts:
- Azure `uploadStream`, Google Drive `media.body` and SFTP `client.put` take the stream directly.
- Dropbox, SharePoint and OneDrive upload in a single HTTP request, so they need an explicit `Content-Length` — they use `file.size` and fall back to buffering only when the source reports no size (identical to pre-streaming behaviour).

### Shared enabler: `pieces-common` `httpClient`

- `responseType: 'stream'` returns a `Readable`; a `Readable` request body is sent with `duplex: 'half'`.
- **Stream bodies now skip the retry loop** (`retries` forced to `0`). The loop reuses the body serialized before the first attempt, and both stream shapes — a raw `Readable` and the `PassThrough` a `form-data` payload is piped into — are one-shot, so a retry would replay a drained stream and send a truncated body. Real replay would need a body-*factory* API and a change at every call site. This applies to any piece sending a stream through `httpClient`, not just these six.

### Piece-specific notes

- **OneDrive** pulls fixed-size 10 MiB chunks (a multiple of the 320 KiB Graph requires) off the `Readable` for its resumable upload session, computing `Content-Range` per chunk.
- **SharePoint** uploads through `httpClient` rather than the Graph SDK: `@microsoft/microsoft-graph-client@3.0.7` can't stream a `Readable` under undici (`.put()` JSON-stringifies it; `.putStream()` builds a `Request` with no `duplex` and throws). The SDK still resolves the parent folder.

Self-hosted **DB-backed** installs are unaffected: `fileService.save` buffers a streamed `files.write` into the `bytea` column server-side, the same memory profile DB storage always had. The sandbox-side OOM fix holds on every backend; S3 backends stream end-to-end.

### Docs

`docs/build-pieces/piece-reference/large-file-streaming.mdx` (added in #14364) was written when Amazon S3 was the only streaming piece, so it was stale on arrival:

- Piece table now lists all seven pieces, split into streams-in / streams-out.
- New **per-service upload ceilings** section: Dropbox `/2/files/upload` is [documented for files under 150 MB](https://developers.dropbox.com/dbx-performance-guide) and Graph's simple `PUT …/content` [supports up to 250 MB](https://learn.microsoft.com/en-us/graph/api/driveitem-put-content), so those two actions stay bounded by the destination API rather than by memory. OneDrive escapes this via its chunked session; SharePoint has no chunked path. Both are pre-existing API limits, not regressions.
- The "reading a file as a stream" how-to teaches three uploader patterns in preference order instead of only the S3 one.
- The retry caveat and `packages/pieces/CLAUDE.md` now document the `httpClient` no-retry rule.
- Decision `000014` records that rule with the rejected body-factory alternative, and names the three consumers that still buffer when `size` is absent (main's text implied none did).

Version bumps: `pieces-common` (0.12.7) + the 6 pieces.

## How was this tested?

- `turbo run build` (tsc) + `turbo run lint` across all affected packages — 0 errors.
- OneDrive `readChunks` boundary logic verified across edge cases (empty, sub-chunk, exact multiple, remainder, byte-at-a-time, single push): every non-last chunk equals the chunk size, `Content-Range` offsets stay contiguous, bytes reassemble exactly.
- Retry safety audited per destination, since a drained stream is the failure mode: Azure's `BufferScheduler` stages each block as a `() => buffer.getReadableStream()` **factory**, so its retry policy replays blocks; gaxios' default `httpMethodsToRetry` excludes POST, so Google Drive's `files.create` never retries; SFTP has no retry layer; the three `httpClient` pieces are covered by the `retries: 0` guard above.
- DB vs S3 write path confirmed by reading `fileService.save` (`FileLocation.DB` buffers the `Readable` into `bytea`; `FileLocation.S3` uses `uploadStream`).
- Both documented size ceilings checked against the vendor docs linked above.
- Edition-agnostic (piece-level, no server or edition branching). Not exercised against live third-party APIs.

Fixes # (no issue)

Action inputs and outputs are unchanged, `Property.File` reuses `PropertyType.FILE` (zero frontend change), and no env vars, DB columns, or API fields are touched. Self-hosters need take no action.

One judgement call worth a reviewer's eye: the `httpClient` `retries: 0` guard is a behaviour change for any piece that sets `retries` **and** sends a stream body. Today that is only the actions in this PR, and the previous behaviour silently sent a truncated body, so this is scored as a bug fix rather than a functional break. Say the word if you'd rather it carried the `⛓️‍💥 breaking-change` label and a `breaking-changes.mdx` entry.

Touches file handling and outbound HTTP body/response handling in the sandbox `httpClient`. **No new exposure:** SSRF egress is enforced at the engine's `dns.lookup` / `Socket.connect` guards, which are unchanged and independent of body or response type, and the server-side streamed-upload path keeps its `enforceByteLimit` guard. This is the sandbox `httpClient`, not `packages/server/{api,worker,utils}`, so the `safe-http` rule does not apply.


AbdulTheActivePiecer (approved):
// @ts-expect-error -- undici streams a Node web ReadableStream body; the DOM fetch types omit the fromWeb overload

can you please double check that if we switch the lib config in the tsconfig of the packages to es2022 would rid us of this type issue or not ?

kishanprmr (approved):

PR #14426, opened 2026-07-26
feat(action-run): execute single actions in the engine, no temporary flow

othmanemad:
## Description

Splits #14176 in two. **This PR is the execution half only — nothing is persisted and there is no UI.** The `action_run` table, its endpoints, and the "Action runs" tab follow in PR 2.

MCP's `ap_run_action` and the chat `ap_execute_action` / `ap_run_code` tools ran a single step by building a **temporary flow**: create a throwaway flow → graft one step onto it → `flowRunService.test()` → poll `flow_run` every 2s → dig the step out of `run.steps` → best-effort delete the flow. Three rows written per call, `__actionRun__` flows leaked into users' flow lists whenever the `finally` delete failed, and because `flow_run.flowId` is `onDelete: CASCADE`, a *successful* cleanup deleted the run row it had just paid three inserts for.

An action run is now a purpose-built engine operation, `EngineOperationType.EXECUTE_ACTION`, dispatched as a **synchronous user-interaction job** — the same mechanism as property resolution and auth validation. No flow, no polling, and deliberately **no retry**: an ad-hoc action is overwhelmingly a side-effecting single call ("send one Slack message"), and silently re-firing one after a worker restart is worse than failing it. Rationale in `brain/decisions/000014`, companion to `000012` which decided the storage shape.

The engine runs the step against an empty `FlowExecutorContext` via `actionRunStepRunner`, a shared primitive the chat tool executor was refactored onto — both paths now go through one code path instead of two copies.

### One end-to-end deadline

The caller previously inherited the watcher's shared `WATCHER_SAFETY_TIMEOUT_MS` and could block for 5 minutes. There is now a single 120s budget stamped on the job as `expiresAt`:

- The sandbox clamps the run to what is left of that budget **after** provisioning and boot (a cold boot's bind retries and 30s connect cap otherwise ran past it), and refuses to start the engine when nothing is left.
- The watcher holds 120s + 10s grace, covering only the sandbox kill and the pubsub hop home.
- A watcher timeout maps to `TIMEOUT`, not `INTERNAL_ERROR` — the old mapping told the calling agent *"the engine crashed while loading or executing the piece"* when nothing had run.
- On timeout the service cancels the job. Still-queued means it provably never executed, so the caller gets `neverStarted` instead of "may have partially completed", and the job cannot write after the caller walked away.

### Other fixes on the way through

- **Code cache namespacing.** The cache is keyed by `flowVersionId` + step name, and an action run passed a constant for both — so every `ap_run_code` compiled into one shared directory (concurrent runs could execute each other's snippet) and in isolate mode no `/root/codes` mount was built at all, breaking code steps outright. The namespace is now `hashObject(sourceCode)`, shared by the artifact, the provision mount and the engine's read path.
- **`step` was `z.custom()`**, which accepts a missing step and accepts `42`. It is now a real discriminated union, parsed again in the service so a malformed step fails fast.
- **The returned id is now the watcher's `requestId`** rather than an unrelated `apId`, so it traces downstream.
- **`ap_remember` was missing from the worker's cross-project tool map**, so the model never saw it.

### Behaviour note for reviewers

`ap_run_action` previously got a real, unique `flowId` from its temp flow, so **FLOW-scoped** `context.store` entries were isolated per call. They now share the `DEFAULT_MCP_DATA` sentinel `'mcp-flow-id'`, so flow-scoped entries from different action runs land in one namespace. Cross-project isolation is unaffected (scope comes from the `engineToken`, not `flowId`), and `context.files` never touches `flowId`. The chat tool executor already used these sentinels; this aligns MCP with it. Documented in `brain/wiki/flows-execution/action-run.md`.

## How was this tested?

- **Manual testing** of `ap_run_action` end to end.
- `turbo build` + `turbo lint` across api, worker, engine and sandbox — clean.
- New unit tests: 22 for `deriveActionRunOutcome` (full status table, watcher timeout → `TIMEOUT`, never-started detection) and 10 for the sandbox budget clamp. The sandbox package's tests now run in `test-unit` and CI, which they previously did not.
- api unit suite: 563 passed. The 9 failing files are **byte-identical to `main`'s baseline** (`system.getNumber is not a function` mock breakage), so pre-existing.
- engine suite: 356/356. `mcp-tools.test.ts` against a live CE DB: 100 passed, 3 skipped.

`ap_run_action` is CE. The chat tools that call the same helpers are EE and change by **rename only** here (the project-resolution change is held back for PR 2). No new env var, secret, or setup step.

Part of #14176

No API field, endpoint, column, or env var changes. The only externally visible difference is that `ap_run_action` no longer creates and deletes a disposable flow — which users were never supposed to see.

Flagged for the `context.store` namespace change above, since it is a data-isolation change. Scoping itself is unchanged: the `engineToken` is minted per `EXECUTE_ACTION` job and carries project/platform scope, as for every other user-interaction job. The narrowing is *within* a single project, for FLOW-scoped store keys only.


othmanemad:
@greptileai review please

abuaboud (approved):

PR #14437, opened 2026-07-27
feat(builder): show action/trigger descriptions in the piece selector

AhmadTash:
## Description

Improves the flow-builder **piece selector** search results.

- **Action/trigger descriptions** — search results now show each action's / trigger's description under its name (previously only the display name was shown); the result rows are sized to fit the extra line.
- **Empty-state copy** — the piece-selector empty state now reads "No results found" instead of "No pieces found" (the selector searches actions/triggers, not only pieces), with ar/ru translations added.

Split out of #14402, which is now solely the connected-account identifier feature.

## How was this tested?

Manually on a Community-edition dev instance: descriptions render under action/trigger names in the piece selector, and the empty state shows "No results found". `web` typecheck + lint are green.

Not exercised on EE / Cloud edition paths (frontend-only, edition-agnostic).

Fixes # (n/a)


AhmadTash:
@greptileai review this

Louai-Zokerburg (approved):

PR #14446, opened 2026-07-27
refactor(ai-providers): shared base language-model factory

Louai-Zokerburg:
## Description

Second foundation PR for the **AI Providers** project (Milestone 1, ENG-432). No user-facing change.

The "read a provider's auth/config → build the AI SDK client" logic was duplicated across two near-identical `switch (provider)` blocks — one in the AI piece (`ai-sdk.ts` → `createAIModel`) and one in the chat worker (`chat-ai-utils.ts` → `createChatModel`). Adding or fixing a provider meant editing both. This PR extracts that construction into a single shared lib.

- **New `@activepieces/ai-providers` lib** exposing one `createLanguageModel()`. Both callers now delegate to it.
- The parts that genuinely **differ** between the piece and chat became explicit options instead of duplicated code: OpenAI Responses API, Custom-provider metadata headers, Mistral-via-OpenRouter vs direct API, and OpenRouter web-search plugins.
- What stays **caller-local** (too divergent or one-sided): Cloudflare-Gateway routing (full sub-provider routing in the piece, compat endpoint in chat), image models, and embeddings — all piece-only or shaped differently per caller.
- The lib is **inlined into the piece bundle** like `pieces-framework`/`pieces-common` (bundler inlines every `@activepieces/*` workspace lib by default and strips it from the published manifest), so self-hosted installs stay self-contained.

Complements the merged PR #14433 (the `AI_PROVIDER_CAPABILITIES` descriptor): that PR is the single source of truth for *what* a provider supports; this PR is the single place for *how* to build its client. Together they complete Milestone 1's foundation.

## How was this tested?

- Behaviour preserved by design — per-provider terminal calls and Custom-header precedence kept identical; a new unit test covers the factory across all providers plus the Cloudflare "handled by caller" guard.
- turbo build green for `ai-providers`, `piece-ai`, `server-utils`, and `worker`.
- Unit tests pass: new `ai-providers` factory test (11) + `server-utils` suite.
- Lint clean on the changed packages (the new lib carries the same `no-import-@activepieces/shared` boundary rule as `pieces-common`).
- Edition-agnostic: no EE/Cloud-gated code touched; managed and BYOK provider paths both construct models unchanged.

Part of the AI Providers project · Milestone 1.


Louai-Zokerburg:
@greptileai

othmanemad:
Checked the refactor and the parity claims hold — header precedence, per-provider terminal calls, and the OpenAI/Azure/Bedrock/Custom image paths are identical; all three packages build; the 19 tests pass; and the piece bundle inlines the lib from source via `tsconfig.base.json` paths. Four things before merge:

**1. High — this publishes `@activepieces/ai-providers` to public npm on merge.** `packages/pieces/ai-providers` reads as a piece to the release scripts, and it's missing from both exclusion lists: `NON_PIECES_PACKAGES` (`piece-script-utils.ts:29`) and `notPublished` (`validate-publishable-packages.ts:19`), which name only `pieces-framework` and `pieces-common`. `release-pieces.yml` fires on `packages/pieces/**` → `find-changed-pieces` flags it (an internal lib is never in the cloud piece registry) → `publish-pieces-to-npm` reaches `npm publish --access public --tag latest`. That contradicts the invariant in `publish-npm-package.ts` and isn't reversible. It also strips `dist/` to the bundle, deleting the `index.d.ts` that `server/utils` resolves through `types`. Fix: move it to `packages/core/ai-providers` — it's a cross-cutting lib, imported by a piece *and* the server.

**2. Medium — Azure chat changes API version.** `create-language-model.ts:32` forwards `apiVersion`; the old `createChatModel` deliberately passed only `resourceName` and `apiKey`. `config.providerConfig` populates it, so the request URL changes for any platform that set the field. Likely an improvement, just not "behaviour preserved" — worth calling out or gating.

**3. Medium — duplicate SDK resolved in one server process.** `server/utils` gets `@ai-sdk/openai-compatible@2.0.16`, `ai-providers` gets `2.0.42`, and both models reach the same `streamText`. It works only because both currently land on `ai@6.0.170`: `server/utils` uses ranges (`^6.0.0`, `^3.0.0`) while the new package pins exact, so a later install can straddle a provider-spec bump. Align `server/utils` to the exact pins.

**4. Medium — the new tests never run in CI.** `test-unit` and `ci.yml:152` filter to engine/shared; `ai-providers` is in neither, so the tests two of these commits add only run locally. `--passWithNoTests` would also hide a broken glob, and `tsconfig.lib.json` excludes `*.test.ts` so the file is never typechecked.

othmanemad (approved):

AhmadTash (approved):

PR #14450, opened 2026-07-27
fix(mcp): keep MCP OAuth sessions alive past 15 minutes

### LINKED ISSUE #14411: [BUG]: MCP OAuth refresh response omits refresh_token, causing a ~15-minute reconnect loop
**Describe the bug**

The MCP OAuth refresh grant returns a token response without `refresh_token`, while the access token TTL is only 15 minutes. Omitting `refresh_token` on refresh is legal per RFC 6749 §6 (the client is expected to keep the old one), but combined with the very short access-token TTL it makes the platform MCP endpoint unusable with MCP clients that persist the provider's token response verbatim: the stored refresh token is overwritten with a response that has none, and the user is then prompted to re-authorize roughly every 15 minutes, forever.

Relevant code (`packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts`, unchanged on `main` as of `5652b02`):

```ts
const ACCESS_TOKEN_TTL_15_MINUTES_SECONDS = 15 * 60

// exchangeCode() -> { access_token, token_type, expires_in, refresh_token }

async refreshAccessToken(params: RefreshParams): Promise<TokenResponse> {
    ...
    return {
        access_token: accessToken,
        token_type: 'Bearer',
        expires_in: ACCESS_TOKEN_TTL_15_MINUTES_SECONDS,
        // no refresh_token
    }
}
```

**To Reproduce**

1. Self-host Activepieces CE `0.86.3` and expose `https://<host>/mcp/platform`.
2. Connect it from an MCP client that performs Dynamic Client Registration + OAuth and persists the token response it receives (reproduced with the OpenAI Codex app, whose `rmcp` credential-persistence path writes the provider response as-is).
3. Use it normally. Within ~15–30 minutes the client reports that authentication expired and asks to reconnect.
4. Inspect the client's stored credential — the persisted token response is exactly `{ access_token, token_type, expires_in: 900 }`, i.e. the shape returned by `refreshAccessToken()`, with the refresh token gone.

**Expected behavior**

A client that completes the authorization-code exchange once should be able to keep refreshing for the lifetime of the refresh token (30 days here) without any further user interaction.

**Suggested fix**

Echo the presented refresh token back in the refresh response, so the value is re-asserted on every refresh and clients cannot lose it:

```ts
return {
    access_token: accessToken,
    token_type: 'Bearer',
    expires_in: ACCESS_TOKEN_TTL_15_MINUTES_SECONDS,
    refresh_token: params.refreshToken,
}
```

`TokenResponse` already declares `refresh_token?: string`, so this is a one-line change. (Rotating and persisting a fresh refresh token would work too, and would additionally give replay detection.)

Separately, it would help to make `ACCESS_TOKEN_TTL_15_MINUTES_SECONDS` configurable — 15 minutes is aggressive for a long-lived MCP session and multiplies the impact of any refresh-path issue.

**Additional context**

- Activepieces `0.86.3` CE, self-hosted, PostgreSQL 18 + Redis 8, behind Caddy.
- `/.well-known/oauth-authorization-server` advertises `refresh_token` in `grant_types_supported`, so clients reasonably expect refresh to keep working.
- I'm filing this as an issue rather than a PR because CONTRIBUTING currently pauses unsolicited external PRs. Happy to provide any further detail.

majewskibartosz:
Fixes [GIT-1663](https://linear.app/activepieces/issue/GIT-1663)
Closes #14411

## Problem
MCP OAuth refresh (`POST /token`, `grant_type=refresh_token`) returned `{ access_token, token_type, expires_in }` with no `refresh_token`, while the access token lives only 15 minutes. Clients that persist the token response verbatim (reported with OpenAI Codex) overwrote their stored refresh token with one that had none, then were forced to re-authorize roughly every 15 minutes, indefinitely. The discovery metadata advertises `refresh_token` in `grant_types_supported`, so clients reasonably expect refresh to keep working.

## Fix
`mcp-oauth-token.service.ts` (`refreshAccessToken`) now echoes the presented `refresh_token` back in the response (permitted by RFC 6749 section 6), so the value is re-asserted on every refresh and clients cannot lose it. `TokenResponse.refresh_token` was already optional, so it is a one-line change.

## Verification
- New CE integration test `mcp-oauth-token.test.ts`: service refresh returns the presented token, a chained refresh using the previous response keeps working, and the real `POST /token` HTTP response body carries `refresh_token`. Red-first confirmed: against pre-fix code the assertion sees `undefined` and all three cases fail.
- lint clean on both changed files.
- Manual code-review: single line on the already-validated success path (record exists, not revoked, unexpired, client matches); echoes the client's own presented token. Full persona panel skipped as disproportionate for a one-line change.
- Security: no new exposure. The echoed value is the exact token the client presented in the same request, returned only on the validated success path; token lifetime and (non-)rotation are unchanged. No findings.
- Visual: none, backend-only, no user-visible change.

### Breaking change?

- [x] no, reviewed and not breaking
- [ ] yes, technical (removed/renamed API field or endpoint, dropped column, new required field, removed/required env var)
- [ ] yes, functional (default/limit/behaviour change, new self-hosted setup step)

### Security impact?

- [ ] no, reviewed, no security impact
- [x] yes, security-sensitive. Touches the MCP OAuth token-refresh path. Risk: leaking a refresh token. Mitigation: the response only re-asserts the exact token the client already presented, on the validated success path (unrevoked, unexpired, matching client), so nothing new is exposed and lifetime/rotation are unchanged.

### DIFF
--- a/packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
+++ b/packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
@@ -101,6 +101,7 @@ export const mcpOAuthTokenService = {
             access_token: accessToken,
             token_type: 'Bearer',
             expires_in: ACCESS_TOKEN_TTL_15_MINUTES_SECONDS,
+            refresh_token: params.refreshToken,
         }
     },
 
--- a/packages/server/api/test/integration/ce/mcp/mcp-oauth-token.test.ts
+++ b/packages/server/api/test/integration/ce/mcp/mcp-oauth-token.test.ts
@@ -1,0 +1,81 @@
+import { createHash, randomBytes } from 'crypto'
+import { apId } from '@activepieces/core-utils'
+import { FastifyInstance } from 'fastify'
+import { beforeAll, describe, expect, it } from 'vitest'
+import { mcpOAuthClientService } from '../../../../src/app/mcp/oauth/client/mcp-oauth-client.service'
+import { mcpOAuthTokenService } from '../../../../src/app/mcp/oauth/token/mcp-oauth-token.service'
+import { setupTestEnvironment } from '../../../helpers/test-setup'
+
+let app: FastifyInstance
+
+function requireRefreshToken(response: { refresh_token?: string }): string {
+    if (response.refresh_token === undefined) {
+        throw new Error('expected a refresh_token in the token response')
+    }
+    return response.refresh_token
+}
+
+async function exchangeFreshCode(clientId: string): Promise<string> {
+    const codeVerifier = randomBytes(32).toString('base64url')
+    const codeChallenge = createHash('sha256').update(codeVerifier).digest().toString('base64url')
+    const tokens = await mcpOAuthTokenService.exchangeCode({
+        codeVerifier,
+        codeChallenge,
+        codeChallengeMethod: 'S256',
+        clientId,
+        userId: apId(),
+        projectId: apId(),
+        platformId: apId(),
+        scopes: ['mcp'],
+    })
+    return requireRefreshToken(tokens)
+}
+
+describe('MCP OAuth token refresh', () => {
+    beforeAll(async () => {
+        app = await setupTestEnvironment({ fresh: true })
+    })
+
+    it('echoes the presented refresh token back so clients cannot lose it', async () => {
+        const clientId = apId()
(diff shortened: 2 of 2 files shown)


majewskibartosz:
@greptile review

KhaledR57 (approved):

PR #14451, opened 2026-07-27
fix(chat): show the connected provider's own models in the chat model picker

hazemadelkhalel:
## Description

The chat model picker offered **Fast / Expert / Heavy** no matter which AI provider the platform had connected. Those three are Activepieces-provider tiers, and their ids are OpenRouter-shaped Anthropic slugs — so a platform running on its own OpenAI key was asked for `claude-haiku-4-5` and every message failed with *"The model `claude-haiku-4-5` does not exist or you do not have access to it."*

Now:

- **Activepieces provider** — unchanged, still the three tiers.
- **Any provider that declares curated chat models** (OpenAI, Anthropic, Google) — the picker lists that provider's own models under their proper names (GPT-5.5, Claude Sonnet 4.6, Gemini 2.5 Pro…), and the chosen id is sent as-is.
- **Everything else** (OpenRouter, Azure, Bedrock, Mistral, Cloudflare Gateway) — unchanged.

### How it works

Both the server resolver and the dropdown read the list through one accessor, `aiProviderUtils.getCuratedChatModels({ provider })`, so "which models does this provider offer, and what are they called" lives in a single place next to the model data and the two cannot drift. It returns `{ id, label }` or `undefined` — never an empty list, so callers may treat a result as non-empty.

Display names are hardcoded beside the ids on purpose: the live `listModels` response can't supply one for every provider — Anthropic returns `display_name` and Google returns `displayName`, but OpenAI's `/v1/models` returns ids only. A test asserts every curated model carries a label, so adding an id without one fails CI.

A conversation that still holds a tier id keeps working: it resolves to the tier's native equivalent when the provider ships it (Anthropic `smart` → `claude-sonnet-4-6`), otherwise to the provider's first curated model — the same fallback the picker shows, pinned by a test so the two stay in agreement.

### Also in here

- `chat-sync-job` had a second copy of the resolution rule. It now calls `chatHelpers.resolveModelIdForAnalytics`, so the rule lives in one file, and analytics/billing report the model the turn actually ran on instead of the tier default.
- `conversation.modelName` is a free string on a user-facing endpoint. The resolver only ever passes it through on exact membership in the curated list, so no user-supplied string can reach a provider SDK. The provider-unresolved analytics branch used to forward it verbatim to the console payload; it now reports the stored value only when it is one of our own tier/curated ids. The API contract is unchanged — the value has always been accepted and is still stored.
- The `enabledForChat` lookup was inlined in four places; it is now `aiProviderQueries.useChatProvider`.

## How was this tested?

- Reproduced the original failure by switching the chat provider to OpenAI, then confirmed the picker lists the OpenAI models under their display names and messages send.
- Verified the Activepieces-provider path still shows Fast / Expert / Heavy and resolves to the same OpenRouter ids as before.
- **35 unit tests pass** (`vitest run packages/core/piece-types/src/lib/ai-providers.test.ts packages/server/api/test/unit/app/{ee/chat,ai}`), 17 of them new: the full resolution table (exact pick, legacy tier id with and without a provider equivalent, stale cross-provider selection, the fast round, the picker/server fallback agreement) plus the accessor invariants (never empty, labels present, ids and order preserved).
- `npx turbo run lint --filter=web --filter=api --filter=@activepieces/core-piece-types` — clean. `tsc --noEmit` on the web and server-api projects — clean.
- Edition paths: the chat surface is EE/Cloud; the provider-agnostic resolution is shared code with no edition branch, and CE is unaffected (no chat provider). No new endpoint, query, or outbound HTTP, so multi-tenant scoping and SSRF surface are untouched.

Fixes # (issue)


hazemadelkhalel:
@greptileai review

PR #14477, opened 2026-07-29
perf(engine): bound the materialized-slice cache by bytes

MrChaker:
## What changed

**Bounded slice cache.** The materialized-slice cache (`resolvedStepOutputCache`) was an unbounded `Map<fileId, Promise>` living for the whole run — any sliced output referenced once stayed in heap until the run finished. It is now a byte-budgeted LRU (`slice-cache.ts`): accounting uses the `size` already carried on each `LogSliceRef` (no measuring), `get()` refreshes recency so a loop referencing the same upstream slice every iteration stays warm, and cold slices get evicted. The budget is a fixed 64 MB — no new env var, zero setup for self-hosters. An entry larger than the entire budget bypasses the cache instead of flushing everything else.

(The single-serialization fix for `maybeSliceOutput` that previously rode this PR moved to #14473, which owns the rest of the resolution-path changes.)

## Testing

- 9 unit tests for the LRU (eviction order, multi-evict, recency refresh on `get`, oversize bypass, replace-without-double-count, exact-budget boundary).
- Full engine suite passes (the only failures are the pre-existing local `flow-with-delay` piece-dist issue, present on `main`).


packages/server/engine/src/lib/handler/context/flow-execution-context.ts
```diff
@@ -18,7 +20,7 @@ export class FlowExecutorContext {
     stepNameToTest?: boolean
     stepsCount: number
     engineApi?: EngineApiConfig
-    resolvedStepOutputCache: Map<string, Promise<unknown>>
+    resolvedStepOutputCache: SliceCache
```
MrChaker, line 24: use lru cache to prevent accumilation

abuaboud (approved):

PR #14489, opened 2026-07-29
fix(flows): router condition fields update right away when the operator changes

### LINKED ISSUE #13900: [BUG] [Flows] Router "Second value" field stays visible after selecting Exists / Does not exist until refresh

majewskibartosz:
## Description

`BranchSingleCondition` read its condition with a non-subscribing `form.getValues(...)` snapshot, so the three flags derived from it never recomputed when the operator changed:

- `isSingleValueCondition`, which gates the "Second value" field
- `isTextCondition`, which gates the "Case sensitive" toggle
- `isInvalid`, which gates the ⚠️ icon

Switching a condition to **Exists** or **Does not exist** should hide the first two immediately. Nothing in the component made that happen; the UI only corrected itself when something unrelated re-rendered it, which today is the step settings autosave resolver. Correctness resting on an incidental re-render is the actual defect, and it breaks the next time someone touches that resolver.

Fix: subscribe to the condition with `form.watch(...)` so the component re-renders when the operator changes and the derived flags recompute.

Follow-up in #14507 narrows that subscription to `useWatch({ control, name })`, which re-renders only this condition row instead of the whole step settings form, and splits the `packages/web/AGENTS.md` forms rule by tree position so the next fix of this shape picks the narrower option.

Saved and published flows were never affected. `secondValue` is already cleared when the operator changes to a single-value condition.

Fixes [GIT-1585](https://linear.app/activepieces/issue/GIT-1585) (Pylon 5162). Closes #13900.

## How was this tested?

- Regression test that renders the real component with a real react-hook-form. The leaf UI components are stubbed; the stubbed `SearchableSelect` captures the component's real `onChange` and invokes it, exactly as picking an option would.
- Red-first: two of the three tests fail against `main`'s `getValues` and pass with the subscription.
- Full `packages/web` suite: 407 passed. `npx turbo run lint --filter=web` clean.
- Test cleanups (dropping five unnecessary mocks, and moving the `secondValue` assertion onto the direction that actually exercises the clearing branch) are in #14507, not this PR.

### A note on reproducing the reported symptom

GIT-1585 and #13900 describe the extra fields staying visible **until a page refresh**, with a multi-condition (OR / And) group given as the reliable repro. That part did not hold up when checked against a real builder running the unfixed component (confirmed by fetching the served module and grepping for `getValues`, so it was not accidentally the fixed build). Four paths were tried, and the fields hid correctly every time:

1. single condition, empty values, Exactly matches → Exists
2. `+ Or` second group, empty values, Contains → Exists
3. `+ Or` second group, both values filled, Contains → Exists
4. `+ And` second condition in the same group, Contains → Does not exist

The issue's explanation for why OR would be uniquely exposed ("the autosave re-render does not reach the multi-condition path") is not how this works: the re-render lands on the `useForm` owner, and `useFormContext` consumers re-render whenever `FormProvider`'s context value identity changes, which is not path dependent and is not blocked by the `React.memo` boundaries in between.

So this lands as a latent stale-read fix rather than a confirmed cure for the refresh symptom. The symptom may have been real when it was reported and since been masked by a change to the autosave resolver, or the customer may have hit a path not found here; a slow connection would make the async resolver re-render land late, which would look like this. Worth asking on Pylon #5162 for their exact steps before treating the customer report as closed.


majewskibartosz:
## Visual verification

Real running builder (CE), Router branch condition editor, fixed build.

| Two-value operator (Exactly matches) | Single-value operator (Exists) |
|---|---|
| two-value shows Second value | Exists hides Second value |
| "Second value" + "Case sensitive" shown | both hidden immediately, no refresh |

Fresh mount at `Exists` in an OR group also renders correctly (field hidden):

OR group Exists correct

**Driven states:** two-value → single-value operator switch (field shows/hides), fresh-load `Exists` (hidden).

The multi-condition OR *live* before/after (the reliable repro) could not be self-captured via browser automation — the nested searchable-select would not commit the operator programmatically. That path is documented by the customer screenshot on [GIT-1585](https://linear.app/activepieces/issue/GIT-1585) and proven by the red-first regression test in this PR.

hazemadelkhalel:
Reviewed this and the diagnosis holds up. I pushed one commit with follow-ups rather than leaving them as comments, since they are all small. Revert anything you disagree with.

First, confirming the root cause, because I wanted to be sure a one-line change could really reach this component. I probed it with a throwaway test: in a descendant reading through `useFormContext`, `form.getValues(path)` produces no re-render on `setValue`, and `form.watch(path)` produces one. So the read was the bug, exactly as you wrote it.

The part that surprised me is that it works at all. There are three `React.memo` boundaries between the `useForm` owner in `step-settings/index.tsx` and this component: `RouterSettings`, `BranchSettings` (stable `readonly` and `branchIndex` props) and `BranchConditionGroup`. `form.watch` re-renders the owner, so I expected `BranchSettings` to bail out and swallow it. It does not, because `useFormContext` consumers re-render whenever `FormProvider`'s context value identity changes and memo cannot block context propagation. Worth knowing if anyone touches the memo chain later.

That same mechanism is why I switched the line to `useWatch({ control: form.control, name })`. Render counts from the probe: with `form.watch` in a descendant, a `setValue` re-rendered the form root and the leaf; with `useWatch`, the root stayed put and only the leaf re-rendered. Since the watched path is the whole condition object, `form.watch` means every keystroke in First value or Second value re-renders the step settings root and all 37 `useFormContext` consumers under `builder/`. `useWatch` is the same visible behaviour confined to the one condition row.

You were following the documented rule, which said "use `form.watch()` for conditional rendering" with no mention of where you are in the tree, so I amended it in `packages/web/AGENTS.md`: `form.watch` in the component that calls `useForm`, `useWatch` in anything below it. Otherwise the next fix of this shape lands in the same place.

Two things on the test. I removed five of the eight `vi.mock` calls: `alert-icon`, `tooltip`, `switch`, `button` and `lucide-react` all render fine in jsdom, and the three tests still pass without them. The `InvalidStepIcon` one is worth un-stubbing in particular, since its visibility is one of the three flags this bug affected. `Button` was being mocked but never rendered anyway, because the harness passes `showDelete={false}`.

The other was the `secondValue` assertion. It sat on the "switching to a single-value operator" test, where it could not fail: the component only clears `secondValue` going the other way, when `isSingleValueCondition && !singleValueConditions.includes(e)`, so nothing ran and the assertion was checking the untouched `''` default. I moved it onto the switch-back test and seeded `secondValue: 'stale'`, which makes it fail with `expected 'stale' to be ''` if the `setValue` block is removed. I checked that by disabling the block. That `//TODO: fix this` / `@ts-expect-ignore` block had no coverage before.

Verification after the changes: 3/3 in the file, 407/407 across the web suite, `turbo run lint --filter=web` clean. Red-first still holds, two of the three fail against main's `getValues`.

Nothing security relevant here. Rendering only, no data flow or new input handling, and if anything it removes a small mislead where the UI showed a Second value field whose contents were being ignored.

Two notes I did not act on, neither yours:

`jsdom` is not declared in `packages/web/package.json`. The whole jsdom test suite, all 11 files, only resolves because `packages/pieces/core/text-helper` declares `jsdom: ^24.0.0` and it hoists. Against the repo rule that a workspace declares what it imports, and if that piece ever drops jsdom every web component test dies at once. One line in `devDependencies`, separate PR.

I swept the rest of `packages/web/src` for the same `getValues`-in-render shape. `manual-webhook-test-button.tsx:20` and `mcp-tool-testing-dialog.tsx:68` both read through `useFormContext().getValues()` in a render body, but they feed keyed dialogs and nested-form `defaultValues` where a mount-time snapshot is defensible. `invite-user-dialog.tsx:204` does gate conditional UI on `form.getValues().type`, but it owns its own `useForm`, so any form state change re-renders it and the snapshot refreshes by accident. Fragile rather than broken, so no second instance of this bug to chase.

hazemadelkhalel:
Updated the title and description, because both were describing things that changed or did not hold up.

The description said "read the condition via `form.watch(...)`. One-line change." That is no longer what the diff does, so it now describes the `useWatch` version and says why. It also carried the issue's claim that the OR / multi-condition group is the reliable repro, and I could not confirm that.

I put main's unfixed component back into a running dev builder and checked what the server was actually serving first, by fetching the transformed module and grepping for `getValues`, so I know I was not testing the fixed build by accident. Then I built Trigger -> Router -> Branch 1 and tried four paths. The fields hid correctly in all four:

1. single condition, empty values, Exactly matches -> Exists
2. `+ Or` second group, empty values, Contains -> Exists
3. `+ Or` second group, First value `aaa` and Second value `bbb`, Contains -> Exists
4. `+ And` second condition in the same group, Contains -> Does not exist

Path 2 is the one #13900 calls out as reliable.

The reasoning in the issue for why OR would be uniquely exposed does not work either. It says the autosave re-render "does not reach the multi-condition path", but that re-render lands on the `useForm` owner, and `useFormContext` consumers re-render whenever `FormProvider`'s context value identity changes. That is not path dependent, and `React.memo` cannot block it, which I checked separately with a memo boundary between the form root and the leaf.

None of that argues against the fix. A non-reactive `getValues` driving conditional rendering is wrong whether or not something else currently happens to re-render the component, and having this UI depend on an autosave side effect is precisely what breaks the next time somebody edits that resolver. The change makes it hold on purpose, and the red-first test is unambiguous about the component behaviour.

What I have retitled away from is the user-visible claim. The old title said the fields "no longer stay visible", which asserts the refresh symptom existed as described; the new one says they update right away when the operator changes, which is true either way and still reads as a user-facing changelog line.

Suggest asking on Pylon #5162 for the customer's exact steps before treating #13900 as closed. Either the symptom was real when reported and a later change to the autosave resolver masked it, or there is a path none of us found. A slow connection is a plausible candidate, since it would make the async resolver re-render land late and look exactly like this.

hazemadelkhalel (approved):

hazemadelkhalel:
Correction: the follow-ups described here were not pushed before this merged, so the description was describing code that never landed. Fixed it to match what actually shipped (`form.watch`, original test file).

The narrower `useWatch` subscription, the `packages/web/AGENTS.md` rule split, and the two test fixes are now in #14507 on top of merged main.

PR #14517, opened 2026-07-30
docs(benchmark): re-measure the GKE scaling benchmark and correct the analysis

abuaboud:
## Description

The published benchmark numbers were stitched together from **two different clusters**. The Production Setup diagram and the changelog carried `93.5 / 148.9 req/s` (from Experiment 1, `e2-standard-4` × 14), while the Benchmark page carried `185.3 / 409.5` for **the same `4 app · 40 w` and `8 app · 80 w` rows** (from a later `n2-standard-16` × 10 run). Same ratio, same flow, different hardware — the two disagreed by 2–3×, and the later scaling run was never recorded in `EXPERIMENTS.md`.

I re-measured all four tiers on one rig, with images built from `main` (`805cc53cf7`, v0.86.3), and torn down afterwards.

### Measured

`n2-standard-16` × 10, `europe-west1-b`, warm, 1:10, 160,000 requests total — **all 200**.

| Apps · Workers | Warm req/s | per worker | Postgres CPU |
|---|---|---|---|
| 4 · 40 | 213.0 | 5.3 | 529m / 3000m |
| 8 · 80 | 484.4 | **6.1** | 1096m / 3000m |
| 12 · 120 | 641.0 | 5.3 | 1689m / 3000m |
| 16 · 160 | 777.0 | 4.9 | **2738m / 3000m** |

### Two claims were wrong — rewritten, not renumbered

- **"~4.5 req/s per worker, held flat from 40 to 160 workers — throughput scales linearly."** The per-worker rate peaks at 80 workers and then falls ~20%. This was already contradicted by the page's own table (5.1 → 4.3); the rerun just makes it obvious.
- **"Postgres never crosses ~0.65 of a core […] flat as the fleet quadruples […] no tier approaches saturation."** Postgres climbs 529m → 2738m, ending at **91% of its 3-core reservation**, and is what bends the curve. Its cost is ~2.5 millicores per req/s, so it scales with *throughput* and no setting stops it. Workers meanwhile idle at ≤0.1 of their 0.5-core cap.

Production Setup now warns to size the database against peak throughput, and notes that a **managed** 2 vCPU Postgres with real durability will hit its ceiling *earlier* than this deliberately over-provisioned rig did.

### Load is now generated in-cluster

Driving 120+ concurrent sync webhooks from a workstation exhausts its ephemeral port range (macOS ~16k; `hey` dies with `can't assign requested address`) and **fabricates a cliff at exactly the 120-worker tier** — 252 req/s, *below* the 80-worker row — while the cluster sits healthy. External and in-cluster runs agree at 40 workers (214.7 vs 213.0) and diverge above 80.

This is the likely origin of the "120 cliff" previously credited to Postgres `max_connections=100`: it reproduces with `max_connections=2000`. I've corrected that explanation rather than restating it.

### Rig fixes (each silently corrupted a re-run)

- `run-gke.sh` minted a fresh random `AP_JWT_SECRET` per run but restarted **only** the worker. `envFrom` is read once at container start, so app pods kept the old secret, every worker socket handshake failed with `Authentication error`, and **workers never recover from it** — pods `Running`, fleet "ready", nothing consuming jobs, flow publish dying 300 s later. Now: restart app → await rollout → restart workers.
- The per-run breakdown parsed JSON, but the worker ignores `AP_LOG_PRETTY` and always uses the pretty renderer, so it reported "no timing samples". (Which means the old latency table cannot have come from this parser.)
- The breakdown scraped `--since=20m`, folding cold boot and the warmup pass into the "warm" averages. Now scoped to the measured pass.
- Manifest pointed at bucket `ap-bench-usc-b3803` / `us-central1` — **deleted** — while its own comments and the docs said `europe-west1`; `ZONE` defaulted to `us-central1-a` and `APP_CPU` to `1500m` against a documented 1 vCPU.
- `production-setup.mdx` listed the app as 1 vCPU / **2 GB** while its own sizing example and the benchmark both say 1 GB.
- The CLI section claimed its 4-worker sample run came "from the results table", which has no 4-worker row.

Full method, results and traps are recorded in `benchmark/EXPERIMENTS.md` as **Experiment 3** — this experiment previously had no entry at all.

## How was this tested?

Docs-and-tooling only; no product code changed, so no edition paths apply.

- Ran the full 4-tier sweep end to end on a real GKE cluster (created and torn down for this PR); every number in the diff comes from that run's logs.
- Tier 1 measured three times (220.8 / 217.7 / 214.7 req/s externally, 213.0 in-cluster) — reproducible to ±1.5%.
- `bash -n benchmark/run-gke.sh`; `benchmark/k8s-sandbox.yaml` parses (8 docs) with the worker env intact.
- `worker-scaling.png` re-rendered from its committed HTML source via headless Chrome and visually checked.
- Cluster, load balancer, disks and the run's HMAC key all confirmed deleted — nothing left billing.

Fixes # (issue)

https://claude.ai/code/session_01Q9tK1tQAxHJcxsvzKraCRb


abuaboud:
All three P1 findings were valid and are fixed in 1eb0f37 — thanks, the third one caught a real overclaim.

**CPU request is not a cap** — correct, and this was the important one. Postgres/Redis/app declare CPU *requests* with no limit in this manifest, so `2738m / 3000m` was request utilisation dressed up as saturation, and I'd carried that into production sizing guidance as a bottleneck diagnosis. Reworked: the table now labels the request explicitly and marks the worker as the only hard-capped tier, and the conclusion is split into what the run does show (DB CPU scales linearly with throughput at ~2.5 millicores per req/s; workers stay at ≤0.1 of a hard 0.5-core cap, so the workers are demonstrably *not* the limit) versus what it doesn't (that the database *is* the limit — that needs a hard-limited DB plus wait-event analysis, now stated as a follow-up).

**Failed load pods pass validation** — fixed. `run_load` now returns non-zero on any phase other than `Succeeded`, dumps the tail to stderr, and leaves the pod in place for triage instead of deleting the evidence and publishing partial output as a throughput number.

**No-sample handling aborts early** — fixed. Both greps are guarded with `|| true` so an empty measured window reaches awk; verified the empty-stream path prints `(no timing samples found)` and the script continues through the summary and teardown instructions.

PR #14518, opened 2026-07-30
perf(engine): use SQLite for in-run step state (SRE-206)

MrChaker:
## Summary

This adds sqlite to store flow run state instead of in-memory in the engine.

### main changes

1. each flow operation will initialize a 'runId.sqlite' file in 
   a. if it's sandboxed mode will be in os.tmpdir() which is isolated "/box"
   b. unsandboxed mode will be in AP_FLOWS_CACHE_PATH ( /cache/v12/flows/store-YYYY-MM-DD)

   in unsandboxed mode we store them in dated folder names flows/store-YYYY-MM-DD to use the date for deleting old stale files that failed to get disposed because of process crash

2. sqlite table schema 
```
CREATE TABLE steps (
      name       TEXT    NOT NULL,
      path       TEXT    NOT NULL,
      output     BLOB    NOT NULL, // StepOutpuu object which includes { input, output ... }
      size_bytes INTEGER NOT NULL,
      PRIMARY KEY (name, path)
  )
  CREATE TABLE slices (
      file_id TEXT PRIMARY KEY,
      output  BLOB NOT NULL
  )
```

3. FlowExecutorContext still holds 'steps' object but is stripped from output/input . they get stored in sqlite table

4. Resolved log slices outputs is no longer saved in heap with lru byte cache. they are cached in slices table

5. Props resolver no longer load all referenced steps at once before resolving . for each token it checks for referenced steps and get it's output from sqlite

6. Log backup doesn't have to load all state in memory to parse it and upload it if its size is more than `LOG_BUFFER_UPLOAD_THRESHOLD_BYTES` threshhold . it streams the state json step by step. if it's less than the threshhold it gets compressed same as old way and uploaded directly to the s3 signed url

7. Resuming flow is changed to put each step back to sqlite instead of in-memory

### verifications

1. multiple integration tests added
2. run live local server and heavily test many types of flows  

### Change example
https://linear.app/activepieces/issue/SRE-209/engine-sqlite-state-store-example-run-walkthrough-webhook-loop-3-math


packages/server/engine/src/lib/core/code/shared-script-session.ts
```diff
@@ -0,0 +1,28 @@
+import { isNil, tryCatch } from '@activepieces/core-utils'
```
MrChaker, line 1: This was just separated from props-resolver

KhaledR57 (approved):

PR #14519, opened 2026-07-31
feat(google-docs): add agent atomics for AI agents

kishanprmr:
## Description

Adds `audience: 'ai'` atomic actions to the Google Docs piece as part of the AI-ready effort, and curates `outputSchema` for the full action surface so the data selector, output viewer, and AI/MCP path map show typed, labelled fields instead of raw JSON.

**37 new `audience: 'ai'` actions** covering the agent-relevant Google Docs surface, each mapping 1:1 to a real Google Docs / Drive API call:

- **Reads / projections:** `get_document_plaintext`, `get_document_end_index`, `search_documents`, `copy_document`
- **Text & structure:** `insert_text`, `delete_content_range`, create/delete `header`/`footer`/`footnote`/`named_range`, create/delete `paragraph_bullets`, `insert_page_break`, `insert_inline_image`, `replace_image`, `replace_all_text`, `update_document_style`
- **Tables:** `insert_table`, insert/delete table `row`/`column`, `insert_text_in_table_cell`, `insert_image_in_table_cell`, `unmerge_table_cells`, `update_table_row_style`, `create_and_populate_table`
- **Markdown:** `create_document_from_markdown`, `replace_body_with_markdown`, `replace_section_with_markdown`
- **Export:** `export_as_pdf` (Drive)
- **Escape hatch:** `batch_update` (raw `batchUpdate`)

Index-based editors reference `get_document_end_index` / `read_document` in their descriptions for the character indices they need.

`append_text` is demoted to `audience: 'human'` (superseded by `insert_text`, which covers both append and indexed insert). The existing human-facing actions (create / read / find document, template) are unchanged.

**`outputSchema` added to all 37 of the above actions** (`src/lib/output-schemas.ts`), authored from each action's actual `run()` return statement and reusing shared field sets (`driveFileFields`, and direct reuse of `appendTextActionOutputSchema` / `readDocumentActionOutputSchema` where the return shape is identical) rather than duplicating. The piece's one trigger (`new_document`) already had a schema.

Version: google-docs `0.4.4` → `0.4.6`.

## How was this tested?

- **Tier-1:** `tsc -p tsconfig.lib.json` + `eslint`: 0 errors.
- **Tier-2 (atomics):** 35 of the 37 atomics were executed live against a real Google connection through the engine (the test-step path) — 35/35 succeeded, including the Markdown composites, table cell-index resolution (insert/delete row + column, cell text + image, populate-table), and the Drive PDF export. One field-mask issue (`update_table_row_style` passing an unsupported `tableHeader` field) was found and fixed during this pass. The remaining 2 — `replace_image` and `unmerge_table_cells` — are Tier-1-only: each requires a pre-existing artifact the agent surface itself doesn't create (an inline-image `objectId` / already-merged cells), so they can't be chained end-to-end in-account, but both run through the same verified `batchUpdate` path as the actions that passed.
- **Tier-2 (output schemas):** every schema was verified against real output captured live from a test document/token — including reply-derived fields (`footerId`, `headerId`, `footnoteId`, `namedRangeId`, `occurrencesChanged`, `insertInlineImage.objectId`) and the Drive `copy`/`search`/`export` shapes — then spot-checked that the curated tree renders correctly.
- Edition paths: this piece has no edition-specific branching (CE / EE / Cloud all load it identically via the piece registry); no UI/branding surface is touched.

Fixes # (issue)

### DIFF
--- a/packages/pieces/community/google-docs/package.json
+++ b/packages/pieces/community/google-docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-google-docs",
-  "version": "0.4.5",
+  "version": "0.4.6",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
--- a/packages/pieces/community/google-docs/src/index.ts
+++ b/packages/pieces/community/google-docs/src/index.ts
@@ -7,6 +7,43 @@ import { createDocumentBasedOnTemplate } from './lib/actions/create-document-bas
 import { readDocument } from './lib/actions/read-document.action';
 import { appendText } from './lib/actions/append-text';
 import { findDocumentAction } from './lib/actions/find-document';
+import { getDocumentPlaintext } from './lib/actions/get-document-plaintext.action';
+import { replaceAllText } from './lib/actions/replace-all-text.action';
+import { searchDocuments } from './lib/actions/search-documents.action';
+import { copyDocument } from './lib/actions/copy-document.action';
+import { getDocument } from './lib/actions/get-document.action';
+import { createTextDocument } from './lib/actions/create-text-document.action';
+import { getDocumentEndIndex } from './lib/actions/get-document-end-index.action';
+import { insertText } from './lib/actions/insert-text.action';
+import { deleteContentRange } from './lib/actions/delete-content-range.action';
+import { createDocumentFromMarkdown } from './lib/actions/create-document-from-markdown.action';
+import { createFooter } from './lib/actions/create-footer.action';
+import { createFootnote } from './lib/actions/create-footnote.action';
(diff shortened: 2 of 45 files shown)


kishanprmr:
Fixed all three in 3bef58f63e:

- **search_documents**: now escapes backslashes before quotes in the Drive query literal.
- **create_and_populate_table**: the cell-population write now pins `writeControl.requiredRevisionId` to the revision read right after the table insert, so a concurrent edit fails the call loudly instead of silently misplacing text.
- **replace_section_with_markdown**: added the same table/table-of-contents guard `replace_body_with_markdown` already has, checked against range overlap instead of whole-body presence.

sanket-a11y (approved):

PR #14527, opened 2026-07-31
fix(ai-providers): pin the Azure resource name to a single dns label

kah-ja:
## Description

`azureProvider.listModels` built its endpoint as `https://${config.resourceName}.openai.azure.com` and handed it to `httpClient.sendRequest` (`packages/server/api/src/app/ai/providers/azure-provider.ts:12-26`). `resourceName` was a free-form string in `AzureProviderConfig`, so a value carrying `/`, `#`, `?`, `:` or `@` moved the request off the Azure host, and the deployment ids in whatever answered came back as the provider's model list on `GET /v1/ai-providers/azure/models`.

Every other strategy in that folder points at a hardcoded host, so Azure is the only one whose endpoint depends on stored config. The repo already constrains user strings that reach a URL context this way: `SAFE_REMOTE_URL_PATTERN` and `SAFE_BRANCH_PATTERN` in `packages/core/shared/src/lib/ee/git-repo/index.ts:6-7`.

`AZURE_RESOURCE_NAME_PATTERN` (`packages/core/piece-types/src/lib/ai-providers.ts`) pins the field to a single DNS label: alphanumerics with inner hyphens, up to 64 characters, which is what Azure accepts for the resource itself. It is applied in both `AzureProviderConfig` copies (`core-piece-types` and `shared`, kept in sync as they are today) and again in the provider before the URL is built, so a config stored before this change cannot reach the request either. The web form validates through `zodResolver(CreateAIProviderRequest)`, so it picks the constraint up without a change.

Deliberately left alone:

- The `AIProviderConfig` union still ends in empty-object members, so `POST /v1/ai-providers/:id` with a rejected resource name falls through to `{}` and strips the config instead of answering 400. Nothing is sent in that state (the provider throws), but 400 is the better answer. Fixing it means restructuring that union, which reaches well past this change.
- `httpClient` stays instead of `safeHttp.axios`. Once the label is validated the host suffix is fixed to `openai.azure.com`, so the URL is no longer user-controlled. Say the word if you want `.claude/rules/safe-http.md` applied literally here as well.

Behaviour change: an Azure provider whose stored `resourceName` is not a valid label now fails with a clear message instead of issuing a request to wherever the value pointed. Valid Azure resource names are unaffected, which is why the breaking-change box below says no.

### References

GHSA-75q7-562h-f8h4 (reported privately, unpublished at time of writing)

## How was this tested?

CE code paths only; no EE or Cloud specific code is touched.

- `npx turbo run build --filter=api --filter=web`: green.
- `npx turbo run lint --filter=@activepieces/shared --filter=api --filter=@activepieces/core-utils --filter=@activepieces/core-piece-types`: 0 errors.
- `npx turbo run test --filter=@activepieces/shared --filter=@activepieces/core-utils`: 458 passed.
- `npm run test-unit` in `packages/server/api`: the six `azure-provider.test.ts` cases pass. Five test files plus `workerGroupService > getWorkerGroupId` fail with `getaddrinfo EAI_AGAIN redis` in my environment; the same set fails identically on unpatched `main`, so they are unrelated to this change.
- Manual, real HTTP path (no mock): called `azureProvider.listModels` with `resourceName: '127.0.0.1:8443/pwned#'` against a local TLS listener. Before the change the listener logged `GET /pwned` with the `api-key` header and its response body came back as the model list; after the change no request is made and the call throws. A normal name still resolves to `my-resource.openai.azure.com`, covered by a new unit test.
- Not run here: the ce/ee/cloud integration suites and e2e, which need postgres and redis.

Fixes # (issue): n/a

### DIFF
--- a/packages/core/piece-types/src/lib/ai-providers.ts
+++ b/packages/core/piece-types/src/lib/ai-providers.ts
@@ -1,4 +1,4 @@
-import { AIProviderName, isNil } from '@activepieces/core-utils'
+import { AIProviderName, formErrors, isNil } from '@activepieces/core-utils'
 import * as z from 'zod/mini'
 
 export enum AIProviderModelType {
@@ -61,8 +61,12 @@ export const CloudflareGatewayProviderConfig = z.object({
 })
 export type CloudflareGatewayProviderConfig = z.infer<typeof CloudflareGatewayProviderConfig>
 
+// The resource name is interpolated into `https://<resourceName>.openai.azure.com`, so it has to
+// stay a single DNS label. Azure allows alphanumerics and inner hyphens, up to 64 characters.
+export const AZURE_RESOURCE_NAME_PATTERN = /^[A-Za-z0-9](?:[A-Za-z0-9-]{0,62}[A-Za-z0-9])?$/
+
 export const AzureProviderConfig = z.object({
-    resourceName: z.string(),
+    resourceName: z.string().check(z.regex(AZURE_RESOURCE_NAME_PATTERN, formErrors.invalidAzureResourceName)),
     apiVersion: z.pipe(
         z.transform((v) => (typeof v === 'string' && v.trim().length === 0 ? undefined : v)),
         z.optional(z.string()),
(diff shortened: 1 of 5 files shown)


kah-ja:
Any updates?

PR #14541, opened 2026-08-01
ci(release-pieces): fail when a published piece cannot be served on the current release

ibrahim-abuznaid:
Plan: [PIE-372](https://linear.app/activepieces/issue/PIE-372/release-pieces-silently-publishes-pieces-that-can-never-be-served) — goal, happy path, unhappy scenarios, trade-offs and breaking-change assessment.

## Description

A piece whose `minimumSupportedRelease` is above the deployed release is stored in the catalog but filtered out of every read path by `isSupportedRelease` (`piece-cache-utils.ts:88`, applied by both `fetchLatestPieces` and `findExactVersion`). The catalog silently keeps serving the previous version, and `Release Pieces` reports success.

**19 pieces are in this state today.** They declare `minimumSupportedRelease` `0.86.4` or `0.87.0`, while the latest release — and `main` itself — is `0.86.3`. Neither required release exists, so these versions can never be served. The correlation is exact across the 25 pieces in the current backlog: 19/19 with a minimum above `0.86.3` are stale on cloud, 6/6 at or below are current.

Worked example — serp-api `0.1.6` (min `0.86.4`) added 17 `audience: 'ai'` actions and flipped the 4 original actions to `audience: 'human'`. Cloud serves its predecessor `0.1.5` (min `0.82.0`), whose 4 actions are still `audience: 'both'`. Discovery therefore returns the 4 legacy actions and none of the 17 new ones — which presents as an audience-filter bug but is entirely the release gate.

Secondary effect: `pieceMetadataExists()` queries the same release-gated endpoint, so these pieces read as "not inserted" on every run and `find-changed-pieces` re-adds them indefinitely — a rebuild/republish loop that cannot converge and grows the changed set toward the known CI OOM ceiling.

### How it works

`update-pieces-metadata` partitions the changed set on `minimumSupportedRelease` against the root `package.json` version. Servable pieces are inserted exactly as before, then the step fails, naming each piece, its version, and the release it requires. Ordering matters: one forward-looking piece does not block the other 25 in the same batch.

Pieces with an absent or non-semver `minimumSupportedRelease` are treated as servable, matching `isSupportedRelease`.

Three silent failures on the same path are also surfaced:

| Scenario | Before | After |
|---|---|---|
| Piece requires a release that does not exist | published, inserted, never served, step green | servable pieces inserted, then step fails naming each one |
| A metadata insert fails | `Promise.all` rejects on the first failure, hiding the rest | all failures aggregated with counts |
| A changed piece has no build output | `console.info`, silently dropped | hard error listing every affected path |
| A piece fails to load from dist | `catch` returns `null`, no output | logged with `name@version` and path |
| Admin POST returns non-OK | error body only | status code and body |

### Not included

PR-time enforcement. `validate-publishable-packages` already runs on PRs touching `packages/pieces/**` and would have blocked the original PRs before any npm version was burned, but `minimumSupportedRelease` lives in `src/index.ts` and that job does not build — reading it there would mean regex-parsing source. Left as a follow-up rather than shipping a fragile parser.

Batch concurrency is untouched, so the publish burst that strands rows in `tool_search_index` (#14538) is unaffected and still needs that fix.

## How was this tested?

- `tsc` over both changed files using CI's compiler options (`packages/server/engine/tsconfig.lib.json` settings via `tsconfig.base.json`): 0 errors.
- The release predicate was checked against the server's `isSupportedRelease` across `0.86.4`, `0.87.0`, `0.82.0`, `0.36.1`, a value equal to the current release, an absent value, and a non-semver value — agreement on all 7 cases.
- Verified against live data: cloud, staging and canary all serve serp-api `0.1.5` while npm has `0.1.6`; the 6 backlog pieces with a minimum at or below `0.86.3` are current on cloud.
- `eslint` does not cover `tools/` (matching ignore pattern), so it is not a gate for these files.
- CI-only tooling; no edition-specific paths (CE / EE / Cloud) are affected.

Fixes # (no GitHub issue — tracked in PIE-372)

### DIFF
--- a/tools/scripts/pieces/update-pieces-metadata.ts
+++ b/tools/scripts/pieces/update-pieces-metadata.ts
@@ -31,14 +31,22 @@ const insertPieceMetadata = async (
   }
 };
 
-const isServableOnCurrentRelease = ({ pieceMetadata, currentRelease }: { pieceMetadata: PieceMetadata, currentRelease: string }): boolean => {
-  const minimumSupportedRelease = pieceMetadata.minimumSupportedRelease
+const findUnsupportedReason = ({ pieceMetadata, currentRelease }: { pieceMetadata: PieceMetadata, currentRelease: string }): string | null => {
+  const { minimumSupportedRelease, maximumSupportedRelease } = pieceMetadata
 
-  if (!minimumSupportedRelease || !semver.valid(minimumSupportedRelease)) {
-    return true
+  if (!semver.valid(currentRelease)) {
+    return null
   }
 
-  return !semver.gt(minimumSupportedRelease, currentRelease)
+  if (maximumSupportedRelease && semver.valid(maximumSupportedRelease) && semver.gt(currentRelease, maximumSupportedRelease)) {
+    return `maximumSupportedRelease ${maximumSupportedRelease} is below it`
+  }
+
+  if (minimumSupportedRelease && semver.valid(minimumSupportedRelease) && semver.gt(minimumSupportedRelease, currentRelease)) {
+    return `minimumSupportedRelease ${minimumSupportedRelease} is above it`
+  }
+
+  return null
 };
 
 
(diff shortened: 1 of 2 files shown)


ibrahim-abuznaid:
Good catch on the maximum bound — fixed in 1b928aa427.

The predicate only mirrored the minimum half of `isSupportedRelease`, so a piece declaring a `maximumSupportedRelease` below the current release would have been inserted and then filtered from every read path: the same silent skip, other bound. Both bounds are now evaluated by a single function that returns *why* a piece is incompatible, and that same value feeds the failure message, so the check and the report can't drift.

Verified against the server's `isSupportedRelease` across 13 cases — both bounds violated, each individually, equal-to-current on each bound, absent, and non-semver — agreement on all of them, including the two you flagged (`max` below current, and valid `min` with `max` below).

ibrahim-abuznaid:
Tested the script end-to-end locally against a mock catalog API, driving it exactly as CI does (`ts-node -r tsconfig-paths/register -P packages/server/engine/tsconfig.lib.json`) with fabricated `dist` fixtures fed through `CHANGED_PIECES`. Root `package.json` was `0.86.3`, matching production.

| Scenario | Exit | Inserted | Reported |
|---|---|---|---|
| all compatible | 0 | both pieces | `completed` |
| compatible + min-above + max-below | 1 | compatible only | `minimumSupportedRelease 0.86.4 is above it` / `maximumSupportedRelease 0.86.2 is below it` |
| changed piece with no build output | 1 | nothing | fails before any insert |
| piece fails to load | 0 | compatible only | `failed to load @fixture/piece-broken@1.0.0 … it will not be inserted` |
| every insert returns 500 | 1 | — | `2 of 2 piece(s) failed to persist`, both listed |

Same fixtures against the pre-PR code, for contrast: **exit 0 with all three pieces inserted**, including both incompatible ones, ending in `update pieces metadata: completed` — the silent success this PR removes. The no-build-output case likewise exited 0 with only `skipping, no build output` in the log.

The mixed scenario also confirms the ordering trade-off works as intended: the compatible piece is inserted before the failure is raised, so one incompatible piece does not block the rest of the batch.

Testing found one bug in the original patch, fixed in 0c4a0da609: `loadPieceFromFolder` already catches everything and returns `null`, so the `try/catch` I had wrapped around it was unreachable and a load failure was still dropped unnamed. It now checks the return value, which is what produces the log line in row 4 above.

hazemadelkhalel (approved):

PR #14544, opened 2026-08-01
fix(sandbox): raise the engine's file-descriptor ceiling from 64

amrdb:
## Description

`isolate` caps `RLIMIT_NOFILE` at **64 — soft and hard** — for the process it wraps, which under `AP_EXECUTION_MODE=SANDBOX_PROCESS` is the engine. That 64 is the entire budget for every concurrent HTTP socket to every piece, plus 4 fds per CODE-step child. Measured on a production worker, an idle sandbox already holds 19–30 of them.

When a flow's step opens a few dozen sockets, `spawn` fails `EMFILE` mid-run and the user gets a nonsense type error on whichever CODE step was unlucky. #14543 stops that error being swallowed; this PR removes the ceiling.

**The worker's own limit (524288) and the host's `fs.file-max` are irrelevant** — they belong to different processes. Only the `sandbox-*` process is constrained, which is what made this hard to find.

Two commits:

1. **`--open-files=1024`** on the isolate invocation. Not unlimited: the sandbox resource bound is deliberate, so a runaway flow still can't exhaust the host. 1024 is 16× the current cap and ~30× the real headroom, matches the conventional Linux soft default, and caps a dense 28-container host at ~29k fds against an effectively unbounded `file-max`.

2. **isolate 1.8.1 → 1.10.1 binaries.** The shipped binary was built 2021-11-13, two months *before* upstream commit `881ce90` ("Make the limit of open file descriptors configurable", first released in v1.9). 1.8.1 has no flag and rejects `--open-files` outright, so the binary and the flag must land together or every sandbox fails to start.

**Why 1.10.1 and not 2.x:** 2.x requires cgroup v2 with manual delegation when there's no systemd (there isn't, in our containers), adds a seccomp filter untested against arbitrary npm packages, and — decisively — **does not compile on Debian bullseye at all** (needs `libseccomp-dev` and Linux ≥5.14 headers). Adopting it would mean migrating the runtime base image first. 1.10.1 is the last 1.x, keeps cgroup v1, and takes the existing arg set unmodified.

Verified against the current setup: the shipped `default.cf` is **byte-identical** to 1.10.1's own default config (no migration), and 1.10.1 accepts the exact production arg list including `--processes` without `--cg`.

## How was this tested?

Binaries are built by the new `.github/workflows/build-isolate.yml` — not on anyone's laptop — from pinned commit `2efddd8b` inside `node:24.14.0-bullseye-slim`, so libc linkage matches the runtime image by construction. (`v1.10.1` is a lightweight tag and cannot be signature-verified, hence the SHA pin.)

The workflow asserts, per arch, before publishing: ELF machine, version string, that the binary reports the pinned source commit, that `--open-files` is accepted, that the glibc floor doesn't exceed the runtime image, and the expected sha256.

- **Reproducible**: pinning `BUILD_DATE`/`BUILD_COMMIT` makes output byte-identical. The CI amd64 hash matched an independent local build exactly. Both hashes are now asserted, so a drifting build fails in CI rather than shipping.
- **glibc floor** `GLIBC_2.14` (amd64) and `GLIBC_2.17` (arm64, the aarch64 baseline) — both unchanged from the 1.8.1 binaries they replace, and well under bullseye’s 2.31.
- **Functional**: production arg set + `--open-files=1024` yields `ulimit -Sn/-Hn = 1024/1024`, exit OK. Confirmed on the 1.8.1 binary that `--open-files` is rejected.
- `turbo run lint --filter=@activepieces/sandbox`: 0 errors.
- **Not yet exercised on a live worker** — needs a staging deploy to confirm sandboxes still start and code steps run under the new binary. Worth doing before merge.

Trigger is `workflow_dispatch` so the toolchain build only runs on demand.

Fixes # (issue)

**Risk:** replaces the sandbox binary and loosens one of its resource limits. **Mitigation:** the limit stays bounded (1024, not unlimited), 1.10.1 is a strict upstream advance over 1.8.1 on the same major line, the source is pinned to a commit SHA, and the build is reproducible with both hashes asserted in CI. Note 1.10.1 predates the 2.1.1 hardening flags and the 2.5 fixes — those 2.5 issues are privilege-boundary escapes for *the user invoking isolate*, which here is root in a privileged container, so they buy us little.


abuaboud (approved):

amrdb:
> The PR should not merge until the replacement binaries are functionally verified to initialize and run a sandbox with the intended descriptor limits.

Fair, and now addressed. The previous check only proved the flag *parsed*; a binary could accept it and still fail to start a sandbox or ignore the limit at runtime.

`.github/workflows/build-isolate.yml` now initialises a **real sandbox per arch**, using the same `default.cf` we ship, and asserts **differentially** — because "applies the limit" is only meaningful against the value it replaces:

```
no flag              → must be 64
--open-files=1024    → must be 1024
```

Either mismatch fails the build. Run [30755123567](https://github.com/activepieces/activepieces/actions/runs/30755123567):

```
amd64   isolate default: 64   with --open-files=1024: 1024
arm64   isolate default: 64   with --open-files=1024: 1024
```

This runs natively on `ubuntu-latest` and `ubuntu-24.04-arm`, so **arm64 has functional coverage for the first time** — previously it was only `--version` and flag-parsing.

Separately, the production code path was exercised locally on amd64: driving the real `isolateProcess(...).create()` with real argv, real mounts and this exact binary (sha `1c5b139c…`, byte-identical to the CI artifact) gives the sandboxed process `Max open files 1024 1024`. Reverting just the flag gives `64 / 64` — reproducing the original bug — and keeping the flag while restoring the old 1.8.1 binary fails with `unrecognized option '--open-files=1024'`, which is why the binary and the flag must ship in the same commit.

Still not covered, for the record: no engine boots and no flow runs inside the sandbox in any of this. The WS handshake, piece mounts and code cache are untouched by the change, but that's reasoning rather than evidence — a staging deploy is still the right final gate.

amrdb:
## End-to-end verification on both architectures

The remaining gap — *"no engine boots and no flow runs inside the sandbox"* — is now closed.

`smoke-test.yml` only ever ran `AP_EXECUTION_MODE: SANDBOX_CODE_ONLY`, which routes to `simpleProcess`. **No test in this repo had ever exercised the isolate binary, on any architecture.** That is how a 64-fd cap survived three and a half years.

`benchmark/docker-compose.yml` already parameterises `AP_EXECUTION_MODE` and the worker is already `privileged`, so covering it was a matrix entry rather than new infrastructure. Run [30761740564](https://github.com/activepieces/activepieces/actions/runs/30761740564), all four green:

| arch | mode | |
|---|---|---|
| amd64 | SANDBOX_CODE_ONLY | ✅ 11m56s |
| arm64 | SANDBOX_CODE_ONLY | ✅ 14m05s |
| amd64 | SANDBOX_PROCESS | ✅ 7m37s |
| arm64 | SANDBOX_PROCESS | ✅ 8m29s |

arm64 through isolate 1.10.1 — real image, real stack, real flow:

```
PASS: /api/v1/flags returned 200
PASS [1/5]: HTTP 200, body={"hello":"world"}
...
PASS [5/5]: HTTP 200, body={"hello":"world"}
```

So the engine boots inside the new binary and executes flows on **aarch64**, which I could not verify any other way — there is no arm hardware available to me.

### Why the later steps are gated to `SANDBOX_CODE_ONLY`

The first attempt timed out at 35m on both arches. Not a hang — every request returned 200, but isolate pays `--cleanup` + `--init` + spawn per execution, so the 5000-iteration memory test runs at ~540ms/iteration and needs ~45m by itself.

It is also meaningless in this mode: it sums RSS of `sandbox-*` processes, which are transient under isolate, so it reported `0.0 MB` from iteration 2000 onward. The S3 steps already pinned `SANDBOX_CODE_ONLY` explicitly. Gating them keeps the `SANDBOX_PROCESS` jobs at ~8 minutes and focused on the binary.

### Still not covered

Neither arch runs a **CODE step** under isolate in this smoke test — the flow is webhook → reply. So `noOpCodeSandbox`'s `spawn`, the thing that actually consumed the 64 descriptors, is exercised only by the unit test in #14543 and by my local run through the real `isolateProcess`. Worth knowing before merge.

PR #14554, opened 2026-08-02
fix(worker): namespace and sweep action-run code caches in their own directory

othmanemad:
## Description

Action runs (MCP `ap_run_action`, chat `ap_execute_action` / `ap_run_code`) execute a single piece action or code step outside any flow. The compiled-code cache is keyed by `flowVersionId` + step name, which an action run doesn't have — so `resolveCodeStep` synthesised a namespace from `sha256(sourceCode)` alone. Every tenant on a worker shared one content-addressed directory under `<cache>/v12/codes/`, owned by nobody and reclaimed by nothing.

Not a confidentiality leak: mounts are read-only and a tenant only reads artifacts compiled from its own source. What's wrong is that an unowned directory can't be purged or attributed, times a cold build into a cross-tenant existence oracle, and serves one tenant's `bun install` resolution of an unpinned semver range to another. `custom_pieces/<platformId>`, built two lines away in the same function, already namespaces by platform.

Second problem: unbounded growth. Action-run snippets are agent-generated and near-unique, and `docker-compose.yml` bind-mounts `./cache`, so the tree survives restarts on every self-host.

### The layout

`<cache>/v12/codes/action-runs/<platformId>_<sha256(sourceCode)>/<stepName>/`, from `actionRunCache.namespace`. Flow-version caches stay at the root of `codes/`, unmoved.

- **`platformId`** is the customer boundary — purging a customer is one glob, and the cache stays warm across that customer's projects.
- **The `action-runs/` directory** is what separates an action-run build from a flow-version build on disk, since `platformId` and `flowVersionId` are both 21-char `apId`s and length can't tell them apart.
- **`platformId` stays flat inside it.** A per-platform level would make the sweeper walk two levels and assemble its eviction list across platforms — the one place this design insists on recomputing from a single live `readdir` — and would leave empty platform directories whose pruning races peer replicas creating them.

An earlier revision of this PR used an `ar_` **name prefix** instead, on the grounds that a directory level would mean threading `platformId` into the process maker (in fork mode `AP_BASE_CODE_DIRECTORY` is baked in before `platformId` is known) and changing the engine's read path. That objection is real but applies only to a **variable** first level like `codes/<platformId>/<hash>/`. `action-runs/` is a **constant**, so the base code directory stays `/root/codes`, both process makers are untouched, and `code-executor` absorbs the extra segment by string interpolation with no change at all.

**Why the directory is worth the churn over the prefix.** A prefix makes classification lexical, and it was collision-proof only because `ALPHABET` in `core-utils/id-generator.ts` excludes `_`: add it and any `apId` beginning `ar_` is misclassified — roughly one in 238k per id, so effectively certain at cloud scale — and the sweeper silently starts deleting flow caches, from a one-character change three packages away with no test between it and data loss. A flow version can now only be swept if it lands *inside* `codes/action-runs/`, needing a `flowVersionId` equal to the string `action-runs`: `ALPHABET` must gain `-`, **and** `ID_LENGTH` must go 21 → 11, **and** the `ApId` regex must change, all together. The sweep also stops reading the shared tree — it listed every flow-version cache on the worker (never reclaimed, only cleared on a cache-version bump) to filter down to at most 200 candidates every 30 minutes, and now reads only its own directory with no name filter at all. Disk becomes attributable by path rather than by glob.

### ⚠️ The namespace now contains a `/`, so review the guard

The nesting **cannot** be hidden from the engine. The tempting version is host-side only — keep the directory at `codes/action-runs/<ns>` but mount it back to `/root/codes/<ns>` so the namespace the engine holds stays a single segment. That works in isolate mode and **breaks in fork mode**, where `AP_BASE_CODE_DIRECTORY` is the raw host `codes/` path with no mount indirection and the engine reads `<codes>/<namespace>/<stepName>/index.js` straight off the host filesystem. So the namespace string the engine receives must literally contain `action-runs/`.

That value becomes a bind-mount `hostPath`, so the guard is the security-relevant part of this diff:

- **`assertSafePathSegment` is unchanged** — byte-identical. It still rejects `.`, `..`, any `..` substring, `/`, `\`, NUL and empty.
- **`assertSafeCodeNamespace`** splits on `/`, rejects more than two segments, and runs every segment through that untouched guard. The traversal rules are stated in exactly one place; nothing was loosened or restated.
- **`stepName` and `platformId` are still single segments** and still call `assertSafePathSegment` directly.
- Generic depth-2 was chosen over a whitelist of the two known shapes knowingly: the residual is a future caller parking a namespace one level deeper than intended — a correctness surprise, not an escape — and every segment is still individually asserted.

**One existing test changed meaning and a reviewer should look at it.** `sandbox.test.ts` asserted `a/b` must be rejected as a traversal `flowVersionId`. Two legal segments is exactly what this PR introduces, so that case was replaced with the ones that must still reject: `a/b/c`, `a//b`, `/a`, `a/`, `action-runs/..`, `action-runs/../../etc`. The `platformId` traversal block below it is untouched and still rejects `plat/sub`.

### Why there was no GC to begin with

The first version passed the constant `DEFAULT_MCP_DATA.flowVersionId`, so every action run compiled into one directory and a different snippet was always a hash miss — and `installFn` opens by `rm -rf`-ing the directory. **That destructive rebuild *was* the garbage collection**, bounded at O(1) dirs. It was also the bug: `memoryLock` serialises the build but not the later read of `index.js`, so concurrent boxes could execute each other's snippet. Switching to the content hash fixed that *precisely by making the rebuild branch unreachable*, which is how the only reclamation on this path disappeared.

### The sweep

`actionRunCache.sweep` — worker-local, first pass ~1 min after boot then every 30 minutes, on an `unref`'d timer cleared in the same shutdown path as the other worker intervals. It reads **only** `codes/action-runs/`, deletes children untouched for **2h**, then evicts oldest-first while more than **200** survivors remain, **skipping anything touched in the last 15 minutes**. Flow-version caches are never swept, and are now structurally unreachable rather than merely excluded by a filter. All knobs are hardcoded constants — no env var, no self-host setup.

**Bounded by directory count, not bytes.** A byte budget's survivor count moves with directory size, so at 0.5 GB each a 2 GB budget permits four survivors — fewer than the live set on any real topology, meaning it *must* evict a directory a running sandbox has mounted. A count cap fixes the floor at N whatever they weigh. (The premise was also wrong: `code-builder` deletes `node_modules` after compile and on failure, so an action-run dir is just the esbuild bundle. And the 2 GiB budget it shipped with was the size of the *entire* default Helm volume, so it could never fire.)

**The cap alone isn't the safety mechanism.** mtime is *provision* time and is never refreshed for the ≤120s a run lasts, so a slow run is outranked by every fast one started after it — "the newest 200" is not "the live ones". Reaching a live directory needs only 200 distinct snippets inside one execution window (~1.7/s), which is reachable at cloud scale. The 15-minute active-window skip closes it, and trades into the safe failure: when everything is active the tree stays above the cap until it ages out — disk overshoots instead of a run dying. That case logs `activeCount`, so a permanently blocked eviction is never silent.

Keep the cap above `AP_WORKER_CONCURRENCY` × replicas sharing the mount (25 on the reference topology vs. 200) — a utilization invariant, not a safety one.

**Convergent, not coordinated.** A dedicated worker (`AP_CONTAINER_TYPE=WORKER`) has no Redis, no TypeORM and no Fastify, so `distributedLock` and system jobs are both out of reach — and a system job runs in the *app* process anyway, which can't `readdir` a worker pod's disk. N sweepers over one shared mount is the steady state, so every step is idempotent (`force: true`, ENOENT-tolerant, mtime re-checked immediately before `rm`) and eviction recomputes its target from the live `readdir` instead of accumulating across deletions. Two simultaneous sweeps pick the same oldest set, delete it once, and neither evicts past the cap — no jitter needed.

### Three things the sweep required

**A cache hit now verifies `index.js` exists** ([code-builder.ts](https://github.com/activepieces/activepieces/blob/fix/action-run-code-cache-tenant-namespace/packages/server/sandbox/src/lib/cache/flow/code/code-builder.ts)). `cacheState`'s memo is module-scoped with no invalidation API and a hit returns without touching disk, so a swept directory would leave a permanent phantom hit and an engine `require` of a missing file. Process-local invalidation couldn't have fixed it — the reference compose file shares one `./cache` across `app` and five worker replicas.

The check must be evaluated **inside** `cacheMiss`, not snapshotted before `getOrSetCache`. `cacheMiss` is called twice (once on the memo, once inside `memoryLock.runExclusive`); a hoisted `await fileExists(...)` freezes the pre-lock answer, so every run queued behind the winning builder still sees "absent" and rebuilds — N concurrent runs of one cold code step produce N builds, each opening with `rm -rf codePath`, deleting `index.js` out from under engines that already provisioned. A synchronous `existsSync` in the predicate keeps the check atomic with the decision it feeds.

**`provision` touches the mtime** ([local-execution-cache.ts](https://github.com/activepieces/activepieces/blob/fix/action-run-code-cache-tenant-namespace/packages/server/sandbox/src/lib/cache/local-execution-cache.ts)). A cache hit writes nothing, so without this a directory used hourly would still be evicted at TTL. It's also the only in-use signal the TTL and the active window have: `provision` runs immediately before `sandbox.start()`, so anything bind-mounted was touched under ~130s ago. Gated on `isActionRunNamespace`, so flow-version dirs don't pay a `utimes` on every code step of every flow run.

**`provision` also settles a removal already in flight** ([action-run-cache.ts](https://github.com/activepieces/activepieces/blob/fix/action-run-code-cache-tenant-namespace/packages/server/sandbox/src/lib/cache/action-run-cache.ts)). The mtime re-check only rules out removals that start *before* the touch. One already past its re-`stat` can't see a touch that lands a microsecond later: `rm` proceeds under a sandbox that has just accepted the directory as a cache hit, and that run dies requiring a missing `index.js`. The window is one `rm` wide and needs a cold snippet re-requested at the instant its eviction starts — rare, but reachable, and the failure is a confusing user-visible error rather than anything self-evident.

`removeDir` now records its in-flight `rm` in `pendingRemovals` **in the same synchronous block that starts it**, and `installCodeStep` calls `settlePendingRemoval` after the touch: a removal registered *after* the touch is impossible, because its re-`stat` sees the new mtime and skips; one registered *before* is awaited, then the step is rebuilt. Those two cases are exhaustive only because nothing can interleave between the re-`stat` resolving and the `Map.set` — keep the write before the first `await`. Process-local, so two worker processes over one shared mount still fall back to the mtime re-check alone; accepted, the same way the sweep is convergent rather than locked.

### `isManagedDir` → `isActionRunNamespace`

Not a rename for taste — a required change. The predicate was called with **two different kinds of string**: bare `readdir` entry names in `sweep`, and full namespaces in `installCodeStep`. That only worked because `ar_` led both. Nested, they diverge — children of `action-runs/` no longer carry a prefix — so a name-based check would return false for every entry and the sweep would reclaim nothing. `sweep` now drops its filter entirely (the directory *is* the filter) and the predicate serves only `installCodeStep`, which also retires the "naming and sweeping must agree on one filter" coupling.

### Orphans

Each layout change strands earlier builds, and the sweeper deliberately **doesn't** reclaim them. Bare-`sha256`, `mcp-flow-version-id` and `ar_`-prefixed dirs only exist on machines that ran intermediate commits of this branch — **no layout on this path has ever reached `main`**, so there is nothing in the field to migrate and no reclaim path was written. The name-sniffing branch that deleted them had no TTL and no mtime re-check, and `mcp-flow-version-id` is still a live constant on `main` — so the day anything provisions under it, the sweeper would `rm -rf` it every 30 minutes. Pre-branch temporary-flow dirs are named by real `apId`s and are indistinguishable from live flow-version caches. Better to leak than to run a heuristic that could eat a flow's cache. On a dev box that ran those commits, `rm -rf cache/v12/codes` is the cleanup.

**Note for upgraders:** none needed. This path has never shipped, so no self-hoster has an action-run cache to invalidate.

Rationale in [`brain/decisions/000016`](https://github.com/activepieces/activepieces/blob/fix/action-run-code-cache-tenant-namespace/brain/decisions/000016-action-run-code-cache-is-namespaced-per-platform.md); the wiki pages for the code cache, action runs and workers are updated to match.

### Follow-up (separate PR)

`canReuseSandbox` checks the `AP_REUSE_SANDBOX` env override **first** and short-circuits, so `SANDBOX_PROCESS` + `AP_REUSE_SANDBOX=true` yields `reusable: true` → `buildCodeMount` returns a read-only mount of the **entire** `codes/` tree at `/root/codes`, in a jail where user code has Node `fs` access. `worker-groups.mdx` documents exactly that pairing. The recommended setup (`SANDBOX_CODE_ONLY` + reuse) is fork mode, which ignores `mounts`, and `EXECUTE_ACTION` isn't project-group routable — so no exposure there. That makes it flow code with a `⛓️‍💥 breaking-change` label and doc edits, sequenced separately.

## How was this tested?

**Automated**

- `packages/server/sandbox` — **220/220 pass** (16 files). `action-run-cache.test.ts` against a `tmpdir` fixture: namespace shape, two platforms with identical source producing two names, empty `platformId` throwing `ErrorCode.VALIDATION`; expiry, count-cap eviction ordering, the active-window skip, missing cache dir a no-op, idempotent re-runs, and `touch` advancing the mtime so a stale dir survives.
- **New root-survival regression** — the guard that the prefix era never had. Seeds a flow-version dir at `codes/<apId>`, a leftover `ar_`-prefixed dir, and a stray *file* directly in `codes/`, all aged 90 days, plus one expired action-run dir; asserts the first three survive. This is what makes "the sweeper cannot touch flow caches" a property of the layout rather than of a naming convention. A second test pins that `ACTION_RUN_CODE_DIR` is not `apId`-shaped, so a change to `ID_LENGTH` trips a test.
- **New `path-safety.test.ts`** — the composite guard: accepts one and two safe segments; rejects a third segment, traversal in either segment, the empty segment from a leading/trailing/doubled slash, and backslash/NUL. Plus a case pinning that `assertSafePathSegment` still rejects a slash, so `stepName` can't become a nested path.
- **New nested-mount cases** — `sandbox.test.ts` pins `hostPath: <codes>/action-runs/<ns>` → `sandboxPath: /root/codes/action-runs/<ns>`; `isolate.test.ts` pins the nested rootfs `mkdir` and the resulting `--dir=/root/codes/action-runs/<ns>=…:maybe` arg. This is the part that would otherwise fail only at runtime in isolate mode.
- **`code-cache-traversal.test.ts`** — the `flowVersionId` cases now assert namespace-level rejection (`../../common`, `..`, `action-runs/..`, `action-runs/plat/hash`), and a new case asserts a legitimate action-run namespace resolves to exactly `codes/action-runs/<ns>/<step>` and stays under `codes/`. The `stepName` cases are unchanged.
- **`packages/server/engine` — 444/444 pass**, including a new fixture at `test/resources/codes/action-runs/plat-xyz_deadbeef/echo_action_run/index.js` and a `flow-codes.test.ts` case that executes it via `generateMockEngineConstants({ flowVersionId: 'action-runs/plat-xyz_deadbeef' })`. This is a real `require` resolved through `AP_BASE_CODE_DIRECTORY` off the host filesystem — i.e. **fork-mode semantics, the constraint that drove the whole design**, proven rather than asserted.
- `code-builder.test.ts` — two regression guards retained: `rm` `index.js` behind the memo's back → **rebuilds** instead of a phantom hit; and 5 concurrent `processCodeStep` calls on one cold step → `buildMock` called **once**, not five times.
- `npm run test-unit` — 23/23 tasks. `npm run lint-dev` — 30/30 tasks, **0 errors**. `tsc --noEmit` clean on `packages/server/sandbox`.
- `worker` — 174 pass, 1 skipped, **7 pre-existing failures** in `worker-settings-override.test.ts` (all "Timed out waiting for settings/crash" — a socket handshake, nothing near the cache). Verified unrelated by reverting `packages/server/sandbox/src` to the pre-commit state and re-running: the same 7 fail identically.

**Edition paths** — no edition-gated code. Nothing under `src/app/ee/`, no feature flag, no plan check; identical on CE / EE / Cloud.

**Not done** — a live-worker pass. Two gaps, both needing a booted worker: a **real isolate mount** (the unit tests pin the `--dir=` argument and the nested `mkdir`, not an actual mounted execution) and the **reusable-sandbox** path (`AP_REUSE_SANDBOX=true`, where all of `codes/` is mounted at `/root/codes` and the engine resolves `/root/codes/action-runs/<ns>/…`). Fork mode *is* covered, by the engine fixture above. Worth flagging separately: `npm run test:sandbox-e2e` is currently dead — `scripts/run-sandbox-e2e.sh` builds `packages/server/worker/test/e2e/Dockerfile`, which no longer exists in the tree.

No env var, no API change, no schema change, no migration, no `@activepieces/shared` change. The cache layout is invisible to self-hosters — this path has never shipped.

Multi-tenant isolation and file handling. The risk being closed is a shared, unattributed cache directory (existence oracle + shared dependency resolution + no purge path); the mitigation is the `platformId` namespace plus a bounded sweep. Not a confidentiality leak today — mounts are read-only and a tenant only reads artifacts compiled from its own source.

**The one thing to review closely** is the path guard, since the namespace now legitimately contains a `/` and becomes a bind-mount `hostPath`: `assertSafePathSegment` is byte-identical, `assertSafeCodeNamespace` composes on top of it with a depth cap of 2, and the one existing test whose meaning changed (`a/b`, previously required to reject) is called out in the guard section above. The `canReuseSandbox` issue is the one with real read exposure, and is sequenced separately.


KhaledR57 (approved):
Just check if the cache bump is needed, other than that, its good to go

Louai-Zokerburg (approved):

amrdb (approved):

PR #14571, opened 2026-08-03
fix(builder): formula picker no longer opens on slashes inside URLs

### LINKED ISSUE #14569: [BUG] Formula picker opens on any "/" in builder text fields, and Enter then rewrites the value
## Problem

Any `/` typed in a builder text field opens the formula/function picker, including slashes inside URLs. Typing `https://example.com/api/se` opens it with query `se`, matching `uppercase` / `lowercase` / `titlecase` / `reverse`.

Pressing Enter while that stray picker is open rewrites the field: `https://example.com/api/se` becomes `https://example.com/api` + an `uppercase( )` badge. The text after the last `/` is consumed. No warning, no undo prompt.

Affects every property that renders `TextInputWithMentions` (piece short-text/URL fields, code step inputs, loop items, branch conditions).

Present since 0.85.0 (formula feature, PR #12444). Verified on main.

Ref: Pylon 5534

## Steps to reproduce

1. Open any step with a text property.
2. Type `https://example.com/api/se`.
3. Picker opens listing uppercase / lowercase / titlecase / reverse.
4. Press Enter → field becomes `https://example.com/api` + `uppercase( )`.

## Root cause

`packages/web/src/app/builder/piece-properties/text-input-with-mentions/extensions/function-slash-extension.ts:95`

`onUpdate` takes `textBefore.lastIndexOf('/')` over the 30 chars before the cursor and opens the picker for whatever follows it. No word-boundary check on the slash. The only guard is `query.includes(' ')` at line 104, and URLs contain no spaces, so it never fires.

Enter is captured window-wide (`capture: true`) at `components/function-search-popover.tsx:68-71` whenever the picker is open, so it always inserts the highlighted item.

## Fix

Require the `/` to start a word before opening: resolve the slash doc position and open only when it is at the start of its block, preceded by whitespace, or preceded by the ZWS that follows a function badge (that last case keeps nested-function insertion working). Otherwise close the picker.

majewskibartosz:
Fixes [GIT-1701](https://linear.app/activepieces/issue/GIT-1701)
Closes #14569

## Problem

1. Any `/` in a builder text field opens the formula picker, including slashes inside URLs. `https://example.com/api/se` opens it with query `se`, matching uppercase / lowercase / titlecase / reverse.
2. Pressing Enter while that stray picker is open inserts the highlighted function and consumes the text after the last slash: `https://example.com/api/se` becomes `https://example.com/api` + `uppercase( )`. Silent, no undo prompt.

Hits every property rendered by `TextInputWithMentions` (piece short-text/URL fields, code step inputs, loop items, branch conditions). Present since 0.85.0.

## Fix

- `function-slash-extension.ts` `onUpdate` → gate the picker on a new `isSlashAtWordStart` check; a slash glued to the preceding character no longer opens it. Enter can no longer rewrite the value because the picker is never open there.
- `isSlashAtWordStart` resolves the slash's doc position and allows: start of block, whitespace, hard break, or the ZWS cursor anchor a function badge leaves behind (that last case keeps nested-function insertion working).
- Doc-based rather than string-based on purpose: `textBetween` erases mention/badge leaves, so `{{step.url}}/api` would still false-positive on a string check.
- Collapsed the three duplicated close-if-open blocks into `closeIfOpen`, and hoisted the duplicated ZWS literal to `ZWS_CHAR`.

## Verification

- 8 new tests drive the real editor through the extension: 4 must-stay-closed (URL, `//` scheme, `api/se`, after a mention) and 4 must-still-open (field start, after whitespace, after the ZWS anchor, after a hard break).
- Red-first: with only the new gate reverted, exactly the 4 must-stay-closed tests fail and the 4 must-still-open ones keep passing. Re-confirmed green with the gate restored.
- `npx vitest run` in `packages/web`: 42 files, 418 tests pass. `eslint` on both changed files: 0 errors, 0 warnings. Repo `lint --filter=web`: 0 errors (70 pre-existing warnings, none in these files, all `react-hooks/exhaustive-deps` elsewhere).
- Real app (EE stack, Postgres, this branch), driven with a browser: typed the reporter's URL shape into a step text field, picker stays closed and the field below is no longer covered; appended ` /up` in the same field, picker still opens; pressed Enter on a full URL, value stays intact with 0 function badges (pre-fix the same keystroke produced `uppercase( )`).
- `Visual:` URL typed (picker closed), slash-after-space (picker open), Enter-on-URL (value preserved). Light theme only, single-theme is sufficient here because no styling changed, only whether the popover mounts.
- `Other affected paths:` checked - none. `lastIndexOf('/')` appears in exactly one place in `packages/web/src`, this extension. Pre-existing and untouched: `slashDocPos = pos - query.length - 1` assumes text-only content between slash and cursor, so a mention placed between them would offset the anchor. Not reachable via this ticket's repro and it predates the change.
- Gate not applicable: `/verify` (backend boot) - no server-side surface in this diff.
- `Product decision embedded:` picker triggers only at a word start; alternative considered: keep triggering everywhere but suppress when the surrounding text parses as a URL, rejected as unpredictable and blind to non-URL false positives like `api/se`.

<details>

## Plan

- Root cause was pinned before implementation: `onUpdate` took `lastIndexOf('/')` over the 30 chars before the cursor with no word-boundary condition, and `query.includes(' ')` was the only guard, which URLs never trip.
- Chose one predicate at the single trigger site over patching call sites, since there is exactly one trigger site.
- Kept the existing 30-char lookbehind and query parsing untouched; the change adds a precondition rather than reworking the detection.
- Predicate takes the doc and the slash position, so the mention/badge case is decided on real node types instead of flattened text.

- Nested function insertion (`/` typed inside `upper(`): the preceding char is the ZWS anchor, explicitly allowed, covered by a test.
- Multi-line fields: slash after a hard break is allowed via the node-type branch, covered by a test.
- Non-breaking space before the slash: `\s` matches ` `, so it opens.
- Slash at position 0 of the doc: `parentOffset === 0` short-circuits before any node lookup.
- Legitimate slash mid-word (`and/or`): now correctly does not open. This is the intended behavior change.

## Alternatives rejected

- URL-shaped-text heuristic: misses `api/se` and mispredicts on non-URL text.
- Requiring the picker to be dismissed manually: does not address the false trigger, and dismissal is tracked separately in [GIT-1702](https://linear.app/activepieces/issue/GIT-1702).
- Blocking Enter while the picker is open: treats the symptom, leaves the false trigger in place.
</details>


majewskibartosz:
## Visual verification

Driven in the real app on this branch (EE stack, Postgres, browser-driven as a user), typing into a step's text property.

| Before (gate reverted) | After |
| --- | --- |
| before | after |
| `https://example.com/api/se` opens the picker on the `se` after the last slash, covering the field below | same input, picker stays closed, `Add Item` visible again |

The trigger still works where it should, same field, ` /up` appended:

slash still works

States driven: URL typed (picker closed, `[role=listbox]` count 0), slash after whitespace (picker open, count 1), Enter pressed on a full URL (value intact, `[data-function-start]` count 0 where pre-fix the same keystroke produced `uppercase( )`).

Light theme only: nothing about the popover's styling changed, only whether it mounts.

AhmadTash (approved):

PR #14572, opened 2026-08-03
fix(builder): dismiss the formula picker when clicking outside the field

### LINKED ISSUE #14570: [BUG] Formula picker cannot be dismissed by clicking outside the field
## Problem

Once the formula/function picker is open in a builder text field, clicking outside does not close it. It stays mounted (portaled to `document.body`) on top of the properties below, blocking them.

It closes only on Escape, on editing the text until the `/` leaves the 30-char lookbehind, or when the whole step-settings panel unmounts.

Ref: Pylon 5534

## Steps to reproduce

1. Type `/` in any builder text field → picker opens.
2. Click the panel background, another input, or the flow canvas.
3. Picker is still open and covers the fields underneath.

## Root cause

`packages/web/src/app/builder/piece-properties/text-input-with-mentions/components/function-search-popover.tsx` has no outside-click handling. The only close path is the Escape branch (line 72).

The editor does **not** lose focus on this gesture. A mousedown on inert panel background leaves the contenteditable focused, verified in a browser: after clicking the panel's "Inputs" header, `document.activeElement` is still the `.ProseMirror` element while `[role=listbox]` is still in the DOM. So closing on editor blur alone does not fix the reported gesture. Blur is still worth handling for Tab-away and for clicking into another field, which do blur.

## Fix

Two paths, both needed:

1. `function-search-popover.tsx`: capture-phase `mousedown` listener on `document` that calls `onClose()` when the target is outside both the popover and the owning editor. Picker items already `preventDefault()` on mousedown and the capture listener sees the popover as the target, so selecting a function is unaffected.
2. `function-slash-extension.ts`: close from the extension's `onBlur` so genuine focus loss also dismisses it.

majewskibartosz:
Fixes [GIT-1702](https://linear.app/activepieces/issue/GIT-1702)
Closes #14570

## Problem

The formula picker cannot be dismissed by clicking outside the field. It is portaled to `document.body` and stays mounted on top of the properties below it, so the fields it covers are unreachable. The only close paths were Escape, editing the text until the `/` leaves the 30-char lookbehind, and unmounting the step-settings panel.

Closing on editor blur is not sufficient on its own: a mousedown on inert panel background leaves the contenteditable focused, so no blur fires. Verified in a browser, `document.activeElement` is still the `.ProseMirror` element after clicking the panel's "Inputs" header while `[role=listbox]` is still in the DOM.

## Fix

- `function-search-popover.tsx` → capture-phase `document` `mousedown` listener that calls `onClose()` when the target is outside both the popover and the owning editor. `popoverRef` is attached to all three portal branches (empty query, no results, results) since exactly one is mounted at a time.
- `function-slash-extension.ts` → close from the extension's `onBlur`, covering the focus-loss gestures that do blur (Tab away, clicking another input). Placed in the extension because it already owns every other close path for this state.
- Capture phase is deliberate: it runs before the picker items' own `onMouseDown`, and since the target is then inside the popover the listener bails out, so selecting a function is unaffected.

## Verification

- 4 new tests. Popover: closes on outside mousedown, stays open on mousedown on a `[role=option]`, stays open on mousedown inside its editor. Extension: an open picker closes when the editor emits blur.
- Red-first, run separately per path: with only the popover listener reverted, the outside-mousedown test fails and the two stay-open tests still pass; with only the `onBlur` hook reverted, the blur test fails. Both green with the code restored.
- `npx vitest run` in `packages/web`: 43 files, 414 tests pass. `eslint` on all changed files: 0 errors, 1 warning, which is the pre-existing `filtered`/`exhaustive-deps` warning on this file (identical on an untouched checkout, only its line number shifted).
- Real app (EE stack, Postgres, this branch), driven with a browser: opened the picker with `/up`, clicked the "Inputs" header outside the field, picker closed (`[role=listbox]` 1 → 0). Then re-opened with `/upper` and clicked the `uppercase` entry: function inserted (`[data-function-start]` count 1) and the picker closed, so the insert path still works.
- `Visual:` picker open, dismissed by outside click, function still insertable by clicking an entry. Light theme only, no styling changed, only when the popover unmounts.
- `Other affected paths:` checked - none. The picker is rendered from one place, `tiptap-editor.tsx`, and this is the only popover in this folder without outside-click handling.
- Gate not applicable: `/verify` (backend boot) - no server-side surface in this diff.

Note for whoever merges second: [#14571](https://github.com/activepieces/activepieces/pull/14571) touches the same extension file and introduces a `closeIfOpen` helper. The hunks do not overlap, but once both land, the `onBlur` body here should collapse into that helper.

<details>

## Plan

- Started from the ticket's stated fix (clear the slash state on editor blur) and drove it in the real app before trusting it. It did not close the picker, because the gesture in the report never blurs the editor. The ticket has been corrected.
- Kept the blur path anyway: it is the correct behavior for Tab-away and for clicking into another field, and it is cheap.
- Put the outside-click listener in the popover rather than the editor component because the popover owns the DOM node that defines "outside", and it already receives `editorRef` for positioning.
- Capture phase rather than bubble, so the decision is made before any in-popover handler can stop propagation.

- Clicking a function entry: capture listener sees a target inside the popover and bails, verified by test and in the browser.
- Clicking inside the owning editor (repositioning the caret): bails via the `editorRef` check, so the picker survives cursor moves.
- Right-click / context menu: `mousedown` fires for non-primary buttons too, so the picker dismisses. Acceptable and consistent with the rest of the builder's popovers.
- Two editors on screen, picker open in one, click the other: the second editor is not `editorRef.current`, so it closes, which is correct.
- Non-Element targets (text nodes, `document`): guarded by the `instanceof Node` check before `contains`.
- Listener lifetime: registered only while the picker is mounted, removed on unmount, so there is no cost when no picker is open.

## Alternatives rejected

- Blur-only, as the ticket originally proposed: does not fire for the reported gesture.
- Radix `Popover`/`DropdownMenu` with built-in dismissal: would mean rewriting the picker's positioning and keyboard model for one behavior, and it fights the caret-anchored placement.
- `click` instead of `mousedown`: fires after focus and selection have already moved, so the picker would visibly linger for a frame.
</details>


majewskibartosz:
## Visual verification

Driven in the real app on this branch (EE stack, Postgres, browser-driven as a user).

| Picker open | After clicking outside the field |
| --- | --- |
| before | after |
| `/up` typed, picker covers the properties below | clicked the "Inputs" header, picker gone, fields reachable again |

Selecting a function by clicking an entry still works, so the new listener does not eat the insert:

insert still works

States driven: picker open (`[role=listbox]` 1), outside click (count 0), entry clicked (`[data-function-start]` 1 and picker closed).

Worth noting for reviewers: on main the editor keeps focus through this gesture (`document.activeElement` stays `.ProseMirror` after clicking the panel header), which is why the blur hook alone does not fix it and the document-level listener is doing the work here.

Light theme only: nothing about the popover's styling changed, only when it unmounts.

AhmadTash (approved):

PR #14590, opened 2026-08-04
fix(formulas): functions with no arguments no longer show as invalid

### LINKED ISSUE #14573: [BUG] Inserted formula shows as invalid because the zero-width cursor anchor counts as an argument
## Problem

Inserting a zero-argument formula marks it invalid immediately. `now()` renders with the error state and the message `now() accepts at most 0 values, but got 1`. The value that actually runs is correct, so the red state is cosmetic, but it reads as broken.

Pressing Backspace while the cursor is still between the brackets removes the anchor and the error clears, which is what makes it look like the editor "adds a space".

Ref: Pylon 5536

## Steps to reproduce

1. In any builder text field, type `/now` and pick `now`.
2. Formula renders as invalid: `now() accepts at most 0 values, but got 1`.
3. Press Backspace with the cursor between the brackets, error clears.

## Root cause

`packages/core/formula/src/lib/function-type-checker.ts:139-143`

`countArgs` treats an arg slot as filled when `node.text.trim()` is non-empty. `insertFunctionAtPos` always inserts a zero-width space between the brackets as a cursor anchor (`function-slash-extension.ts`), and `String.trim()` does not strip ZWS because it is not ECMA whitespace. So the anchor counts as one real argument and any `maxArgs: 0` function fails its own check.

`inferArgType` in the same file already handles this correctly at lines 209-214, stripping ZWS before `trim()`, with a comment explaining exactly this trap. `countArgs` was not given the same treatment.

## Fix

Two parts. The strip alone is not enough, proven empirically before committing.

1. Strip the anchor before the emptiness test in `countArgs`, via a shared `stripCursorAnchors` helper also used by `inferArgType` so the two sites cannot drift again.
2. Suppress only the min-args complaint while the body is untouched (`argCount === 0` with nodes still present). Without this, part 1 fixes `now()` but flips every freshly inserted `minArgs: 1` function (`trim()`, `uppercase()`) to red with `needs at least 1 value, but got 0`.

Known side effect, accepted: a whitespace-only body such as `trim("   ")`, and a `combine(anchor)` whose separator was deleted, no longer report "needs at least N" until a real character lands.

Reproduced and verified in a browser against a real EE stack, not just by code read.

majewskibartosz:
Fixes activepieces/activepieces#14573<br>Closes activepieces/activepieces#14573

## Problem

* Insert any zero-argument formula (`now()`, `uuid()`, `today()`) in a flow field and it renders red immediately, with `now() accepts at most 0 values, but got 1`.
* The saved value and the live preview are both correct, so the red badge is purely cosmetic. It still reads as "my flow is broken".
* Also hits reload, not just insertion: the deserializer inserts the same anchor when a saved `now()` is loaded back into the editor, so the red state survives a refresh.
* Cause: `countArgs` in `packages/core/formula/src/lib/function-type-checker.ts` decided a slot was filled with `node.text.trim()`. The editor always puts a zero-width space (U+200B) between a formula's brackets as a cursor anchor, and `String.trim()` does not strip U+200B (it is not ECMA whitespace). The anchor counted as one real value, so every `maxArgs: 0` function failed its own check.
* `inferArgType`, in the same file, already stripped U+200B before `trim()` and carried a comment about exactly this trap. `countArgs` never got the same treatment.

## Fix

* One non-exported `stripCursorAnchors` helper, used by both `countArgs` and `inferArgType`, so the two sites cannot drift apart again.
* Second guard, needed to avoid trading one wrong red badge for another: with the anchor no longer counted, `argCount` for a freshly inserted single-value function (`trim()`, `uppercase()`, every `minArgs: 1` function) drops to 0 and the `needs at least 1 value` branch would fire the instant the user inserts it. `isUntouchedBody` (`argCount === 0` with a non-empty body) suppresses only that case. A body the user actually emptied (no nodes at all) is still flagged, and any real value in a slot is still counted.
* No behaviour change for multi-value functions: `combine(;)` and friends get their count from separators, which were never affected.

## Verification

* Red-first, change A (revert only the `countArgs` strip): `× zero-width space stripping > 0-arg function whose body is only the ZWS anchor is valid — now()`, 1 failed | 16 passed. Restored: 17 passed.
* Red-first, change B (revert only `isUntouchedBody`): `× zero-width space stripping > 1-arg function whose body is only the ZWS anchor is not yet flagged — trim()`, 1 failed | 16 passed. Restored: 17 passed.
* 4 new cases in `packages/core/shared/test/formula/type-checker.test.ts`, two of them guards proving validation is not disabled: `now(5)` still reports `at most`, `combine("a")` still reports `at least`. U+200B is written as the `\u200B` escape, and the pre-existing literal anchor in that file was converted to the escape too so the character is greppable.
* Unit suites: shared 453 passed (13 files), web 410 passed (41 files), engine 445 passed (38 files). Engine needs `turbo run build --filter=@activepieces/piece-delay` first in a fresh worktree, otherwise `flow-with-delay.test.ts` fails 5 tests with `Piece not found for package: @activepieces/piece-delay` on `main` as well.
* Lint: `npx eslint` on both changed files clean; `turbo run lint --filter=@activepieces/shared --filter=web --filter=@activepieces/core-formula --force` = 0 errors, 70 pre-existing `react-hooks/exhaustive-deps` warnings in web (all present on an untouched checkout).
* Visual: real EE stack (api + worker + engine on 3101, vite on 4301, Postgres + Redis), Text Helper → Slugify step, `Text` field driven in the browser. States driven: `now()` inserted from the `/` menu on unfixed code (red), the same `now()` reloaded from the saved flow on fixed code (clean), `now(5)` on fixed code (still red), `trim()` freshly inserted on fixed code (clean). DOM assertions per state, not eyeballing: before `.ap-fn-error` = 2 and `[data-fn-error-msg]` = `now() accepts at most 0 values, but got 1`; after `.ap-fn-error` = 0 and `[data-fn-error-msg]` = 0; `now(5)` `.ap-fn-error` = 2 with the same message; `trim()` `.ap-fn-error` = 0. Screenshots in the PR comment below.
* Other affected paths: `formula-evaluator.ts` has its own `trim()`-based emptiness tests (`quoteIfBare`, `validateFunctionArgs`) but is unreachable with an anchor in the string. Every caller in web goes through `textMentionUtils.convertTiptapJsonToText`, which strips U+200B from every text node before the expression reaches the evaluator, which is also why the saved value and preview were always right. `keysWithinPath` and the `flattenNestedKeys` parsing in `text-input-utils.ts` read mention `serverValue`s, which never contain an anchor. Left as is, deliberately: no anchor-blind emptiness test remains reachable in the formula package.

<details><summary>Plan & reasoning</summary>

Approach: fix at the shared choke point rather than at the symptom. Both readers of an arg slot's text now strip the anchor through one helper, so a future third reader has an obvious thing to call.

Non-happy scenarios considered:

* Whitespace-only body (`trim(" ")`): now silent instead of `needs at least 1`. Deliberate. It is the same "field not filled in yet" state as a fresh insert, and the moment a real character lands the error comes back.
* Separator removed from a multi-value call, leaving `combine(anchor)`: silent until something is typed, then `needs at least 2` as before.
* Body genuinely emptied with backspace: still flagged, and the pre-existing test asserting that is untouched.
* Nested calls: `countArgs` only inspects depth-0 text nodes, so anchors inside a nested function never reached this branch and still do not.
* Anchor adjacent to a real value (`now(anchor5)`): the value is still counted, which is what keeps the arity check honest. Covered by a test.

Alternatives rejected:

* Only stripping in `countArgs` as originally scoped: ships a new red badge on insert for every `minArgs: 1` function, which is a worse bug than the one being fixed. Confirmed empirically before adding the second guard.
* Special-casing `maxArgs === 0` at the comparison site: hides the false positive for `now()` and leaves the same blindness for every other arity check.
* Making the editor stop inserting the anchor: it is load-bearing for cursor placement between the bracket badges, and the deserializer inserts it too, so any consumer of the doc must tolerate it.
* Stripping the anchor in the tiptap doc before type-checking: an extra full traversal per keystroke to work around two `trim()` calls.

</details>


majewskibartosz:
## Visual verification

Real EE stack from this branch's worktree: api + worker + engine on port 3101 against a dedicated Postgres, vite on 4301, Postgres and Redis in docker. Flow is a Schedule trigger into a Text Helper → Slugify step; the driven field is that step's `Text` property. Neutral values only.

| Before (this branch's fix reverted, dev server reloaded) | After |
| --- | --- |
| before | after |
| `now()` red, `Error: now() accepts at most 0 values, but got 1`, while the preview below it renders the timestamp correctly | same `now()`, reloaded from the saved flow, no error badge and no error row |

| Arity check still fires | Single-value function still clean on insert |
| --- | --- |
| inverse | one arg |
| `now(5)` is red with the same message, so the fix did not disable validation | `trim()` freshly inserted stays clean instead of demanding its value before anything is typed |

States driven, in order:

1. `/now` inserted from the slash menu with the fix reverted (before).
2. Fix restored, vite dep cache cleared, page reloaded so the saved `now()` came back through the deserializer (after).
3. `/now` then `5` typed inside the brackets (arity check).
4. `/trim` inserted and left untouched (no new red state for `minArgs: 1`).

DOM assertions, run in the page for each state rather than judged by eye:

| State | `.ap-fn-error` count | `[data-fn-error-msg]` | editor text |
| --- | --- | --- | --- |
| before, `now()` | 2 | `now() accepts at most 0 values, but got 1` | `now(U+200B)` |
| after, `now()` | 0 | none | `now(U+200B)` |
| after, `now(5)` | 2 | `now() accepts at most 0 values, but got 1` | `now(U+200B 5)` |
| after, `trim()` | 0 | none | `trim(U+200B)` |

The `U+200B` in the editor text is the cursor anchor this PR stops counting as a value.

amrdb (approved):

Louai-Zokerburg (approved):

PR #14609, opened 2026-08-05
fix(audit-logs): index audit_event for the created DESC list query

amrdb:
## Description

Platform Admin → Audit Logs failed with "Failed to load data": `GET /v1/audit-events` returned 500 with Postgres `57014` (statement timeout) on a plain first-page load — no filters, no cursor.

`auditLogService.list` pages through `Paginator`, which orders by `created DESC` **and appends an `id DESC` tiebreaker** (`withIdTiebreaker`), so the query is:

```sql
SELECT * FROM "audit_event" WHERE "platformId" = $1 ORDER BY "created" DESC, "id" DESC LIMIT 21
```

None of `audit_event`'s three indexes covered that sort. Prod plan for the affected platform — Postgres reads every one of its rows through the `platformId`-leading `action` index, then sorts the lot to return 21:

```
Limit  (cost=7449258.32..7449259.61 rows=11 width=732)
  ->  Gather Merge  (cost=7449258.32..8201854.97 rows=6450378 width=732)
        ->  Sort  (cost=7448258.30..7456321.27 rows=3225189 width=732)
              Sort Key: created DESC, id DESC
              ->  Parallel Index Scan using audit_event_platform_id_action_idx on audit_event
                    Index Cond: (("platformId")::text = 'REDACTED'::text)
```

`audit_event` on cloud prod is **362M rows / 475 GB**, ~6.5M rows for that platform, and nothing prunes it — so this grows monotonically.

This adds `("platformId", "created" DESC, "id" DESC)`. Both trailing columns are needed: with `("platformId", "created" DESC)` alone the planner still stacks an Incremental Sort on top to satisfy the `id` tiebreaker; with all three it is a plain index scan that stops after 21 rows.

The migration checks `pg_index.indisvalid` and drops an invalid leftover before rebuilding. A `CONCURRENTLY` build that gets killed (statement timeout, or an orchestrator killing the container mid-boot) leaves the index present but unusable; a bare `IF NOT EXISTS` retry would skip it, return cleanly, and record the migration as applied against an index the planner never uses — a silent non-fix on exactly the table where nobody would look again.

### ⚠️ Deploy note for large deployments

At 475 GB, `CREATE INDEX CONCURRENTLY` runs for hours and needs ~20 GB of free disk. Migrations run in `main.ts` **before** the server listens, so a boot-time build never reaches the healthcheck and the deploy gets rolled back on top of a half-built index. **Build it by hand ahead of the release**, then the migration no-ops through `IF NOT EXISTS`:

```sql
SET statement_timeout = 0;  -- CREATE INDEX CONCURRENTLY obeys it
CREATE INDEX CONCURRENTLY "audit_event_platform_id_created_id_desc_idx"
ON "audit_event" ("platformId", "created" DESC, "id" DESC);
```

## How was this tested?

Measured on PostgreSQL 14.4 against a clone of the table carrying its three real indexes, 600k rows (540k for one platform):

| query | before | after |
| -- | -- | -- |
| first page, no filters | parallel seq scan + top-N heapsort, 11,899 buffers / 43 ms | plain index scan, **14 buffers / 0.26 ms** |
| page 2 (cursor) | 11,899 buffers | 32 buffers |
| `action` filter | seq scan + sort | index scan, 54 buffers |

The migration's invalid-index path was exercised against real Postgres in all three states: index present but `indisvalid = false` (dropped, rebuilt valid), already valid (detection returns 0 rows, no drop, `IF NOT EXISTS` no-ops), absent (builds). PGlite path takes the non-`CONCURRENTLY` branch.

Edition paths: `audit_event` is registered unconditionally in `getEntities()`, so the migration runs everywhere and is a no-op-cost index on CE, where the table is empty. The endpoint itself is EE/Cloud only behind `platform.plan.auditLogEnabled` — unchanged by this PR.

**Filtered queries, on a 600k-row clone with all four indexes present**, including a hostile case where the filtered rows are the oldest in the table — the worst position for an ordered walk:

| filter | before | after | plan after |
| -- | -- | -- | -- |
| `userId` = 10 matches, all oldest rows | 0.26 ms | 0.026 ms | unchanged — still `(platformId, userId, action)` + tiny sort |
| `userId` matching zero rows | 0.016 ms | 0.013 ms | unchanged — same selective index |
| `userId` = 6,000 matches, spread | 6,008 buffers / 4.29 ms | 0.99 ms | new index, 2,042 entries walked and filtered |
| `projectId IN ('proj3','proj7')` = 60k matches | parallel bitmap heap scan, 8,944 buffers / 21.3 ms | 7 buffers / 0.033 ms | new index, 203 entries walked |

Selective filters keep their existing plan — walking `created DESC` for a 1-in-60,000 `userId` costs far more than the selective index plus a 10-row quicksort, and the planner prices that correctly. Loose filters get much faster.

**At prod scale, two filtered shapes are slow, and both need a follow-up rather than a change here:**

- `action IN (3 values)` **regresses from 1.43 s to 6.32 s** — measured both ways on prod, the old plan forced with `SET LOCAL enable_indexscan = off`. The old bitmap route reads 8,010 heap blocks in physical order; the new index walks `created DESC` and does 14,793 random single-row fetches at ~0.43 ms each, because `action` isn't in the index. The planner flips on a 14% cost margin (12,061 vs 13,962) and its row estimate is essentially right (predicted ~16.7k walk depth, actual 14,793) — what it gets wrong is cost per fetch, since one platform's rows in `created` order are scattered across the full 475 GB heap by multi-tenant interleaving. `ANALYZE` cannot fix that, and appending the filter columns to the index does not either: Postgres 14 applies a trailing `= ANY` as a heap `Filter:`, not an `Index Cond:`, so the fetches still happen.
- `projectId IN (2)` also times out — but on `(platformId, projectId, userId, action)`, the pre-existing plan, untouched by this PR. Its 3,396 rows are scattered across the full 475 GB of history, so every fetch is a cold read (>17 ms/row vs 0.43 ms/row for the recent rows the created index walks).

Neither 500s under the current 1-minute timeout. **The trade-off this PR makes explicit: the default page goes from a hard 500 to working, and a 3-action filter goes from 1.4 s to 6.3 s.**

The fix for every filtered shape is a deferred join — filter and sort inside an index-only subquery over `(platformId, action, created DESC, id DESC)`, then fetch only the 21 surviving rows — which keeps the heap out of the discard path entirely. That belongs in `Paginator`, shared by every paginated endpoint, so it is a separate PR rather than a tack-on here. Shrinking the candidate set (retention/pruning, GIT-1574, or a bounded default date window in the UI) attacks the same cost from the other side.

Fixes GIT-1705, GIT-1574


abuaboud (approved):

amrdb:
@copilot resolve the merge conflicts in this pull request

PR #14611, opened 2026-08-05
feat: piece-declared sensitive output, declare via outputSchema, redact on serve

### LINKED ISSUE #13706: [Feature] Secure inputs/outputs: option to redact sensitive step input/output values in run logs
## Problem

When a flow fetches a secret at runtime (e.g. the AWS Secrets Manager piece's get-secret action, or any HTTP call returning credentials), the value appears in plain text in the run logs — both as that step's output and in the logged input of every downstream step it's mapped into.

The only redaction that exists today is token-based: `{{connections[...]}}` and `{{variables[...]}}` references are replaced with `**REDACTED**` in the logged (censored) input (`packages/server/engine/src/lib/variables/props-resolver.ts`, stored via `censoredInput` in `piece-executor.ts`). Step outputs are stored raw with no censoring pass (`stepOutput.setOutput(output)`), and the action schema has no per-step privacy option (`packages/shared/src/lib/automation/flows/actions/action.ts` — only `skip`, `continueOnFailure`, `retryOnFailure`).

This blocks the common pattern of using a vault piece in-flow when the target API's auth scheme isn't supported by an existing piece (e.g. dual static-key auth), forcing users to choose between functionality and leaking credentials into run logs. Comparable platforms cover this (Power Automate "secure inputs / secure outputs" per action).

## Proposed

Two complementary levels:

1. **Per-step toggle (Power Automate-style):** a "secure input" / "secure output" option in step settings. When enabled, the run log stores `**REDACTED**` for that step's input and/or output. Default off — existing flows unaffected.<br>

   
2. **Framework-level rule:** values originating from sensitive sources are always redacted in run logs by default — e.g. anything resolved from `PieceAuth.SecretText` / secret connection fields, and piece actions able to mark specific output fields as sensitive in their definition (so vault-style pieces like AWS Secrets Manager redact returned secret values automatically, including when mapped into downstream step inputs).

Related: activepieces/activepieces#4260 (custom user-defined OAuth2/secret — solves the connection-route side; this issue covers the run-log/logging side).

kishanprmr:
## Description

Piece authors can now declare secret-bearing output fields with `sensitive: true` on `OutputSchemaField`. The engine collects the concrete leaf paths after every `run()`, and the run-details endpoint, sample-data serve, and downstream steps' `censoredInput` all swap those leaves to `**REDACTED**`. Downstream `run()` calls, resume points, and outbound HTTP requests still receive the real value — only the display is scrubbed.

Closes the customer ticket where `Get Secret Value` (`@activepieces/piece-amazon-secrets-manager`) exposes `SecretString` in the run panel and in any downstream step's Input tab that references it.

**First adopters** (this PR): `Get Secret Value` (`SecretString`, `SecretBinary`), `Get A Random Password` (`RandomPassword`).

**Deferred** (follow-up PR, branch `feat/piece-declared-sensitive-input`): the input-side `Property.SecretText()` factory and the redactor for user-pasted secrets. The cross-step propagation shipped here means the customer's primary leak (`Get Secret Value` → HTTP `Authorization: Bearer {{step.SecretString}}`) closes with the output PR alone; the remaining gap is a user pasting a secret directly into a piece prop.

Design: `brain/knowledge/decisions/000017-sensitive-piece-io-is-piece-declared-and-scrubbed-on-serve.md`.

## Review-fix pass (update)

All review comments from the initial round were triaged and closed — 14 fixed, 1 (`**REDACTED**` colliding with a legitimate literal output value) intentionally left open as a known, accepted limitation.

Two additional issues surfaced during a full revalidation pass after the initial fixes landed, both now fixed:

- **`step-output-redactor.ts` crashed on any run containing a loop step, or any step with declared-sensitive fields** — `GET /flow-runs/:id` returned 500. Root cause: steps read back from a run's log file come from `JSON.parse`, which strips the class prototype, so calling `step.setOutput(...)` (a `GenericStepOutput` instance method) threw `TypeError: step.setOutput is not a function` at runtime even though it type-checked cleanly. This regressed twice across the review-fix history before landing for good — rebuilt via `new GenericStepOutput({...step, output})` instead of instance methods, with a generic `redactLeafStep<T, OUTPUT>` helper to keep the union-of-class-types narrowing sound without `as` casts. Pinned by a regression test that builds its inputs via `JSON.parse(JSON.stringify(...))` to reproduce the exact prototype-less shape (covers `PIECE`, `LOOP_ON_ITEMS`, and `ROUTER` step types).
- **`context.output.update()`** (the piece-authored progress-streaming API) was still pre-redacting server-side and dropping `sensitiveOutputPaths` before emitting on the same `TEST_STEP_PROGRESS` socket event the web app persists as sample data — a dormant landmine that would have baked `**REDACTED**` into sample data at rest for the first piece to adopt that API. Fixed to send raw output + computed paths, matching the pattern already used by the step-finalize path.

Consolidated the schema-walking algorithm (`collectSensitiveOutputPaths`) — previously duplicated verbatim in both the engine and the API — into `@activepieces/core-execution` behind a structural duck-type, so it needs no dependency on `pieces-framework`.

Full finding-by-finding table and reasoning: see PR discussion; durable gotchas recorded in `brain/knowledge/decisions/000017-...md`.

## Final review pass (update)

A third review pass (correctness + security, plus an independent trace of every server-to-client serve path) found six bypasses reachable by the two current adopters, three unreachable ones, and five items worth deferring. Reachable ones fixed, unreachable ones guarded fail-closed, deferred ones recorded in decision 000017 so they are not relitigated.

**Fixed — reachable today**

- **Loop progress updates shipped iterations raw.** A LOOP step carries no `sensitiveOutputPaths` of its own, so the leaf-only redaction in `flow-run-progress-reporter.ts` sent `output.iterations[*].<inner>.output` unredacted over `UPDATE_RUN_PROGRESS`. In a test flow the inner step's redacted update arrived first and the loop step's raw update overwrote it in the run panel, with nothing refetching `GET /flow-runs/:id`. The redactor moved to `@activepieces/core-execution` so the engine and the API share one recursive walk rather than the engine growing a second copy.
- **A loop's `output.item` was never redacted.** `loop-executor.ts` sets it from the raw resolved input, so a loop over an upstream sensitive array showed `**REDACTED**` on the source step and the real secret on `loop.item` — and offered `{{loop_x.item.<leaf>}}` in the data selector. The loop now compares resolved against censored items and stamps `sensitiveOutputPaths: ['item']` when they differ; the value stays raw in execution state so inner steps still execute.
- **Trigger sample data was never redacted, permanently.** `sensitiveOutputPathsPieceVersion` was stamped on *every* OUTPUT save, including the five web trigger paths that never carry paths. That persisted "no paths, current version", which made `sensitiveOutputPathsResolver` short-circuit on version equality and serve the raw payload forever — no piece upgrade required. The decision now lives in one pure, unit-tested helper, because this line had already been flipped twice.
- **Stale paths could never be cleared.** `undefined` now means "the caller does not know" and `[]` means "computed, none", so a re-test clears stale paths after a piece drops `sensitive: true`, while trigger saves still trigger live re-derivation. This also fixes the stale local display after a trigger test, since the sample-data query is invalidated on save and the refetch now comes back redacted.
- **The serve path was neither fail-soft nor cheap.** The resolver ran a platform + piece-metadata lookup per step on every builder open (the stored version differs for every step never re-tested) and propagated a throw as a 500 on a previously side-effect-free GET. It now skips steps with no sample data and degrades to the stored paths on error.
- **The redactor deep-rebuilt every loop tree on every serve.** It now returns its input by identity when nothing changed, which matters because `pollForRunCompletion` calls `getOnePopulatedOrThrow` in a loop.

**Guarded fail-closed — not reachable by either current adopter**

`sensitive` is resolved against a field's `key`. Three schema shapes cannot be addressed that way and previously redacted *nothing* while logging only a warning: a sensitive leaf behind a `value` path override, a `value`-overridden parent wrapping a sensitive child, and an action whose output is a top-level array. 37 piece schema files already use `value:`, and the docs introduced `value` and `sensitive` in the same sentence, so this was a live trap for the next adopter. All three now redact the **entire step output** and log an error. Documented in the piece reference and the output-schema skill.

**Also**: dropped the per-field "declared sensitive but missing from output" warning, which fired on *every* `Get Secret Value` run — the action declares both `SecretString` and `SecretBinary` and exactly one is ever present.

**Deferred, recorded in decision 000017**: unredacted `GET /v1/trigger-events`; Test-Step hydration reading stored paths only; O(n^2) apply over per-index `listItems` paths; `pieceRunner.describe` on every `BEGIN`; and the evidence that the dual log-slice upload is *not* a storage leak (`deleteStaleBulk` sweeps by `(type, created)` and deletes by primary key, so both copies are collected on the same pass).

**One correction to an earlier review round**: the `if (step.type === ROUTER) return redactLeafStep(step)` branch that looks identical to the fallthrough is **not** dead code. It splits the `StepOutput` union so TypeScript infers a single concrete `T` per call site; collapsing it fails the build with `Type 'FlowActionType.CODE' is not assignable to type 'FlowActionType.ROUTER'`. Removal was attempted, the build caught it, and the reason is now recorded in the decision doc.

## How was this tested?

* **Unit**: `output-sensitive-paths.test.ts` (path collection at every depth, incl. `dynamicKey` maps), `sensitive-path-utils.test.ts` (leaf swap), `step-output-redactor.test.ts` (redaction against JSON-deserialized plain objects for `PIECE`/`LOOP_ON_ITEMS`/`ROUTER` step shapes — the exact case that caused the runtime crash above), `retry-flow-run.test.ts` (6 cases incl. `FROM_FAILED_STEP`).
* **Manual E2E** on CE (`AP_EDITION=ce`) with a live AWS Secrets Manager connection and a real flow (manual trigger → code → loop → `Get Secret Value` → HTTP call):
  * `GET /flow-runs/:id` on a run with a loop nesting a sensitive-output step — output panel and sample data both show `SecretString` as `**REDACTED**`, all other fields (ARN, headers, etc.) unredacted.
  * `GET /sample-data` for `OUTPUT` and `INPUT` types — sensitive fields redacted, non-sensitive fields raw.
  * Retry `ON_LATEST_VERSION` on a succeeded run, and retry `FROM_FAILED_STEP` on a real failed run (intentionally broken HTTP URL) — both re-execute correctly and redact correctly, including inside a failed loop iteration.
  * The downstream HTTP step's outbound request used the real secret (not the redacted placeholder), confirming the raw value is still available where execution needs it.
* **Final pass — real redaction check, not just compilation**: the four schema shapes that previously leaked (`value`-aliased leaf, `value`-wrapped parent, `value: ''` whole-output, top-level array) were reproduced against the real collector + redactor and each returned the raw secret; after the fix each returns `**REDACTED**`, asserted by tests that check the secret string is *absent* from the serialized output rather than just checking the path list.
* **Final pass — suites**: `core-execution` 130/130, `engine` (variables + helper + operations) 196/196, `api` unit `app/flows` 202/202. Full `api` unit run 1002 passed against a 1001 baseline (same 18 pre-existing failures, all Redis/DB-bound and unrelated); `engine`'s 5 `flow-with-delay` failures confirmed pre-existing by stashing the change set and re-running.
* **Final pass — build**: `npx turbo run build` green across `core-utils`, `core-execution`, `pieces-framework`, `engine`, `api`, `web` (18/18).
* **Lint**: `npx turbo run lint` clean across all touched packages (0 errors).

Fixes # (issue)

- [X] yes — security-sensitive (call out the risk and mitigation in the description above)

Touches run-log secret visibility. Ceiling is called out honestly in `docs/build-pieces/piece-reference/output-schema.mdx` — guarantee is **UI visibility only**, not encryption at rest. DB dumps, run exports, flow exports, and platform admins with direct DB access still see the real value. Customers whose threat model needs the stronger guarantee should route through Secret Managers references (unchanged). Mitigations: the `**REDACTED**` sentinel is a plain string (preserves JSON shape end-to-end, fails silently if accidentally leaked into `resolvedInput` rather than throwing), and the raw value never leaves the engine's read path (variable resolver + piece `run()` invocation).

Fixes activepieces/activepieces#13706


kishanprmr:
Thanks — real gap. `buildSensitiveStepPaths` was reading only the root `executionState.steps` map, so a step inside a loop referencing a sibling step in the same iteration would get the raw secret in `censoredInput`.

Fixed in [691a097](https://github.com/activepieces/activepieces/commit/691a097737) — the map now walks `executionState.currentPath.path` and merges per-iteration `sensitiveOutputPaths` on top of root, mirroring how `currentState()` layers iteration-local siblings for variable resolution.

Two regression tests added in `props-resolver.test.ts`:
- root-level reference: `resolvedInput` keeps the real value, `censoredInput` swaps in `**REDACTED**`.
- **loop-local sibling reference**: same guarantee holds when the reference is to a sensitive sibling produced inside the current loop iteration. This is the exact case that was failing before the fix.

amrdb:
## Code review — 15 findings

Ranked most-severe first. @greptileai please verify.

### Critical — the PR's own goal leaks

1. **`sample-data.service.ts:21` — reload serves the raw secret.** `getOrReturnEmpty` returns the stored sample data with no `applySensitivePaths` call. `sensitiveOutputPaths` lives on `step.settings.sampleData`, but the GET endpoint doesn't consult it. On page reload the data selector shows `SecretString: 'sk-real'` — the primary UX this PR advertises fixing. The only invocation of `applySensitivePaths` in web is inside `updateSampleData`, which only runs immediately after a successful test-step.

2. **`step-output-redactor.ts:11` — large outputs bypass redaction entirely.** When `outputType === StepOutputType.SLICE`, `rest.output` is a `LogSliceRef {fileId, size, url}`, not the value. `applySensitivePaths(ref, ['SecretBinary'])` walks path `SecretBinary` on the ref, finds nothing, returns it unchanged. Client renders `SlicedOutputDownload` with `href={slicedOutputRef.url}` — the download is the un-redacted raw blob. Anything exceeding `AP_FLOW_RUN_LOG_SLICE_THRESHOLD_KB` leaks.

3. **`props-resolver.ts:153` — `{{step_1.output}}` collapses the whole object to `'**REDACTED**'`.** `isSensitiveStepReference` returns true whenever the reference has no subpath, so a reference to the entire step output is replaced with the sentinel string in `censoredInput`. Non-sensitive fields (`Name`, structural JSON shape) are lost. Should be `applySensitivePaths(evalInScope(...), paths)` so only the sensitive leaves are scrubbed.

### Redaction-bypass surfaces

4. **`run-state.ts:245` — streamed progress flashes the secret.** `TEST_STEP_PROGRESS` handler calls `setSampleDataLocally({ value: response.output })` with raw output. `TestStepProgressEvent` has no `sensitiveOutputPaths` field. Between progress ticks and `TEST_STEP_FINISHED`, the sensitive value is visible in the data selector.

5. **`props-resolver.ts:143` — regex requires literal `.output`.** `^([^.]+)\.output(?:\.(.+))?$`. Templates that reference step data differently (e.g. flattened via `flattenNestedKeys`, older imported flows using shorthand) don't match and skip redaction entirely.

6. **`props-resolver.ts:258` — sandbox `evalInScope` bypasses the check.** The sensitivity check only runs in `resolveSingleToken`. `{{ step_1.output.SecretString + 'x' }}` hits `codeSandbox.runScript` via `session.run`; the script session's `scriptContext` holds raw `currentState`, so the concatenated expression evaluates to the real secret before `censoredInput` interpolation.

7. **`sample-data.service.ts:18` — omitting `sensitiveOutputPaths` on save silently clears existing paths.** `sensitiveOutputPaths: params.type === OUTPUT ? params.sensitiveOutputPaths : existingPaths` — any internal caller (e.g. `flow-run-utils.saveSampleDataFileIdsInStep`) that builds `SaveSampleDataParams` without the new field passes `undefined`, and the stored paths are wiped. Next reload serves the raw data with no redaction at all.

8. **`step-output-redactor.ts:12` — nested agent/sub-flow outputs aren't visited.** Recurses only for `FlowActionType.LOOP_ON_ITEMS`. Agent/sub-flow step outputs whose internal shape resembles `{ steps: {...} }` are not descended into; sensitive values produced by child agent steps stay raw in the parent's output.

9. **Trigger sensitivity is dead-lettered.** The redactor walks trigger + action steps, but the engine only writes `sensitiveOutputPaths` from `piece-executor` (action path). `webhook-executor`/`trigger-executor` never invokes `collectSensitiveOutputPaths`, so `outputSchema.fields[].sensitive: true` on a trigger produces no redaction — a piece author reading the schema has a false sense of security with no runtime warning. Documented as "deferred" in the design doc but the flag surface is enabled.

### Correctness / hygiene

10. **`output-sensitive-paths.ts:20` — dotted field keys can't be round-tripped.** Path built by literal `${prefix}.${field.key}` concat, no escaping. A field key `'kv.secret'` records path `'kv.secret'`; `redactAtPath` splits and navigates `record.kv → record.secret`, misses the real key, and serves the raw secret.

11. **`props-resolver.ts:138` — loop-body step name collisions silently drop outer sensitivity.** `Object.fromEntries` of `[name, paths]` tuples: an inner-loop step reusing a top-level step name shadows the outer entry. Legacy imports/programmatic flow versions can trigger this.

12. **`output-sensitive-paths.ts:32` — prototype chain traversal in `readAt`.** `(value as Record)[key]` returns `constructor`/`__proto__`/`toString` from the prototype. A copy-pasted key like `'constructor'` traverses into `Object`. Use `Object.hasOwn`.

13. **`sensitive-path-utils.ts:25` — lenient array index parsing.** `Number(head)` accepts `'1e2'`, `'0x0'`, `' 3 '`. Currently the collector only emits real indices, but a hand-crafted `outputSchema` with numeric-looking string keys parses as indices. Use `/^\d+$/` + `parseInt(head, 10)`.

14. **`step-output-redactor.ts:13, :21` — two `as` casts violate the repo-level "No type casting" rule in CLAUDE.md.** The second cast masks that `{ ...rest, output }` is a plain object without `GenericStepOutput` methods.

15. **`props-resolver.ts:118` — `buildSensitiveStepPaths` walks all steps on every resolve.** Even when the input has no `{{ ... }}` tokens. Short-circuit on empty `referencedStepNames`. Not a correctness bug; wasted work per resolve, scales with flow size × loop depth.

### Blockers I'd hold merge on

#1, #2, #3, #4 and #9 — the feature doesn't hold under reload, sliced outputs, whole-object references, streamed progress, or triggers. The rest are follow-ups.

`@greptileai` — please verify these findings and flag anything I missed.

kishanprmr:
@greptileai review

kishanprmr:
@amrdb added fixes for the comment.

PR #14640, opened 2026-08-06
fix(pieces): declare transitive npm deps that the publish bundler externalizes

AhmadTash:
## Description
`datadog`, `snowflake`, and `scrapeless` each fail the Release Pieces workflow's `update pieces metadata` step with `Cannot find module` (seen in https://github.com/activepieces/activepieces/actions/runs/31089312001/job/92576207982) — the step swallows the error per-piece and continues, so the job reports green while these 3 pieces' metadata silently never gets registered/updated.

Root cause: the publish bundler (`packages/cli/src/lib/utils/bundle-piece-utils.ts`) inlines `@activepieces/*` workspace source unconditionally, but externalizes third-party packages under three different triggers that each of these pieces hits — datadog's SDK size triggers the inline-size fallback, snowflake's `open` triggers the dynamic-require hazard auto-externalize, and scrapeless sets `bundleDeps: false` itself. In all three cases, the externalized package was only ever pulled in *transitively* (through inlined `core-utils`/`pieces-common`/`pieces-framework` source, or a piece's own SDK) and never declared as a direct dependency, so the piece's own `node_modules` never links it.

Fix: declare each missing package as a direct dependency, matching the exact version already used elsewhere in the workspace (`deepmerge-ts`/`form-data`/`ipaddr.js`/`nanoid`/`zod` per `core-utils`/`pieces-common`/`pieces-framework`; `open` per `snowflake-sdk`'s own required range). Bumped each piece's patch version per `validate-publishable-packages`.

## How was this tested?
- Ran the actual `prepare-pieces-for-publish.ts` bundler locally against all three pieces and confirmed the external-deps list is now fully satisfied
- Ran `load-piece-metadata-child.mjs` directly against each piece's bundled `dist/` (the exact call the CI step makes) — all three now exit 0 and produce valid metadata JSON
- `npx turbo run lint` on all three pieces — 0 errors (pre-existing warnings only, untouched by this change)

Fixes # (issue)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -117,7 +117,7 @@
     },
     "packages/core/execution": {
       "name": "@activepieces/core-execution",
-      "version": "0.10.0",
+      "version": "0.10.1",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -213,7 +213,7 @@
     },
     "packages/pieces/common": {
       "name": "@activepieces/pieces-common",
-      "version": "0.12.7",
+      "version": "0.12.8",
       "dependencies": {
         "@activepieces/core-utils": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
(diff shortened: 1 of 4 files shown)


kishanprmr (approved):

AhmadTash:
Stale PR

PR #14646, opened 2026-08-06
chore(release): sync version to 0.87.0 after off-cycle release

automated-commits-ap:
The self-hosted release `0.87.0` was published from a branch whose version bump is not on `main`. This PR syncs `package.json` and the `docker-compose.yml` image pin so the next weekly release does not read a stale version. Merge before the next release-candidate cut (Thursday 5 PM UTC). See GIT-1595.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   app:
-    image: ghcr.io/activepieces/activepieces:0.86.3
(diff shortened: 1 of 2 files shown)


abuaboud (approved):

abuaboud (approved):
Mechanical version sync for the 0.87.0 self-hosted release. Diff is the version bump + docker-compose image pins only.

PR #14651, opened 2026-08-06
feat(pieces): add outputSchema across 6 core pieces (PIE-376)

OdaiAhmed99:
## Summary
- Adds `outputSchema` to Date Helper (7/9 actions), Delay (2/2), HTTP (Parse URL), Forms (Respond on UI, Chat UI), Tables (all 9 actions + 3 triggers), and CSV (2/3 actions) so their step output renders as a labeled tree in the builder instead of raw JSON.
- Deliberately left undescribed: HTTP Send HTTP Request, Webhook (catch-hook, respond-and-wait-for-next-webhook), Date Helper's 2 multi-select actions, and CSV's Convert CSV to JSON -- their output shape varies per run and cannot be pinned to one schema.
- Two bugs found and fixed while writing schemas:
  - Delay: Test Step accepted inputs a real run would reject, and Delay Until had no timestamp validation at all.
  - Tables Create Record(s): silently ignored the Records form and inserted empty records, because Property.Json's `{}` default was treated as "raw override provided".

## Test plan
- [ ] Vitest suites (Tables, CSV) assert every schema path resolves against real run() output
- [ ] Typecheck + lint clean on all six touched pieces
- [ ] Test Step each touched action/trigger in the builder and confirm the Friendly View renders the schema instead of raw JSON

Refs PIE-376


ibrahim-abuznaid:
Hey Odai — went through this one. Two things to fix, then smaller items:

**1. Tables: `children` under the `dynamicKey` `cells` field feeds agents wrong paths**

Both UI consumers ignore `children` when `dynamicKey` is set — the viewer excludes it explicitly (`output-field-row.tsx:34`, `isDescribedObject = !isDynamicMap && ...`) and the data selector drills the live value generically (`utils-schema.ts:135`). The only consumer that actually reads it is the MCP flattener (`mcp-utils.ts`), which recurses into `children` before checking `dynamicKey`, so `ap_get_piece_props` serves agents:

```
cells.fieldName
cells.value
cells.created (datetime)
cells.updated (datetime)
```

None of these resolve — the real paths are `cells['<fieldId>'].fieldName` — and the `(dynamic key)` hint is never emitted because the children branch returns early. Dropping `children` from the field (keep `dynamicKey`, `labelKey`, `description`) loses nothing in the builder and makes MCP emit `cells (dynamic key)` correctly. It's the shared `recordFields`, so this affects 4 actions + 3 triggers.

**2. Delay Until: the past-timestamp throw is a product change, not a schema change**

The parse fix is right — `NaN` silently becoming a 1 ms wait with `success: true` was a real bug, keep it. But "past timestamp resumes immediately" was intended behavior: the old code's own `// resume immediately` branch, and the action's aiMetadata documented it. "Delay until 9am" running at 9:05 used to proceed; now it fails the run. Existing flows are safe (steps are pinned to exact versions since the v19 wildcard-strip migration), but every new flow gets the throwing behavior under a patch bump — and the piece-versioning doc calls changed behavior a major. Can you split the past-timestamp change into its own PR so it gets a deliberate decision? This one keeps the parse fix + schemas.

**3. Smaller things**

- Tables jumps 0.4.0 -> 0.4.2; npm's latest is 0.4.0, so 0.4.1 is free — same as the Pinterest version skip last time.
- The PR body dropped the template's "Breaking change?" / "Security impact?" checkbox sections — please restore them. The PR also needs labels: `🌟 feature` + `🧩 area/core-pieces`.
- Chat trigger: `files` has `format: 'url'` on an array field — same situation as the Pinterest thumbnail array you left format-less; formats apply to leaf values, so either describe the items or drop the format.

The skip decisions (multi-select date actions, form_submission, Send HTTP Request, CSV-to-JSON) all check out against the renderer's real behavior — the "empty"-row rationale matches `output-table-view.tsx`. The path-resolution test pattern is a genuinely good idea; I'd like to make it the standard for schema PRs.

ibrahim-abuznaid (approved):

AhmadTash (approved):

PR #14657, opened 2026-08-06
fix(pieces): disabled piece no longer appears in the AI & Agents builder tab

### LINKED ISSUE #14655: [BUG] Disabled piece in a Piece Set still shows in the AI & Agents builder tab
## Problem

Disabling a piece in a Piece Set does not hide it from the flow builder's AI & Agents tab. The list endpoint respects the policy (piece disappears from Apps/Explore), but the AI tab fetches `@activepieces/piece-ai` by direct name and renders all its actions.

Ref: Pylon 5579

**Update (escalated):** the cascade is worse than the initial report — with the AI piece disabled in the piece set, the Ask AI step is still addable from the AI tab; once added, its dynamic Provider/Model dropdowns error `Unexpected error, please retry` at option-resolve time. The step lands in the flow poisoned. Customer note: default selectors don't reflect the hidden piece; workaround is to clear that default. The fix in this ticket closes the primary path (picker no longer offers the piece); the option-resolve failure on already-added steps and default-value leakage may need follow-ups.

## Root cause

Direct-by-name fetch bypasses the piece-set filter because `projectId` never reaches the visibility policy:

1. `packages/web/src/app/builder/pieces-selector/ai-tab-content/index.tsx:19` — `usePiece({ name: '@activepieces/piece-ai' })` passes no `projectId`.
2. `packages/web/src/features/pieces/hooks/pieces-hooks.ts:92-106` — `usePiece` doesn't accept or forward `projectId` to `piecesApi.get`. `usePieces` at line 178-192 does, so the two hooks are inconsistent.
3. `packages/server/api/src/app/pieces/metadata/piece-metadata-controller.ts:100` reads `req.query.projectId` (undefined), then `packages/server/api/src/app/ee/pieces/filters/piece-filtering-utils.ts:15-17` returns `null` and no policy is applied.

Latent hole: `packages/server/api/src/app/pieces/metadata/piece-metadata-service.ts:80-81` — `getOrThrow` destructures `{ version, name, platformId, locale }` and calls `this.get({ version, name, platformId })`, silently dropping `projectId` even when the caller supplies it.

## Fix

1. `usePiece` accepts `projectId` and forwards it to `piecesApi.get` (mirror `usePieces` at line 187).
2. `AITabContent` passes `authenticationSession.getProjectId()` to `usePiece`.
3. `getOrThrow` forwards `projectId` to `this.get`.

The "AI piece is not available for this platform" error state (from activepieces/activepieces#13767) already renders when `pieceModel` is undefined, so no new UI needed.

majewskibartosz:
Fixes [GIT-1719](https://linear.app/activepieces/issue/GIT-1719)
Closes #14655

### Breaking change?

### Security impact?

## Problem
Disabling a piece in a Piece Set (e.g. AI) still let the piece show up in the flow builder's AI & Agents tab. The `/v1/pieces` list respected the piece-set policy (piece hidden from Apps/Explore), but the AI tab fetches `@activepieces/piece-ai` by direct name and rendered all its actions. Adding the step from there produced dropdowns that errored `Unexpected error, please retry` at option-resolve time, poisoning the flow.

## Fix
- `packages/web/src/features/pieces/hooks/pieces-hooks.ts`: `usePiece` accepts optional `projectId` and forwards it to `piecesApi.get` (opt-in per caller; matches `usePieces` at line 187). React Query cache-key includes `projectId` so opt-in callers don't share cache with the unfiltered path.
- `packages/web/src/app/builder/pieces-selector/ai-tab-content/index.tsx`: passes `authenticationSession.getProjectId()` to `usePiece`.
- `packages/server/api/src/app/pieces/metadata/piece-metadata-controller.ts`: `GET /v1/pieces/:name` and `/:scope/:name` now throw `ENTITY_NOT_FOUND` (404) when the piece is invisible under the resolved policy, instead of returning the piece with only its components filtered. Existing callers that don't pass `projectId` skip the policy and get the piece back (unchanged), so existing-step lookups on running flows are untouched.

## Verification
- `npx vitest run test/integration/ee/pieces/piece-component-filtering.test.ts` — 16/16 pass, including 3 new direct-fetch cases.
- Red-first: without the controller change, `direct fetch returns 404 for a hidden piece when projectId is passed` fails with `expected 200 to be 404`; with the change all pass.
- `npx turbo run lint --filter=api --filter=web` — 0 errors (existing warnings only).
- `npx tsc --noEmit` on both packages — clean.
- Visual: existing "AI piece is not available for this platform" fallback state (shipped in GIT-1570) — no new UI. This change only rewires when that state is reachable; the rendered tab in the disabled case is identical to what already renders when the server returns 404. Live drive skipped: EE dev stack for piece-sets was booted but not run through the UI due to time (the stack was torn down after this check).
- Other affected paths: checked - none. Grep of `piecesApi.get` and `usePiece` callers: admin `piece-component-visibility-sheet.tsx` (must show all pieces to admins - stays unfiltered), `usePieceModelForStepSettings` and other existing-step lookups (must resolve existing steps regardless of piece-set - stay unfiltered). Only AITabContent opts in.

<details>

## Plan
- Frontend hook made project-aware via opt-in `projectId` param; only the picker surface (AITabContent) passes it. Existing usePiece callers keep the current unfiltered behavior — critical for hooks that read pieces already present in a flow (step editors, connection reconnect, run details) after a piece was later hidden.
- Server enforces rejection at the controller (not the service) so `pieceMetadataService.getOrThrow` continues to serve internal callers (app-connection, oauth2, flow-version validator, migrations) that must resolve a piece regardless of piece-set visibility.
- Footprint: 3 product files (~15 lines) + 1 test file. Rejected alternative: forwarding `projectId` inside `pieceMetadataService.getOrThrow` — cascades filtering into 28+ internal callers and breaks existing connections/flows.

- User has no active project (login/switching): `authenticationSession.getProjectId()` returns null → `usePiece` sees `projectId: undefined` → controller resolves policy = null → returns piece (fail-open by design; no project means no piece-set to enforce yet).
- Piece disabled AFTER a flow already uses it: existing-step lookups do not pass `projectId` → still resolve. This ticket does not cover already-added steps whose option resolvers cascade into `Unexpected error`; that is a separate ticket (default-value / already-placed-step leak).
- Piece-set uses EXCLUDE_ALL mode with the piece listed: `isPieceVisible` returns true → piece stays visible, as intended by the config.

## Alternatives rejected
- Filter inside `getOrThrow` — would cascade into every internal server caller of the piece metadata service (28+ call sites: oauth2 refresh, connection creation, flow validation, migrations). Would break existing connections and running flows.
- Wire `projectId` implicitly inside `usePiece` (auto-pick from `authenticationSession`) — silently changes behavior for every existing caller; would hide pieces from step editors and connection dialogs on flows that reference already-hidden pieces.
- New `AITabContent` fallback UI — unnecessary; GIT-1570 already ships the "AI piece is not available for this platform" state, reached via `isError || isNil(pieceModel)`.
</details>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4301,7 +4301,7 @@
     },
     "packages/pieces/community/hubspot": {
       "name": "@activepieces/piece-hubspot",
-      "version": "0.8.8",
+      "version": "0.8.9",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -4605,7 +4605,7 @@
     },
     "packages/pieces/community/jira-cloud": {
       "name": "@activepieces/piece-jira-cloud",
-      "version": "0.3.9",
+      "version": "0.3.10",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -4624,7 +4624,7 @@
     },
     "packages/pieces/community/jira-data-center": {
       "name": "@activepieces/piece-jira-data-center",
-      "version": "0.0.5",
+      "version": "0.0.6",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -8125,7 +8125,7 @@
     },
     "packages/pieces/community/sign-now": {
       "name": "@activepieces/piece-sign-now",
-      "version": "0.0.5",
+      "version": "0.0.6",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
--- a/packages/server/api/src/app/pieces/metadata/piece-metadata-controller.ts
+++ b/packages/server/api/src/app/pieces/metadata/piece-metadata-controller.ts
@@ -78,7 +78,7 @@ const basePiecesController: FastifyPluginAsyncZod = async (app) => {
                 locale: req.query.locale as LocalesEnum | undefined,
             })
             const policy = await resolveVisibility({ platformId, projectId: req.query.projectId, log: req.log })
-            const visiblePiece = isNil(policy) ? piece : policy.filterPieceComponents(piece)
+            const visiblePiece = applyVisibilityPolicy({ policy, piece })
             return filterModelActionsByAudience(visiblePiece, req.query.audience)
         },
     )
@@ -98,7 +98,7 @@ const basePiecesController: FastifyPluginAsyncZod = async (app) => {
                 locale: req.query.locale as LocalesEnum | undefined,
             })
             const policy = await resolveVisibility({ platformId, projectId: req.query.projectId, log: req.log })
-            const visiblePiece = isNil(policy) ? piece : policy.filterPieceComponents(piece)
+            const visiblePiece = applyVisibilityPolicy({ policy, piece })
             return filterModelActionsByAudience(visiblePiece, req.query.audience)
         },
     )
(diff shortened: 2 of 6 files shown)


othmanemad:
Expanded the fix to hide the tab if the piece or all of it's actions are hidden

othmanemad (approved):

PR #14658, opened 2026-08-06
feat(connections): connection-to-piece binding enforcement

KhaledR57:
## Summary
- New env var `AP_ENFORCE_CONNECTION_PIECE_BINDING` (default `false`, CE+EE) locks a connection to the piece that created it.
- Pieces with no fixed destination (HTTP, GraphQL, Webhook, self-hosted/DB-style pieces) are blocked from connection access outright.
- Code/Loop/Router steps have no piece identity, so they lose all connection access under this flag. "Read Connection" gets no exemption and goes non-functional.

## Test plan
- [x] Engine unit tests: `connection-resolver.test.ts`, `props-resolver.test.ts` (query param threading, error mapping, both message branches)
- [x] Integration test: `engine-services.test.ts` — flag off/on, match/mismatch, omitted identity, generic-destination block, Read Connection no-exemption
- [x] `npm run lint-dev` clean (no errors from this change)

## Breaking change?
- [x] No — default `false`, zero-setup, existing behavior unchanged unless an admin opts in.


packages/core/execution/src/lib/workers/index.ts
```diff
@@ -83,6 +83,7 @@ export const ConsumeJobRequest = z.object({
     jobData: JobData,
     attempsStarted: z.number(),
     engineToken: z.string(),
+    internalEngineToken: z.string().optional(),
```
abuaboud, line 86: Why another one?

packages/server/api/src/app/app-connection/generic-destination-pieces.ts
```diff
@@ -0,0 +1,19 @@
+export const GENERIC_DESTINATION_PIECE_NAMES: ReadonlySet<string> = new Set([
```
abuaboud, line 1: Why some pieces different?

packages/server/engine/src/lib/handler/context/engine-constants.ts
```diff
@@ -232,6 +238,7 @@ type SharedFieldsSource = {
     publicApiUrl: string
     internalApiUrl: string
     engineToken: string
+    internalEngineToken?: string
```
abuaboud, line 241: Duplicate?

abuaboud (changes requested):

packages/core/execution/src/lib/workers/index.ts
KhaledR57: Job runs multiple pieces per job, so identity can't live in the job level `engineToken`. `internalEngineToken` is the second signal that proves "this call came from engine code, not from a piece using its own `context.server.token`", it's never exposed to piece code. Will fix one thing: gate the mint behind `AP_ENFORCE_CONNECTION_PIECE_BINDING` so it's not generated when the flag's off.

packages/server/api/src/app/app-connection/generic-destination-pieces.ts
KhaledR57: These pieces (HTTP, GraphQL, Webhook, DB connectors) have no fixed backend, the flow config decides the real destination. Binding buys no security there, so we block them outright. Checked `PieceCategory.CORE` as a possible substitute, it's over inclusive (tags store, delay, crypto, etc), so no existing metadata fits. Hardcoded list is intentional.

packages/server/engine/src/lib/handler/context/engine-constants.ts
KhaledR57: (`SharedFieldsSource` vs `EngineConstantsParams`, predates this PR). my change just adds one field following that pattern.

PR #14685, opened 2026-08-09
feat(auth): move the OTP primitive into core and give it an attempt budget

AhmadTash:
## Description

**1 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

Email sign-in needs a login code, and the `otp` table that already backs password reset and email verification is the natural place to put one. That means Community has to be able to serve it, so the module moves out of `ee/` into the core authentication folder. Reasoning recorded in `brain/decisions/000016-email-sign-in-is-a-typed-code-on-the-existing-otp-primitive.md`.

What lands here:

- **`EMAIL_LOGIN` OTP type** and its email template.
- **`attempts` column**, so a code can be guessed five times and is then discarded. Counted with a single `UPDATE ... RETURNING`, not a read-modify-write: concurrent guesses cannot race, and a wrong guess does not refresh `updated` and quietly extend the credential's life.
- **Consumption is the `DELETE` itself.** Only the request whose delete actually removed the row is told the code was good, so two overlapping confirmations cannot both be honoured.
- **`POST /v1/otp` no longer accepts `EMAIL_LOGIN`**, so the public endpoint cannot mint a login code for an address the caller does not control.
- **Name derivation helpers** — turning an email or a typed full name into first/last and a platform name. Used by the next PR.

Nothing calls the new type yet; the endpoints arrive in PR 2.

## How was this tested?

`signup-names.test.ts` (11 unit cases). Full `api` suite green on the stack: CE 746, EE 126, Cloud 500. `check-migration-rollback` and `check-release-migrations 0.87.0` both pass.


AhmadTash:
@greptile both fixed.

**Attempt exhaustion racing consumption — you were right, and the fix needed to be deeper than the delete.** Guarding only the delete would still have let parallelism outrun the budget: every concurrent request read the row before any increment landed, so N simultaneous guesses each got evaluated against `attempts` values that were all stale. Claiming an attempt and reading the row are now one statement:

```sql
UPDATE "otp" SET "attempts" = "attempts" + 1
WHERE "identityId" = $1 AND "type" = $2 AND "attempts" < $3
RETURNING "id", "value", "state", "updated", "attempts"
```

Row-level locking serialises concurrent updates on that row, so each request gets a distinct attempt number and the sixth finds nothing to claim. A correct code presented after the budget is spent no longer matches the `WHERE`, which is the hole you described. Covered by a test verified to fail without it (`expected true to be false`).

I dropped a second test I had written alongside it — a 20-way parallel guess storm — because it passed with and without the fix. A concurrency test that has not been shown to fail without the fix is not evidence, and sequential exhaustion is already covered in the passwordless PR.

**Recurring reconciliation — this one was mine, and it was a real regression.** `scheduleRecurringReconcile()` is restored.

Worth recording how it got there, because it is a hazard for any stacked split: this PR was carved out of #14653 by checking whole files out of the feature branch onto a newer `main`. `app.ts` was a file **both** sides had touched, so taking my version wholesale silently reverted a line `main` had added in the intervening 10 commits. Nothing conflicted and nothing failed — the line just quietly disappeared.

I have since diffed every one of the 64 files this split touches against current `main` for the same pattern. Two were affected: `app.ts` (this) and `bun.lock`, which was regenerated from `main`’s baseline rather than carried over. The rest are clean.

AhmadTash:
@greptile pushing back on the re-raised P1 — I believe it is a false positive on the current code, and I had it independently verified two ways before saying so.

**The budget gate is the `WHERE` clause, not the row's existence.** `spendAttempt` is a single `UPDATE "otp" SET "attempts" = "attempts" + 1 WHERE "identityId" = $1 AND "type" = $2 AND "attempts" < $3 RETURNING ...`. The unique index on `(identityId, type)` means it targets exactly one tuple, so two concurrent spends serialise on the row lock; the loser's EvalPlanQual re-fetch re-evaluates `attempts < 5` against the committed value. Successful spends over the row's lifetime therefore return exactly 1,2,3,4,5 — never a sixth.

The interleaving you describe resolves the other way:

- B1..B5 take slots 1-5. A's `UPDATE` then matches nothing, `spendAttempt` returns null, and `confirm` returns false at the `isNil` guard **without ever comparing the code**. It never reaches `consumeOtp`, so the timing of B5's cleanup delete is irrelevant.
- The case where A does succeed is: B1..B4 take slots 1-4, A takes slot 5 with the correct code. That is a correct code occupying one of five budgeted evaluations — the specification, not a bypass.

`consumeOtp` genuinely does not re-check `attempts`, and that is deliberate: adding the check would reject a correct code that legitimately claimed slot 5, silently turning a five-guess budget into four.

Two things that came out of verifying this, both worth having on the record:

- The named impact is not reachable regardless. `otp-generator.ts` emits a 6-digit code **only** for `EMAIL_LOGIN`; `EMAIL_VERIFICATION` and `PASSWORD_RESET` use `randomUUID()` — 122 bits. Brute force is not the threat model for the two flows your finding names.
- One verifier ran the exact statement under 480 concurrent executions against Postgres: exactly 5 rows returned, `attempts` 1..5, zero duplicates.

The one genuine race left is fail-closed: if A holds slot 4 with the right code while B5 exhausts the budget and deletes the row first, A's `DELETE ... RETURNING` removes nothing and `confirm` returns **false** for a correct code. A false negative, never a false accept.

Also checked while I was in there, since both would have been real findings had they gone the other way: `createAndSend` re-sends the *same* code for a pending non-expired OTP, so resend cannot reset the budget; and `updated` is a TypeORM `updateDate` set in JS rather than by a trigger, so the raw `UPDATE` does not bump it and wrong guesses cannot extend the 10-minute expiry.

Happy to be shown an interleaving I have missed — but it needs to be one where a confirmation returns true after six evaluations, and I do not think one can be constructed against this statement.

AhmadTash:
@greptile review

AhmadTash:
@greptile review

AhmadTash:
Rebased onto current `main` (bafa44d5c6) — conflicts resolved, and one of them needed more than a merge.

**`packages/core/shared/package.json`** — `main` moved to 0.128.3 with patch bumps while this branch carried 0.129.0. Kept 0.129.0: still ahead, and a minor bump is what new exports call for.

**`postgres-connection.ts`** — not just a merge. `main` has taken slot `1820000000000` for `AddAuditEventPlatformIdCreatedIdIndex`, which is where this migration sat, so keeping both entries would have left two migrations sharing a timestamp and an undefined order between them. Renumbered to **`1821000000000-AddAttemptsToOtp`** — file, class, import and array entry — so it sorts after everything on `main`. `release` stays `0.87.0`, which still matches the root version.

This is the second time this migration has had to move (1819 → 1820 → 1821); each time `main` claimed the slot first. Worth knowing if the stack sits unmerged much longer.

**`bun.lock`** was regenerated from `main`'s baseline rather than hand-merged, so the diff is 4 lines: `input-otp` plus the shared version. The last time I merged a lockfile by hand it silently upgraded a dozen `@ai-sdk/*` packages.

Both migration gates pass (`check-migration-rollback`, `check-release-migrations 0.87.0`). All nine PRs in the stack are back to `MERGEABLE`, and the suites are green on the rebased tip: CE auth+platform 43, EE 126, Cloud 500, build/lint/typecheck 21/21.

@greptile review

packages/server/api/src/app/authentication/otp/otp-controller.ts
MrChaker, line 10: @AhmadTash This endpoint should be rate limited no ?

packages/server/api/src/app/authentication/otp/otp-service.ts
```diff
@@ -0,0 +1,121 @@
...
+        await repo().upsert(newOtp, ['identityId', 'type'])
+        await emailService(log).sendOtp({
+            platformId,
+            userIdentity,
+            otp: newOtp.value,
+            type: newOtp.type,
+        })
+    },
+
+    // Serialised per credential, because the budget and the single-use guarantee
```
MrChaker: @AhmadTash remove comment

packages/server/api/test/integration/ce/authentication/otp-service.test.ts
```diff
@@ -0,0 +1,68 @@
...
+afterAll(async () => {
+    await teardownTestEnvironment()
+})
+
+beforeEach(async () => {
+    await databaseConnection().getRepository('otp').createQueryBuilder().delete().execute()
+    await databaseConnection().getRepository('user_identity').createQueryBuilder().delete().execute()
+})
+
+describe('otpService#confirm attempt budget', () => {
```
MrChaker: @AhmadTash why this has only 1 test ?

MrChaker (changes requested):

MrChaker (approved):

othmanemad (approved):

MrChaker (approved):

PR #14686, opened 2026-08-09

AhmadTash:
## Description

**2 of 6 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one above it</summary>

| | PR | Area |
|---|---|---|
| 2 | #14686 — sign in with an emailed code | server/api |
| 3 | #14687 — first-platform creation single-use and self-healing | server/api |
| 4 | #14688 — building blocks for the auth screen | web |
| 5 | #14689 — one screen for sign in and sign up | web |
| 6 | #14690 — ask new members their name in the card | web |

Split out of #14653, which exceeded the review-size budget. The tip of this stack is byte-identical to that branch across all 64 files it touched.

</details>

## How was this tested?


AhmadTash:
Closing and reopening against the correct base. This branch now sits on top of the new name-derivation layer, but GitHub refuses to retarget a PR that is part of a detected stack (`Cannot change the base branch because the pull request is part of a stack`), via both REST and GraphQL — so the base here is stale and its diff double-counts the layer below.

Replacement: see the new PR linked below. Your three findings are answered there.

PR #14687, opened 2026-08-09
fix(auth): make first-platform creation single-use and self-healing

AhmadTash:
## Description

**4 of 7 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|
| 3 | #14694 — sign in with an emailed code | server/api |
| 4 | #14687 — first-platform creation single-use and self-healing | server/api |
| 5 | #14688 — building blocks for the auth screen | web |

Split out of #14653, which exceeded the review-size budget. #14686 was an earlier
incarnation of PR 3, closed because GitHub will not retarget a PR inside a stack.

</details>

Creating an identity's first platform was replayable and is not atomic, so one identity could end up with duplicate tenant resources or stuck part-way. Found by review of PR 2, fixed here.

**The guard goes in `createPlatformWithProject`, where every caller passes through.** A lock in the sign-up route alone was not enough: `POST /v1/platforms` accepts the same `ONBOARDING` principal, so a client calling both still provisioned twice.

**Deduplication is opt-in per call site via `isFirstPlatform`, not unconditional** — on cloud an existing signed-in user is *allowed* to create additional platforms (`platform.controller` only refuses that on ce/ee). An unconditional "identity already has a platform, return it" guard would have silently broken multi-platform creation while fixing the replay.

Provisioning writes the user, the platform link and the project separately, and each gap left state a retry read wrongly:

| interrupted after | stale read | before | now |
|---|---|---|---|
| user insert | user has null `platformId` | second user created | unlinked user reused |
| platform save | user still unlinked, platform already names it owner | **500 forever** (unique owner constraint) | platform adopted |
| owner link | platform has no project | **403 forever** (`getProjectAndToken` throws) | project created |

Two of those wedged the identity permanently — every retry hit the same error.

**What this is not:** the three writes are still not atomic. One transaction across `userService`, `platformService` and `projectService`, each owning its own `repoFactory` repository, means threading an `EntityManager` through all three and touching every existing caller of platform creation — wider than this stack should carry. What is here makes retries *converge* from any interruption point. The non-atomic sequence itself predates this work.

## How was this tested?

Five integration cases, **each verified to fail without its fix** (`expected 2 to be 1`, `expected 500 to be 200`, `expected 403 to be 200`, and both routes returning the same `platformId`).

### DIFF
--- a/packages/server/api/src/app/authentication/passwordless-auth.service.ts
+++ b/packages/server/api/src/app/authentication/passwordless-auth.service.ts
@@ -122,6 +122,7 @@ export const passwordlessAuthService = (log: FastifyBaseLogger) => ({
             identityId,
             name: signupNames.platformNameFromPerson({ firstName, email: identity.email }),
             invalidatePreviousTokens: false,
+            isFirstPlatform: true,
         })
     },
 })
--- a/packages/server/api/src/app/database/seeds/dev-seeds.ts
+++ b/packages/server/api/src/app/database/seeds/dev-seeds.ts
@@ -50,6 +50,7 @@ const seedDevUser = async (): Promise<void> => {
         identityId: response.id,
         name: 'dev\'s Platform',
         invalidatePreviousTokens: true,
+        isFirstPlatform: true,
     })
 
     log.info({ email: DEV_EMAIL, password: DEV_PASSWORD }, '[devSeeds#seedDevUser] Dev user and platform created')
--- a/packages/server/api/src/app/platform/platform.controller.ts
+++ b/packages/server/api/src/app/platform/platform.controller.ts
@@ -39,6 +39,7 @@ export const platformController: FastifyPluginAsyncZod = async (app) => {
             identityId,
             name: req.body.name,
             invalidatePreviousTokens: isOnboarding,
+            isFirstPlatform: isOnboarding,
         })
     })
 
--- a/packages/server/api/src/app/platform/platform.service.ts
+++ b/packages/server/api/src/app/platform/platform.service.ts
@@ -282,11 +282,6 @@ function personalProjectName(platformName: string): string {
 }
 
 async function finishExistingPlatform({ user, platformId, name, invalidatePreviousTokens, identityId, log }: FinishExistingPlatformParams): Promise<AuthenticationResponse> {
-    if (invalidatePreviousTokens) {
-        await userIdentityRepository().update(identityId, {
-            tokenVersion: nanoid(),
-        })
-    }
     const hasProjects = await projectService(log).userHasProjects({
         platformId,
         userId: user.id,
@@ -300,6 +295,11 @@ async function finishExistingPlatform({ user, platformId, name, invalidatePrevio
             platformId,
             type: ProjectType.PERSONAL,
         })
+    if (invalidatePreviousTokens) {
+        await userIdentityRepository().update(identityId, {
+            tokenVersion: nanoid(),
+        })
+    }
     return authenticationUtils(log).getProjectAndToken({
         userId: user.id,
         platformId,
--- a/packages/server/api/src/app/platform/platform.service.ts
+++ b/packages/server/api/src/app/platform/platform.service.ts
@@ -84,7 +84,7 @@ export const platformService = (log: FastifyBaseLogger) => ({
                 const existingUsers = isFirstPlatform ? await userService(log).getByIdentityId({ identityId }) : []
                 const linkedUser = existingUsers.find((user) => !isNil(user.platformId))
                 if (!isNil(linkedUser) && !isNil(linkedUser.platformId)) {
-                    return finishExistingPlatform({ user: linkedUser, platformId: linkedUser.platformId, name, invalidatePreviousTokens, identityId, log })
+                    return finishExistingPlatform({ user: linkedUser, platformId: linkedUser.platformId, name, invalidatePreviousTokens: false, identityId, log })
                 }
                 const unlinkedUser = existingUsers.find((user) => isNil(user.platformId))
                 const orphanedPlatform = isNil(unlinkedUser) ? null : await platformRepo().findOneBy({ ownerId: unlinkedUser.id })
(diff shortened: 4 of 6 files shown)


AhmadTash:
@greptile one fixed, one a stated limitation.

**Token rotation bypass — correct, fixed.** `finishExistingPlatform` returned a user token straight from the existing platform and skipped the rotation the caller had asked for, so an ONBOARDING principal retrying `POST /v1/platforms` kept a live onboarding JWT after a user token had been issued. Both recovery paths now take `invalidatePreviousTokens` and rotate before minting, exactly as the provisioning path does.

Note the sign-up route deliberately still passes `false`, so a double-submitted name step returns the same session rather than a 401. The duplicate-tenant risk is closed at the database level by the guard in this PR, and that is the protection that matters; rotating there would turn an ordinary double-click into a failed request without removing a capability an attacker has. The controller path asks for rotation and now gets it in every branch.

**Non-transactional writes — acknowledged, and deliberately not fixed here.** `userService`, `platformService` and `projectService` each own their own `repoFactory` repository, so one transaction means threading an `EntityManager` through all three and touching every existing caller of platform creation. That is a refactor with its own blast radius, not a line in a sign-in stack, and the non-atomic sequence predates this work.

What this PR does instead is make every interruption point recoverable, which is the property that protects the user: a retry after any of the three gaps now converges rather than duplicating, 500-ing or 403-ing forever. I would rather ship convergent retries now and raise the transaction separately than leave three permanent wedges in place waiting for a bigger change. Say the word if you want it tracked as an issue.

AhmadTash:
@greptile confirmed and fixed — and this one was my regression, introduced by the rotation fix you asked for in the previous round.

Rotating first meant `finishExistingPlatform` bumped `tokenVersion` and *then* did the fallible work: `userHasProjects`, the project repair, and finally `getProjectAndToken`. A failure in the repair left the caller holding a dead onboarding token, so the retry answered `SESSION_EXPIRED` and the member had to start again from the email — destroying precisely the convergence property this PR exists to provide. Worse, it broke it on the recovery path, which is only ever reached because something already went wrong.

Rotation now happens after the repair and immediately before the replacement token is minted, matching the order the provisioning path already used. The residual window is the mint itself, which is what the provisioning path has too.

Test added and verified to fail without the fix (`expected '7DXOdf…' not to be '7DXOdf…'`): a successful recovery through `POST /v1/platforms` must still rotate. That is the regression risk of moving the call — that rotation gets dropped rather than reordered — so it is the property worth pinning.

I did try to test the failure direction directly and could not do it honestly: forcing a deterministic failure between rotation and mint needs an injected fault, and my first attempt (an over-long platform name) did not fail because the repair uses the *request* name, not the stored one. Rather than ship a test that passes for the wrong reason, I dropped it. The reorder itself is a five-line relocation and reads as obviously correct.

Non-atomic recovery writes: position unchanged, and you accepted it above — tracked as a follow-up rather than done here.

AhmadTash:
@greptile review

AhmadTash:
@greptile confirmed and fixed — my rotation fix from the previous round caused this, and the two findings pointed opposite ways, so I split the behaviour by what the call actually does rather than flip between them.

Rotating on *every* recovery meant a second call that provisioned nothing still invalidated the session the first call had already returned, so that caller's next request died with `SESSION_EXPIRED`.

Rotation now follows the work:

- **provisioning** rotates — it completes the transition the onboarding credential authorised;
- **adopting a platform whose owner link never landed** rotates, for the same reason: that call is the one finishing an interrupted provision;
- **returning a session for an already fully linked platform** provisions nothing, so it leaves `tokenVersion` alone.

The security property you asked for in the earlier round survives: whichever call completes the transition retires the onboarding token. What is gone is the destructive side effect on an idempotent path.

Verified with a test that fails without it (`expected 'gtHNZF…' to be 'rrCdZp…'`): two concurrent `createPlatformWithProject` calls, then assert every issued token's `tokenVersion` claim still matches the identity's current value.

Worth flagging how long that took to pin, because it shaped what I trust: my first two attempts asserted through `app.inject` and **passed with and without the fix** — inject serialises the requests, so the second one 401s at the door and never reaches the recovery path at all. Only calling the service directly reproduces the interleaving. Both earlier versions were deleted rather than kept as decoration.

AhmadTash:
@greptile review

AhmadTash:
@greptile stopping here, and saying why rather than making a fourth change.

The current summary asks for three things that cannot all hold with one `tokenVersion` field:

1. the linked-user recovery branch **must** rotate, so a still-valid onboarding token gets retired;
2. recovery **must not** rotate, because a duplicate call would strand the session an earlier call returned (your finding from the previous round, which I fixed);
3. rotation **must not** precede token issuance, because a failure in the mint would leave the caller with no usable credential.

(1) and (2) are direct opposites on the same branch. `tokenVersion` is identity-scoped, so "retire the onboarding credential" and "invalidate every session for this identity" are the same operation — satisfying both at once is not available in this design.

I have now rewritten this rotation three times across three rounds, each time to satisfy the newest finding, and each rewrite has been correct about the finding it answered and wrong about the one before it. That oscillation is worse than either endpoint, so here is the reasoning I am settling on and the trade I am making:

**Rotation follows the work.** Provisioning rotates. Adoption of a half-created platform rotates — that call completes the transition. Returning a session for an already-linked platform does not, because it provisioned nothing.

**On (1) specifically:** yes, a retry that lands on the linked-user branch after a pre-rotation failure leaves the onboarding token valid until its own expiry. That grants no capability to anyone who did not already hold the token — it is the client's own bearer credential, and the normal path *does* retire it. Against that, (2) breaks an honest double-click by killing the session the first call legitimately issued, mid-signup. I weigh a stranded real user above a token that stays valid slightly longer in the hands of whoever already had it.

**On (3):** unavoidable without a transaction. The window is `getProjectAndToken` — two reads and a JWT signing — and the provisioning path has always had exactly the same shape.

**The design that would satisfy all three** is a single-use onboarding credential of its own — consumed on use — rather than overloading `tokenVersion`, which cannot distinguish "retire this one token" from "invalidate this identity everywhere". That is a larger change than this stack should carry, and it is the right follow-up. Adding it to the list alongside the provisioning transaction, the prune-on-expiry job, and the two missing tests.

Each behaviour above is pinned by a test verified to fail without it, including the concurrent one, which took three attempts to write honestly — the first two asserted through `app.inject` and passed with and without the fix, because inject serialises the requests and the second never reaches the recovery path.

AhmadTash:
@greptileai update the Confidence Score

AhmadTash:
Closing in favour of #14702. Same code, reordered.

The provisioning guard now lands **before** the passwordless feature that relies on it, rather than after. That was the right call and it was not mine — it came out of the question "why are these two not one PR?", and the honest answer was that they should not be one PR, they were simply in the wrong order.

The old order made this PR read as shipping a live replay bug that a later PR quietly fixed, which is exactly why the reusable-onboarding-token finding kept coming back here: reviewed alone, the mitigation genuinely was not present. `createPlatformWithProject` already exists on `main` with callers in `platform.controller` and `dev-seeds`, so the guard never depended on passwordless at all — only its tests did, and those now mint an ONBOARDING principal directly through `getOnboardingResponse` instead of borrowing the passwordless endpoints.

Both halves stay inside the budget in the new order (273 and 564 of 600), so nothing needed to be dropped or bypassed. The reasoning agreed in this thread carries over to the replacement; the tip of the stack is byte-identical to what it was before the reorder, apart from the two test files being re-split along the new seam.

PR #14689, opened 2026-08-09
feat(web): one screen for sign in and sign up

AhmadTash:
## Description

**7 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

Sign-in and sign-up stop being two pages with two forms. The visitor lands on one card, types an email, and gets a code; whether they already have an account is something we work out, not something they have to declare up front.

- **Google and SSO sit above the email field**, because for most people they are the shorter path.
- **No password unless the deployment has no mail configured** — the card falls back to the password form when SMTP is absent, so a self-hoster needs no new setup.
- **The card sits on a facsimile of the product** rather than a logo wall, which is what the sign-in page is actually selling.
- Trust logos are replaced by terms and privacy links.

**A brand-new identity verifies its code and continues to the existing `/create-platform` route.** The inline name step that replaces that hop is PR 6 — keeping it separate is what holds this diff inside the review-size budget, and the fallback is the route that already serves onboarding principals today.

`AuthFormTemplate` is removed here; `AuthLayout` stays. That is why the two auth routes and `sign-in-form` move in this PR rather than the previous one — `SignInForm` gains a required `onForgotPassword` prop, so it cannot land before its callers.

## How was this tested?

`turbo build lint typecheck --filter=web` clean. Browser-verified on a Cloud-edition instance with real SMTP and Google sign-in configured.

**Not yet exercised on CE or EE**, which is part of why the stack is in draft. CE is the path that matters most: the OTP module moving out of `ee/` in PR 1 is what makes email codes reachable there at all.


AhmadTash:
@greptile the first one is a real bug and a serious one — fixed. The second is a deliberate seam in this PR.

**Fresh no-SMTP install could never create its first user.** Confirmed exactly as you traced it: `/sign-up` now redirects to `/sign-in`, the card falls back to the password form with no SMTP, and `SignInForm` redirected to `/sign-up` whenever `USER_CREATED` was absent. Straight loop, and the first user on a self-hosted install is precisely the person who hits it.

Two changes:

- the `!userCreated → /sign-up` redirect is gone. It existed to push first-time visitors at a separate sign-up page; that page no longer exists, so the redirect could only ever bounce.
- the card now opens on the **create-account** form when no user exists, and hides the mode switch there. Falling back to a sign-in form on an installation with no accounts was wrong even without the loop — there is nothing to sign in to.

This is the self-hosting rule biting exactly where the handbook says it does: developed against a cloud instance where SMTP and a user already exist, so the first-run path stayed invisible. Good catch.

Honest limitation on the fix: it is **code-verified, not browser-verified**. I tried to stand up a fresh CE instance with no SMTP to drive it end to end and did not get one running, so I have not watched the first user register with my own eyes. Worth a CE smoke test before this merges rather than taking my reasoning for it.

**Onboarding token not resuming on `/sign-in` — intended here, resolved in #14690.** `signedIn` deliberately excludes onboarding tokens, because treating them as signed in is what caused the redirect loop this stack fixed earlier: `AuthLanding` bailed, `DefaultRoute` sent the pre-platform session to `/create-platform`, and that bounced back to `/`. In this PR a new identity continues to `/create-platform` after verifying, so the setup does resume — through the route rather than the card. #14690 replaces that hop with the in-card name step and restores resume-in-place via `authenticationSession.isOnboarding()`.

AhmadTash:
@greptile the loop fix is confirmed on your side, and the second point is the intended seam for this PR.

`signedIn` excludes onboarding tokens on purpose: treating them as signed in is what produced the redirect loop this stack fixed earlier — `AuthLanding` bailed, `DefaultRoute` sent the pre-platform session to `/create-platform`, and that bounced back to `/`.

In this PR a new identity does resume: verification continues to `/create-platform`, through the route rather than the card. #14690 replaces that hop with the in-card name step and restores resume-in-place via `authenticationSession.isOnboarding()`, so the behaviour you want lands one PR up.

Still standing by the caveat from my last comment: the first-run CE path is code-verified only. A "first user, no SMTP, registers successfully" smoke test is the right gate before merge.

AhmadTash:
@greptile review

AhmadTash:
@greptile the first-user comparison was wrong and is fixed. The other point I am holding, with reasoning.

**`userCreated === false` never matched.** You are right about the mechanism, and it is worse than a style nit: `flagService.getAll` reads rows with `findBy({ id: In([...]) })`, and `USER_CREATED` has no row until `flagService.save` writes it when the first account is created. A fresh install therefore **omits** the flag rather than publishing `false`, so my check could not fire on the one deployment it exists for. I confirmed the mechanism against a database with users — the row is there, and it is there only because they are.

Now `userCreated !== true`. Flags come through a suspense query, so `undefined` here means missing rather than still loading, and there is no flash of the wrong mode on an established instance.

To be exact about what was broken, since the finding says the first account cannot be created: the **loop** is gone — `SignInForm` no longer renders `<Navigate to="/sign-up" />`, so the two routes cannot bounce. What survived was that a fresh install opened on the sign-in form, with the mode switch as the only way across to sign-up. Usable, but wrong, and it defeated the fix I had claimed.

This is the second time this specific path has been wrong, and both times it was invisible to me because my environment has users and SMTP. The **CE first-run smoke test remains the outstanding item on this stack** — "fresh install, no SMTP, first user registers" — and I would not merge the web layers without it.

**On onboarding sessions not resuming from `/sign-in`: intentional here, resolved in #14690.** `signedIn` excludes onboarding tokens because treating them as signed in is what caused the redirect loop this stack fixed earlier — `AuthLanding` bailed, `DefaultRoute` sent the pre-platform session to `/create-platform`, and that bounced back to `/`. In this PR a new identity does resume, through the `/create-platform` route rather than in the card. #14690 replaces that hop with the in-card name step and restores resume-in-place via `authenticationSession.isOnboarding()`. Holding the seam rather than reintroducing the loop one PR early.

@greptile review

AhmadTash:
@greptile the remaining P1 does not hold — the resume you are looking for exists, it just is not a redirect.

`auth-drawer-body.tsx:84` seeds the step from the token:

```ts
const [step, setStep] = useState<Step>(
  authenticationSession.isOnboarding() ? "name" : "method",
);
```

So a visitor holding an onboarding token who reloads `/sign-in` opens **on the name step**, not on a fresh card, and does not authenticate again. `auth-drawer-body.tsx:229` renders that step ahead of the `!passwordlessAvailable` branch at 243, so it survives a deployment with no SMTP too — that ordering is the fix for the earlier first-user report on this same file.

Redirecting to `/create-platform` from `AuthLanding` is the one thing that must **not** happen here, which is what the comment at `auth-landing.tsx:21-26` records: `redirectAfterLogin()` sends a platform-less session to the default route, and `default-route.tsx:21` sends it straight back to the card. That is the loop. The redirect belongs on the protected routes, and it is already there:

- `default-route.tsx:21` — onboarding token → `/create-platform`
- `allow-logged-in-user-only-guard.tsx:26` — same, for any guarded route

So an onboarding token resumes on the name step on the auth screen, and resumes at `/create-platform` anywhere behind the guard. Both paths are covered; neither goes through `signedIn`.

Please re-score — the 4/5 rationale is only this item.

AhmadTash:
@greptile you are right and my last comment was wrong — I quoted `auth-drawer-body.tsx:84` from the aggregate branch I test the whole stack on, not from this PR head. At `feat/unified-auth-screen` the line really is `useState<Step>('method')`. The onboarding seeding arrives one layer up in #14690. Sorry for the bad citation.

The finding still is not a blocker, for a different reason: **this PR preserves the behaviour of `main` on that exact path, and the layer above improves it.**

On `main` today, `/sign-in` is `AuthFormTemplate` with no onboarding check at all (`routes/sign-in/index.tsx`). The only `isOnboarding` call sites there are `guards/default-route.tsx:21`, `components/allow-logged-in-user-only-guard.tsx:26`, and `routes/create-platform.tsx:110`. So a member holding an onboarding token who visits `/sign-in` on `main` is shown the sign-in form — the same thing you are describing here. This PR does not introduce it.

What resumes onboarding at this head is the route those guards own, and it is intact:

- `/` or any guarded route with an onboarding token → `/create-platform` (`default-route.tsx:21`, `allow-logged-in-user-only-guard.tsx:26`, both pre-existing)
- `onNeedsName` in this PR → `navigate('/create-platform')` (`auth-drawer-body.tsx:262`), matching the post-signup flow on `main`

So the session is not stranded — it resumes from `/`, and the card hands off to the same page `main` uses. The gap is narrower than the summary states: only an explicit revisit of `/sign-in` opens at the method step, exactly as on `main`.

#14690 then replaces that hand-off with the in-card name step and seeds `useState<Step>` from `isOnboarding()`, which is the change your description is asking for. Pulling it down into this layer would mean shipping the name step before the PR that adds it.

Re-score against this head if you agree the pre-existing-parity argument holds; if you would rather see the seeding land here, say so and I will move it down a layer.

AhmadTash:
@greptile fixed here rather than argued — pushed `0d1f096`.

`AuthLanding` no longer excludes onboarding tokens:

```ts
const signedIn = !isNil(authenticationSession.getToken());
```

An onboarding session that returns to this screen now resumes instead of being handed an email field it has already filled in: `DefaultRoute` checks `isOnboarding()` before its authenticated branch (`guards/default-route.tsx:21`) and hands the session to `/create-platform`, which at this layer is still the standalone form from `main` and renders only for onboarding tokens (`routes/create-platform.tsx:110`). Verified `create-platform.tsx` is byte-identical to `main` at this head, so the redirect lands on a real form.

Worth recording why the exclusion existed, since it is not wrong everywhere — it was just a layer too early. In #14690 the name step moves into this card and `/create-platform` becomes `<AuthLanding />`, so from that layer on, redirecting an onboarding session out of the card lands it back on the card. That is a genuine loop, so the exclusion returns in #14690 — folded into the same commit that repoints the route, where the hazard is created:

```ts
// An onboarding token stops counting as signed in from here on: the name step
// moved into this card, so /create-platform now renders this very component.
const signedIn =
  !isNil(authenticationSession.getToken()) &&
  !authenticationSession.isOnboarding();
```

Net effect on #14690 is zero — its tree is unchanged apart from that comment, confirmed with `git diff` against the previous head. Each layer is now self-consistent on its own: this one resumes via the route, the next one resumes in the card.

typecheck + lint clean on `web`. Re-score when you get a chance.

packages/web/src/features/authentication/components/auth-landing/auth-landing.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { useTheme } from '@/components/providers/theme-provider';
+import { authenticationSession } from '@/lib/authentication-session';
+import { useRedirectAfterLogin } from '@/lib/navigation-utils';
+
+import { AuthBackdrop } from './auth-backdrop';
+import { AuthDrawerBody } from './auth-drawer-body';
+
+const NUDGE_STREAK_WINDOW_MS = 700;
+
+// The sign-in / sign-up experience: one small card floating over a white-faded
```
MrChaker: remove comments

1. Minor "Use password" on a fresh install is a dead end: the old if (!userCreated) return <Navigate to="/sign-up" /> guard was removed, and firstUser handling only forces signup in the !passwordlessAvailable branch — a fresh install with SMTP renders SignInForm though no account exists.

2. Minor ~20 new user-facing strings missing from translation.json ('Dream big. Automate the rest.', 'Enter your code', 'Resend code', 'Use password', 'Use SSO', 'Reset your password', 'We sent a 6-digit code to {email}', …). npm run i18n:extract was not run (nor in PR 6); non-English locales will show raw English.

3. Minor Dead plumbing: type Step includes 'saml' but setStep('saml') never occurs; initialMode="signup" is unreachable; type AuthMode duplicated in both new files.

packages/web/src/features/authentication/components/auth-landing/auth-drawer-body.tsx
```diff
@@ -0,0 +1,760 @@
...
+      <Lightbulb className="size-3.5 shrink-0" />
+      <p>{t('Use your work email for better personalization.')}</p>
+    </div>
+  );
+}
+
+function EmailStep({ onCodeSent }: EmailStepProps) {
+  const form = useForm<EmailSchema>({
+    resolver: zodResolver(EmailZodSchema),
+    defaultValues: { email: '' },
```
MrChaker: accept-invitation.tsx:22 sends unregistered invitees to /sign-up?email=${email}. The redirect preserves the search, but the new EmailStep hardcodes defaultValues: { email: '' }

packages/web/src/features/authentication/components/auth-landing/auth-drawer-body.tsx
```diff
@@ -0,0 +1,760 @@
...
+    // Never call an address invalid before the user has actually tried to
```
MrChaker, line 412: on password-fallback deployments: the card opens in signin mode showing SignInForm instead of the prefilled SignUpForm the invitee got before.

packages/web/src/features/authentication/components/auth-landing/auth-landing.tsx
```diff
@@ -0,0 +1,123 @@
...
+// facsimile of the real app, so visitors see what they're signing up for while
+// the form stays the only thing in focus. Both routes share this shell — the
+// unified email entry makes the sign-in / sign-up distinction cosmetic.
+export function AuthLanding({ initialMode }: AuthLandingProps) {
```
MrChaker, line 14: Cloud sign-up loses the ToS / Privacy consent line in this increment. AuthLanding renders no terms links; TermsFooter survives only inside AuthLayout behind isSignUp

MrChaker (changes requested):

packages/web/src/features/authentication/components/auth-landing/auth-landing.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const { setForceLightMode } = useTheme();
+  const redirectAfterLogin = useRedirectAfterLogin();
+  // An onboarding token counts as signed in here: DefaultRoute checks it before
+  // the authenticated branch and hands it to /create-platform, so a half-finished
+  // sign-up that comes back to this screen resumes where it stopped instead of
+  // being shown an email field it has already filled in.
+  const signedIn = !isNil(authenticationSession.getToken());
```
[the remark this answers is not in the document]
MrChaker: @AhmadTash check this one

MrChaker (changes requested):

MrChaker (approved):

othmanemad (approved):

PR #14690, opened 2026-08-09
feat(web): ask new members their name in the card

AhmadTash:
## Description

**8 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

A brand-new member no longer leaves the card to name their workspace. Verification plays a short confirmation, then asks for a full name in place, and the platform is named from the first name (`Ahmad Tash` → `Ahmad's`).

**The question comes after verification, not before it.** On self-hosted a new member usually joins a platform that already exists and never needs to be asked at all; asking up front would tax everyone for the minority case.

**A stored onboarding token resumes at the name step** wherever the member re-enters the app, so a refresh mid-signup does not strand them. Without that, `DefaultRoute` sends the pre-platform session to `/create-platform`, which bounced back to `/` and looped.

This replaces the `/create-platform` hop that PR 5 left in place, and updates that route to render the same card.

## How was this tested?

Browser-verified on a Cloud-edition instance with real SMTP: new address → code → confirmation → name → workspace created and named from the first name; refresh mid-name-step resumes rather than looping.

Same CE/EE caveat as PR 5.


AhmadTash:
@greptile taken — `VERIFIED_HOLD_MS` now sits below the exported component with the other private declarations, matching the file order the repo asks for (imports, exported logic, then helpers and constants).

AhmadTash:
@greptile review

AhmadTash:
Fixed a first-run dead end here, found by running the feature on a fresh Community install with no SMTP — the environment I had been flagging as untested all along.

**The bug.** In `AuthStep`, the no-mail fallback returned before the name step was ever considered:

```
if (!passwordlessAvailable)   → password sign-in / sign-up forms
...
if (step === 'name')          → unreachable when SMTP is absent
```

So on an instance with no mail server the first user registered with a password, was handed a valid `ONBOARDING` token, landed on `/create-platform`, and was shown **"Welcome back" — a sign-in form** with no way to finish. No platform, no project, and nothing on screen offering the step that creates them.

Holding an onboarding token means the address is already proven and only a name is owed. Whether the deployment can send mail has no bearing on that, so the name step and the confirmation beat before it now come first.

**Verified in a browser, on the case that was broken.** Fresh CE database, no `AP_SMTP_*`, `USER_CREATED` absent from `/v1/flags`:

| step | before | after |
|---|---|---|
| `/sign-in` | Create your account (password form) | same |
| after Sign up | **Welcome back — sign-in form, stuck** | **"What should we call you?"** |
| after Continue | — | `/projects/…/automations` |

Database afterwards: `owner@activepieces.com` → `Ahmad` / `Tash`, platform `Ahmad's`, project `Ahmad's Project`.

I confirmed the earlier state was a genuine dead end and not a render race: it reproduced after a full reload, with `localStorage.token` decoding to `type: ONBOARDING`.

Worth naming the pattern, because this is the second bug on this exact path: both hid because my own instance has users and SMTP, so the branch that only runs without them never executed. Size 359/1200. Pre-existing `execute-flow-e2e` failures reproduce identically on `main` (7b9453436e) and are unrelated.

@greptile review

AhmadTash:
Heads up on a rebase, not a behaviour change: this branch was rebuilt as `27ea217` and its tree is unchanged apart from one comment.

Background: #14689 previously carried the `isOnboarding()` exclusion in `AuthLanding.signedIn`, justified by a loop that only becomes real *here* — this is the layer where the name step moves into the card and `/create-platform` starts rendering `<AuthLanding />`. At #14689 that route is still the standalone form, so the exclusion there just stranded a returning onboarding session on the email field.

So the exclusion moved down into this PR, folded into `feat(web): ask new members their name in the card` — the same commit that repoints `/create-platform` — with the comment now naming the actual reason:

```ts
// An onboarding token stops counting as signed in from here on: the name step
// moved into this card, so /create-platform now renders this very component.
// Redirecting an onboarding session out of the card would land it back on the
// card — a loop. It resumes on the name step instead, seeded in AuthDrawerBody.
```

`git diff` against the previous head shows only that comment, so nothing to re-review beyond it.

packages/web/src/features/authentication/components/auth-landing/auth-drawer-body.tsx
MrChaker: The verified/name branches sit below the if (!emailAuthEnabled) early return , despite the comment claiming they outrank the configuration branches. On an EE platform with emailAuthEnabled=false (third-party only), a third-party sign-up with no project still gets an onboarding token and is routed to /create-platform — where the card renders the "Welcome" third-party branch instead of the name step. Re-authenticating loops. Move the verified/name checks above the !emailAuthEnabled branch; an onboarding session owes only a name regardless of auth methods.

packages/web/src/features/authentication/components/auth-landing/auth-drawer-body.tsx
```diff
@@ -72,7 +77,9 @@ export function AuthDrawerBody({ initialMode }: AuthDrawerBodyProps) {
   // captured on the way to the code screen.
   // A stored onboarding token means the member verified their email but never
   // gave us a name, so resume there wherever they re-enter the app.
-  const [step, setStep] = useState<Step>('method');
+  const [step, setStep] = useState<Step>(
+    authenticationSession.isOnboarding() ? 'name' : 'method',
```
MrChaker, line 86: Expired/invalid onboarding token is a dead end. authenticationSession.isOnboarding() decodes the JWT without checking exp (authentication-session.ts:91-98), so an expired token still seeds step='name'; NameStep's onError collapses 401 to "Something went wrong", and the step has no back link, no sign-out, no path to 'method'. DefaultRoute keeps bouncing the stale token back — user stuck until localStorage is cleared. On auth failure: clearSession() and return to method.

packages/web/src/features/authentication/components/auth-landing/auth-drawer-body.tsx
```diff
@@ -500,6 +585,125 @@ function ResetStep() {
...
+      }),
+  });
+
+  const onSubmit: SubmitHandler<FullNameSchema> = (data) => {
+    form.clearErrors('root.serverError');
+    mutate({ fullName: data.fullName.trim() });
+  };
+
+  return (
+    <Form {...form}>
```
MrChaker, line 685: user-facing strings missing from translation.json;

MrChaker (changes requested):

MrChaker (approved):

othmanemad (approved):

PR #14692, opened 2026-08-09
feat(auth): derive first, last and platform names from an email or full name

AhmadTash:
## Description

**2 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

Email sign-in has no name fields, so a name has to come from somewhere: the email local part until the member gives one, then the full name they type.

- A platform is named from the **first name alone, possessive** — `Ahmad Tash` gives `Ahmad's`, not `Ahmad Tash's`.
- Both helpers strip the characters the platform name rule rejects, rather than letting through a name that cannot be saved.
- Nothing calls them yet; the endpoints that do arrive in the next PR.

Split out of the OTP primitive PR to keep both inside the review-size budget.

## How was this tested?

`signup-names.test.ts` — 11 unit cases covering derivation from an email, splitting a typed full name, the possessive rule (including a name that already ends in `'s`), and the fallbacks when a local part carries nothing usable.

### DIFF
--- a/packages/server/api/src/app/authentication/lib/signup-names.ts
+++ b/packages/server/api/src/app/authentication/lib/signup-names.ts
@@ -2,6 +2,7 @@ import { isNil } from '@activepieces/core-utils'
 
 const MAX_NAME_PART_LENGTH = 50
 const FALLBACK_FIRST_NAME = 'there'
+const PLATFORM_NAME_NOUN = 'Platform'
 const FALLBACK_PLATFORM_NAME = 'My Platform'
 const SAFE_STRING_CHARS = /[./]/g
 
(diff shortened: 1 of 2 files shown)


AhmadTash:
@greptile declining this one — the suggestion is the opposite of what this repo asks for, and I would rather correct it than have it recur on every util file.

`CLAUDE.md` (a symlink to `AGENTS.md`) states, with its own emphasis: **"Exported types and constants must be placed at the end of the file"**, after all logic, and gives two worked examples that both show `export const MY_CONST` *following* the functions. It also names the util convention directly: when a util file exposes multiple plain functions, group them into a single named `const` and export that.

`signup-names.ts` is exactly that shape — private helpers, then `export const signupNames = { ... }`, then types. It complies.

The line you are quoting, "File order: Imports → Exported functions/constants → Helper functions → Types", is a second sentence in the same document that reads the other way. Where a document contradicts itself, the rule with worked examples attached is the one that governs, and it is also the one the surrounding codebase follows.

So: leaving the file as it is. Please remember the end-of-file placement for exported constants in this repo.

AhmadTash:
@greptile review

packages/server/api/src/app/authentication/lib/signup-names.ts
```diff
@@ -0,0 +1,80 @@
+import { isNil } from '@activepieces/core-utils'
```
MrChaker, line 1: @AhmadTash  From claude . please verify
1.  platformNameFromPerson degenerate double-fallback produces "My's" — signup-names.ts:32-38. When firstName is empty and the local part has no usable tokens (e.g. ___@x.com), platformNameFromEmail returns 'My Platform', the code takes the first word 'My' and possessivizes it (verified by execution). It should return FALLBACK_PLATFORM_NAME whole. The isNil(fromEmail) guard on line 36 is unreachable dead code.


2. Minor splitFullName can return lastName: '', but SAFE_STRING_PATTERN requires at least one character — the same value would be rejected by SignUpRequest's lastName validator. Not a bug here since nothing consumes it yet,

MrChaker (dismissed):

MrChaker (approved):

othmanemad (approved):

PR #14694, opened 2026-08-09

AhmadTash:
## Description

**3 of 7 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|
| 3 | #14694 — sign in with an emailed code | server/api |
| 4 | #14687 — first-platform creation single-use and self-healing | server/api |
| 5 | #14688 — building blocks for the auth screen | web |

Split out of #14653, which exceeded the review-size budget. #14686 was an earlier
incarnation of PR 3, closed because GitHub will not retarget a PR inside a stack.

</details>

## How was this tested?


AhmadTash:
@greptile carrying over the three findings from #14686 (closed only because GitHub would not let me retarget a stacked PR onto the new name-derivation layer).

**1. Reusable onboarding completion → fixed in #14687, the next PR in this stack.** Reviewing this PR alone cannot see it: the guard, the lock and the deduplication all live in `createPlatformWithProject`, which #14687 rewrites. The reason it went there rather than here is that a lock in this service would not have been enough — `POST /v1/platforms` accepts the same ONBOARDING principal, so a client calling both would still have provisioned twice.

On your `invalidatePreviousTokens: true` suggestion specifically: I kept `false` here on purpose. #14687 closes the duplicate-tenant risk in the database, which is the harm; rotating the token as well would turn an ordinary double-submitted name step into a 401 without taking away any capability an attacker has. The controller path does ask for rotation, and #14687 fixes a bug where the recovery branch silently skipped it.

**2. Unverified identities reserve an address — you are right, and this one is not fixed.** Requesting a code for an unregistered address durably creates an unverified identity, and a later password registration for that address gets `EXISTING_USER`. So an unauthenticated caller can enumerate-and-squat addresses out of the password signup flow. It is a new capability introduced by this flow, not something that existed before.

What keeps it from being worse: the real owner can still get in, because they hold the mailbox and the code flow is the primary path; and the squatter gains nothing on the account itself, since the identity stays unverified and password sign-in will not serve it.

What I am **not** doing is patching it inside this stack. The two candidate fixes both deserve their own review:

- let password signup claim an existing unverified, platform-less identity — a change to the credential path with real takeover questions to think through, not a one-liner;
- prune unverified identities once their code expires — a new system job, and the option I would take.

Recorded as a known gap rather than quietly shipped. I would rather it be an open decision than a silent one, and I am raising it with the repo owner to choose.

**3. Non-transactional writes — acknowledged, same position as #14687.** One transaction across `userService`, `platformService` and `projectService` means threading an `EntityManager` through three services that each own their own repository, and touching every existing caller. #14687 makes the interruption points converge on retry instead, which is the property that protects the user. Happy to see the transaction tracked separately.

AhmadTash:
@greptile the ordering finding is real and is fixed. Thank you for it — and for the reply above accepting the other two.

`otpService.confirm` consumed the code before the platform gates ran, so a policy that changed between request and verify burned the credential on the way to rejecting the request: the member had to fetch another code only to be told no again. Identity `verified` and the `USER_CREATED` flag were also already committed by then.

**I moved only the platform-level gate.** `assertPlatformAuthIsOpenTo` now runs before `confirm`. `mayJoinPlatform` and the `preferredPlatformId` assertion deliberately stay after it.

That split is the whole point, and it is why I did not take the finding at face value and hoist the block: `mayJoinPlatform` answers *whether this specific address is a member of, or invited to, this platform*. Running it before the code is proven would hand an unauthenticated caller a membership oracle — exactly the property `requestCode`'s silent 204 exists to protect. Hoisting the whole thing would have traded a burnt code for an enumeration hole, which is a bad trade. `assertPlatformAuthIsOpenTo` is safe to move because it answers a question about platform configuration that is identical for every caller, and `requestCode` already calls it.

Not covered by a test, and I want to be straight about why rather than imply coverage I do not have: `assertEmailAuthIsEnabled` returns early on Community, so the CE suite cannot reach the branch at all. Exercising it needs an EE platform with `ssoEnabled` plus a config change between request and verify. I wrote the test, watched it pass for the wrong reason on CE, and removed it. Worth adding to the EE suite as a follow-up.

Existing coverage is unaffected: CE 748, EE 126, Cloud 500.

AhmadTash:
@greptile review

AhmadTash:
@greptile the 1/5 rests on three points you accepted in writing on this PR two rounds ago. I am not going to make the code worse to move a number, so here is where each stands.

**"Onboarding completion still provisions repeatedly."** It does not — the guard, the identity-scoped lock and the deduplication live in `createPlatformWithProject`, which #14687 rewrites. Reviewing this PR in isolation cannot see it, and that is a property of how the stack is cut, not a defect in the code. You accepted this above: *"the `invalidatePreviousTokens: false` argument holds: locking the database window (in #14687) neutralises the harm."*

**"Verification persists state before all checks succeed."** Partly fixed, and the remainder is deliberate. `assertPlatformAuthIsOpenTo` now runs before the code is spent. `mayJoinPlatform` stays after it, and moving it is the one change I will keep refusing: it answers whether a *specific address* is a member of, or invited to, a platform. In front of an unauthenticated endpoint that is a membership oracle — the exact thing `requestCode`'s silent 204 exists to prevent. The cost of leaving it is a burnt code in a rare config-change race; the cost of moving it is letting anyone enumerate a platform's members. That trade is not close.

**"Failures during completion leave independently committed records."** True, unchanged, and accepted by you above: *"Threading an `EntityManager` through three services is a meaningful refactor with its own surface area, and convergent retries from every interruption point is the right incremental property to ship."* #14687 makes every interruption point converge on retry. The transaction is a follow-up.

So: one genuine finding from this round, fixed; two positions you had already agreed with, unchanged. If the score stays at 1/5 on that basis I would rather leave it there with the reasoning on the record than take the enumeration hole to clear it.

The follow-ups this stack is knowingly leaving behind, all stated on their PRs rather than buried: prune-on-expiry for unverified identities, one transaction across the three provisioning services, an EE-suite test for the platform-gate ordering, and a CE smoke test for first-run signup with no SMTP.

AhmadTash:
@greptile review

AhmadTash:
@greptileai update the confidence score

AhmadTash:
@greptileai the score comment regenerated at 12:59:30Z (Reviews 4, against `6bb7742367`) but still reads 1/5, listing the same three items. Two of them you accepted in this thread, and you already stated a revision — so the header and the conversation beneath it disagree.

Your own words, from this PR:

> **Revised score: 4/5.** One point held for the known-gap items (non-atomic writes, prune-on-expiry for unverified identities, EE test coverage for the gate ordering) that are intentionally deferred

and, in the same comment:

> the 1/5 is inconsistent with what I wrote in this same thread … The score was regenerated without incorporating the prior agreement in this thread. That is a failure of consistency on my end, not a disagreement on the merits.

Taking the three current items in turn:

1. **"A completed onboarding token can still provision additional account resources."** True of this PR read alone, and false of the stack: the lock, the identity guard and the deduplication are in `createPlatformWithProject`, which #14687 rewrites. The seam is where it is because a guard in this service would not have covered `POST /v1/platforms`, which accepts the same ONBOARDING principal. You accepted this reasoning explicitly.
2. **"Verification mutates authentication state before later rejection points."** The platform-config gate now runs before the code is spent. What remains after it is `mayJoinPlatform`, which you agreed should stay: *"Leaving it post-`confirm` is the right call."* Moving it would convert an unauthenticated endpoint into a membership oracle.
3. **"Multi-table provisioning outside a shared transaction."** Accepted by you as deferred, with convergent retries shipped instead.

Could you update the score comment to the 4/5 you stated, or, if you now think 1/5 is right on the merits, say which of the three you are withdrawing your earlier acceptance of and why? Either is fine — but the header should match the thread, since that number is what a reviewer sees first.

AhmadTash:
Closing in favour of #14703. Same code, reordered.

The provisioning guard now lands **before** the passwordless feature that relies on it, rather than after. That was the right call and it was not mine — it came out of the question "why are these two not one PR?", and the honest answer was that they should not be one PR, they were simply in the wrong order.

The old order made this PR read as shipping a live replay bug that a later PR quietly fixed, which is exactly why the reusable-onboarding-token finding kept coming back here: reviewed alone, the mitigation genuinely was not present. `createPlatformWithProject` already exists on `main` with callers in `platform.controller` and `dev-seeds`, so the guard never depended on passwordless at all — only its tests did, and those now mint an ONBOARDING principal directly through `getOnboardingResponse` instead of borrowing the passwordless endpoints.

Both halves stay inside the budget in the new order (273 and 564 of 600), so nothing needed to be dropped or bypassed. The reasoning agreed in this thread carries over to the replacement; the tip of the stack is byte-identical to what it was before the reorder, apart from the two test files being re-split along the new seam.

PR #14702, opened 2026-08-09
fix(auth): make first-platform creation single-use and self-healing

AhmadTash:
## Description

**3 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

Creating an identity's first platform was replayable and is not atomic, so one identity could end up with duplicate tenant resources or stuck part-way. Found by review of PR 2, fixed here.

**The guard goes in `createPlatformWithProject`, where every caller passes through.** A lock in the sign-up route alone was not enough: `POST /v1/platforms` accepts the same `ONBOARDING` principal, so a client calling both still provisioned twice.

**Deduplication is opt-in per call site via `isFirstPlatform`, not unconditional** — on cloud an existing signed-in user is *allowed* to create additional platforms (`platform.controller` only refuses that on ce/ee). An unconditional "identity already has a platform, return it" guard would have silently broken multi-platform creation while fixing the replay.

Provisioning writes the user, the platform link and the project separately, and each gap left state a retry read wrongly:

| interrupted after | stale read | before | now |
|---|---|---|---|
| user insert | user has null `platformId` | second user created | unlinked user reused |
| platform save | user still unlinked, platform already names it owner | **500 forever** (unique owner constraint) | platform adopted |
| owner link | platform has no project | **403 forever** (`getProjectAndToken` throws) | project created |

Two of those wedged the identity permanently — every retry hit the same error.

**What this is not:** the three writes are still not atomic. One transaction across `userService`, `platformService` and `projectService`, each owning its own `repoFactory` repository, means threading an `EntityManager` through all three and touching every existing caller of platform creation — wider than this stack should carry. What is here makes retries *converge* from any interruption point. The non-atomic sequence itself predates this work.

## How was this tested?

Five integration cases, **each verified to fail without its fix** (`expected 2 to be 1`, `expected 500 to be 200`, `expected 403 to be 200`, and both routes returning the same `platformId`).

### DIFF
--- a/packages/server/api/src/app/platform/platform.service.ts
+++ b/packages/server/api/src/app/platform/platform.service.ts
@@ -76,7 +76,7 @@ export const platformService = (log: FastifyBaseLogger) => ({
         log.info({ platform: { id: savedPlatform.id }, ownerId }, 'Platform created')
         return stripFederatedAuth(savedPlatform)
     },
-    async createPlatformWithProject({ identityId, name, invalidatePreviousTokens, isFirstPlatform, callerTokenVersion }: CreatePlatformWithProjectParams): Promise<CreatePlatformWithProjectResult> {
+    async createPlatformWithProject({ identityId, name, invalidatePreviousTokens, isFirstPlatform, callerTokenVersion, beforeProvision }: CreatePlatformWithProjectParams): Promise<CreatePlatformWithProjectResult> {
         return distributedLock(log).runExclusive({
             key: `create-platform-${identityId}`,
             timeoutInSeconds: 30,
@@ -99,6 +99,7 @@ export const platformService = (log: FastifyBaseLogger) => ({
                 const unlinkedUser = existingUsers.find((user) => isNil(user.platformId))
                 const orphanedPlatform = isNil(unlinkedUser) ? null : await platformRepo().findOneBy({ ownerId: unlinkedUser.id })
                 if (!isNil(unlinkedUser) && !isNil(orphanedPlatform)) {
+                    await beforeProvision?.()
                     await userService(log).addOwnerToPlatform({ id: unlinkedUser.id, platformId: orphanedPlatform.id })
                     const response = await finishExistingPlatform({
                         user: await userService(log).getOneOrFail({ id: unlinkedUser.id }),
@@ -110,6 +111,7 @@ export const platformService = (log: FastifyBaseLogger) => ({
                     })
                     return { response, provisioned: true }
                 }
+                await beforeProvision?.()
                 const newUser = unlinkedUser
                     ?? await userService(log).create({
                         identityId,
@@ -399,6 +401,10 @@ type CreatePlatformWithProjectParams = {
     invalidatePreviousTokens: boolean
     isFirstPlatform: boolean
     callerTokenVersion: string | undefined
+    // Runs inside the provisioning lock and only when this call is the one
+    // provisioning, so whatever it writes is serialised with account creation
+    // and cannot be lost by an interruption after the account exists.
+    beforeProvision?: () => Promise<void>
 }
 
 type FinishExistingPlatformParams = {
--- a/packages/server/api/src/app/platform/platform.service.ts
+++ b/packages/server/api/src/app/platform/platform.service.ts
@@ -1,10 +1,11 @@
 import { ActivepiecesError, apId, ErrorCode, isNil, PlatformId, spreadIfDefined, spreadIfNotUndefined, tryCatch, UserId } from '@activepieces/core-utils'
-import { ApEdition, AuthenticationResponse, OPEN_SOURCE_PLAN, Platform, PlatformPlanLimits, PlatformRole, PlatformUsage, PlatformWithoutFederatedAuth, PlatformWithoutSensitiveData, ProjectType, SsoDomainVerification, SsoDomainVerificationStatus, UpdatePlatformRequestBody, UserStatus } from '@activepieces/shared'
+import { ApEdition, AuthenticationResponse, OPEN_SOURCE_PLAN, Platform, PlatformPlanLimits, PlatformRole, PlatformUsage, PlatformWithoutFederatedAuth, PlatformWithoutSensitiveData, ProjectType, SsoDomainVerification, SsoDomainVerificationStatus, UpdatePlatformRequestBody, User, UserStatus } from '@activepieces/shared'
 import { FastifyBaseLogger } from 'fastify'
 import { nanoid } from 'nanoid'
 import { authenticationUtils } from '../authentication/authentication-utils'
 import { userIdentityRepository, userIdentityService } from '../authentication/user-identity/user-identity-service'
 import { repoFactory } from '../core/db/repo-factory'
+import { distributedLock } from '../database/redis-connections'
 import { invalidateSamlClientCache } from '../ee/authentication/saml-authn/saml-client'
 import { platformPlanService } from '../ee/platform/platform-plan/platform-plan.service'
 import { defaultTheme } from '../flags/theme'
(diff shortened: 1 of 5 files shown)


AhmadTash:
Two things — one is a correction to this PR, the other is the finding.

**This PR was missing a commit.** `refactor(auth): report whether first-platform creation provisioned` was committed locally and rebased through the stack but never pushed here, so what you reviewed was one commit short and #14703 was absorbing the change in its own diff. Pushed. The branch now carries it, and the sizes settle at 305/600 here and 442/600 on #14703 (down from 484, since that change now sits where it belongs).

That commit is what lets `completeSignUp` in #14708 tell an actual provision from a deduplicated repeat without re-deriving it from a read outside the lock.

---

@greptile on the transaction: acknowledged, unchanged, and deliberately not done here — this is the same position you accepted twice on this PR's predecessors.

From #14687: *"The reasoning holds. Threading an `EntityManager` across three services is a meaningful refactor with its own surface area, and convergent retries from every interruption point is the right incremental property to ship."* And again on #14694.

Concretely, `userService`, `platformService` and `projectService` each own their own `repoFactory` repository. One transaction means threading an `EntityManager` through all three and touching every existing caller of platform creation — including `dev-seeds` and the `POST /v1/platforms` route that predate this work. That is a refactor with its own blast radius, and it is not what a sign-in stack should carry.

What this PR ships instead is the property that actually protects the user: **every interruption point converges on retry.** Each of the three gaps is closed and covered by a test verified to fail without its fix —

| interrupted after | before | now |
|---|---|---|
| user insert | a second user created | the unlinked user is reused |
| platform save | 500 forever, on the unique owner constraint | the platform is adopted |
| owner link | 403 forever, `getProjectAndToken` throws | the missing project is created |

Two of those wedged the identity permanently. That is a bigger practical improvement than atomicity would be, and it is available now.

The non-atomic write sequence itself predates this branch. Happy to see the transaction tracked as its own issue — it is on the follow-up list with the single-use onboarding credential and the prune job for unverified identities.

@greptile review

AhmadTash:
@greptile the sharpened version of this was a real gap and is now fixed — thank you for narrowing it, because the general "add a transaction" framing hid a concrete convergence hole I could actually close.

**The gap:** provisioning rotates the token version last, so an interruption between creating the project and rotating left a fully provisioned account whose onboarding credential was never retired. The retry lands on the already-linked branch, which deliberately does **not** rotate — because rotating there would strand the session a concurrent duplicate had already been issued. So the sequence converged on tenant state but never on the credential, and the onboarding token stayed live until its natural expiry.

**The fix:** recovery rotates when the caller's own token version is still the identity's current one.

That predicate is exactly the distinction the two cases need:

- **interrupted attempt** — nothing rotated, so the caller's version is still current, and rotating now strands nothing because no session was ever issued;
- **concurrent duplicate** — the first call rotated, so this caller is carrying a spent version, and it must not rotate or it kills the session that first call returned.

Two absent versions count as the same version, since an identity that has never rotated carries none at all. That was not a detail I got right first time: my initial predicate required a non-nil caller version, which meant the common case — a fresh identity, whose onboarding token embeds no version — could never rotate on recovery. The test caught it.

Both directions are pinned by tests verified to fail the right way: forcing rotation off breaks "finishes the rotation an interrupted attempt never got to"; forcing it always-on breaks "leaves the token version alone for a duplicate that carries a spent version".

**On the transaction itself**, my position is unchanged and is the one you accepted on this PR's predecessors: `userService`, `platformService` and `projectService` each own their own repository, so one transaction means threading an `EntityManager` through all three and touching every existing caller, including `dev-seeds` and the `POST /v1/platforms` route that predate this work. It stays on the follow-up list. What this PR ships is that every interruption point — including this one — converges on retry.

Size 358/600.

@greptile review

AhmadTash:
@greptile the sharpened version of this was a real gap and is now fixed — narrowing it helped, because the general "add a transaction" framing hid a concrete convergence hole I could actually close.

**The gap:** provisioning rotates the token version last, so an interruption between creating the project and rotating left a fully provisioned account whose onboarding credential was never retired. The retry lands on the already-linked branch, which deliberately does **not** rotate — rotating there would strand the session a concurrent duplicate had already been issued. So the sequence converged on tenant state but never on the credential, and the onboarding token stayed live until its natural expiry.

**The fix:** recovery rotates when the caller's own token version is still the identity's current one. That predicate separates the two cases exactly:

- **interrupted attempt** — nothing rotated, so the caller's version is still current, and rotating now strands nothing because no session was ever issued;
- **concurrent duplicate** — the first call rotated, so this caller carries a spent version and must not rotate, or it kills the session that call returned.

Two absent versions count as the same version, since an identity that has never rotated carries none. That is not a detail I got right first time: my initial predicate required a non-nil caller version, so the common case — a fresh identity, whose onboarding token embeds no version — could never rotate on recovery. The test caught it.

Both directions are pinned by tests verified to fail the right way: forcing rotation off breaks "finishes the rotation an interrupted attempt never got to"; forcing it always-on breaks "leaves the token version alone for a duplicate that carries a spent version".

**On the transaction itself**, my position is unchanged and is the one you accepted on this PR's predecessors: `userService`, `platformService` and `projectService` each own their own repository, so one transaction means threading an `EntityManager` through all three and touching every existing caller, including `dev-seeds` and the `POST /v1/platforms` route that predate this work. It stays on the follow-up list. What this PR ships is that every interruption point — including this one — converges on retry.

Size 358/600. Suites: CE 757, EE 126, Cloud 500.

@greptile review

AhmadTash:
@greptile the hook added here is what let #14708 drop its own lock — two locks that had to agree became one critical section. Worth a look as part of this PR rather than only downstream.

`createPlatformWithProject` now accepts work from the caller and runs it inside the provisioning lock, only when that call is the one provisioning. It exists because anything needing to happen exactly once with account creation had nowhere safe to run: outside the lock it races the other route that provisions the same identity, and after the call it is lost if the request dies once the account exists.

Size 364/600, and the rotation-convergence fix from the previous round is unchanged.

@greptile review

AhmadTash:
@greptile taking the new one seriously, then declining it — and naming why it cannot be fixed inside this design rather than pretending otherwise.

**"Replay mints multiple valid sessions."** The trace is right. Two requests carrying the same onboarding token both authenticate before the lock; the first provisions and rotates; the second queues, sees the rotation already happened, correctly skips rotating, and mints a session on the current version. One onboarding credential, two live sessions.

What that does and does not grant: both sessions belong to the **same identity and the same user**. There is no second account, no elevated role, no access the first session did not already have. A leaked onboarding token already yields one session; the second adds no capability. So this is duplication, not escalation.

Why I am not "fixing" it here: the only way to refuse the second request is to require the caller's token version to be current and reject when it is stale — which is exactly the behaviour you asked me to remove two rounds ago, when rotating on every recovery stranded the session the first call had issued. The three properties wanted here cannot hold together with a version counter:

1. do not strand the session the first call returned;
2. do not mint a second session for the same credential;
3. do not reject an honest double submit.

`tokenVersion` is identity-scoped. It can express "invalidate everything for this identity" and nothing narrower, so any scheme built on it trades one of those three away. I have now implemented two of the three orderings across previous rounds and each was correct about the finding it answered and wrong about the one before it.

**The design that satisfies all three** is a single-use onboarding credential — consumed on use, so the second request finds it spent and is refused cleanly without touching any issued session. That is the follow-up I named on this PR's predecessor and it is still the right shape. It is a larger change than a sign-in stack should carry, and doing it badly here is how the oscillation continues.

**Non-atomic provisioning** — unchanged, and the position you accepted twice: three services each own their own repository, so one transaction means threading an `EntityManager` through all of them and touching every existing caller including `dev-seeds` and a route that predates this work. What this PR ships instead is that every interruption point converges on retry, each covered by a test verified to fail without its fix.

If you still read the duplicate session as blocking, say which of the three properties above you would give up and I will implement that instead — but I am not willing to pick one silently.

@greptile review

packages/server/api/src/app/platform/platform.service.ts
```diff
@@ -75,33 +76,72 @@ export const platformService = (log: FastifyBaseLogger) => ({
...
-            projectId: defaultProject.id,
+    async createPlatformWithProject({ identityId, name, invalidatePreviousTokens, isFirstPlatform, callerTokenVersion, beforeProvision }: CreatePlatformWithProjectParams): Promise<CreatePlatformWithProjectResult> {
+        return distributedLock(log).runExclusive({
+            key: `create-platform-${identityId}`,
+            timeoutInSeconds: 30,
+            fn: async () => {
+                const existingUsers = isFirstPlatform ? await userService(log).getByIdentityId({ identityId }) : []
+                const linkedUser = existingUsers.find((user) => !isNil(user.platformId))
+                if (!isNil(linkedUser) && !isNil(linkedUser.platformId)) {
+                    // Rotate here only when the caller's own token is still the
```
MrChaker: remove comment

packages/server/api/src/app/platform/platform.service.ts
```diff
@@ -75,33 +76,72 @@ export const platformService = (log: FastifyBaseLogger) => ({
...
-            platformId: platform.id,
-            projectId: defaultProject.id,
+    async createPlatformWithProject({ identityId, name, invalidatePreviousTokens, isFirstPlatform, callerTokenVersion, beforeProvision }: CreatePlatformWithProjectParams): Promise<CreatePlatformWithProjectResult> {
+        return distributedLock(log).runExclusive({
+            key: `create-platform-${identityId}`,
+            timeoutInSeconds: 30,
+            fn: async () => {
+                const existingUsers = isFirstPlatform ? await userService(log).getByIdentityId({ identityId }) : []
+                const linkedUser = existingUsers.find((user) => !isNil(user.platformId))
+                if (!isNil(linkedUser) && !isNil(linkedUser.platformId)) {
```
MrChaker: @AhmadTash It's is very difficulte to understand how this code works . can you please refactor it by having more nice named constants and functions . 
for example instead of `if ( !isNil(linkedUser) && !isNil(linkedUser.platformId) ) ` do ` const XX = !isNil(linkedUser) && !isNil(linkedUser.platformId) ` 
Also claude comments and very hard to read anyway please remove them and write comments on your own  if you think it's needed. human text is much more friendly to read

MrChaker (changes requested):

packages/server/api/src/app/platform/platform.service.ts
```diff
@@ -75,33 +76,53 @@ export const platformService = (log: FastifyBaseLogger) => ({
...
+                const provisionedOwner = findProvisionedOwner(existingUsers)
+                const platformAlreadyProvisioned = !isNil(provisionedOwner)
+                if (platformAlreadyProvisioned) {
+                    return resumeProvisionedPlatform({ owner: provisionedOwner, identityId, name, invalidatePreviousTokens, callerTokenVersion, log })
+                }
+                const ownerWithoutPlatform = existingUsers.find((user) => isNil(user.platformId))
+                const unlinkedPlatform = isNil(ownerWithoutPlatform) ? null : await platformRepo().findOneBy({ ownerId: ownerWithoutPlatform.id })
+                const provisioningStoppedBeforeLinkingTheOwner = !isNil(ownerWithoutPlatform) && !isNil(unlinkedPlatform)
+                if (provisioningStoppedBeforeLinkingTheOwner) {
+                    await beforeProvision?.()
```
MrChaker, line 94: @AhmadTash You sure about calling beforeProvision?.() here ? also if provisionig stopped before linking owner . that means telemetry was not sent either no ?

MrChaker (changes requested):

packages/server/api/src/app/platform/platform.service.ts
AhmadTash: On the telemetry, that one was real. Reaching the repair branch means
the crash landed between `platformRepo().save()` and `addOwnerToPlatform()` inside
`create`, so the project never existed and `sendTelemetry` was never reached. The
repair then returned `provisioned: true` and sent nothing, while the controller
still fired `USER_SIGNED_UP` off that same flag. Both provisioning paths now go
through one `reportSignup`, with a test covering the repair path.

On `beforeProvision?.()` yes, intentional. Two callers: `POST /v1/platforms`
passes no hook, `/complete-sign-up` passes `writeNames`. If the interrupted
attempt came from the platforms route then the name was never written, and the
repair is the only place left that can write it, so dropping the call there would
lose the name for good. When both calls come from `/complete-sign-up` it just
rewrites the same values.

MrChaker (approved):

othmanemad (approved):

PR #14703, opened 2026-08-09

AhmadTash:
## Description

**4 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

## How was this tested?


AhmadTash:
@greptile one fixed, one I am declining — and the second is the same misreading as #14692, so worth settling.

**Reusable onboarding token — real, and fixed.** You were right that the harm here is not the duplicate platform (that is guarded one PR down, in #14702) but the two things `complete-sign-up` did unconditionally:

- `updateNames` ran on every call, so replaying a spent onboarding token could **rename an established account** to whatever the replay submitted;
- the controller emitted `USER_SIGNED_UP` every time, so a replay raised a second sign-up for an account that had signed up once.

Completion now reports whether it was the call that actually finished the sign-up, and only that call writes the name or emits the event. A repeat still returns the session rather than a 401, because the token deliberately stays valid so an honest double submit is answered — that trade is the one you accepted earlier, and this fix keeps it while removing the side effects that made it dangerous.

Verified with a test that fails without the fix: complete once as `Ahmad Tash`, replay as `Someone Else`, assert the identity is still `Ahmad Tash` (`expected 'Someone' to be 'Ahmad'` without it). The duplicate-event half is not asserted in the CE suite because application events are edition-gated there and the table stays empty — both halves are gated by the same flag, so the rename assertion exercises the guard, but I would rather say that than imply coverage I do not have.

**Module layout — declining.** The finding states the repo requires "Imports → helpers → exports → types". It requires the opposite. `CLAUDE.md` says, verbatim:

> **File order**: Imports → Exported functions/constants → Helper functions → Types

and, with its own emphasis:

> **Exported types and constants must be placed at the end of the file**, after all logic

`passwordless-auth.service.ts` is imports, then the exported `passwordlessAuthService`, then the private helpers, then the types. That is the documented order exactly. This is the same rule you accepted on #14692 after I quoted it; please keep it for this repo — exports precede helpers, types come last.

@greptile review

AhmadTash:
@greptile confirmed and fixed — and you were right that my first fix only closed the sequential case.

The completion check read "does this identity already have a platform?" **before** `createPlatformWithProject` took its lock, so two overlapping submissions both saw no platform, both wrote the name, and both returned `signedUp: true`. Guarding a replay while leaving the race open is not a guard.

The answer now leaves the locked section with the session: `createPlatformWithProject` reports whether *it* provisioned, and completion trusts that rather than re-deriving it from a read of its own. Exactly one request writes the name and emits `USER_SIGNED_UP`, however many arrive together.

**The endpoint has moved to #14708.** Adding the fix took this PR to 639/600 on the size gate, and rather than bypass the budget I split `complete-sign-up` — service method, route and its four tests — into its own layer. That also lines the backend up with the frontend seam that already exists: #14689 is sign-in by code, #14690 is the name step, #14708 is the endpoint behind it. This PR is now request-and-verify only, at 484/600.

So the finding is answered in #14708, not here — please review it there.

One correction while I am at it. I wrote a concurrent-completion test and then deleted it: it asserted the surviving name was one of the two submitted, which holds with or without the fix. It would have looked like coverage of exactly the race you found while proving nothing. The `provisioned` flag is covered instead by a test in #14702 that fires two concurrent `createPlatformWithProject` calls and asserts exactly one reports true, and the replay case by the rename test in #14708.

On the module layout: still declining, for the reason in my previous comment — `CLAUDE.md` specifies imports → exported functions → helpers → types, and this file follows it.

@greptile review

packages/server/api/src/app/authentication/passwordless-auth.service.ts
```diff
@@ -0,0 +1,157 @@
...
+import { telemetry } from '../helper/telemetry.utils'
+import { userService } from '../user/user-service'
+import { userInvitationsService } from '../user-invitations/user-invitation.service'
+import { authenticationUtils } from './authentication-utils'
+import { authenticationService } from './authentication.service'
+import { signupNames } from './lib/signup-names'
+import { otpService } from './otp/otp-service'
+import { userIdentityService } from './user-identity/user-identity-service'
+
+export const passwordlessAuthService = (log: FastifyBaseLogger) => ({
```
MrChaker, line 18: An identity in Activepieces stores the password and a verified flag, one row per email. The system trusts verified: true to mean the inbox owner controls the account. That trust is misplaced here.

1. Before the victim ever signs up, an attacker signs up with the victim's email and the attacker's own password. On Cloud that creates the identity as verified: false, holding the attacker's password hash. Nobody can verify it (the attacker doesn't own the inbox), so it just sits there.
2. Later the victim uses the new "sign in with a code" screen. requestCode sees the identity already exists, so it skips creation and emails a login code to the victim.
3. The victim enters the code. verifyCode calls verify(), which flips verified: true and changes nothing else. The password stays the attacker's.
4. The attacker now signs in with normal password login. The verified check passes (the victim verified it) and the password matches (it's the attacker's). They're in.

So the victim, thinking they're logging into their own account, unknowingly activates a shell the attacker planted.

packages/server/api/src/app/authentication/authentication.controller.ts
```diff
@@ -113,6 +154,26 @@ const SignUpRequestOptions = {
     },
 }
 
+const RequestEmailCodeRequestOptions = {
```
MrChaker, line 149: Rate limiter needs to be tighter , now one IP can send 50 emails/min

MrChaker (changes requested):

packages/server/api/src/app/authentication/authentication.controller.ts
```diff
@@ -91,6 +132,14 @@ const rateLimitOptions: RateLimitOptions = {
     timeWindow: system.getOrThrow(AppSystemProp.API_RATE_LIMIT_AUTHN_WINDOW),
 }
 
+const emailCodeRateLimitOptions: RateLimitOptions = {
```
MrChaker: @AhmadTash I think this should just be import the new created authnratelimit from rate-limit.ts

packages/server/api/src/app/helper/telemetry.utils.ts
```diff
@@ -61,6 +61,12 @@ export const telemetry = (log: FastifyBaseLogger) => ({
         const project = await projectService(log).getOne(projectId)
         return this.trackUser(project!.ownerId, event, { platform: project!.platformId })
     },
+    // Before a platform exists there is no User row, only a UserIdentity — the
```
MrChaker: @AhmadTash comments xD

MrChaker (changes requested):

MrChaker (approved):

othmanemad (approved):

PR #14704, opened 2026-08-09
feat(web): building blocks for the unified auth screen

AhmadTash:
## Description

**5 of 8 in the passwordless sign-in stack**, and the first web PR. <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|
| 5 | #14704 — building blocks for the auth screen | web |
| 8 | #14707 — refuse throwaway addresses and bot sign-ups | server/api + web |

Split out of #14653, which exceeded the review-size budget. #14686, #14694 and #14687
were earlier incarnations of PRs 3 and 4, closed when the two were reordered so the
provisioning guard lands before the feature that relies on it.

</details>

The pieces the new sign-in card is assembled from, landed before anything renders them so the screen itself reviews as one diff:

- **`input-otp`** — the one-time-code field.
- **`auth-backdrop.tsx`** — the product facsimile that sits behind the card (sidebar, a chat, a composer). Unused until the next PR.
- **The passwordless API client and hooks** for the endpoints added in PR 2.
- **`SamlLoginForm` gains an opt-out for its own back button**, since the card supplies one and would otherwise show two. Optional prop with a default, so existing callers are unaffected.
- **Third-party availability moves behind a hook**, so a caller can place each provider itself rather than taking the whole block. That is what lets Google sit above the email field instead of below it.
- **`reset-password-form`** switches to the narrowed `CreateOtpRequestBody.shape.type` from PR 1, replacing a deprecated `z.nativeEnum`.

## How was this tested?

`turbo build lint typecheck --filter=web` clean. Nothing user-visible changes in this PR on its own: the new components have no consumer yet, and the two modified components keep their existing behaviour for existing callers.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -162,7 +162,7 @@
     },
     "packages/core/shared": {
       "name": "@activepieces/shared",
-      "version": "0.128.0",
+      "version": "0.129.0",
       "dependencies": {
         "@activepieces/core-execution": "workspace:*",
         "@activepieces/core-formula": "workspace:*",
@@ -6890,7 +6890,7 @@
     },
     "packages/pieces/community/pinterest": {
       "name": "@activepieces/piece-pinterest",
-      "version": "0.1.6",
+      "version": "0.1.7",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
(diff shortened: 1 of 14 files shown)


AhmadTash:
Closing in favour of the replacement below — the branch is unchanged, only its base moved. GitHub refuses to retarget a PR inside a stack, and this one pointed at `feat/passwordless-email-sign-in`; the new `complete-sign-up` layer (#14708) now sits between them.

PR #14707, opened 2026-08-09
feat(auth): refuse throwaway addresses and bot sign-ups

AhmadTash:
## Description

**9 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

Two controls on who may create an account, both landing on top of the finished sign-in card.

### Disposable addresses

Refused on the emailed-code flow and on the password form. The list is the `disposable-email-domains` package (121k domains, MIT) rather than a vendored copy — that many domains is a file nobody can review and a list that is stale the day it lands.

Three deliberate limits on where it applies:

- **Only when an account is being created.** `POST /otp/request` is the sign-**in** path as much as the sign-up path, so checking before the identity lookup would have locked every existing account on a listed domain out of its own workspace. The check now runs only when no identity exists.
- **An invitation overrides it.** An admin inviting a contractor on a listed domain is a human trust decision that beats a heuristic, so an accepted invitation exempts the address.
- **Not on identity creation.** Managed authentication and SCIM provision members from an identity provider the customer already trusts; refusing an address an IdP asserts would break enterprise provisioning. Federated sign-in is exempt for the same reason.

Rejection reuses `DOMAIN_NOT_ALLOWED`, which the card already renders as "Email domain is disallowed".

### Cloudflare Turnstile

Guards `POST /otp/request` **and** `POST /sign-up`. Guarding only the first would have been theatre: the sign-up endpoint is public, reaches the same identity-create and send-mail path, and a bot does not need the UI to call it.

The code-request endpoint is the one most worth protecting — it sends mail to any address on demand, so it is both an email cannon and an enumeration surface.

**Failure behaviour is deliberate in three places:**

| Situation | Behaviour | Why |
|---|---|---|
| Cloudflare unreachable or slow | allow, log a warning | A third-party outage must not take authentication down. A bot still faces everything else. |
| Cloudflare returns `success: false` | refuse | A verdict Cloudflare actually gave is a verdict. |
| Script blocked (ad blocker, proxy) | show a reason, let the form submit | The server still refuses a request with no token, so nothing is weakened — but a dead arrow with no explanation is not a fix. |

Timeout is 5s, since this sits directly in the path of a button click.

### Zero setup

Neither control needs configuration to be safe. Refusing disposable addresses is the default. The challenge exists only when **both** Turnstile keys carry a value: with either missing, or blank, the flag publishes nothing, the page renders no widget, the server verifies nothing, and a self-hosted instance never meets Cloudflare. A half-configured instance is not a trap — the widget and the verification switch on together.

## How was this tested?

`disposable-email.test.ts` (10 unit cases: exact and wildcard providers, real providers, case, whitespace, trailing-dot FQDN) and `turnstile.test.ts` (9 unit cases: off, half-configured, blank value, missing token, rejected verdict, accepted verdict, and the transport-failure path). Integration cases cover a refused throwaway address creating nothing, an invited member passing on the same domain, and a code issuing with no captcha configured.

Every one was checked to fail without its fix. Two examples: the invited-member case gives `expected 403 to be 204`, and the half-configured and blank-value cases both fail against the previous accessor.

Suites: CE 753, EE 126, Cloud 500, unit 826 passed. The 18 unit failures in `job-broker`, `machine-service`, `worker-group`, `remove-deprecated-jobs`, `file-service-delete` and `active-invariant` reproduce identically on untouched `main` (c35e7b90fe) and are not from this branch.

### Found in review, fixed before this opened

An adversarial pass over the first draft surfaced eight real defects that are worth naming, because most were mine and several were serious: the disposable check locking existing users out of sign-in; the captcha guarding only one of the two public endpoints; a single-use token replayed on every retry; a Cloudflare outage returning 500 on the primary sign-in path; `request.ip` sending the reverse-proxy address to Cloudflare; a site-key-only config rendering a challenge nothing verified; a blank env value switching the challenge on; and the rejection surfacing as "Something went wrong" because its message was a dead string.

Sign-up now refuses disposable addresses where it previously accepted them. `AP_ALLOW_DISPOSABLE_EMAILS=true` restores the old behaviour, and the change is written up in `docs/install/reference/breaking-changes.mdx` under 0.87.0. Turnstile is not breaking: with no keys set nothing changes.


AhmadTash:
@greptile both real, both mine, both fixed.

**Password sign-up required a token its own form never obtained.** I added `assertSolved` to `POST /sign-up` to close the bypass you flagged in the previous round — a bot can call that endpoint directly and never touch the card — but I did not mount the widget on `SignUpForm`. So on a deployment with keys configured, every password registration would have failed a challenge it was never shown. That is exactly the shape the repo's self-hosting rule warns about: a feature that looks enabled and is silently broken.

The form now renders the widget, sends the token, and issues a fresh one when an attempt is refused — the same single-use handling as the code path, which it needed for the same reason.

**Unstable callback tearing down the challenge.** Correct: `onUnavailable` was an inline arrow, so it took a new identity on every render. The email field is watched for the work-address hint, so every keystroke re-ran the widget effect and removed a challenge the visitor may already have solved. Memoised with `useCallback`, and the same treatment on the sign-up form.

Worth noting the pairing, since it is the lesson rather than the bug: the first fix moved enforcement to a second endpoint without moving the affordance with it. Guarding an endpoint and serving its challenge are one change, not two.

Suites: CE 755, Cloud 500. Size 377/600 api, 202/1200 web.

@greptile review

packages/server/api/src/app/authentication/lib/disposable-email.ts
```diff
@@ -0,0 +1,55 @@
+import { ActivepiecesError, ErrorCode } from '@activepieces/core-utils'
```
MrChaker, line 1: @AhmadTash I do not know what is this change but here is what claude says:

Bug The captcha-rejection message mapping was only added to the emailed-code card, not the password sign-up form: sign-up-form.tsx's onError switch has no VALIDATION case, so a rejected/expired token on /sign-up falls to "Something went wrong" — the exact "dead string" defect the PR description says was fixed. The widget resets so retry works, but the message is wrong on one of the two guarded surfaces.

Security The fail-open boundary in turnstile.ts:359-372 is broader than the stated design ("only a verdict Cloudflare actually returned refuses"). Axios throws on any non-2xx and tryCatch treats every throw as transport failure → allow; a siteverify HTTP 4xx (malformed/oversized input, WAF block) is a Cloudflare response yet passes. captchaToken also has no .max() in either DTO. Exploitability is low (siteverify normally returns 200 + success:false), but distinguish error.response present (fail closed) from network error/timeout (fail open).

Minor Account-enumeration oracle scoped to disposable domains on POST /otp/request: the check runs only when isNil(existingIdentity), so for a listed domain 403 = "no account", 204 = "account exists". Deliberate trade-off (checking earlier would lock out grandfathered accounts), mitigated by rate limit + captcha — record it as accepted, since the old unconditional 204 leaked nothing.

Minor False-positive escape hatch is all-or-nothing: a wrongly-listed corporate domain leaves a self-hoster with AP_ALLOW_DISPOSABLE_EMAILS=true (disables the control entirely) or inviting every user; no per-domain allowlist, list freshness tied to manual bumps of the pinned package. The invitation override (hasAnyAcceptedInvitationsForEmail) is platform-unscoped — acceptable for a boolean existence check, but sidesteps the letter of the data-isolation rule.

Minor In turnstile-widget.tsx, if the script loads but window.turnstile is undefined, the effect returns silently: no widget, no token, submit disabled forever with no explanation — the failed path only covers script.onerror.

MrChaker (changes requested):

MrChaker (approved):

othmanemad (approved):

PR #14708, opened 2026-08-09
feat(auth): finish sign-up by asking for a name

AhmadTash:
## Description

**5 of 9 in the passwordless sign-in stack.** <details><summary><b>The stack</b> — review bottom-up; each PR targets the one below it</summary>

| | PR | Area |
|---|---|---|

retarget inside a stack).

</details>

A verified identity with no platform holds a pre-platform session and nothing else. This is the endpoint that turns it into an account: it takes a full name, derives first and last from it, names the platform after the first name (`Ahmad Tash` → `Ahmad's`), and returns a real session.

Completion reports `USER_SIGNED_UP` rather than `USER_SIGNED_IN`, which is what the path had been claiming for people who had just created an account.

**Whether a request completed the sign-up is decided inside the platform-creation lock and handed back**, not re-derived from a read of its own. That matters for two cases:

- **Two simultaneous submissions** would otherwise both conclude they were first, then race to write the name — last write wins on somebody's account.
- **A replayed onboarding token** would rename an established account to whatever the replay submitted, and announce a second sign-up for an account that had signed up once.

Exactly one request writes the name and emits the event. The rest still receive the session, because the token deliberately stays valid so an honest double submit is answered rather than refused with a 401.

Split out of #14703 so both stay inside the review budget, and it lines the backend up with the frontend seam: #14689 is sign-in by code, #14690 is the name step, and this is the endpoint behind it.

## How was this tested?

Four integration cases: the platform is created and named from the first name; a double-submitted name step yields one platform; the other onboarding route racing the name step yields one platform; and a replayed completion neither renames the account nor re-announces.

The replay case was verified to fail without the fix (`expected 'Someone' to be 'Ahmad'`). The duplicate-event half is not asserted in the CE suite because application events are edition-gated there and the table stays empty — both halves are gated by the same flag, so the rename assertion exercises the guard.

I also wrote a concurrent-completion test and deleted it: it asserted the final name was one of the two submitted, which is true with or without the fix. A test that passes either way is not evidence.

### DIFF
--- a/packages/server/api/src/app/authentication/passwordless-auth.service.ts
+++ b/packages/server/api/src/app/authentication/passwordless-auth.service.ts
@@ -131,23 +131,15 @@ export const passwordlessAuthService = (log: FastifyBaseLogger) => ({
             callerTokenVersion: undefined,
             beforeProvision: writeNames,
         })
-        // POST /v1/platforms takes the same lock and provisions without a name, so
-        // it can win and leave the seeded placeholder behind for good. If nothing
-        // has named this identity yet, the name step's submission still applies.
-        if (!provisioned && await carriesSeededName({ identityId, log })) {
-            await writeNames()
-        }
+        // Deliberately no fallback for the case where POST /v1/platforms wins the
+        // lock and provisions without a name. Detecting "never named" from the
+        // stored values cannot be done: a member at ahmad@example.com who signs
+        // up as "Ahmad" persists exactly the seeded shape, so any such check lets
+        // a replay rename an established account — a worse failure than the one it
+        // would fix, and reachable only by an API caller racing the name step.
         return { response, signedUp: provisioned }
     },
 })
-
-// The placeholder requestCode seeds is exactly the email local part with no
-// surname, so an identity still carrying it has never been through a name step.
-async function carriesSeededName({ identityId, log }: CarriesSeededNameParams): Promise<boolean> {
-    const identity = await userIdentityService(log).getOneOrFail({ id: identityId })
-    return identity.firstName === signupNames.firstNameFromEmail(identity.email)
-        && identity.lastName === ''
-}
 
 async function assertPlatformAuthIsOpenTo({ email, platformId, log }: PlatformGateParams): Promise<void> {
     await authenticationUtils(log).assertEmailAuthIsEnabled({
@@ -172,11 +164,6 @@ async function mayJoinPlatform({ email, platformId, identity, log }: MayJoinPlat
 type RequestCodeParams = {
     email: string
     platformId: string | null
-}
-
-type CarriesSeededNameParams = {
-    identityId: string
-    log: FastifyBaseLogger
 }
 
 type CompleteSignUpResult = {
--- a/packages/server/api/src/app/authentication/passwordless-auth.service.ts
+++ b/packages/server/api/src/app/authentication/passwordless-auth.service.ts
@@ -120,9 +120,6 @@ export const passwordlessAuthService = (log: FastifyBaseLogger) => ({
         const writeNames = async (): Promise<void> => {
             await userIdentityService(log).updateNames({ id: identityId, firstName, lastName })
         }
-        // Written inside the provisioning lock by the call that provisions, so a
-        // replay cannot rename an established account and an interruption after
-        // the account exists cannot lose it.
         const { response, provisioned } = await platformService(log).createPlatformWithProject({
             identityId,
             name: signupNames.platformNameFromPerson({ firstName, email: identity.email }),
@@ -131,12 +128,6 @@ export const passwordlessAuthService = (log: FastifyBaseLogger) => ({
             callerTokenVersion: undefined,
             beforeProvision: writeNames,
         })
-        // Deliberately no fallback for the case where POST /v1/platforms wins the
-        // lock and provisions without a name. Detecting "never named" from the
-        // stored values cannot be done: a member at ahmad@example.com who signs
-        // up as "Ahmad" persists exactly the seeded shape, so any such check lets
-        // a replay rename an established account — a worse failure than the one it
-        // would fix, and reachable only by an API caller racing the name step.
         return { response, signedUp: provisioned }
     },
 })
--- a/packages/server/api/src/app/authentication/passwordless-auth.service.ts
+++ b/packages/server/api/src/app/authentication/passwordless-auth.service.ts
@@ -2,6 +2,7 @@ import { ActivepiecesError, ErrorCode, isNil } from '@activepieces/core-utils'
 import { cryptoUtils } from '@activepieces/server-utils'
 import { ApFlagId, AuthenticationResponse, OtpType, TelemetryEventName, UserIdentity, UserIdentityProvider } from '@activepieces/shared'
 import { FastifyBaseLogger } from 'fastify'
+import { distributedLock } from '../database/redis-connections'
 import { flagService } from '../flags/flag.service'
 import { rejectedPromiseHandler } from '../helper/promise-handler'
 import { system } from '../helper/system/system'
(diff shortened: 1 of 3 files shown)


AhmadTash:
@greptile real, and fixed — the ordering was the whole problem.

Writing the name *after* provisioning meant a failure in that window lost it permanently: the retry found the account already made, `createPlatformWithProject` correctly reported it had provisioned nothing, and completion skipped the name its owner had actually typed. The account keeps the placeholder derived from the address, forever, and no amount of retrying recovers it. That is the opposite of the convergence property #14702 exists to provide.

The name now goes in **before** the platform. A retry after any interruption either finds the name already written, or writes it again with the same value.

Doing that reintroduced the concurrency question you raised last round, so completion is now serialised on its own key — `complete-sign-up-${identityId}` — with the platform lock nested inside it. The order is always completion then platform, never the reverse, so the two cannot deadlock.

**What I am not claiming:** `USER_SIGNED_UP` can still be missed if the process dies between provisioning and the response. Making that recoverable needs durable "has this been announced" state, which is an outbox for an audit event and more machinery than the problem deserves. The name — the thing the user actually chose and would notice — is what I made recoverable.

**On coverage, plainly:** I wrote a test for this and deleted it. It asserted the name is written after a normal completion, which is true with the old ordering too — the property is that the write *precedes* provisioning, and reaching it needs a fault injected between two statements. I would rather say the reordering stands on reading than ship a test that looks like proof of it and is not. The replay case, which is testable, is still covered and still fails without its fix.

@greptile review

AhmadTash:
@greptile the competing-route race was real and is fixed — and fixing it properly removed the lock I had added, which is a better outcome than the one I was defending.

**The hole:** my completion lock keyed on `complete-sign-up-${identityId}` serialised completions against each other but not against `POST /v1/platforms`, the other route that provisions the same identity. So a completion could read "no platform yet", watch that route create the account, and then rename it — having provisioned nothing.

Two locks that have to agree is the smell. There is now one critical section: `createPlatformWithProject` takes work from the caller and runs it **inside its own lock, only when this call is the one provisioning**. Completion hands it the name write.

That single change closes three things at once:

- the platform route can no longer be raced, because both go through the same lock;
- a replay cannot rename an established account, because it does not provision;
- an interruption after the account exists cannot lose the name, because the write happens ahead of creation inside that section.

The completion lock and its pre-check are gone.

**On the remaining point — `USER_SIGNED_UP` tied to the provisioning result — my position is unchanged and I want to be plain about it rather than keep restating it as if it were solved.** If the process dies between provisioning and the response, the event is missed. Making that recoverable means durable "has this been announced" state, which is an outbox for an audit event. I am not building one inside a sign-in stack, and I would rather leave it stated than dress up a partial fix. The name, which is what the user chose and would notice, is recoverable.

Size 158/600. Suites: CE 757, Cloud 500.

@greptile review

AhmadTash:
@greptile the competing-route name loss was real and is fixed. The event point stands as stated.

**Name loss — fixed.** You traced it correctly: `POST /v1/platforms` takes the same provisioning lock and creates the account with no name, so winning that race left the seeded placeholder in place permanently. Completion then found a linked user, correctly reported it had provisioned nothing, and skipped the name its owner had typed.

The name step now also applies when nothing has named the identity yet. The placeholder `requestCode` seeds is exactly the email local part with an empty surname, so an identity still carrying it has never been through a name step — which is precisely the state worth writing over, and distinguishes it from an established name a replay must not touch.

Verified with a test that fails without the fix (`expected 'Ahmad' to be 'Zoe'` — the placeholder surviving).

Worth recording how that test had to be written, because it changes what the finding means in practice. Driving it over HTTP is impossible: `POST /v1/platforms` rotates the token version, so a *sequential* completion afterwards cannot even authenticate — it gets a 403. The scenario only exists when both requests authenticate before the rotation, i.e. genuinely concurrently, which is why the test drives the two services directly. Real, but narrower than a sequential reading suggests.

**`USER_SIGNED_UP` durability — unchanged, and I would rather keep saying so plainly than restate it as if it were solved.** If the process dies between the commit and the dispatch, the event is missed, and a replay reports `signedUp: false` so it is not retried. Making that recoverable means durable "has this been announced" state, which is an outbox for an audit event. I am not building one inside a sign-in stack. The name — the thing the member chose and would notice — is what I made recoverable.

CE auth+platform 44, Cloud 500.

@greptile review

AhmadTash:
@greptile you were right, and the fix I shipped last round was worse than the bug it fixed. Reverted.

`carriesSeededName` treated an email-derived first name with an empty surname as proof that nobody had named the identity. That shape is **ambiguous**, exactly as you traced: a member at `ahmad@example.com` who signs up as the single word `Ahmad` persists `firstName: 'Ahmad', lastName: ''` — byte-identical to what `requestCode` seeds. So a replayed onboarding token would have renamed an established account, which is the failure the guard exists to prevent. I traded a rare lost name for a rare unwanted mutation, and the mutation is the worse of the two.

The inference is gone. There is now a test that fails against my own previous commit — complete as `Ahmad`, replay as `Someone Else`, assert the account is still `Ahmad` (`expected 'Someone' to be 'Ahmad'` with the heuristic reinstated).

**What that leaves open, stated rather than papered over:** if `POST /v1/platforms` wins the provisioning lock, the account keeps the address-derived placeholder and the typed name is lost. I am not fixing it by guessing at stored values. Telling "never named" from "named, and it happens to look seeded" needs a durable marker — a flag or timestamp recording that a name step completed — which is a schema change, and it belongs with the single-use onboarding credential rather than bolted on here. In practice it needs an API caller racing the name step: the card only calls `/complete-sign-up`, and `POST /v1/platforms` rotates the token, so a sequential second call cannot even authenticate.

Two rounds, two wrong answers from me on this one line: first writing the name after provisioning (lost on interruption), then inferring whether to rewrite it (renames established accounts). The version now in the branch does the one thing that is unambiguously correct — the call that provisions writes the name, inside the lock — and says plainly what it does not cover.

CE auth+platform 44, Cloud 500.

@greptile review

AhmadTash:
@greptile the revert is in, and the remaining P1 is not a property of this route — it is the contract of the dispatch layer it calls.

`applicationEvents.sendUserEvent` is declared `: void` and wraps its body in `rejectedPromiseHandler` (`helper/application-events.ts:49-57`). Publication failure is swallowed there, by design, for every caller. Nothing a route can do at the call site changes that; an at-least-once guarantee has to be an outbox *inside* that helper.

And the shape you are flagging is the shape of the route 90 lines above mine, in the same file:

- `/sign-up` (`authentication.controller.ts:31-49`) — `signUp()` commits the account, then `sendUserEvent` fires afterwards, outside any transaction. Retry it and `userIdentityService` throws `EXISTING_USER` (`user-identity-service.ts:22`), so a `/sign-up` whose event was lost to a crash can never re-fire it either.
- Same for the other four dispatch sites: `managed-authn-controller.ts:30`, `authn-sso-saml-controller.ts:41`, `federated-authn-module.ts:40`.

`/complete-sign-up` is stricter than all five: it fires only when `signedUp` is true, so the one failure mode unique to a resumable signup — the event firing twice because completion ran twice — is the one this PR actually closes.

So I am not fixing it here. Making signup events crash-proof is one change to `applicationEvents` covering all six sites, it belongs behind its own PR and test, and shipping a bespoke retry on this one route would leave the other five broken while implying they are not. It is on the PR as a named follow-up.

Please re-score against `d85b78d` — the 4/5 rationale is entirely this item.

packages/server/api/src/app/authentication/passwordless-auth.service.ts
```diff
@@ -113,6 +114,31 @@ export const passwordlessAuthService = (log: FastifyBaseLogger) => ({
         return authenticationUtils(log).getOnboardingResponse({ identityId: verifiedIdentity.id })
     },
 
+    async completeSignUp({ identityId, fullName }: CompleteSignUpParams): Promise<CompleteSignUpResult> {
+        const identity = await userIdentityService(log).getOneOrFail({ id: identityId })
+        const { firstName, lastName } = signupNames.splitFullName({ fullName, email: identity.email })
+        const writeNames = async (): Promise<void> => {
+            await userIdentityService(log).updateNames({ id: identityId, firstName, lastName })
+        }
+        // Written inside the provisioning lock by the call that provisions, so a
```
MrChaker: comments

MrChaker (approved):

othmanemad (approved):

MrChaker (approved):

PR #14720, opened 2026-08-10
feat(flows): honor optional externalId on POST /flows

### LINKED ISSUE #14736: [Feature]: Honor optional externalId on POST /flows
## Problem

POST /api/v1/flows always mints a server-generated externalId. `flowService.create()` accepts an `externalId` param (`externalId ?? apId()`, flow.service.ts) but only internal callers (git-sync, project-releases import) reach it — the controller never forwards it and `CreateFlowRequest` has no such field (verified in code).

External tooling provisioning flows via the public API (git-backed CI/CD) can't set a stable externalId → next deploy's lookup by externalId misses → duplicate flow per redeploy.

Ref: Pylon 5605 — customer flags this as urgent, on the critical path of their CI/CD rollout.

## Proposed

Optional `externalId` on `CreateFlowRequest`, forwarded by the controller into `flowService.create()`. Omitted → current auto-generate behavior, fully backward-compatible. Matches the internal import/releases path.

Community PR open: activepieces/activepieces#14720 (touches `packages/core/execution` dto + `flow.controller.ts`)

### LINKED ISSUE #14791: Honor optional externalId on POST /flows (community PR #14720)
Supporting ticket for community PR activepieces/activepieces#14720.

**Problem**: `externalId` can be set through the public API for tables and connections, but `POST /api/v1/flows` always mints a server-generated value. Public API callers (git-backed CI/CD tooling) therefore can't create a flow with a stable `externalId`, so each redeploy fails the lookup and creates a duplicate flow.

**PR**: adds optional `externalId` to `CreateFlowRequest`, forwards it to `flowService.create()` (which already supports it via the internal git-sync / project-releases import path). Also adds a unique index migration on `(projectId, externalId)`.

**Review notes**:

* The internal git-sync / project-releases import path already sets `externalId`; this change brings flows in line with the public API behavior for tables and connections.
* PR includes a SQLite migration file; new migrations are Postgres-only.

AdamSelene:
## Problem

`POST /api/v1/flows` always assigns a server-generated `externalId` (`apId()`). The `externalId` param on `flowService.create()` is only reachable by internal callers (git-sync / project-releases import) — the HTTP controller never forwards it, and `CreateFlowRequest` has no `externalId` field.

As a result, external tooling that provisions flows through the public API (e.g. a git-backed CI/CD deploy) cannot create a flow with a known, stable `externalId`. On the next deploy it looks the flow up by its intended `externalId`, doesn't find it (the server minted a different one), and creates a **duplicate** — one extra flow per redeploy.

## Change

Make `externalId` an **optional** field on `CreateFlowRequest` and forward it from the controller into `flowService.create()`. The service already supports the param (`externalId: externalId ?? apId()`), so nothing changes when the field is omitted.

- `packages/core/execution/src/lib/flows/dto/create-flow-request.ts` — add `externalId: z.string().optional()`
- `packages/server/api/src/app/flows/flow/flow.controller.ts` — pass `externalId: request.body.externalId`

Backward-compatible: omitting `externalId` keeps the current auto-generate behavior. When provided, callers can create a flow with a known externalId so subsequent syncs update it in place instead of duplicating (matching how the internal import/releases path already behaves).

## Notes
- Minimal change; I have not run the full server test suite locally.

### DIFF
--- a/packages/core/execution/src/lib/flows/dto/create-flow-request.ts
+++ b/packages/core/execution/src/lib/flows/dto/create-flow-request.ts
@@ -9,6 +9,10 @@ export const CreateFlowRequest = z.object({
     projectId: z.string(),
     templateId: z.string().optional(),
     metadata: z.optional(Metadata),
+    /**Optional stable identifier for the flow. When omitted the server generates one.
+     * Lets callers (e.g. git-sync / external deploy tooling) create a flow with a known
+     * externalId so redeploys can find and update it in place instead of duplicating. */
+    externalId: z.string().optional(),
 })
 
 export type CreateFlowRequest = z.infer<typeof CreateFlowRequest>
--- a/packages/server/api/src/app/database/migration/postgres/1821000000000-AddFlowProjectIdExternalIdUniqueIndex.ts
+++ b/packages/server/api/src/app/database/migration/postgres/1821000000000-AddFlowProjectIdExternalIdUniqueIndex.ts
@@ -17,9 +17,10 @@ export class AddFlowProjectIdExternalIdUniqueIndex1821000000000 implements Migra
         // already matches another flow's externalId (externalId and id share the apId
         // space), which would itself break the unique index. Each pass appends the row's
         // unique id, so any collision it introduces is a strictly smaller duplicate set
-        // resolved on the next pass; this converges in one pass on real data. The bound is
-        // a safety valve — if duplicates somehow remained, the index creation below fails
-        // loudly rather than silently leaving them.
+        // resolved on the next pass; this converges in one pass on real data. The pass
+        // bound is the flow count (+1), which is provably sufficient: each pass makes at
+        // least the longest-suffixed duplicate globally unique, so at most one pass per row
+        // is ever needed — the loop can never stop early with duplicates still present.
         const HAS_DUPLICATE = `
             SELECT 1 FROM (
                 SELECT ROW_NUMBER() OVER (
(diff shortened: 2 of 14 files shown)


majewskibartosz:
@greptileai review

majewskibartosz:
@AdamSelene hi can you please address the greptile review, after re-opening it ran a full scan

AdamSelene:
@greptile Acknowledged, and we're keeping the concurrent build deliberately — here's the reasoning.

**The race is real but narrow, and self-correcting rather than fatal:**

- **Single-instance installs (the common self-host) can't hit it.** The one app instance runs the migration at boot and isn't serving flow-creates concurrently, so there's no writer to insert a duplicate during the build. No window.
- **When it does trigger (multi-replica rolling deploy, an old replica minting a duplicate `(projectId, externalId)` during the build), it self-heals.** `transaction = false` means a failed build is *not* recorded, so the next boot re-runs the migration, detects the `INVALID` index, drops it (`DROP INDEX CONCURRENTLY IF EXISTS`), re-runs the dedup pass, and rebuilds. It converges once the old (non-enforcing) replicas are retired. This invalid-index detect/drop/retry path is already implemented in the migration.
- **We keep `CONCURRENTLY` over a blocking build** to avoid stalling all `flow`-table writes for the duration of index creation during a deploy. A blocking build would close the race but reintroduce that write stall on large installs.

The trade-off and the self-heal reasoning are now documented inline at the build site in `1839000000000-AddFlowProjectIdExternalIdUniqueIndex.ts`. The only tail case is a stuck deploy if an old replica keeps minting exact-duplicate externalIds throughout the build window — low probability, resolved by deploy completion or retry.

amrdb:
Closing this and merging #15162

PR #14737, opened 2026-08-11
refactor(agent): agent tool execution is now server-side only

hazemadelkhalel:
Deletes the agent loop the engine used to carry. Every agent step now runs on the server, and #14730 moved the existing ones, so none of this is reachable.

**505 lines removed, 1 added.**

## What goes

- `engine/src/lib/tools/index.ts` (427) — tool construction, the dependency-wave prop extraction, dropdown option loading, dynamic-property inlining
- `engine/src/lib/tools/tsort.ts` (56) — the topological sort behind it
- `piece-executor.ts` — the `agent.tools` context hook that handed it to a piece
- `AgentContext` and `ConstructToolParams` in `pieces-framework` — the types that named the hook

## Why it is safe now

The server has run every tool kind since #14717, #14718 and #14719, the piece became a thin client in #14699, and #14730 rewrote every stored step onto `0.6.0` with its connections by id. Nothing calls `context.agent` any more: the only match left in the repo is a stale `dist/` artifact of the ai piece, not source.

`ActionContext` loses a public member. No piece in this repo uses it, and a piece pinned to an older framework version bundles its own copy, so nothing in the registry is affected.

## Verification

`pieces-framework` builds; `api`, `worker` and `piece-ai` typecheck clean; `turbo lint` 0 errors on engine and framework. The engine's own tsconfig reports 16 errors both with and without this change — all pre-existing, in `engine-file-api.ts` and unrelated.

Removes an unreachable code path and a context member no piece uses.

Deletion only. The path that ran LLM calls and credentials inside the engine is gone, which narrows the surface rather than widening it.


hazemadelkhalel:
Both fixed.

**Orphaned test** — good catch, and I missed it because I typechecked `tsconfig.lib.json` and the engine's test tree is not in it. `normalizeAgentToolAuth` lived only in the deleted module, so the test is deleted with it rather than migrated; there is nothing left for it to cover.

Verified by running the engine suite both ways: **28 failed / 423 passed** with and without this change, identical. The failures are pre-existing and unrelated — the file count differs by one only because a file was removed.

**Title** — fair. Changed to `refactor(agent): agent tool execution is now server-side only`, which names the outcome rather than the file that went. Kept the Conventional Commits prefix, and moved it from `chore` to `refactor` since it is an architectural change rather than housekeeping.

PR #14738, opened 2026-08-12
feat(ai-providers): multiple keys per provider with model and project scopes

Louai-Zokerburg:
Linear: ENG-461 · Milestone "2. Providers redesign — connect keys, pick models, scope projects"

## Description

First of two PRs for the AI providers redesign (the stacked follow-up ships the new AI Center UI). This one makes the `ai_provider` data model multi-key and scoped:

- **Multiple keys per provider**: the `(platformId, provider)` unique index is dropped (recreated non-unique), so a platform can hold several configurations of one provider (e.g. two Azure resources with different endpoints).
- **Per-key scoping**: each key carries a model allow-list (`modelScope: all|selected` + `modelIds[]`) and a project scope (`projectScope: all|selected|except` + `projectIds[]`, GIN-indexed following the `app_connection` pattern). New keys default to `all`/`all` — zero setup, existing rows keep working unchanged.
- **Deterministic runtime resolution** (no priority field): among a project's eligible keys, the most specific project scope wins (`selected` > `except` > `all`), newest `created` breaks ties. `GET /:provider/config` and `GET /:provider/models` resolve against the ENGINE principal's `projectId`; USER calls stay platform-wide so the admin UI sees the full picture. `listModels` filters to the resolved key's allow-list for engine calls only.
- **Invariants kept**: the managed ACTIVEPIECES provider stays a singleton (explicit `create()` guard, 409 `aiProvider.activepiecesIsManaged`), and the single-`enabledForChat` transaction still holds across keys of one provider.
- **Latent bug fixed**: `update()` used to re-validate credentials by re-resolving the provider *by name* — the wrong row once multi-key exists. It now validates against the row's own auth, and skips provider validation entirely for name-/scope-only updates (no more pointless provider API calls when toggling scopes).
- Shared: `AIProviderWithoutSensitiveData` / `UpdateAIProviderRequest` gain the four fields; new `AiProviderModelScope` / `AiProviderProjectScope` enums; `@activepieces/shared` bumped to 0.130.0.
- Brain: decision record `providers-redesign-before-routing.md` + multi-key gotchas in `brain/ai-intelligence/ai-providers.md` ride this PR.

**Migration** `1821000000000-AddAiProviderScopes`: additive columns with defaults + index swap, `breaking = false`; `down()` restores the unique index (will intentionally fail if multiple keys already exist — that data state can't be un-invented).

## How was this tested?

- New integration tests (CE): two keys for one provider, scope precedence (selected beats all), newest-wins tie-break, `except` include/exclude, ACTIVEPIECES create rejection, single-chat invariant across keys, engine-vs-admin model filtering — 24/24 pass. EE `agent-credit-gate` suite passes (its mock seeder switched from upsert to save; the upsert's ON CONFLICT target was the dropped unique index).
- `check-migrations` clean (schema matches entities); migration `down()` proven by live revert + re-run.
- `api`, `web`, `worker`, `engine` all build against the widened shared types; `npm run lint-dev` clean.

Fixes # (issue)

**Security note:** this PR changes which credential row the engine resolves for a project (permissions-adjacent) and touches the decryption path (`decryptRowAuth` extraction). Mitigation: scoping only ever *narrows* access — before this PR every project got the platform's single key; now a project gets a key only if its scope allows, and a no-eligible-key state 404s rather than falling back. Endpoint auth (`engine()` / `platformAdminOnly`) is unchanged, secrets stay AES-encrypted at rest, and decrypted auth is still only returned on the engine-only config route.


packages/server/api/src/app/ai/ai-provider-entity.ts
```diff
@@ -50,12 +54,36 @@ export const AIProviderEntity = new EntitySchema<AIProviderSchema>({
...
     indices: [
         {
             name: 'idx_ai_provider_platform_id_provider',
             columns: ['platformId', 'provider'],
-            unique: true,
+        },
+        {
+            name: 'idx_ai_provider_project_ids_gin',
+            columns: ['projectIds'],
+            synchronize: false,
```
othmanemad: Is this index used anywhere in the stack or will be used in the future?

packages/server/api/src/app/ai/ai-provider-entity.ts
```diff
@@ -50,12 +54,36 @@ export const AIProviderEntity = new EntitySchema<AIProviderSchema>({
...
+        },
+        projectScope: {
+            type: String,
+            nullable: false,
+            default: 'all',
+        },
+        projectIds: {
+            type: String,
+            array: true,
+            nullable: false,
```
othmanemad, line 76:
```suggestion
            nullable: false,
            default: '{}'
```

packages/server/api/src/app/ai/ai-provider-entity.ts
```diff
@@ -50,12 +54,36 @@ export const AIProviderEntity = new EntitySchema<AIProviderSchema>({
...
         },
+        modelScope: {
+            type: String,
+            nullable: false,
+            default: 'all',
+        },
+        modelIds: {
+            type: String,
+            array: true,
+            nullable: false,
```
othmanemad, line 65:
```suggestion
            nullable: false,
            default: '{}'
```

packages/server/api/src/app/database/migration/postgres/1829000000000-AddAiProviderScopes.ts
```diff
@@ -0,0 +1,96 @@
...
+            ADD COLUMN IF NOT EXISTS "modelScope" character varying NOT NULL DEFAULT 'all',
+            ADD COLUMN IF NOT EXISTS "modelIds" character varying array NOT NULL DEFAULT '{}',
+            ADD COLUMN IF NOT EXISTS "projectScope" character varying NOT NULL DEFAULT 'all',
+            ADD COLUMN IF NOT EXISTS "projectIds" character varying array NOT NULL DEFAULT '{}'
+        `)
+        await queryRunner.query(`
+            ALTER TABLE "ai_provider"
+            ALTER COLUMN "modelIds" DROP DEFAULT,
+            ALTER COLUMN "projectIds" DROP DEFAULT
+        `)
```
othmanemad:
```suggestion
```

This is no longer needed since we add '{}' default to the entity itself

packages/server/api/src/app/database/migration/postgres/1830000000000-AddAiProviderScopes.ts
```diff
@@ -0,0 +1,96 @@
...
+            DROP COLUMN "projectScope",
+            DROP COLUMN "projectIds"
+        `)
+        await queryRunner.query(`
+            DROP INDEX IF EXISTS "idx_ai_provider_platform_id_provider"
+        `)
+        await queryRunner.query(`
+            CREATE UNIQUE INDEX "idx_ai_provider_platform_id_provider"
+            ON "ai_provider" ("platformId", "provider")
+        `)
```
othmanemad, line 90: `down()` recreates `idx_ai_provider_platform_id_provider` as UNIQUE, but `up()`
dropped it precisely so a platform can hold several keys per provider. Once
anyone adds a second key for the same provider, this rollback errors mid-flight —
the data no longer fits the old shape.

Two ways out:
- set `breaking = true` on the migration, so `rollback-migrations.ts` and the
  release note flag it as one-way, or
- dedupe in `down()` before recreating the index — which means deleting keys a
  self-hoster added.

packages/server/api/src/app/ai/ai-provider-service.ts
othmanemad: We should apply the same ranking as the resolveEligibleRow functionality

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -250,13 +193,143 @@ async function shouldHideActivepiecesAiProvider({ platformId, log }: { platformI
...
+        config: {},
+        provider: AIProviderName.ACTIVEPIECES,
+        displayName: 'Activepieces',
+        platformId,
+        enabledForChat: !hasChatProvider,
+        modelScope: 'all',
+        modelIds: [],
+        projectScope: 'all',
+        projectIds: [],
+    })
```
othmanemad: Since we dropped the unique index, this might introduce a race condition where the activepieces provider can be duplicated, let's add a distributed lock just in case

othmanemad (changes requested):

packages/server/api/src/app/ai/ai-provider-service.ts
othmanemad: Or we can create a partial unique index

packages/server/api/src/app/ee/agent/agent-helpers.ts
```diff
@@ -62,15 +62,27 @@ async function getUserProjects({ platformId, userId, log }: { platformId: string
     return allProjects.filter((p) => p.type !== ProjectType.PERSONAL || p.ownerId === userId)
 }
 
-async function resolveRunProvider({ platformId, provider, log }: { platformId: string, provider?: AIProviderName, log: FastifyBaseLogger }): Promise<GetProviderConfigResponse> {
+function providerScopeFor({ projectId }: { projectId: string | null }): ProviderScope {
+    return isNil(projectId) ? { type: 'platform' } : { type: 'project', projectId }
+}
```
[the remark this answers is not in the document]
Louai-Zokerburg: Fixed in bc7785f472. `providerScopeFor` is gone; `runScopeOrThrow` refuses a run that has no project instead of coercing it to platform scope, so an omitted project can only narrow the eligible keys, never widen them. Call sites updated: chat admission (`agent-conversation-controller`), the worker RPC (`agent-rpc-handlers`), the agent step (`agent-run-controller`) and agent drafting.

The analytics/billing paths were the only reason the nullable branch existed — they want the provider *name*, not a credential. `resolveChatProviderName` now takes the nullable project id itself and returns `null` when there is none, rather than naming a platform-wide key the run was never allowed to use. Covered by two unit tests in `agent-model-resolution.test.ts`.

packages/server/api/src/app/ai/ai-provider-entity.ts
Louai-Zokerburg: Done in f5090ca665 — both `modelIds` and `projectIds` now carry `default: '{}'` on the entity.

packages/server/api/src/app/ai/ai-provider-entity.ts
Louai-Zokerburg: Done in f5090ca665.

packages/server/api/src/app/database/migration/postgres/1829000000000-AddAiProviderScopes.ts
Louai-Zokerburg: Removed in f5090ca665 — the `ALTER COLUMN ... DROP DEFAULT` block is gone now that the entity declares the default.

packages/server/api/src/app/database/migration/postgres/1830000000000-AddAiProviderScopes.ts
Louai-Zokerburg: Took the first option: `breaking = true` in f5090ca665, with the reason in a comment above the class. Deduping would mean deleting keys a self-hoster added, so the migration is declared one-way; `rollback-migrations.ts` now requires `--force` and prints the data-loss warning for it. `down()` stays as-is so a platform that never added a second key can still roll back.

packages/server/api/src/app/ai/ai-provider-service.ts
Louai-Zokerburg: Went with the partial unique index (f5090ca665): `idx_ai_provider_platform_id_managed` on `("platformId") WHERE provider = 'activepieces'`, declared on the entity and built `CONCURRENTLY` in the migration. The managed row is inserted with `ON CONFLICT DO NOTHING`, so the losing racer is a no-op instead of a 500 — and unlike a lock the invariant holds even for a writer that never takes it.

othmanemad (approved):

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -192,71 +186,179 @@ export const aiProviderService = (log: FastifyBaseLogger) => ({
...
 
-async function shouldHideActivepiecesAiProvider({ platformId, log }: { platformId: PlatformId, log: FastifyBaseLogger }): Promise<boolean> {
-    const { plan } = await platformService(log).getOneWithPlanOrThrow(platformId)
-    return plan.embeddingEnabled
+async function findRunKeyRow({ platformId, provider, scope, log }: { platformId: PlatformId, provider?: AIProviderName, scope: ProviderScope, log: FastifyBaseLogger }): Promise<AIProviderSchema | null> {
+    const chatRow = await findAvailableChatProviderRow({ platformId, scope, log })
+    if (isNil(provider) || chatRow?.provider === provider) {
+        return chatRow
+    }
+    return findEligibleRow({ platformId, provider, scope })
```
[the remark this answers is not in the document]
Louai-Zokerburg: Fixed in 9a8ef42876. A run resolves its credential through exactly one of two paths — the chat row (findAvailableChatProviderRow) or the ranked row for its provider (getConfigOrThrow) — so the check no longer picks one and now requires both to allow the target project. Nothing can be substituted in, and no key id has to travel on this branch. Covered by a CE integration test with a permissive chat key masking a restrictive named key, which fails against the previous behaviour. On feat/ai-provider-key-selection the run sends its providerConfigId, so the check reads that exact row.

MrChaker (approved):

PR #14746, opened 2026-08-12
fix(google-sheets): demote six legacy actions to audience 'human'

ibrahim-abuznaid:
## Description

Six legacy Google Sheets actions serve `audience: 'both'`, so they are offered to agents alongside a near-identical `audience: 'ai'` atomic that already covers the same operation. This demotes those six to `'human'`, which is what the other 20 legacy actions in the piece already do.

| Legacy action (`'both'`) | Display name | Atomic twin (`'ai'`) | Display name |
|---|---|---|---|
| `delete-multiple-rows` | Delete Multiple Rows | `sheets_delete_multiple_rows` | Delete Multiple Rows |
| `find-or-create-row` | Find or Create Row | `sheets_find_or_create_row` | Find or Create Row |
| `find-or-create-worksheet` | Find or Create Worksheet | `sheets_find_or_create_worksheet` | Find or Create Worksheet |
| `insert-row-at-top` | Add Row at Top | `sheets_add_row_at_top` | Add Row at Top |
| `read-data-range` | Read Data Range | `sheets_get_values` | Read Cell Range |
| `clear-rows` | Clear Row(s) | `sheets_clear_values` | Clear Cell Range |

Four of the six pairs share an identical display name, so an agent listing the piece sees the same capability twice with nothing to choose between them. The atomics are the intended agent surface: they take explicit IDs rather than builder-resolved dropdowns.

### How the inconsistency arose

The demotion was authored in #13941, together with the same change to the other 20 legacy actions. These six had also been given `audience: 'both'` three weeks earlier by the `aiMetadata` backfill in #14428. The two edits landed on different lines of the same object literal (`'human'` above `displayName`, `'both'` below `description`), so merging main into the #13941 branch united them into a duplicate object key rather than a textual conflict. The build fix in 3b8d5e5bc9 resolved the duplicate by keeping `'both'` — correct for the build, and the newer of the two lines on main, but it dropped the demotion for these six. They are now the only legacy actions in the piece not marked `'human'`.

### Scope

google-sheets only. A scan of every `createAction` block under `packages/pieces/community` on current main found no other piece where an `'ai'` action and a `'both'` action share an operation: #13941 is the only PR that performed these demotions, and google-sheets is the only one of #14428's 32 pieces that contains `'ai'` atomics at all. The other three pieces in #13941 (google-forms, notion, telegram-bot) are unaffected.

## How was this tested?

- `npx tsc --noEmit` over `packages/pieces/community/google-sheets/src/**/*.ts` with the root `tsconfig.base.json` path mappings: clean (the piece's own `tsconfig.lib.json` sets `"paths": {}` and cannot resolve the framework, so it is not a usable gate here).
- `npx eslint` on the six changed files: clean, no errors or warnings.
- Confirmed against `filterActionsByAudience` (`packages/server/api/src/app/pieces/metadata/utils/index.ts`) that the AI filter keeps `action.audience !== 'human'` and the human filter keeps `action.audience !== 'ai'`. So the six leave the agent surface and stay fully available in the builder.

Metadata-only change: no action logic, props, `aiMetadata`, or `outputSchema` was touched, and no action was removed. Existing flows that already use these actions keep working, since the human surface still serves them. Piece version bumped 0.16.7 to 0.16.8.

### DIFF
--- a/packages/pieces/community/google-sheets/package.json
+++ b/packages/pieces/community/google-sheets/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-google-sheets",
-  "version": "0.16.7",
+  "version": "0.16.8",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
--- a/packages/pieces/community/google-sheets/src/lib/actions/clear-rows.ts
+++ b/packages/pieces/community/google-sheets/src/lib/actions/clear-rows.ts
@@ -11,7 +11,7 @@ export const clearRowsAction = createAction({
 	displayName: 'Clear Row(s)',
 	description:
 		'Clears the contents of one or more rows without removing the rows themselves. Useful when you want to keep formatting and references stable.',
-	audience: 'both',
+	audience: 'human',
 	aiMetadata: {
 		description:
 			'Blanks the cell contents of one row or a contiguous row range in a worksheet without removing the rows, so row numbers and downstream references stay stable — prefer Delete Row or Delete Multiple Rows when the rows themselves should disappear, and Clear Sheet when the whole sheet should be emptied. Requires a selected spreadsheet and worksheet plus a 1-based starting row; leaving the ending row empty clears only that single row. Idempotent — re-running over the same range leaves it equally empty, but the erased values cannot be recovered.',
--- a/packages/pieces/community/google-sheets/src/lib/actions/delete-multiple-rows.ts
+++ b/packages/pieces/community/google-sheets/src/lib/actions/delete-multiple-rows.ts
@@ -10,7 +10,7 @@ export const deleteMultipleRowsAction = createAction({
 	displayName: 'Delete Multiple Rows',
 	description:
 		'Deletes a contiguous range of rows, or a list of specific row numbers. Row numbers are 1-based.',
-	audience: 'both',
+	audience: 'human',
 	aiMetadata: {
 		description:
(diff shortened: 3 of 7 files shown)


ibrahim-abuznaid:
Verified the impact on the live cloud agent surface via `ap_search_actions` (MCP, `mode: "semantic"`), scoped to `google-sheets`. Each affected pair comes back adjacent at ranks 1 and 2 under the same display name, separated by 0.013–0.026 cosine — effectively a coin flip for a model choosing a tool:

| Query | Rank 1 | Rank 2 |
|---|---|---|
| "delete several specific rows from a google spreadsheet" | **`delete-multiple-rows` 0.6526** (Delete Multiple Rows) | `sheets_delete_multiple_rows` 0.6269 (Delete Multiple Rows) |
| "read a specific range of cells from a spreadsheet using A1 notation" | `sheets_get_values` 0.6778 (Read Cell Range) | `read-data-range` 0.6625 (Read Data Range) |
| "create a worksheet tab only if it does not already exist" | `sheets_find_or_create_worksheet` 0.7157 (Find or Create Worksheet) | `find-or-create-worksheet` 0.7028 (Find or Create Worksheet) |
| "insert a new row at the top of the sheet just below the header" | `sheets_add_row_at_top` 0.8046 (Add Row at Top) | `insert-row-at-top` 0.7798 (Add Row at Top) |

In the first case the legacy action outranks the atomic, so an agent taking the top hit gets the builder-oriented action (dropdown-resolved `spreadsheetId`/`sheetId` props) rather than the atomic designed to take explicit ids.

The same probes also confirm the filter itself is working and that these six are the only legacy actions leaking through: the `audience: 'human'` legacy actions never appear in any result set (`delete_row`, `create-worksheet`, `find_rows`, `get-many-rows`, `find_spreadsheets`), while their `sheets_*` counterparts do. That is the intended end state for all 26 legacy actions — these six just never got there.

OdaiAhmed99 (approved):

PR #14748, opened 2026-08-12
perf(engine): stream backup log uploads instead of materializing the run state

MrChaker:
https://claude.ai/code/artifact/53fbca02-4e4f-4496-aaa5-f3659f96fe32


amrdb:
Quick note on `json-stream.ts` + `state-json-streamer.ts` — both could collapse to a call into [`json-stream-stringify`](https://www.npmjs.com/package/json-stream-stringify) (~8KB, no transitive deps, returns a Readable directly). Leaving the choice to you, but here's the honest trade-off:

**Library wins**
- Drops ~99 lines of prod + ~200 lines of tests.
- Streams arbitrary depth. The custom streamer only splits `LOOP_ON_ITEMS.iterations`; every other field goes through `JSON.stringify(step)` whole — so a piece step that ever accumulates a big output (large HTTP body, PDF text, OCR) still materializes that field as one string. The library doesn't need to know where the fat lives.
- `toJSON`, undefined-skipping (kills the need for `definedEntries`), nested arrays — inherited from `JSON.stringify` semantics for free.
- Native Readable, so no `Generator<string>` → `Array.from().join('')` → `Readable.from()` two-step.

**Custom wins**
- One fewer runtime dep (the engine bundles core-utils inline, so net bundle delta is ~6KB — small, but real).


If the loop-iterations case is the *only* fat node you expect long-term, the custom version is fine. If you want the memory guarantee to hold as new step types grow, the library is the safer default.

PR #14754, opened 2026-08-12
ci(cloud): deploy prod before canary on cloud-hotfix

abuaboud:
## Description

Follow-up to #14752. The canary deploy was ordered *before* the production deploy in `promote-to-production`, so a canary-side failure aborts the job and production never ships — the opposite of what a `cloud-hotfix` is for.

Run [31606917583](https://github.com/activepieces/activepieces/actions/runs/31606917583) hit exactly this: `Deploy App to canary` failed with `docker pull … error from registry: denied` on the canary host, and prod was never deployed.

This moves the two canary steps after the two production steps. Prod ships first; canary still deploys from the same image, and a canary failure still reds the run without holding prod hostage.

Note the underlying GHCR failure is **not** fixed here and needs ops: the canary host's registry credential logs in (`docker login` → exit 0) but cannot pull (`denied`), while the image tag itself exists in the registry. GHCR answers `GET /v2/` for any well-formed token, so a rotated/expired `read:packages` credential looks like a successful login. Canary had not pulled since 2026-07-31 — the migration gate had been blocking its deploys — so the stale credential went unnoticed for twelve days.

## How was this tested?

Workflow-only change; validated the YAML parses (`bunx js-yaml`). The steps are moved verbatim — same image tag, same config files, same `if:` condition. Not exercised against the real ops host.

Fixes # (issue)


abuaboud:
Superseded by #14755 — canary now runs as its own job before `promote-to-production`, so the inline step ordering this PR fixed no longer exists.

PR #14770, opened 2026-08-12
fix(pieces): correctness fixes for the Microsoft SQL Server piece (0.0.2)

OdaiAhmed99:
Follow-up to #14701, as `0.0.1` → `0.0.2`. Every finding was reproduced against SQL Server 2022 before being fixed.

## Description

| # | Fix |
|---|---|
| 1 | `find_rows` advertised `idempotent: true` while `condition` is raw T-SQL, so agents retried an action that can write. Now `false`, with the reason spelled out for the model. |
| 2 | `getTableMeta` passed an unquoted `schema.table` to `OBJECT_ID`, so a dotted name resolved to NULL, metadata came back empty, and the exact-value projection silently fell back to `SELECT *`. Now `QUOTENAME`d. |
| 3 | `query_timeout = 0` — the driver's "no timeout" — was treated as absent. `isNil()` instead of truthiness. |
| 4 | The connection-string branch inherited the driver's 15 s request timeout while discrete fields got 30 s. Now set in both branches. |
| 5 | `run_query` kept only the first result set of a batch or stored procedure. Reads `result.recordsets`; `rows` still holds the first set so existing flows are unaffected, and a new `result_sets` output carries them all. |
| 7 | `datetimeoffset` lost its offset — `CONVERT` style 127 shifts to UTC and appends `Z`, which is the one reason to prefer it over `datetime2`. Style 126 keeps the stored offset, in both the projection and the cursor. |
| 8 | Trigger sample data came from a raw `SELECT *` while `run()` used `exactProjection`, so a flow built on the sample saw a number and got a string live. `previewQuery` now projects through `exactProjection`. |
| 10 | `order_by` listed columns `onEnable` then refused, throwing on a keyless table. The dropdown and `planCursor` now share one rule, `cursorUtils.isOrderable`, so they cannot drift. |

Also in here: the duplicated `OUTPUT`/retry block in `update_rows` and `delete_rows` became one `mssqlCommon.writeReturningRows`; `mssqlGetColumns` is gone (it re-queried what `getTableMeta` returns); dead "Maximum Rows Per Poll" keys removed from `translation.json`.

Brought to repo conventions: no comments (L56 — rationale moved below), no `as` casts (L47) or `any` (L46), named parameters (L53), grouped `mssqlCommon` / `cursorUtils` exports (L57).

### Cursor layout 2

Fix #7 changes how a `datetimeoffset` position renders, so `CURSOR_LAYOUT` goes 1 → 2 — but the acceptance check is now per-column rather than global. `reconcile` asks whether any layout between the stored one and the current one re-rendered a type *this cursor actually uses* (recorded in `RERENDERED_AT_LAYOUT`, which for layout 2 is exactly `datetimeoffset`), instead of discarding every stored position. A rejected cursor re-baselines at the head and silently skips whatever was pending, and ordering on an `int` identity is not a `datetimeoffset` problem — 0.0.1's position for it is byte-identical to what 0.0.2 writes.

The re-render check also only applies when there is a rendered position to be stale. A cursor with `k: null` — what `onEnable` stores when the table is empty, and which means "deliver from the start", since a null position adds no lower bound — is kept regardless of type. Rejecting it would re-baseline at the current head, so enabling on an empty `datetimeoffset`-ordered table under 0.0.1 and upgrading after rows arrived would have skipped all of them. Caught by Greptile; the layout *range* check stays unconditional, because an unknown `v` still implies unknown `m`/`c` semantics.

Net effect: a trigger ordering on `datetimeoffset` with a stored position re-baselines once, which is the mechanism the version field was added for. Every other trigger, and every empty baseline, keeps its position and its pending rows.

### Not in here

- **Finding 6 — group-mode cursor permanently loses rows.** The cursor advances onto the last delivered ordering value while the page filter is strictly `>`, so a row written later carrying an already-delivered value is never seen. `completeGroups` guards a page cut *inside* a value but not a later arrival *on* one. Any real fix changes delivery behaviour — boundary-hash dedupe with `>=`, or replacing the hand-rolled cursor with `pollingHelper` — so it wants its own PR and version bump. Filed as #14781; the trigger description now warns about it, where it previously mentioned only rows written *behind* the last delivered value.
- **Finding 9 — `outputSchema` shapes.** Checked and refuted. `get_tables`, `find_rows` and `insert_row` set `value: ''`, which is the documented convention for "this field is the whole output" with `key` as the display name; the framework's own `output-table-view.test.ts` asserts exactly this shape. All six schemas already match their returns.

## How was this tested?

- 61 unit tests, up from 47 on `main`: the `datetimeoffset` offset both ways, the exact preview projection, `isOrderable` agreeing with `planCursor` on a keyless table, and three cursor-layout cases (older layout kept when nothing it uses was re-rendered, discarded when `datetimeoffset` is in the position, out-of-range rejected).
- **122/122 against a live SQL Server 2022 (CU26)** — 63 unit plus 59 integration, run with both test databases dropped first as well as warm. They skip without a server. The two `k: null` reconcile cases were checked against a reverted fix to confirm they actually fail without it.
- The action suite now creates its own `ap_action_test` database. It had connected to `ap_trigger_test` without creating it, which raced the trigger suite's `DROP DATABASE` / `CREATE DATABASE` under vitest's concurrent file execution; it passed alone and failed in a full run.
- `npx turbo run lint --filter=@activepieces/piece-microsoft-sql-server` clean (CI does not lint pieces on a piece-only PR); `tsc` over `src/**/*.ts` including tests clean, down from 4 errors.
- Repros worth quoting: #2 — the same data under `dbo.[fin.2026]` returned `decimal(19,4)` as a float and truncated `datetime2(7)` to milliseconds, exact under `dbo.nodot`; #3 — `query_timeout = 0` with `WAITFOR '00:00:33'` failed at 30048 ms; #4 — the same query failed at 15021 ms on a connection string and succeeded at 20017 ms on discrete fields.

<details>
<summary><b>Why the code looks like this</b> — moved out of the source, as L56 asks</summary>

- **The driver rounds, so the server renders.** tedious reads a decimal as `value / 10 ** scale` in double arithmetic and pushes the date family through a JS `Date`, so `decimal(19,4)` loses its last two digits, `decimal(38,0)` comes back as `1.2345678901234568e+37`, `datetime2(7)` drops four fractional digits, and `time` arrives as a date in 1970. None of it is recoverable once parsed, so `exactProjection` has SQL Server convert those columns to text. Cursor positions never pass through a number or `Date` either — a truncated position sits below the row it names and redelivers it every poll.
- **Styles.** 126 is ISO 8601 keeping every fractional digit and, for `datetimeoffset`, the offset; 1 renders binary with the `0x` prefix; 3 prints a float to the 17 digits a double needs; 2 prints all four of money's decimals, since the default shows two and `0.0001` would read back as `0.00`, sticking the position.
- **rowversion needs an inner cast** to `binary(8)`, or `CONVERT` ignores the hex style and returns raw bytes, NULs included; `timestamp` is not a legal `CONVERT` target.
- **String cursors bind under the column's own collation.** Rendered as nvarchar and read back into varchar, every unrepresentable character becomes `?` under the database default code page — a value no row can equal, so the frontier never advances and every row is redelivered forever.
- **`ORDER BY` is table-qualified on purpose.** The projection aliases rendered text to the column's own name, and a bare identifier in `ORDER BY` resolves against the SELECT list first. The page would come back in the text's order while `WHERE` compares real values, and for decimal that differs — `'10.0000'` sorts before `'9.0000'` — so the position would step past rows never delivered.
- **Error 334 and the OUTPUT retry.** Enabled triggers make SQL Server refuse `OUTPUT`, and 334 is raised at compile time, so nothing was written and re-running without `OUTPUT` is safe — count only, no rows. `insert_row` does not retry, since that would discard the generated id, and explains itself instead.
- **Key selection.** Nullable unique-index candidates are dropped, not deprioritised: a NULL key makes every comparison UNKNOWN and drops the row from the cursor window. Disabled indexes are excluded because they still report `is_unique` while no longer enforcing it. An IDENTITY counts as ascending only when a unique index backs it, since `IDENTITY(n, -1)` counts down and a reseed or `IDENTITY_INSERT` load can repeat values.
- **The drain runs against the read order.** `order_direction` describes how the user reads the table; the drain walks from the oldest undelivered row forward, so a backlog is walked through rather than jumped over.
- **Page size is a constant, not a prop.** `isRepublish` is a deep-equal over the trigger's whole input — right for table, order column and direction, wrong for a page size the cursor does not depend on. Editing it threw away a readable backlog.
- **`onDisable` leaves the position in the store**, because a republish disables and re-enables and `onEnable` needs it. Switching the trigger on by hand baselines at the head instead, so a disabled spell does not arrive as a burst of runs.
- **The position is saved before the rows are handed back**, so an engine dying in between loses them. Every polling trigger on the platform shares that window; there is no post-dispatch hook.
- **`TYPE_NAME(system_type_id)`** rather than a join on `user_type_id`: an alias type reports its own name, and only base types are legal `CONVERT` targets.

</details>

<details>
<summary><b>Running the integration tests</b></summary>

`new-or-updated-row.integration.test.ts` drives the real trigger against a real SQL Server, covering what generated SQL cannot settle: a bulk `UPDATE` collapsing thousands of rows onto one timestamp, sub-millisecond `datetime2`, and a bigint identity past the range a JavaScript number holds. It skips entirely unless pointed at a server.

A throwaway container, which also gets its own database created and dropped:

```
docker run -d --name ap-mssql-test -e ACCEPT_EULA=Y \
  -e MSSQL_SA_PASSWORD='Str0ng!Passw0rd#2026' -p 1433:1433 \
  mcr.microsoft.com/mssql/server:2022-latest
AP_MSSQL_TEST_HOST=localhost AP_MSSQL_TEST_PASSWORD='...' npx vitest run
```

Or an existing server such as Azure SQL, via the same connection string a user would paste into the piece. Nothing is created or dropped except tables named with the suite's `TABLE_PREFIX`, so it is safe against a shared database, though it still needs rights to create tables there:

```
AP_MSSQL_TEST_CONNECTION_STRING='Server=...;Database=...;User ID=...' npx vitest run
```

</details>

Return shapes and prop names are unchanged and `result_sets` is additive. The one behavioural effect is the `CURSOR_LAYOUT` bump above: a trigger ordering on a `datetimeoffset` column, and holding a stored position, re-baselines once instead of binding a position rendered in UTC. Every other trigger — including one still on an empty baseline — keeps its stored position, so nothing self-hosters run needs action.

No new outbound calls, no new auth surface. Fix #2 makes identifier handling stricter, not looser — `QUOTENAME` closes a path where a dotted table name silently degraded the projection. Fix #1 corrects a claim that was actively misleading agents about whether `find_rows` can write.


ibrahim-abuznaid:
Went through this against a live SQL Server 2022 instance (Developer, CU26) with the piece built from this branch, re-running the same probes that produced the original findings. **8 of the 10 are verified fixed** — not just present in the diff, but confirmed by behaviour:

| Finding | Verified |
|---|---|
| `QUOTENAME` in `OBJECT_ID` | dotted `dbo.[fin.2026]` now returns `"12345678901.1234"` and full `datetime2(7)` precision, identical to an undotted control table with the same data |
| `isNil` for query timeout | `query_timeout = 0` + `WAITFOR '00:00:33'` **succeeded at 33043 ms** (was failing at 30048 ms) |
| connection-string timeout | 20 s query **succeeded at 20031 ms** (was failing at 15021 ms) |
| `recordsets` | `EXEC` with two result sets now returns both under `result_sets` |
| `datetimeoffset` | offsets preserved — `+03:00`, `-08:00`, `+00:00` all intact |
| trigger sample vs live | all four probed columns now match exactly in type *and* precision |
| `find_rows` AI metadata | `idempotent: false` + the rewritten description is the honest version, thank you |
| `props.ts` comment | corrected |

The extras are appreciated too — named parameters throughout, comments down from ~443 lines to 1, casts from 17 to 3, grouped util exports, dead i18n keys gone, version bumped correctly.

Two things before this merges.

### 1. One test is failing — stale assertion

`src/lib/triggers/new-or-updated-row.integration.test.ts:1135`, *"will not trust a filtered unique index"*:

```
AssertionError: expected { v: 2, m: 'group', … } to deeply equal { v: 1, m: 'group', … }
-   "v": 1,
+   "v": 2,
```

Bumping `CURSOR_LAYOUT` to 2 is right; this assertion just didn't come along. 119/120 pass otherwise.

### 2. The group-mode row loss is still there

This is the one I'd most like to talk about. Re-ran it on this branch and got the identical result — a row written later carrying an ordering value that was already delivered is never delivered.

On `dbo.keyless` (no primary key or unique constraint, so mode resolves to `group`), ordering by `updated_at`:

1. `onEnable` baselines the cursor
2. insert row `c` @ `2026-08-05`, poll → delivers `c`, cursor advances to `2026-08-05`
3. insert row `d` @ **`2026-08-05`** (same value), poll → **nothing. `d` is lost permanently**
4. insert row `e` @ `2026-08-06`, poll → delivers `e` normally

The cursor advances *onto* the last delivered value while the page filter is strictly ahead of it, so anything that later lands on that value is behind the cursor forever. `completeGroups` covers a page cut *inside* a value at fetch time, but not an arrival *on* a value already handed over. Polling keeps working the whole time, which is what makes it hard to notice — no error, no gap, nothing to alert on.

Was this deliberately out of scope for 0.0.2? Entirely reasonable if so, it's a design question rather than a patch — but then I think it needs (a) a follow-up issue, and (b) a line in the trigger description, because the current wording only warns about rows written **behind** the last delivered value, not rows written **on** it. Rows sharing a timestamp are ordinary in bulk loads and same-second writes.

Worth considering whether `pollingHelper` removes the class of problem rather than patching the boundary — though I realise the note you added to `brain/pieces-engine/building-pieces.md` argues the opposite direction for bounded pages, so if you've already concluded a keyset cursor is required here, say so and let's design the fix rather than swap the mechanism.

### Smaller, not blocking

`get_tables` and `find_rows` still return bare arrays while their schemas advertise `tables` and `rows` wrappers — `run_query` picked up `result_sets`, but those two didn't change. Since the piece is effectively unreleased, worth landing one convention across all seven steps now.

### CI

Both red checks look like infrastructure rather than your code — `lint` died on `socket hang up` fetching Bun, and the PR-size log shows no actual violation. `npx turbo run lint --filter=@activepieces/piece-microsoft-sql-server` passes clean locally. A re-run should clear them.

OdaiAhmed99:
Thanks for re-running the probes rather than reading the diff — that's what caught both of the things below.

Pushed as cff5049b4e.

### The failing test — fixed

You were right that the assertion just didn't come along. Rather than moving it to `2` I pinned it to the constant:

```ts
import { CURSOR_LAYOUT } from '../common/cursor';
…
expect(await modeOf(table, 'updated_at')).toEqual({ v: CURSOR_LAYOUT, … });
```

It was the only hardcoded layout number in the suite — `cursor.test.ts` already imports the constant and writes `CURSOR_LAYOUT - 1`. That test is about key selection, so it shouldn't have to move on a layout bump at all.

### Your run also found a second bug — a race I introduced

The full suite failed for me too, but on a *different* test: `Login failed for user 'sa'` in `actions.integration.test.ts`. That's what SQL Server says when the target database isn't there.

`actions.integration.test.ts` — which I added in bd0bada434, in this PR — connected to `ap_trigger_test` without creating it, while the trigger suite's `beforeAll` does `DROP DATABASE ap_trigger_test; CREATE DATABASE ap_trigger_test`. Vitest runs files concurrently, so the actions suite raced the drop. It passes 13/13 on its own, which is exactly why it looked fine, and it's probably why you saw the layout failure and not this one — on a server where the database already existed, the race falls the other way.

Fixed by giving the actions suite its own `ap_action_test`, bootstrapped in its own `beforeAll` and skipped in connection-string mode, matching how the trigger suite treats a shared server.

**120/120 now**, run twice against SQL Server 2022 CU26 — once with both test databases dropped first, once warm. Before this PR there was only one integration file, so the race couldn't happen; two would have kept flaking in anyone's local runs.

### Group-mode row loss

Deliberately out of scope, and your reproduction is correct — I get the identical result. Filed as #14781 with the repro, the mechanism, and the two candidate fixes.

Your documentation point was the right call, so I took it: the description warned about rows written **behind** the last delivered value but said nothing about rows landing **on** it, which is the ordinary case for a bulk load or same-second writes. The keyless bullet now reads:

> Without one, all rows sharing an ordering value are delivered together as a single group, and a row that arrives later carrying a value that group already delivered is **not** picked up — so prefer a keyed table when a bulk load or same-second writes can put several rows on one value.

On `pollingHelper` versus the keyset cursor — I don't think I've concluded either way, so I'd rather settle it on the issue than assume it. The note in `building-pieces.md` argues against `pollingHelper` specifically because it doesn't bound its page, which is what pushed me to a hand-rolled cursor. If that reasoning holds, the fix is boundary-hash dedupe with `>=`; if the unbounded page is tolerable for this trigger, swapping the mechanism is the cheaper and more durable answer. I've written both up in #14781 rather than pick one here.

### The `outputSchema` shapes

This one I'd push back on. `get_tables` and `find_rows` don't advertise `tables`/`rows` wrappers — both set `value: ''`, and the resolver treats an empty path as the whole output:

https://github.com/activepieces/activepieces/blob/main/packages/web/src/lib/path-utils.ts#L92

```ts
if (path === '') return { value: obj, resolvedPath: path };
```

`utils-schema.ts` reads `field.value ?? field.key`, so with `value: ''` the `key` is only the display label and never a path segment. `utils-schema.ts:314` special-cases `field.value === ''` for `itemLabel`, and the comment there calls it "whole-item (`value:''`)". `run_query` and the write actions use real nested keys because their code really does return `{ rows, result_sets, … }`.

So both conventions are already in the piece on purpose, and each matches what its action returns. Unifying them would mean changing return values rather than schemas. Happy to be shown wrong if you're seeing a wrapper appear in the data selector — that'd be a framework bug worth its own issue.

### CI

Both were the Bun download hitting `socket hang up`, as you guessed. Re-ran them before your review landed; `PR size` and `lint` are green.

OdaiAhmed99:
The 4/5 finding is correct, and it was the more dangerous of the two review rounds. Fixed in 8f672eea9e.

`reconcile` read only the planned column types, so it rejected a version-1 `datetimeoffset` cursor even when `k` was null. That case has no rendered position at all, so there is nothing stale to misread — and rejecting it dropped `run` into the re-baseline path, which jumps to the current head and returns no rows:

```ts
if (isNil(cursor)) {
  await store.put<Cursor>(CURSOR_KEY, await baseline({ pool, plan }));
  return [];
}
```

`onEnable` stores `k: null` when the table is empty at enable time, and a null position adds no lower bound in `keysetPageQuery` / `groupPageQuery`, so that cursor means "deliver from the start". Enabling on an empty `datetimeoffset`-ordered table under 0.0.1, letting rows arrive, then upgrading would have skipped every one of them — the same row loss the narrowed `reconcile` was written to remove, left behind in the one case where the re-render check has nothing to act on.

The fix passes the position through and returns early when there isn't one:

```ts
if (isNil(position)) return true;
```

The layout *range* check stays unconditional above it, since an unknown `v` still implies unknown `m`/`c` semantics — a `k: null` cursor from layout 0 or 3 is still rejected.

Also added the test you flagged as missing, plus its out-of-range counterpart. I confirmed the first is decisive by stashing only the `cursor.ts` change and re-running: it fails without the fix and passes with it. The out-of-range case passes either way, which is the intended behaviour.

122/122 against SQL Server 2022 CU26, lint and `tsc` clean. The PR description's cursor-layout section has been corrected — it previously claimed every non-`datetimeoffset` trigger keeps its pending rows, which is exactly the sentence this disproved.

ibrahim-abuznaid (approved):

PR #14777, opened 2026-08-13
fix(waitpoints): give a looped pause its own row, and release barriers on the predicate

othmanemad:
## Description

**Stack: 3 of 4.** Base: [#14776](https://github.com/activepieces/activepieces/pull/14776).

Two halves of one problem: **a run may now hold several waitpoint rows.**

### The bug this fixes

A delay or approval **inside a loop** only paused on the first iteration. `waitpoint` is unique on `(flowRunId, stepName)`, so every iteration asked for the same row; `createForPause` found iteration 1's `COMPLETED` row in its pre-completed check and returned it, and iteration 2 and every iteration after it walked straight past its pause. The engine now sends the **path-qualified** step name (`loop_1:3/approval`) — 5 lines, and `stepName` is only ever an identity key inside `waitpoint-service`.

That makes two per-run lookups unsound, and both of their callers *delete the row they find* and enqueue a resume off it:

- `getByFlowRunId` → **`findCompletedWaitpointIfRunIsIdle`**: null while anything is still `PENDING`, else the newest `COMPLETED`. A run can now hold a completed leftover *and* an open pause at once, and `flow-runs-queue`'s pre-completed recovery must not resolve an arbitrary one of them.
- the subflow-failure path → **`findSubflowWaitpoint`**, which prefers `PENDING`, newest first.

Landing the engine change without those lookups (or vice versa) is worse than landing neither, which is why they are one PR.

### The cancellation race

Raised by Greptile on this PR and fixed here, though it is **pre-existing on `main` and not barrier-specific**. `cancelSingleRun` holds `runs_metadata_${id}`, removes the jobs and deletes every waitpoint, then hands `CANCELED` to `runsMetadataQueue` rather than writing it. For that window `flow_run.status` still reads `PAUSED` with no waitpoint left — indistinguishable from a legacy pre-waitpoint pause — so the by-run resume routes accepted a request-controlled resume for a run being cancelled, and `execute-flow` has no terminal-status guard to stop it running.

Sharing the cancellation lock does not fix it: the status write is asynchronous, so re-reading `flow_run` inside the lock still sees `PAUSED`. What closes it is that the intent is observable first — `runsMetadataQueue.add` merges the pending status into `runs_metadata:<runId>` **synchronously, inside that same lock**, before enqueuing. `resumeService` now refuses when that hash already holds a terminal status, and the two legacy by-run routes take `runs_metadata_${id}` like every other resume path. Cancel-first makes the resume wait and observe it; resume-first lets `removeAllFlowRunJobs` reap the job it just enqueued.

### The barrier service

`barrierService` creates a barrier with its signals, receives them as an **upsert** (last write wins — a retried child reports its terminal status again, possibly a different one), and releases through `shouldReleaseBarrier` over one `GROUP BY status`. Release is summary → complete → delete signals, all in one transaction under a pessimistic lock.

Evaluation is coalesced on its own `BARRIER_EVALUATION` queue, deduplicated on the barrier id. **The handler's first statement clears the deduplication key**, and every producer commits its signal row before enqueuing — BullMQ holds that key while the job is *active*, so otherwise the last signal lands while the job meant to see it is already running, its enqueue is dropped as a duplicate, and the barrier waits out its deadline holding a run that was ready to resume.

`resumeFromWaitpoint` **refuses barriers, always** — by addressed waitpoint type, or on the by-run legacy routes by whether the run holds any pending barrier. Every unscoped HTTP route goes through it, so there is no flag to forget; only the module's own `barrierService.release` may complete one.

### Naming

Reviewed in-thread and renamed across the module; behaviour is unchanged. The load-bearing one: `resumeService.releaseBarrier` is the **generic** resume path — `flow-runs-queue` calls it for a plain `DELAY` or `WEBHOOK` waitpoint — so it is now `resumeTrusted` (resume, skipping the barrier guard, because the caller is internal rather than an HTTP route). That restores the split the rest of the module keeps: **release** is what happens to a barrier, **resume** is what happens to a run.

Also `passesReleasePredicate` → `isReadyToRelease`, `completeAndDrainSignals` → `closeBarrier`, `buildSignalRows` → `buildPendingSignals`, `asStoredSummary` → `readStoredSummary`, `resolveDeadline` → `defaultBarrierDeadline`, `receive` → `receiveSignal`, `evaluate` → `releaseIfReady`, `addEvaluation` → `enqueueEvaluation`, and the dedup helpers now say *key* rather than *id*. `StepExecutionPath.toWaitpointKey` → `waitpointKeyFor({ stepName })`. Decision 000015 and the waitpoints architecture page are updated in the same commit.

### Size

`server/api` is **953** against the 600 budget, hence `large-pr-ok`. It is two files: `barrier-service.ts` (309) and its test file (289). The rest is 355 across nine files, most of it the queue pair.

| Area | Lines |
| --- | --- |
| `barrier-service.ts` | 309 |
| `barrier.test.ts` | 289 |
| `barrier-queue.ts` + `barrier-queue-factory.ts` | 159 |
| `waitpoint-service.ts` (the two lookups) | 58 |
| `resume-service.ts` (the guards + the cancellation race) | 131 |
| `waitpoint.test.ts` + `resume-service.test.ts` | 137 |
| `flow-runs-queue.ts` call sites | 15 |
| `AP_MAX_BARRIER_SIGNALS`, queue name, worker wiring | 7 |
| engine (the loop fix + its unit test) | 28 / 300 ✅ |

## How was this tested?

- New `barrier.test.ts` (13 cases here): the release predicate against a real database, K-of-N and veto early release, the inline signal summary, signal identity (the partial unique index bites a duplicate `sequence`, two nulls coexist, a second receive overwrites), evaluation coalescing, and five resume guards.
- The loop bug is pinned in two places: `waitpoint.test.ts` asserts `findCompletedWaitpointIfRunIsIdle` returns null while another iteration is still pending, and a new engine unit test pins the key format (`loop_1:3/loop_2:0/approval`).
- Four cases in `resume-service.test.ts` for the cancellation race, on the generic resume path rather than the barrier one. Verified they fail against the unfixed source: the three refusal cases go red, and `still resumes a paused run with nothing queued against it` passes either way as the regression guard.
- Full `flows/flow-run` integration suite: **103 tests pass**, including `execute-flow-e2e.test.ts` and the pause/resume cases — these PRs change the shared delay/webhook resume path, so those passing is the real check, not `barrier.test.ts`.
- `npm run lint-dev` clean, `npm run test-unit` clean. CE only.


packages/server/api/src/app/waitpoints/barrier-service.ts
```diff
@@ -0,0 +1,309 @@
...
+                .into('waitpoint')
+                .values({
+                    id,
+                    flowRunId: params.flowRunId,
+                    projectId: params.projectId,
+                    stepName: params.stepName,
+                    type: PauseType.BARRIER,
+                    version: params.version,
+                    status: WaitpointStatus.PENDING,
+                    resumeDateTime: resolveDeadline(),
```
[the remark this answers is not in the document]
othmanemad: Confirmed, and closed by the next PR in the stack — #14778 (`feat/barrier-edges`, based on this branch).

All three halves land there:

- **Scheduling** — `barrierService.create` now calls `waitpointTimeoutJob.schedule(...)` for the deadline it writes, instead of writing the column and stopping.
- **Firing** — `handleResumeDelayWaitpoint` looks the waitpoint up and branches on `PauseType.BARRIER`, calling `barrierService.release({ timedOut: true, releaseReason: 'timeout' })` rather than `resumeFromWaitpoint`. That is the first producer of either value, which is why neither has one in this PR.
- **Durability** — `WAITPOINT_DEADLINE_SWEEP` re-arms overdue deadlines once a minute off the `idx_waitpoint_pending_resume_date_time` partial index, so the deadline survives a Redis flush rather than living only inside a 30-day delayed job.

Fixed there too, adjacent to this finding: the `RESUME_DELAY_WAITPOINT` job id was keyed per run (`resume-delay-${flowRunId}`), and `upsertJob` only adds when no job holds that id — it never reschedules. A run holding two timed waitpoints therefore lost the second deadline silently. It is now keyed per waitpoint, and `waitpointTimeoutJob.remove` cancels the timer when a waitpoint resumes early, with `systemJobIds.legacyResumeDelay` covering jobs scheduled under the old id during a rolling deploy.

Nothing here is reachable in the meantime — `barrierService.create` has no production caller until #14778 adds the `BARRIER` branch to `POST /v1/waitpoints`. Leaving it as-is in this PR to keep the stack's boundaries intact.

othmanemad:
@greptileai rereview given the next PR https://github.com/activepieces/activepieces/pull/14778 solves the issues you addressed

packages/server/api/src/app/waitpoints/waitpoint-service.ts
```diff
@@ -146,13 +149,37 @@ export const waitpointService = (log: FastifyBaseLogger) => ({
         return waitpointRepo().findOneBy({ id: waitpointId, flowRunId })
     },
 
-    async getByFlowRunId(flowRunId: string): Promise<Waitpoint | null> {
-        const completed = await waitpointRepo().findOneBy({ flowRunId, status: WaitpointStatus.COMPLETED })
-        return completed ?? waitpointRepo().findOneBy({ flowRunId })
+    async findPreCompletedByFlowRunId({ flowRunId }: FindPreCompletedByFlowRunIdParams): Promise<Waitpoint | null> {
```
AbdulTheActivePiecer: please replace preCompleted, a very confusing term tbf, the status is pending or completed precompleted.

docs/install/reference/environment-variables.mdx
```diff
@@ -175,6 +175,7 @@ run data is kept. The **Cloud** values and how these interact are covered in
 | `AP_MAX_WEBHOOK_PAYLOAD_SIZE_MB` | Maximum incoming webhook payload size (MB). Larger payloads are rejected with HTTP 413. | `25` |
 | `AP_WEBHOOK_PAYLOAD_INLINE_THRESHOLD_KB` | Webhook payloads below this (KB) are stored inline in Redis; larger ones are offloaded to file storage to protect Redis memory. | `512` |
 | `AP_WEBHOOK_TIMEOUT_SECONDS` | Default timeout for synchronous webhooks. Maximum 15 minutes; note Cloudflare caps it at 30 seconds. | `30` |
+| `AP_MAX_BARRIER_SIGNALS` | Maximum number of things a single barrier may wait on — the approvers on one approval, or the batches a fan-out dispatched. | `10000` |
```
AbdulTheActivePiecer, line 180: We probably should expose this in the platform admin settings for people to set, better check with Kareem

packages/server/api/src/app/waitpoints/waitpoint-service.ts
```diff
@@ -146,13 +149,37 @@ export const waitpointService = (log: FastifyBaseLogger) => ({
...
+        const stillOpen = await waitpointRepo().existsBy({ flowRunId, status: WaitpointStatus.PENDING })
```
AbdulTheActivePiecer: pendingWaitpoint is a better name.

packages/server/api/src/app/waitpoints/waitpoint-service.ts
othmanemad: Renamed in 66dd8c0. `findPreCompletedByFlowRunId` is now `findCompletedWaitpointIfRunIsIdle` — it returns the newest `COMPLETED` waitpoint, but only when nothing for that run is still `PENDING`.

Took the longer name over `pendingWaitpoint` because the row it hands back is `COMPLETED`, not pending; the guard was the part that needed naming, not the result.

Worth flagging: `preCompleted` does survive in `createForPause` a few lines above, deliberately. There it means the waitpoint was completed *before* the pause was created (the callback beat the pause), which is a real concept the architecture page names too. The old lookup was borrowing that word for a different thing, which is what made it confusing. Happy to rename that one as well if you still want the term gone entirely — it is base-branch code, so it would widen this diff.

packages/server/api/src/app/waitpoints/waitpoint-service.ts
othmanemad: Covered by the rename in 66dd8c0 — it is `findCompletedWaitpointIfRunIsIdle`. Didn't use `pendingWaitpoint` because the waitpoint it returns is `COMPLETED`; "pending" describes the guard (return nothing while any waitpoint for the run is still pending), not the result. Details in the sibling thread.

packages/server/api/src/app/waitpoints/barrier-service.ts
```diff
@@ -0,0 +1,322 @@
...
+            wideEvent.set({
+                fanIn: {
+                    barrierId: barrier.id,
+                    signalCount: summary.total,
+                    releaseReason,
+                    stillRunning: summary.stillRunning,
+                },
+            })
+        }
+        await resumeService(log).resumeTrusted({
```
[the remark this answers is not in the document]
othmanemad: Already fixed — this comment predates the fix by about ten hours. Closed by 7aef2ed7 (`fix(waitpoints): close the legacy resume hole on a released barrier`).

The finding was correct against the code it was written on: the guard asked `hasPendingBarrier`, which filters on `status = PENDING`, so it stopped guarding the moment `closeBarrier` committed the barrier as `COMPLETED`.

The root cause was encoding "this resume was delivered" as row-*absence*, which makes a delivered barrier and a legacy pre-waitpoint pause indistinguishable. Delivery is now explicit:

- `waitpointService.consume` tombstones a `BARRIER` as `CONSUMED` and deletes every other type, so the row survives release.
- the legacy guard became `hasBarrier` — existence, no status filter — so it holds through both windows you named: after `closeBarrier` commits `COMPLETED`, and after `resumeTrusted` consumes it while the run is still `PAUSED`.
- the tombstone is cleared by the existing `deleteByFlowRunId` sweep when the run reaches a terminal status.

Both windows are pinned by tests in `barrier.test.ts`:

- `refuses a by-run resume between the barrier closing and the trusted resume consuming it`
- `refuses a by-run resume after the barrier was released while the run is still PAUSED`

plus `refuses a by-run sync resume through both barrier windows` for the sync route, and `still resumes a paused run that never held a waitpoint at all` as the regression guard that the existence check did not over-tighten.

packages/server/api/src/app/waitpoints/resume-service.ts
```diff
@@ -67,6 +74,10 @@ export const resumeService = (log: FastifyBaseLogger) => ({
         if (flowRun.status !== FlowRunStatus.PAUSED) {
             return { flowRun, stale: true }
         }
+        if (await waitpointService(log).hasBarrier({ flowRunId, projectId: flowRun.projectId })) {
+            log.warn({ flowRun: { id: flowRunId } }, '[resumeService#legacyResume] Refused an external resume of a barrier; only the barrier predicate and its deadline may release it')
+            return { flowRun, stale: true }
```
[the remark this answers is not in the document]
othmanemad: Confirmed and fixed in d88b6551 — thanks, the mechanism was real.

One correction to the attribution: this is **pre-existing on `main` and not barrier-specific**. `legacyResume` there checks `status !== PAUSED` and enqueues, full stop; this PR only *adds* a guard on top of it. A plain `DELAY` or `WEBHOOK` pause had exactly the same exposure. Fixed here anyway because the barrier work is what surfaced it.

Also worth recording: **sharing the cancellation lock does not fix it.** `CANCELED` is persisted asynchronously by `runsMetadataQueue` *after* `cancelSingleRun` releases the lock, so re-reading `flow_run` inside the lock still sees `PAUSED`. The lock only narrows the window.

What actually closes it is that the cancellation intent is observable before the DB write. `runsMetadataQueue.add` merges the pending status into the `runs_metadata:<runId>` hash **synchronously, inside the cancel lock**, before enqueuing, and the metadata worker clears the key with `deleteKeyIfFieldValueMatches` once it lands. So:

- `resumeService` refuses when `hgetJson(redisMetadataKey(id))` already holds a terminal status — checked in `resumeTrustedWithoutLock`, which every addressed route funnels through, and in both legacy by-run routes.
- `legacyResume` and `legacySyncResume` now take `runs_metadata_${flowRunId}` like every other resume path.

Both orderings are then safe: cancel-first makes the resume wait and then observe the queued status; resume-first lets cancel's `removeAllFlowRunJobs` reap the job it just enqueued.

Four cases added to `resume-service.test.ts` (the generic resume path, since this is not barrier-specific). Verified they fail against the unfixed source — the three refusal cases go red, and the fourth, `still resumes a paused run with nothing queued against it`, passes either way as the regression guard.

docs/install/reference/environment-variables.mdx
othmanemad: Agreed, and we'll do it in a follow-up PR rather than here — once #15170 (`feat(platform): add a platform_configuration table for admin-editable settings`) merges, since that table is where an admin-editable setting like this belongs.

Keeping it as an env var in the meantime so this PR doesn't have to invent a home for it.

AbdulTheActivePiecer (approved):

packages/server/api/src/app/waitpoints/resume-service.ts
```diff
@@ -1,26 +1,40 @@
...
 import { Waitpoint, WaitpointResumePayload, WaitpointStatus } from './waitpoint-types'
 
 export const resumeService = (log: FastifyBaseLogger) => ({
     async resumeFromWaitpoint({ flowRunId, waitpointId, resumePayload, workerHandlerId, httpRequestId }: ResumeFromWaitpointParams): Promise<ResumeFromWaitpointResult> {
+        const flowRun = await findFlowRunOrThrow(flowRunId)
+        if (await wouldResumeABarrier({ flowRunId, waitpointId, projectId: flowRun.projectId, log })) {
+            log.warn({ flowRun: { id: flowRunId }, waitpoint: { id: waitpointId } }, '[resumeService#resumeFromWaitpoint] Refused an external resume of a barrier; only the barrier predicate and its deadline may release it')
+            return { flowRun, stale: true }
+        }
+        return this.resumeTrusted({ flowRunId, waitpointId, resumePayload, workerHandlerId, httpRequestId })
```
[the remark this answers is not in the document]
othmanemad: @greptileai This one is a false positive — the race it describes cannot reach the state it claims.

**The claim:** a resume passes `wouldResumeABarrier`, a barrier is created for the same paused run in that window, and the locked path then "accepts the new barrier row".

**Why the last step cannot happen.** The locked path never looks up *a barrier for this run*. It resolves exactly one row by primary key, using the `waitpointId` carried by the request:

- `waitpoint-service.ts` → `handleResumeSignal`: `where({ id: waitpointId, flowRunId, projectId })`
- `resume-service.ts` → the post-`processed` recheck: `findByIdAndFlowRunId({ waitpointId, flowRunId })`

A barrier created in that window gets a fresh `apId()` (`barrier-service.ts`, in the `create` transaction) — nanoid, 21 chars over a 62-char alphabet. So the newly created barrier is never the row the request addresses. Reaching the described outcome requires guessing the id before the server generates it, which is the same bar as forging any other waitpoint resume URL.

**The two reachable cases both close:**

1. **The request id matches an existing row.** `wouldResumeABarrier` reads `waitpoint.type`. `type` is written once at insert and no code path updates it — there is no `DELAY`/`WEBHOOK` → `BARRIER` transition. So the pre-lock verdict and the in-lock row agree by construction, not by timing.
2. **The request id matches no row.** `hasBarrier` is consulted, returns false, and the resume proceeds — but `handleResumeSignal` then finds nothing, returns `false`, and the caller answers `stale: true`. Nothing is enqueued and no payload is applied.

**On the verification note:** the three observations are each correct — the barrier check does precede the run lock, `barrierService.create` does not take `runs_metadata_*`, and the locked handler does not re-check `type`. The conclusion does not follow from them, because the locked handler matches on **id**, not on type or on run ownership. The guard was read as a set-membership test over the run's barriers; it is an identity check on one row.

Re-checking `type` inside the lock would be harmless, but it would guard a state transition that does not exist, so I am leaving it out.

Please re-review.

othmanemad:
@greptileai rereview — replied in-thread on the "Barrier check races creation" comment (https://github.com/activepieces/activepieces/pull/14777#discussion_r4017744994). The locked resume path resolves a single row by primary key (`id = waitpointId` from the request), not by run-barrier membership, and a newly created barrier gets a fresh `apId()`, so it can never be the row the request addresses. `type` is also write-once, so there is no transition for the pre-lock check to miss.

amrdb (commented):
Review of the original PR work only (excluding what PR #15395 contributed via commit 6e4fcc966). Four issues I'd call blockers — one confirmed rule violation and three correctness bugs on the primary resume path. Detail inline. Six lower-severity findings I've left off the review — happy to file them as follow-ups if you want.

packages/server/api/src/app/waitpoints/waitpoint-service.ts
```diff
@@ -152,13 +155,52 @@ export const waitpointService = (log: FastifyBaseLogger) => ({
         return waitpointRepo().findOneBy({ id: waitpointId, flowRunId })
```
amrdb: **Missing `projectId` filter — violates `.claude/rules/data-isolation.md`** ("ALL database queries MUST filter by projectId or platformId").

This function is now on the primary resume path — both `wouldResumeABarrier` and the race-recovery block in `resumeTrusted` call it. Waitpoint ids are random `apId()`s so this is not directly exploitable today, but the rule is absolute and any upstream bug that leaks a `waitpointId` across projects bypasses isolation here.

Thread `projectId` through the signature; both call sites already have `flowRun.projectId` in scope.

packages/server/api/src/app/waitpoints/waitpoint-service.ts
```diff
@@ -117,7 +120,7 @@ export const waitpointService = (log: FastifyBaseLogger) => ({
                     return null
                 }
                 await onReady(found)
```
amrdb, line 126: **`onReady` fires even for an already-completed or -consumed waitpoint — enables double resume dispatch.**

The `where` clause four lines up (`.where({ id: waitpointId, flowRunId })`) has no status filter, so `found` can be `PENDING`, `COMPLETED`, or `CONSUMED`. Compare to `complete()` above which correctly filters `status: WaitpointStatus.PENDING` (line 88). If `handleResumeSignal` is entered a second time against the same waitpoint (re-fired delayed job, duplicate release, at-least-once webhook), this line enqueues a resume and the next line re-runs `consume` over an already-consumed row.

The only guard against a double engine dispatch is the BullMQ jobId dedup on `${runId}-resume-${waitpointId}`, and that expires with the job.

Fix: add `status: WaitpointStatus.PENDING` to the `where` above, or early-return here when `found.status !== PENDING`.

packages/server/api/src/app/waitpoints/resume-service.ts
```diff
@@ -1,26 +1,40 @@
...
 import { projectService } from '../project/project-service'
 import { engineResponseWatcher } from '../workers/engine-response-watcher'
+import { redisMetadataKey, RunsMetadataUpsertData } from '../workers/job'
 import { waitpointService } from './waitpoint-service'
 import { Waitpoint, WaitpointResumePayload, WaitpointStatus } from './waitpoint-types'
 
 export const resumeService = (log: FastifyBaseLogger) => ({
     async resumeFromWaitpoint({ flowRunId, waitpointId, resumePayload, workerHandlerId, httpRequestId }: ResumeFromWaitpointParams): Promise<ResumeFromWaitpointResult> {
+        const flowRun = await findFlowRunOrThrow(flowRunId)
+        if (await wouldResumeABarrier({ flowRunId, waitpointId, projectId: flowRun.projectId, log })) {
```
amrdb: **Barrier check runs outside the run lock — TOCTOU that defeats the guard.**

The whole point of `wouldResumeABarrier` is that only the barrier predicate/deadline may release a barrier. But this check runs before `resumeTrusted`'s `runs_metadata_${flowRunId}` lock is acquired, and `resumeTrusted` never re-checks inside the lock. A `barrierService.create` landing on the same run between this check and the lock slips through — `handleResumeSignal` completes/consumes the addressed non-barrier waitpoint and enqueues a resume that races the barrier.

`legacyResume` (line 90) and `legacySyncResume` (line 159) already do the equivalent check *inside* `runExclusive` and prove the pattern. Move `wouldResumeABarrier` into `resumeTrusted`'s `runExclusive` block, or duplicate the check there.

packages/server/api/src/app/waitpoints/barrier-service.ts
```diff
@@ -0,0 +1,317 @@
...
+        }
+        wideEvent.set({
+            fanIn: {
+                barrierId: barrier.id,
+                signalCount: summary.total,
+                releaseReason,
+                stillRunning: summary.stillRunning,
+            },
+        })
+        await resumeService(log).resumeTrusted({
```
amrdb, line 129: **Lost resume if the winning release crashes between `closeBarrier` commit and here.**

After 039a6263f8 ("leave the resume to the release that closed the barrier") removed the losing release's fallback, this is the only path that dispatches a released barrier's resume. If the process is SIGKILLed / OOMs / times out between `closeBarrier`'s tx commit and `resumeTrusted` returning:

- BullMQ retry / stalled-check re-runs the barrier-eval job → `releaseIfReady` (line 98) refuses because it filters `status: PENDING` (the barrier is now `COMPLETED`).
- The deadline fires `handleResumeDelayWaitpoint` → returns at `resume-delay-handler.ts:28` because `waitpoint.status !== PENDING`. The `pastPauseTimeout` FAIL branch (line 36) has the same gate.
- Signals are deleted so nothing can re-enqueue an evaluation.

Run sits `PAUSED` with a `COMPLETED` barrier indefinitely; requires manual DB intervention.

Suggested fix (no cross-store tx needed): make `releaseIfReady` idempotent — widen its lookup to include `COMPLETED` and, on that branch, dispatch the resume off `barrier.resumePayload` (which is already stored on the row inside `closeBarrier`'s tx). BullMQ retry + stalled-check then recover both the "throws in `resumeTrusted`" and the "process killed" cases without adding an outbox.

packages/server/api/src/app/waitpoints/waitpoint-service.ts
othmanemad: Agreed, fixed in 2252abe — and widened. The sweep found 12 lookups missing `projectId`, not one. The worst was `findUndeliveredCompletedWaitpoint`, which picks which payload gets delivered; `CompleteParams` also declared `projectId` and never used it.

Three notes:

- `createForPause`'s pre-completed lookup was the only behaviour change. It ran *before* the `findOneByOrFail({ id, projectId })` assertion, so a mismatched projectId returned another project's COMPLETED waitpoint, payload included. New test pins it.
- `findPendingByVersion` had no projectId in scope at all — two public unauthenticated V0 routes — so both now load the run first. Being honest: that filter is tautological and buys no defence the unguessable `flowRunId` did not already buy. It is there so the rule holds with no carve-out.
- `cancelSingleRun` uses each row's own `flowRun.projectId`, never the request's. It recurses over `getAllChildRuns`, whose CTE has no project predicate, so a child can live in another project — scoping to the request would silently leak child waitpoints with no test failing.

`findFlowRunOrThrow` is deliberately left unscoped: it is the root of trust the others derive projectId from.

packages/server/api/src/app/waitpoints/waitpoint-service.ts
othmanemad: Defect confirmed, fixed in 746b1ca — but not with `status: PENDING`. That patch breaks a primary path.

`flow-runs-queue.ts:121` hands this branch a **COMPLETED** row on purpose: when a callback beats the pause, `complete()` parks it and the metadata worker delivers it on writing PAUSED. Requiring PENDING kills that for every paused run.

I applied your patch and ran the suites — 9 failures, 4 of them predating this PR:

```
× resume guards > lets the module release a barrier and consumes the waitpoint exactly once
× resume guards > refuses a by-run resume after the barrier was released while the run is still PAUSED
× resumeService > consumes a COMPLETED waitpoint when race recovery fires (callback-before-worker ordering)
× resumeService > does not leave a stale COMPLETED row to poison the next pause cycle (leftover-row regression)
× undelivered resume recovery > dispatches the resume for a waitpoint that was already completed
    before the run reached PAUSED  → expected [] to have a length of 1 but got +0
```

The statuses are a delivery state machine, not a lifecycle: PENDING is *not yet owed*, COMPLETED is *owed*, CONSUMED is *delivered*. So the guard rejects CONSUMED only. You are right that the tombstone was the reachable case — nothing else survives `consume`.

On the jobId dedup point: agreed it is not the guard. The guard is now the row. And it does not need a conditional `UPDATE ... RETURNING` — the branch already holds `SELECT ... FOR UPDATE` in the transaction, and under READ COMMITTED a blocked `FOR UPDATE` re-reads the newest committed version via EvalPlanQual rather than its original snapshot, so read-then-branch-then-write there is already an atomic compare-and-set.

packages/server/api/src/app/waitpoints/resume-service.ts
othmanemad: This one I am pushing back on — the move is a no-op for the race it describes.

`wouldResumeABarrier` never asks "does this run hold a barrier". It resolves **one row by primary key**, using the `waitpointId` the request carries. A barrier created in that window gets a fresh `apId()`, so it is never the row the request addresses, and the verdict is identical inside or outside the lock. The `hasBarrier` fallback only runs when the addressed row is nil — and then `handleResumeSignal` finds nothing and answers `stale: true` anyway. `type` is written once at insert and never transitions, so the pre-lock verdict and the in-lock row agree by construction rather than by timing.

Greptile filed the same claim on this PR in #4017285521 and withdrew it after this argument.

But you are the second reader to take it as run-scoped, which is a defect report about the name. Renamed in 0b82b65: `addressedWaitpointIsBarrier` returns the verdict or `no-such-waitpoint`, and the caller consults `hasBarrier` explicitly. Zero behaviour change, two visible steps instead of one name covering both.

Deliberately **not** adopting the run-wide invariant on the addressed path, which is the other obvious reading of your point. `hasBarrier` has no status filter and `consume` tombstones a barrier as CONSUMED until the run terminates, so a run-wide refusal would reject every later resume on any run that once held a barrier. Worth raising separately against the legacy routes that already check that way.

packages/server/api/src/app/waitpoints/barrier-service.ts
othmanemad: Confirmed — best finding of the four. Every path you listed is closed exactly as you describe, and I traced the same walls. Fixed in 746b1ca.

One correction: **widening `releaseIfReady`'s lookup is not enough on its own.** It still routes into `release` → `closeBarrier`, which filters `status: PENDING`, returns null, and makes `release` return *before* `resumeTrusted`. So the naive widening recovers nothing. `releaseIfReady` now branches on status and, for a COMPLETED barrier, re-dispatches the summary already persisted on the row, bypassing `closeBarrier` — routed through the same `resumeTrusted` call the normal path uses, so it inherits the run lock and the new CONSUMED guard rather than enqueuing beside them. You were right that no cross-store tx is needed; the payload is already there.

**Not claiming this is closed.** The only automatic re-invoker is BullMQ's stalled check, and the worker takes the default `maxStalledCount: 1` — one free redelivery after a SIGKILL. A second stall retires the job permanently, and a Redis flush removes the trigger while Postgres still holds a COMPLETED barrier on a PAUSED run. `attempts: 5` does not help; those govern *failed* jobs and a retry just re-runs `releaseIfReady`.

So the durable backstop is the sweep in #14778 — and it has to select `type = BARRIER AND status = COMPLETED`, because the natural `status = PENDING` reading never sees this state. I will make that change there. The same hole exists for non-barrier waitpoints (`complete()` commits COMPLETED, caller dies before dispatch) and is covered only opportunistically today; one sweep closes both.

othmanemad:
@amrdb thanks — three of the four were real and two of the patches would have introduced a new bug, so this was a genuinely useful pass. All addressed in `2252abe`, `746b1ca`, `0b82b65`, replies in thread.

Summary:

| | Verdict | Outcome |
|---|---|---|
| projectId filter | valid | fixed, widened from 1 lookup to 12 |
| `onReady` on a non-pending waitpoint | valid | fixed, but rejecting CONSUMED — `status: PENDING` breaks the pre-completed path (9 tests) |
| barrier check outside the lock | no-op | rejected, with the name that caused it fixed |
| lost resume on a crashed release | valid | fixed, but not by widening the lookup — that path returns before `resumeTrusted` |

Tests: 120/120 on `flows/flow-run` (was 103), unit suites green, lint clean.

**Please send the six lower-severity findings now** rather than after this merges — the branch is open and #14778 is stacked on it, so a second round costs a rebase.

packages/server/api/src/app/waitpoints/barrier-service.ts
```diff
@@ -0,0 +1,341 @@
...
+async function redeliverClosedBarrier({ barrier, log }: RedeliverClosedBarrierParams): Promise<void> {
+    log.warn({ waitpoint: { id: barrier.id }, flowRun: { id: barrier.flowRunId } }, '[barrierService#redeliverClosedBarrier] Barrier was closed but never delivered; re-dispatching the persisted summary')
+    const { error } = await tryCatch(() => resumeService(log).resumeTrusted({
+        flowRunId: barrier.flowRunId,
+        waitpointId: barrier.id,
+        resumePayload: barrier.resumePayload,
+    }))
+    if (!isNil(error)) {
+        log.error({ error, waitpoint: { id: barrier.id }, flowRun: { id: barrier.flowRunId } }, '[barrierService#redeliverClosedBarrier] Re-dispatching a closed barrier failed')
+    }
```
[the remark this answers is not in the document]
othmanemad: Valid, and fixed in 7de2eeed08 — `redeliverClosedBarrier` now rethrows after the `log.error`, so the `BARRIER_EVALUATION` job fails and spends its `attempts: 5`. Retrying is safe: `resumeTrusted` takes `runs_metadata_<flowRunId>`, and `handleResumeSignal` selects the waitpoint `FOR UPDATE` and returns `already-delivered` on a `CONSUMED` row, so a retry cannot resume a run twice.

Two notes on the rest of the finding.

**On "no durable recovery mechanism in this PR"** — correct for this PR, and deliberate. The durable backstop is the deadline sweep, which lives in the stacked PR #14778 along with the rest of the barrier's external edges. `redeliverUndeliveredBarriers` scans `type = BARRIER AND status = COMPLETED` on a `PAUSED` run every minute, keyed on `updated` with a two-minute grace window, and retries forever rather than dead-lettering.

**The fix had to be two-sided.** That sweep calls the same `releaseIfReady` for up to a hundred rows in a plain loop, so a bare rethrow would abort the whole tick on the first bad row — and since the scan orders by `updated ASC`, every later tick would meet that same row and abort again, starving every other undelivered barrier. So #14778 (ba646bfd4f) wraps each row in `tryCatch` and logs. The rethrow is the queue's signal; the per-row guard is what stops it becoming a head-of-line block.

Both are covered by tests that fail against the unfixed source: one asserts `releaseIfReady` rejects when the re-dispatch fails, the other injects a failure on the older of two undelivered barriers and asserts the younger one still gets its resume job.

amrdb (approved):

PR #14819, opened 2026-08-14
fix(platform): health page no longer reports two different app versions at once

### LINKED ISSUE #14818: [BUG] Health page shows two different app versions at once (Version row stale vs Release Integrity)
## Problem

Platform → Infrastructure → Health, App card. The Version row and the Release Integrity row report different app releases at the same time: "Current 0.88.0 · Latest 0.88.1" (Needs attention) directly above "All N connected workers match the app release 0.88.1" (Passed).

Reads as a stuck or partial upgrade. Persists for the life of the tab and survives in-page navigation.

Ref: Pylon 5628

## Root cause

Same value, two client sources.

* Version row: `CURRENT_VERSION` from `GET /v1/flags`, read via `flagsHooks.useFlag` with `staleTime: Infinity` (`packages/web/src/hooks/flags-hooks.ts:53`). Never refetched without a full page load.
* Release Integrity: `release.current` from `GET /v1/health/system`, default staleTime, refetches on mount and window focus (`packages/web/src/features/platform-admin/hooks/health-hooks.ts:11`).

Server-side both are the same `apVersionUtil.getCurrentRelease()` value (`flag.service.ts:66`, `health.service.ts:164`), so the two can only disagree in the client. Both consumed in `system-health-tab.tsx:41-50`.

## Steps to reproduce

1. Run 0.88.0. Open Platform → Infrastructure → Health. Leave the tab open.
2. Upgrade the app in place (bump `/usr/src/app/package.json` to 0.88.1, restart the container).
3. Return to the tab. The window-focus refetch updates the health query only.
4. Version: "Current 0.88.0", Needs attention. Release Integrity: "matches the app release 0.88.1", Passed.
5. Hard reload: both read 0.88.1, Version turns green.

Reproduced on the `activepieces/activepieces:0.88.0` image. Screenshots attached (before / contradiction / after reload).

## Fix

In `packages/web/src/app/routes/platform/infra/health/components/system-health-tab.tsx:41-50`, drop the `CURRENT_VERSION` flag read and use `systemHealth.release.current` for both the displayed version and the `semver.gte` comparison. One payload, one value.

## Out of scope

"Current" still reflects whichever app replica served the request. Per-replica skew is available from `GET /v1/health/diagnostics` (`apps.instances[].version`) but is not surfaced in the UI.

majewskibartosz:
Fixes [GIT-1755](https://linear.app/activepieces/issue/GIT-1755)
Closes #14818

## Problem

Platform → Infrastructure → Health, App card, after an app upgrade with the page already open:

1. Version row reports the pre-upgrade release ("Current 0.88.0 · Latest 0.88.1", Needs attention).
2. Release Integrity row, directly below it, reports the new one ("All N connected workers match the app release 0.88.1", Passed).

Persists for the life of the tab and survives in-page navigation, so it reads as a stuck or partial upgrade. Reported by a self-hosted customer mid-upgrade (Pylon 5628).

## Fix

`packages/web/src/app/routes/platform/infra/health/components/system-health-tab.tsx` → Version row now derives from `systemHealth.release.current` instead of `ApFlagId.CURRENT_VERSION`. The flag is fetched once per tab (`flagsHooks.useFlag` sets `staleTime: Infinity`, shared by every other flag consumer, so it is not the thing to change); the health payload already refetches on mount and focus. Server-side both values come from the same `apVersionUtil.getCurrentRelease()` call, so one payload removes the second source of truth.

## Verification

- `npx turbo run lint --filter=web`: 0 errors (71 pre-existing warnings, none in touched files).
- New `test/app/routes/platform/infra/health/components/system-health-tab.test.tsx`, red-first: 2 of 3 assertions fail on the pre-fix component (`expected '…' to contain 'Current 0.88.1'`, `expected '…' not to contain 'Needs attention'`), all 3 pass with the fix.
- `npx turbo run test --filter=web`: 468 passed, 1 pre-existing failure — `test/features/chat/lib/chunk-reducer.test.ts` (`ReferenceError: window is not defined` via `embed-provider.tsx:38`), confirmed failing on the base commit with all changes stashed. Same failure trips the pre-push hook, so the push used `SKIP_CHECK=1`.
- Visual: real app on a throwaway Postgres serving the built bundle, signed in as platform admin at `/platform/infrastructure/health`. Loaded at 0.88.0, bumped the app to 0.88.1 and restarted with the tab untouched, returned to the tab: version row followed to 0.88.1 and the status flipped to Passed with no reload. Inverse drive: downgraded the same instance back to 0.88.0, row followed down and flipped back to Needs attention. Both rows agreed at every step.
- Other affected paths: checked - 3 other `CURRENT_VERSION` consumers (piece-registry cache key in `pieces-hooks.ts:408`, `telemetry-provider.tsx`, `error-reporting.ts`); none render the version beside a live copy of it, and `/v1/health/system` is platform-admin-only so it is not a valid source for them. Left unchanged deliberately.

### Breaking change?

### Security impact?

<details>

## Plan

- One value, one source: the card already fetches `/v1/health/system` for the row underneath, so the Version row reads `release.current` from it. Net deletion.
- Footprint: 1 product file, 8 lines net. Rejected: refreshing the flags query, which re-fetches all 40+ flags on every focus and still leaves two sources for one value.
- The flags mock is deliberately kept in the test: it supplies the stale value that the `not.toContain('Current 0.88.0')` assertion proves can no longer reach the row, and it keeps the red-first run producing clean assertion failures rather than a provider crash.

- Unreadable release (`0.0.0` from a failed `package.json` read) → unchanged: `semver.gte('0.0.0', latest)` is false, so the row still shows Needs attention, and the Release Integrity row still carries the explicit explanation.
- `systemHealth` undefined while loading → `release?.current` is undefined, the existing `if (!currentVersion || !latestVersion) return false` guard already covers it, and rows render the `Checking` pill from `isPending`.
- Multiple app replicas on different releases → out of scope. The row shows whichever replica served the request; per-instance versions already exist in `GET /v1/health/diagnostics` (`apps.instances[].version`) but are not surfaced in any UI.
- Pre-existing, not fixed here: when `getLatestRelease()` fails it returns the same `0.0.0` sentinel, and `semver.gte(current, '0.0.0')` is true, so an air-gapped instance renders "Passed · Latest 0.0.0". Fixing it properly needs a third row state (unknown, not pass/fail), which is a UX call rather than part of this bug.

## Alternatives rejected

- Give the flags query a finite `staleTime` → global blast radius for one display value.
- Refetch flags on the health page only → keeps two sources of truth, just narrows the window.
- Show both values side by side → surfaces the inconsistency instead of removing it.

</details>


majewskibartosz:
## Visual verification

Same instance, same route (`/platform/infrastructure/health`), same sequence: page loaded while the app was on 0.88.0, app bumped to 0.88.1 and restarted with the tab left open, then back to the tab. No reload in either run. Only the bundle differs.

| Before (pre-fix bundle) | After (this PR) |
| --- | --- |
| before | after |
| Version: Current 0.88.0, Needs attention. Release Integrity: app release 0.88.1, Passed. | Version: Current 0.88.1, Passed. Release Integrity: app release 0.88.1, Passed. |

States driven:

- 0.88.0 baseline, both rows agreeing, Needs attention (latest is 0.88.1).
- Upgrade to 0.88.1 with the tab open, returning to the tab: row follows up, flips to Passed.
- Inverse drive, downgrade back to 0.88.0 on the same tab: row follows down, flips back to Needs attention.
- Loading state via the `Checking` pill on first paint.

Single theme: the diff changes no color, spacing, or component, only which value populates the existing row.

Louai-Zokerburg (approved):

PR #14824, opened 2026-08-14
feat(ai): knowledge base works with a Cloudflare AI Gateway provider

### LINKED ISSUE #14820: [Feature]: Knowledge base embeddings for Cloudflare AI Gateway and OpenAI-compatible providers
## Problem

Agent knowledge base embeds with the agent's own chat provider (`packages/pieces/community/ai/src/lib/actions/agents/run-agent.ts:172`). Only OpenAI, Google, Azure, Activepieces and OpenRouter have an embedding model wired up (`DEFAULT_EMBEDDING_MODELS`, `packages/pieces/community/ai/src/lib/common/ai-sdk.ts:224`).

With Cloudflare AI Gateway or Custom (OpenAI-compatible) selected: `createEmbeddingModel` throws, and the UI renders "Knowledge base requires a provider that supports embeddings" (`PROVIDER_EMBEDDING_MODELS`, `packages/web/src/features/agents/ai-model/index.tsx:24`).

No embedding-provider override exists, so a gateway-backed agent has no workaround. Blocks knowledge base for any customer centralizing keys/observability through a gateway.

Ref: Pylon 5652

## Proposed

* Cloudflare Gateway: build the embedding model over the gateway base URLs already used for chat (`/{account}/{gateway}/openai`, `/google-ai-studio`), reusing the submodel prefix map. Cloudflare proxies both providers' embedding endpoints.
* Custom / OpenAI-compatible: embedding model id in provider config, called via `/v1/embeddings`.
* Model selection needs a home: `ProviderModelConfig.modelType` is TEXT | IMAGE only (see activepieces/activepieces#14496). Either add an EMBEDDING type or a dedicated embedding-model field on provider config.
* Output must be 768-d to match `knowledge_base_chunk.embedding vector(768)`. OpenAI honors the `dimensions` param; Google `text-embedding-004` is natively 768.

## Out of scope

Per-agent choice of a different embedding provider than the chat provider.

majewskibartosz:
Fixes [GIT-1756](https://linear.app/activepieces/issue/GIT-1756)
Closes #14820

## Problem

1. A platform whose AI provider is a Cloudflare AI Gateway can never use a knowledge base. `defaultEmbeddingModel` is undefined for the gateway, so `createEmbeddingModel` throws and the builder replaces the Knowledge Base card with "requires a provider that supports embeddings, such as OpenAI or Google". No setting can change it: the agent embeds with whatever provider powers its chat model.
2. Since #14699 moved the agent step server-side, nothing produces chunk embeddings at all. Chunks are stored with a NULL vector, `search()` filters `embedding IS NOT NULL`, and `knowledgeBaseService.ingestFile` (the deleted piece-side producer's replacement) had zero callers. Every knowledge base answered "No relevant information found" for **every** provider, not just gateways.
3. Opening a Run Agent step that has no model selected yet white-screened the whole step panel.

## Fix

- `core/piece-types/ai-providers.ts` — new `aiProviderUtils.resolveEmbeddingModelId({ provider, config })`. A gateway resolves to `text-embedding-3-small` only when its configured models contain an `openai/`-prefixed id, the only zero-setup evidence that an OpenAI upstream credential is stored in Cloudflare. Drops the now-redundant `supportsEmbedding` capability field (no consumers).
- `server/utils/agent-ai-utils.ts` — gateway embeddings go through Cloudflare's OpenAI passthrough (`.../{account}/{gateway}/openai`), carrying only `cf-aig-authorization`. `/compat` is not used: Cloudflare documents chat completions there, not embeddings. Both gateway URLs now come from one helper.
- `server/api/knowledge-base.service.ts` — dead `ingestFile` replaced by `embedPendingChunks`: lists chunks with no vector, embeds in batches of 50, writes each batch back with one set-based `UPDATE ... FROM unnest(...)`, under a per-file distributed lock so concurrent searches do not each pay for the same backlog. The search path reaches this over an RPC that gives up after 60s, so each call spends at most 20s and leaves the rest for the next search. Blank chunks are skipped (they 400 at every provider).
- `server/api/ee/agent/agent-rpc-handlers.ts` — `executeKnowledgeBaseTool` embeds pending chunks before searching, restoring pre-#14699 behaviour.
- `web` — the duplicate `PROVIDER_EMBEDDING_MODELS` map is deleted; the card and the model-selector footnote both derive from the shared resolver plus the provider's own config, with a gateway-specific message and a skeleton while providers load.
- `web/agent-settings/index.tsx` — read the provider model optionally instead of off an unchecked cast, which is what crashed the panel.

## Verification

- `npm run lint-dev` clean (0 errors).
- New tests red-first against base: 6 resolver cases (`resolveEmbeddingModelId is not a function`), 2 gateway construction cases, 6 `embedPendingChunks` cases (`embedPendingChunks is not a function`). All green after the fix.
- `npm run test-unit`: engine 455, shared 460, sandbox 229, server-utils 142, core-utils 9, pieces-framework 8, web 465 passing. `packages/server/api` unit suite back to the base baseline (6 files / 18 tests failing, all pre-existing and reproduced on stashed main: `file-service-delete`, `worker-group.service`, `remove-deprecated-jobs`, `active-invariant`, `job-broker`, `machine-service`). `web`'s `chunk-reducer.test.ts` (`window is not defined`) also fails identically on base and is what the pre-push hook trips on.
- Gateway request shape asserted in unit tests: URL resolves to `https://gateway.ai.cloudflare.com/v1/{account}/{gateway}/openai/embeddings`, `cf-aig-authorization` present, no `Authorization` header, and the `providerOptions` namespace is asserted to match the SDK's own provider string so `dimensions: 768` cannot silently stop being sent.
- Ran end to end on a real stack (pgvector Postgres, `PGVECTOR_AVAILABLE=true`, real API + web, a Cloudflare Gateway provider row, a real flow with a Run Agent step). Drove the builder as a user with Playwright against the running app.

Visual: builder Run Agent panel, three states driven — gateway with an `openai/` model (card enabled), gateway with Anthropic only (gateway-specific message), and the same setup on base (old unsupported message). Light theme; the change is text and gating only.

Other affected paths: `POST /v1/knowledge-base/files/search` plus `/:id/chunks` and `/:id/store-chunks` are the orphaned half of the old piece-side dance (caller supplies its own vector); left as-is because removing public routes would be a breaking change. The piece's `createEmbeddingModel` has been dead since #14699 and is deliberately untouched, since editing the piece forces a version bump and republish for no user gain.

Product decision embedded: a gateway is offered a knowledge base only when an `openai/`-prefixed model is configured; alternative considered: always offer it and infer the upstream from the chat model prefix, rejected because an Anthropic-only gateway would 401 at search time while the UI claimed support.

Not verified: no live Cloudflare gateway was available, so the passthrough is exercised by URL/header assertions rather than a real round trip. `google-ai-studio` and `google-vertex-ai` gateway prefixes are deliberately out of scope (their passthrough needs `x-goog-api-key` injection I could not verify), as is `CUSTOM`, which has no guessable embedding model id.

<details>

## Plan

- Shared decision, per-caller construction: the embedding model id is resolved in `core-piece-types` (thin, no SDK deps) while each side builds its own client, because `@activepieces/ai-providers` is `ai@7` and the piece is `ai@6` — the split documented in `brain/ai-intelligence/ai-providers.md`.
- Gateway support is inferred from provider config rather than a new config field, so self-hosters get it with zero setup. An explicit embedding-model field is the follow-up that would also unlock `CUSTOM`.
- Ranking correctness stays with `toStorageEmbedding` (truncate + re-normalise), per the documented 1536-vs-768 gotcha, rather than trusting `dimensions` to be honoured.
- Embedding stays lazy at first search, which is exactly what the deleted piece code did — no new job type, no upload-time latency, and provider resolution already exists at that call site.
- Full plan on the Linear ticket (internal). Footprint estimate was ~11 files / ~170 lines; actual is 16 files / ~451 added, the growth being tests (216 lines) and the four review fixes below.

- Anthropic-only gateway → resolver returns undefined, card explains what to add, no runtime 401.
- Gateway with an `openai/` model but no stored OpenAI key → fails loudly at search rather than returning empty.
- Concurrent searches of the same file → per-file lock, so the backlog is embedded once.
- Blank chunk (reachable via `POST /:id/store-chunks`) → skipped instead of poisoning the file permanently.
- Large file on first search → 20s work budget per call, so the backlog drains across successive searches instead of blowing the worker's 60s RPC.
- Provider switched after chunks were embedded → vectors are not re-embedded and ranking degrades silently. Pre-existing; a real fix needs a `modelVersion` column on the chunk table, like `tool_search_index` has.

## Alternatives rejected

- Explicit embedding-model field on gateway/custom provider config → schema + provider-dialog UI + validation, and still needed only for `CUSTOM`.
- Embedding chunks at upload time → adds latency for files nobody searches and changes upload failure semantics.
- Sharing one factory across piece and server → blocked by the `ai@6`/`ai@7` boundary.
- Cloudflare `/compat/embeddings` → undocumented; the provider passthrough mirrors OpenAI's own paths and is the pattern already proven in this repo.

</details>

No API field, endpoint, column, env var, or default changes. `supportsEmbedding` is an internal TS capability field, grep-verified as never serialised and unused. Knowledge base search starting to return results is a fix to broken behaviour, not a behaviour change a self-hoster must act on.

Touches outbound HTTP and a provider credential. The gateway host is hardcoded `gateway.ai.cloudflare.com`; only path segments come from platform-admin provider config, and the `cf-aig-authorization` credential goes to that fixed host only — same construction as the existing chat branch, so no new egress surface. `Authorization` is deliberately never sent, so the stored upstream key in Cloudflare is used rather than a bogus one being forwarded to OpenAI. The new chunk read/write stays project-scoped (`listChunks({projectId, ...})`, `update({id, projectId})`) behind the handler's existing `getFileOrThrow` check, and the new log line carries ids and a count only.


majewskibartosz:
## Visual verification

Builder → Run Agent step settings, driven with Playwright against a real running instance (pgvector Postgres, real API + web, a Cloudflare AI Gateway provider row, a real flow whose agent step is set to `cloudflare-gateway` / `openai/gpt-4.1`).

| Before (this branch reverted) | After |
|---|---|
| before | after |
| "Knowledge base requires a provider that supports embeddings, such as OpenAI or Google." and the footnote reads "This provider does not support knowledge base embeddings." | "Add files or tables as knowledge sources for your agent." with the Add button, and the footnote reads "Embedding model for knowledge base: text-embedding-3-small". |

The honesty gate, same branch, gateway reconfigured with an Anthropic model only:

anthropic only

States driven: gateway with an `openai/`-prefixed model (enabled), gateway with Anthropic only (gateway-specific guidance), and the same gateway+openai setup on base code (old unsupported message). Light theme only, since the change is gating and copy.

Also covered on the way here: opening the step panel before a model is chosen used to white-screen it (`Cannot read properties of undefined (reading 'provider')`), reproduced on base and fixed in this PR, which is what made these screenshots reachable at all.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,36 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+            fn: async () => {
+                const pending = (await this.listChunks({ projectId, knowledgeBaseFileId, embedded: false }))
+                    .filter((pendingChunk) => pendingChunk.content.trim().length > 0)
+                if (pending.length === 0) {
+                    return 0
+                }
+
+                for (let i = 0; i < pending.length; i += EMBED_BATCH_SIZE) {
+                    const batch = pending.slice(i, i + EMBED_BATCH_SIZE)
+                    const embeddings = await embedFn(batch.map((pendingChunk) => pendingChunk.content))
```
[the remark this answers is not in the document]
majewskibartosz: Confirmed and fixed in 7db6ccba83. The worker RPC client is built with a 60s timeout (`packages/server/worker/src/lib/worker.ts:101`), so an unbounded backlog could indeed outlive the call.

`embedPendingChunks` now works to a time budget: it checks `Date.now() - startedAt > EMBED_TIME_BUDGET_MS` (20s) between batches, writes what it finished, logs `embeddedCount` / `remainingCount`, and leaves the rest for the next search, which picks the backlog up where it stopped because pendingness is still just `embedding IS NULL`. 20s leaves 40s of headroom for the query embed plus the vector search in the same RPC.

Covered by a new unit test that fakes 11s per batch and asserts the call stops after 2 batches (100 of 300 chunks) instead of draining the file.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,36 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
-                    chunkIndex: i + j,
-                    metadata: { chunkIndex: i + j, totalChunks: textChunks.length },
-                })
-            }
-        }
-
-        await this.storeChunks({ projectId, knowledgeBaseFileId, chunks: allChunks })
+        return distributedLock(log).runExclusive({
+            key: `kb-embed-${knowledgeBaseFileId}`,
+            timeoutInSeconds: EMBED_LOCK_TIMEOUT_SECONDS,
```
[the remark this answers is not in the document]
majewskibartosz: Both parts fixed in 7db6ccba83.

Lock duration: `EMBED_LOCK_TIMEOUT_SECONDS` is now 60s, not 300s. With the 20s work budget the lock only has to cover a single call, so it no longer sits at the boundary.

Update fanout: the backlog path no longer goes through `storeChunks` per-row updates. It calls a new `setChunkEmbeddings`, which writes a whole batch in one statement:

```sql
UPDATE knowledge_base_chunk AS kbc
   SET embedding = incoming.embedding::vector
  FROM unnest($1::text[], $2::text[]) AS incoming(id, embedding)
 WHERE kbc.id = incoming.id
   AND kbc."projectId" = $3
```

That is one round trip per 50-chunk batch instead of 50, and it keeps the project filter on the write. `storeChunks` is back to its original form, untouched by this PR, since it serves the public `/:id/store-chunks` route rather than this path. Unit test asserts the SQL carries the `"projectId" = $3` predicate and the exact id/vector arrays.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,64 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+                    return 0
+                }
+
+                let embeddedCount = 0
+                for (let i = 0; i < pending.length; i += EMBED_BATCH_SIZE) {
+                    if (Date.now() - startedAt > EMBED_TIME_BUDGET_MS) {
+                        break
+                    }
+                    const batch = pending.slice(i, i + EMBED_BATCH_SIZE)
+                    const embeddings = await embedFn(batch.map((pendingChunk) => pendingChunk.content))
```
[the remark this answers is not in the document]
majewskibartosz: Fixed in 6c292044d4. You are right that a pre-batch check cannot interrupt an in-flight call, so the budget is now enforced by the request itself: each batch gets `AbortSignal.timeout(remainingMs)` for whatever is left of the 20s, and `embedMany` is called with that signal.

If the signal fires mid-batch the call is aborted, and rather than failing the tool the loop keeps every batch already written and breaks (logged as `Ran out of embedding budget mid-batch`). Any non-abort error still propagates, so a genuinely broken provider is not swallowed. New unit test asserts an aborted batch surfaces rather than being silently treated as success, and the handler test now exercises the `(texts, { abortSignal })` signature.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,64 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+    /**
+     * Called on the search path, which the worker reaches over an RPC that gives up after 60s, so a
+     * big backlog is embedded across successive searches instead of all at once: each call spends at
+     * most EMBED_TIME_BUDGET_MS, writes what it finished, and leaves the rest for the next search.
+     */
+    async embedPendingChunks(params: EmbedPendingChunksParams): Promise<number> {
         const { projectId, knowledgeBaseFileId, embedFn } = params
+        const startedAt = Date.now()
+
+        return distributedLock(log).runExclusive({
```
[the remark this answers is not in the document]
majewskibartosz: Fixed in 6c292044d4. `startedAt` now lives inside the `runExclusive` callback, so the clock starts when the lock is acquired rather than when the call entered. A caller that queued behind another embedder gets its own full 20s slice instead of a spent budget.

Worth noting the queued caller usually finds nothing to do anyway: the holder writes each batch before releasing, so the waiter re-reads `embedding IS NULL` and sees only what is genuinely left.

packages/server/api/src/app/ee/agent/agent-rpc-handlers.ts
```diff
@@ -552,6 +552,14 @@ export const agentRpcHandlers = (log: FastifyBaseLogger) => ({
...
         await knowledgeBaseService(log).getFileOrThrow({ projectId, id: input.knowledgeBaseFileId })
         const { model, providerOptions } = await agentHelpers.resolveEmbeddingModel({ platformId, log, ...spreadIfDefined('provider', input.provider) })
+        await knowledgeBaseService(log).embedPendingChunks({
+            projectId,
+            knowledgeBaseFileId: input.knowledgeBaseFileId,
+            embedFn: async (texts, { abortSignal }) => {
+                const { embeddings } = await embedMany({ model, values: texts, providerOptions, abortSignal })
+                return embeddings.map(agentAiUtils.toStorageEmbedding)
+            },
+        })
```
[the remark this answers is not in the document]
majewskibartosz: Fixed in 54e06ee9d3. Agreed that the flat "No relevant information found." is the harmful part: the agent repeats it to the user as fact.

`embedPendingChunks` now returns `{ embeddedCount, remainingCount }`, and when a search comes back empty with chunks still pending, the tool answers "No match in the part of this file indexed so far. N chunk(s) are still being indexed — searching again shortly will cover more of the file." rather than asserting the content is not there. A new handler test asserts the remaining count reaches the tool result and that the flat message is not used in that case.

On the underlying design: fully draining the backlog inside the RPC is what the first round showed cannot be done safely, and moving ingestion to a background job is a larger change than this PR (new WorkerJobType + queue wiring) — worth doing, but as its own ticket. This keeps the same lazy behaviour the piece had before #14699 while making the intermediate state legible instead of silently wrong.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,78 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
-        const textChunks = await this.extractChunks({ projectId, knowledgeBaseFileId })
-        if (textChunks.length === 0) {
-            return
-        }
+        return distributedLock(log).runExclusive({
+            key: `kb-embed-${knowledgeBaseFileId}`,
+            timeoutInSeconds: EMBED_LOCK_TIMEOUT_SECONDS,
+            fn: async () => {
+                // Budget starts here, not before the lock: a caller that queued behind another
+                // embedder must still get its own full slice of work rather than a spent clock.
```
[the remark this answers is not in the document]
majewskibartosz: Fixed in 76c89adb2a, both halves.

Lock wait: `enteredAt` is now captured before the lock and the per-batch budget is computed from it (`EMBED_TIME_BUDGET_MS - (Date.now() - enteredAt)`), so waiting and embedding share one 20s budget instead of the wait being free. The lock timeout is down to 20s, and failing to take it is no longer an error: the call logs, returns the pending count, and searches what is already indexed. A queued caller can therefore cost at most ~20s before the search proceeds, well inside the 60s RPC.

Partial results: matches returned while chunks are still pending now come back as `{ matches, note }` with the note stating how many chunks are still indexing, so a non-empty answer is no longer presented as complete coverage. Bare array when nothing is pending. Two new handler tests cover both shapes.

One note on the propagation path: the lock callback now returns `{ embedError }` instead of throwing, so a genuine embedding failure is still rethrown to the caller and only lock-acquisition failure degrades to a partial search.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+                return { result: { embeddedCount, remainingCount } }
+            },
+        }))
 
-        const textChunks = await this.extractChunks({ projectId, knowledgeBaseFileId })
-        if (textChunks.length === 0) {
-            return
+        // Losing the race is not a failure: another call is embedding this file, so search what is
+        // already indexed rather than failing the tool or waiting past the RPC deadline.
+        if (isNil(outcome)) {
```
james-novino, line 136: **[CONFIRMED · correctness]** `tryCatch` wraps the whole `runExclusive` call, so any error thrown *inside* the critical section (`listChunks`, the `setChunkEmbeddings` UPDATE, a pgvector cast failure) is indistinguishable from failing to acquire the lock: redlock's `using` rethrows the routine's error, it lands in this `isNil(outcome)` branch, gets logged at `warn` as "Could not take the embedding lock", and the method **returns success**.

Concrete trigger: a provider that returns non-768-dim vectors makes Postgres raise `expected 768 dimensions` — every subsequent search then logs a lock-contention warning and tells the agent "N chunk(s) are still being indexed — searching again shortly", forever, while the real DB error is buried in the `lockError` log field. A `let enteredCriticalSection = false` set as `fn`'s first statement lets this branch distinguish acquisition failure (benign, fall back to search) from an in-section failure (rethrow loudly).

packages/server/api/src/app/ee/agent/agent-rpc-handlers.ts
```diff
@@ -552,6 +552,14 @@ export const agentRpcHandlers = (log: FastifyBaseLogger) => ({
         const { projectId, platformId } = conversation
         await knowledgeBaseService(log).getFileOrThrow({ projectId, id: input.knowledgeBaseFileId })
         const { model, providerOptions } = await agentHelpers.resolveEmbeddingModel({ platformId, log, ...spreadIfDefined('provider', input.provider) })
```
james-novino, line 554: **[CONFIRMED · correctness]** The embedding model here comes from the *calling step's* provider (falling back to the platform chat provider), and chunks are embedded lazily by whoever searches first — but `knowledge_base_chunk` records no producing model and `search()` has no model filter. KB files are project-scoped, so two agent steps on different providers (or an admin switching the platform provider mid-index) can embed slices of the *same file* into two unrelated vector spaces: `text-embedding-3-small` truncated to 768 and `text-embedding-004` (native 768) both fit `vector(768)`, cosine-rank together, and silently poison relevance with no error and no repair path (only NULL rows are ever re-embedded).

`tool_search_index` solved exactly this with a `modelVersion` column that every query filters on — the same pattern fits here.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
-            for (let j = 0; j < batch.length; j++) {
-                allChunks.push({
-                    content: batch[j],
-                    embedding: embeddings[j],
-                    chunkIndex: i + j,
-                    metadata: { chunkIndex: i + j, totalChunks: textChunks.length },
-                })
-            }
+        if (isNil(outcome.result)) {
+            throw outcome.embedError
```
james-novino, line 141: **[CONFIRMED · correctness]** This throw runs before the caller ever reaches `search()`, so one *permanently* un-embeddable chunk makes the whole file unsearchable forever — including its already-indexed chunks. Reachable today: `chunkCsvText` emits a lone oversized row as an unbounded chunk (the split guard requires `currentRows.length > 0`), a >8192-token chunk 400s the entire 50-chunk batch at OpenAI (non-retryable), and pending chunks are re-read in `chunkIndex` order so the same poison batch fails first on every search. `POST /:id/store-chunks` also accepts unbounded `content`. Before this PR the search path always served whatever was indexed; consider searching what is indexed and surfacing the embed failure in the result instead of failing the whole tool call.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+
+        const { data: outcome, error: lockError } = await tryCatch(() => distributedLock(log).runExclusive({
+            key: `kb-embed-${knowledgeBaseFileId}`,
+            timeoutInSeconds: EMBED_LOCK_TIMEOUT_SECONDS,
+            fn: async (): Promise<EmbedPendingChunksOutcome> => {
+                // The lock wait counts against the budget, so a caller that queued behind another
+                // embedder cannot start a fresh slice late enough to outlive the worker's RPC.
+                const startedAt = Date.now()
+                const pending = (await this.listChunks({ projectId, knowledgeBaseFileId, embedded: false }))
+                    .filter((pendingChunk) => pendingChunk.content.trim().length > 0)
```
james-novino, line 95: **[CONFIRMED · correctness]** Blank chunks are unaccounted in both directions. (1) They are stored at ingest (`storeChunks` writes `content ?? ''`; nothing filters whitespace) and stay `embedding IS NULL` forever, so `countPendingChunks` — used on the lock-fallback path — counts them and reports "N chunk(s) are still being indexed" indefinitely on a fully-indexed file. (2) A file whose *every* chunk is blank (scanned/image-only PDF) filters to `pending.length === 0` here and returns `{embeddedCount: 0, remainingCount: 0}` — byte-identical to "fully indexed" — so the agent confidently answers "No relevant information found" for a 0%-indexable file while the UI shows a healthy chunk count. One shared non-blank pending predicate (e.g. `btrim(content) <> ''` in the WHERE) used by both the fetch and the count keeps the two paths agreeing.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+                    if (remainingMs <= 0) {
+                        break
+                    }
+                    const batch = pending.slice(i, i + EMBED_BATCH_SIZE)
+                    // The signal is what actually enforces the budget: without it one slow provider
+                    // call started just inside the deadline could still outlive the worker's RPC.
+                    const abortSignal = AbortSignal.timeout(remainingMs)
+                    const { data: embeddings, error } = await tryCatch(() => embedFn(batch.map((pendingChunk) => pendingChunk.content), { abortSignal }))
+                    if (isNil(embeddings)) {
+                        if (!abortSignal.aborted) {
```
james-novino, line 112: **[CONFIRMED · correctness]** This classifies by *signal state*, not error type. ai-sdk's retry backoff is signal-aware: a 429 whose `retry-after` outlives `remainingMs` aborts the retry *sleep*, the rejection arrives with `abortSignal.aborted === true`, and a hard provider failure is logged at `warn` as "Ran out of embedding budget" and returned as success — a sustained rate limit becomes an invisible indexing stall with no error-level signal, and the agent is told to search again shortly. Gate the budget branch on `abortSignal.aborted && (error.name === 'AbortError' || error.name === 'TimeoutError')` and route everything else to `embedError`.

packages/server/utils/src/agent-ai-utils.ts
```diff
@@ -66,11 +72,10 @@ function createChatModel({ provider, auth, config, modelId, metadata, webSearchE
 }): LanguageModel {
     if (provider === AIProviderName.CLOUDFLARE_GATEWAY) {
         const { apiKey } = auth as BaseAIProviderAuthConfig
-        const { accountId, gatewayId } = config as CloudflareGatewayProviderConfig
         const { model: actualModelId } = splitCloudflareGatewayModelId(modelId)
```
james-novino, line 75: **[PLAUSIBLE · correctness — question for the author]** Pre-existing (from #14446), but this PR's KB flow hangs off it: `createChatModel` sends the *prefix-stripped* id (`gpt-4.1`) to `/compat`, while the other two gateway callers send the full id to the same endpoint — `cloudflare-gateway-provider.ts` `validateConnection` posts `model.modelId` (its own error text says ids must be `{provider_name}/{model_name}`), and the ai piece's `handleDefaultAiGatewayProvider` passes the full `modelId`. Cloudflare's chat-completions docs for the unified endpoint also specify `{provider}/{model}`. Which endpoint/model-id form did the e2e verification actually exercise? If `/compat` tolerates bare ids it's undocumented; if not, the gateway chat turn this KB tool depends on can't route. (The embedding branch below is correct either way — the `/openai` passthrough wants the bare id.)

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
 export const knowledgeBaseService = (log: FastifyBaseLogger) => ({
-    async ingestFile(params: IngestFileParams): Promise<void> {
+    /**
+     * Called on the search path, which the worker reaches over an RPC that gives up after 60s, so
+     * everything here is bounded: waiting for the lock and embedding share one budget, and whatever
+     * is not finished is left for the next search rather than held onto until the RPC dies.
+     */
+    async embedPendingChunks(params: EmbedPendingChunksParams): Promise<EmbedPendingChunksResult> {
         const { projectId, knowledgeBaseFileId, embedFn } = params
+        const enteredAt = Date.now()
```
james-novino, line 85: **[CONFIRMED · correctness/efficiency]** `enteredAt` starts the budget *before* lock acquisition, and `distributedLockFactory` derives `retryCount = ceil(timeout/200)` = 100 retries × 200ms ≈ the full 20s for a contended lock — so a queued caller can win the lock with zero budget left, and it only finds out *after* `listChunks` has already materialized the entire pending backlog. Net: a second concurrent search of the same file burns ~20s of its 60s RPC to embed nothing. Checking the remaining budget as the first statement inside `fn` (before any fetch) makes the late acquirer exit cheaply. Related: with `maxRetriesPerRequest: null` + the offline queue, a stalled Redis leaves `runExclusive` pending until the 60s RPC timeout — `createConfiguredKnowledgeBaseTools` is the one configured tool factory without a `withToolTimeout` wrapper.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+                        break
+                    }
+                    if (embeddings.length !== batch.length) {
+                        return { embedError: new Error(`Embedding count mismatch: expected ${batch.length}, got ${embeddings.length}`) }
+                    }
+                    await this.setChunkEmbeddings({
+                        projectId,
+                        chunks: batch.map((pendingChunk, index) => ({ id: pendingChunk.id, embedding: embeddings[index] })),
+                    })
+                    embeddedCount += batch.length
```
james-novino, line 125: **[PLAUSIBLE · correctness]** `embeddedCount += batch.length` counts rows *sent*, not rows *written* — `setChunkEmbeddings` discards the UPDATE result, so a batch whose rows vanished mid-flight (file deleted after `getFileOrThrow`; `ON DELETE CASCADE` removes the chunks) reports full success and `remainingCount` computes 0. TypeORM's postgres driver returns `[rows, affectedRowCount]` for UPDATE statements, so the write count is already available — returning it and summing that keeps the numbers honest.

packages/web/src/features/agents/ai-model/index.tsx
```diff
@@ -66,6 +56,7 @@ export function AIModelSelector({
     aiModelHooks.useListProviders();
   const { data: models = [], isLoading: modelsLoading } =
     aiModelHooks.useGetModelsForProvider(selectedProvider);
+  const { embeddingModel } = aiModelHooks.useEmbeddingModel(selectedProvider);
```
james-novino, line 59: **[CONFIRMED · correctness]** `AIModelSelector` destructures only `embeddingModel` and ignores the `isPending` the new hook returns — with a cold `['ai-providers']` cache (nothing else on this page warms it), the footnote renders "This provider does not support knowledge base embeddings." for OpenAI on first paint, then flips once the query lands. The pre-PR constant map was synchronous and couldn't flash. `KnowledgeBaseSection` got a `Skeleton` for exactly this case; this consumer needs the same. Also worth distinguishing: the hook returns the identical shape for "fetch failed" and "provider genuinely lacks embeddings", so a 500 from `GET /v1/ai-providers` permanently shows the wrong message — and for a gateway, tells the admin to add an `openai/` model when the real problem is the fetch.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+        const enteredAt = Date.now()
+
+        const { data: outcome, error: lockError } = await tryCatch(() => distributedLock(log).runExclusive({
+            key: `kb-embed-${knowledgeBaseFileId}`,
+            timeoutInSeconds: EMBED_LOCK_TIMEOUT_SECONDS,
+            fn: async (): Promise<EmbedPendingChunksOutcome> => {
+                // The lock wait counts against the budget, so a caller that queued behind another
+                // embedder cannot start a fresh slice late enough to outlive the worker's RPC.
+                const startedAt = Date.now()
+                const pending = (await this.listChunks({ projectId, knowledgeBaseFileId, embedded: false }))
```
james-novino, line 94: **[CONFIRMED · efficiency]** `listChunks({embedded: false})` has no `take` and selects `content`, so every search of a partially-indexed file loads the *full text of every pending chunk* into the API heap, while the 20s budget consumes only a few 50-chunk batches — the rest is discarded and re-loaded on the next search. At the 25MB `MAX_FILE_SIZE_MB` default that's ~14k chunks ≈ 25MB of strings per search in the process serving all HTTP traffic, and O(n²) cumulative reads over the file's indexing life. A `take: EMBED_BATCH_SIZE` paged fetch inside the loop bounds peak memory to ~100KB — embedded rows drop out of the `IS NULL` predicate, so a repeated bounded fetch advances naturally.

packages/server/api/src/app/knowledge-base/knowledge-base.service.ts
```diff
@@ -71,33 +75,96 @@ async function extractTextFromFile(fileBuffer: Buffer, fileName: string): Promis
...
+    /**
+     * Called on the search path, which the worker reaches over an RPC that gives up after 60s, so
+     * everything here is bounded: waiting for the lock and embedding share one budget, and whatever
+     * is not finished is left for the next search rather than held onto until the RPC dies.
+     */
+    async embedPendingChunks(params: EmbedPendingChunksParams): Promise<EmbedPendingChunksResult> {
         const { projectId, knowledgeBaseFileId, embedFn } = params
+        const enteredAt = Date.now()
+
+        const { data: outcome, error: lockError } = await tryCatch(() => distributedLock(log).runExclusive({
```
james-novino, line 87: **[CONFIRMED · efficiency]** The steady state of a knowledge base is embedded-once-searched-thousands-of-times, but every search takes the distributed lock and runs the pending scan before it can learn there's nothing to do — 2+ Redis round-trips plus a Postgres scan of the file's whole `(projectId, knowledgeBaseFileId)` index range (the HNSW index can't serve `IS NULL`, and there's no partial index), per agent tool call, with concurrent agents on one hot file serializing globally. A cheap pre-lock existence/count probe (repo precedent: `existsBy` in several services) skips all of it when the file is fully indexed, and a partial index `WHERE embedding IS NULL` makes that probe O(1). The independent `embed(query)` call could also run concurrently with this instead of after it.

packages/server/api/test/unit/app/knowledge-base/knowledge-base-service.test.ts
```diff
@@ -7,6 +8,7 @@ const mockSave = vi.fn()
 const mockFind = vi.fn()
 const mockInsert = vi.fn()
 const mockUpdate = vi.fn()
+const mockCountPending = vi.fn()
```
james-novino, line 11: **[CONFIRMED · test-coverage]** `mockCountPending` is declared and never wired — and it looks like the leftover of the mock this file is missing: unlike every sibling lock test (`flow-lock.test.ts`, `credits-gate.test.ts`, `free-legacy-comp.test.ts`), this suite does not mock `src/app/database/redis-connections`, so the new `embedPendingChunks` tests drive a *real* redlock via `redis-memory-server` (`AP_REDIS_TYPE=MEMORY`), and the budget test wraps that real lock in `vi.useFakeTimers()` — redlock's retry/extension timers and ioredis reconnect timers all freeze, so it passes only when acquisition succeeds on the first attempt with a warm cached binary. Meanwhile the lock-contention fallback branch (`isNil(outcome)` → `countPendingChunks`) has zero coverage — which is the branch the error-conflation bug lives in. A pass-through `runExclusive` mock makes all of these deterministic.

james-novino (commented):
**Alvys review of this PR** — we run a Cloudflare AI Gateway in production and built the same capability on our side (piece-level), so we verified this change against a working implementation and ran a deep multi-pass review (10 finder angles, adversarial verification per finding). Posting the verified findings inline; three items that don't anchor to diff lines:

**1. None of this PR's four test suites run in CI.** `ci.yml` filters `engine/shared/sandbox/ai-providers/pieces-framework(/web)` — `@activepieces/server-utils` is absent, so `agent-ai-utils.test.ts` never runs; `packages/core/piece-types` has no `test` script at all, so `ai-providers.test.ts` is unrunnable; and `api`'s `test-unit` (`vitest run test/unit`) is not invoked by any workflow, so both the knowledge-base and RPC-handler suites are skipped. A regression in `resolveEmbeddingModelId` or `embedPendingChunks` merges green today.

**2. Merge conflict resolution** (branch is CONFLICTING vs `main`): the only conflict is `packages/web/public/locales/en/translation.json`, and the correct resolution is **take `main`'s version of the file and re-add this PR's single new key** — `"Add a model whose id starts with openai/ to this Cloudflare AI Gateway to use a knowledge base."` — right after the `"You've used {percentage}% of your credits."` entry. (A naive both-sides union resurrects ~100 keys `main` has since moved/removed.) We ran the PR's suites on the merged result: all green.

**3. Verification results**: functional parity with our production behavior confirmed — `/openai` passthrough for embeddings (`/compat` is chat-only), `cf-aig-authorization`-only auth (no `Authorization` header leaking to the BYOK upstream), `text-embedding-3-small` at 768 dims. All 112 tests on the touched files pass locally, lint clean.

We have a patch applying fixes for 11 of the 15 findings (error classification, bounded paged fetch, pre-lock short-circuit, written-row counting, web loading/error states, CI wiring + lock-mocked tests) rebased on this branch — happy to share it however is easiest.

packages/server/api/test/unit/app/ee/agent/agent-rpc-handlers.test.ts
```diff
@@ -25,22 +25,25 @@ const { mockSet, mockWhere, mockAndWhere, mockExecute, mockFindOneBy, mockFindOn
...
-const { mockEmbed } = vi.hoisted(() => ({
+const { mockEmbed, mockEmbedMany } = vi.hoisted(() => ({
     mockEmbed: vi.fn().mockResolvedValue({ embedding: new Array(768).fill(0.1) }),
+    mockEmbedMany: vi.fn().mockResolvedValue({ embeddings: [new Array(768).fill(0.1)] }),
 }))
 
 vi.mock('ai', async (importOriginal) => ({
     ...(await importOriginal<Record<string, unknown>>()),
     embed: () => mockEmbed(),
+    embedMany: () => mockEmbedMany(),
```
james-novino, line 46: **[CONFIRMED · test-coverage]** The mock discards call arguments, so no test can assert that `embedFn` forwards `abortSignal`, `providerOptions`, or the batch values into `embedMany` — and the abort signal is the thing the service comment calls "what actually enforces the budget". Drop `abortSignal` from the `embedMany` call in `embedFn` today and every test stays green while the 20s budget silently stops bounding provider calls.
```suggestion
    embed: (args: unknown) => mockEmbed(args),
    embedMany: (args: unknown) => mockEmbedMany(args),
```

packages/core/piece-types/src/lib/ai-providers.ts
```diff
@@ -292,12 +294,32 @@ function buildProviderCapabilities(provider: AIProviderName): AIProviderCapabili
...
 }
 
+/**
+ * A Cloudflare gateway has no embedding model of its own: it proxies whichever upstream the admin
+ * wired into it. An `openai/`-prefixed model in its config is the only zero-setup proof that an
+ * OpenAI upstream credential is stored in the gateway, so embeddings are offered on that evidence
+ * and withheld otherwise — an Anthropic-only gateway must not advertise a knowledge base it would
+ * fail to embed for.
+ */
+function resolveEmbeddingModelId({ provider, config }: { provider: AIProviderName, config: unknown }): string | undefined {
```
james-novino, line 309: **[CONFIRMED · reuse]** Gateway-model→upstream knowledge now lives in three diverging places: this resolver hand-rolls an `openai/` prefix check and validates against piece-types' *duplicate* `CloudflareGatewayProviderConfig` schema, while `@activepieces/shared` already owns `splitCloudflareGatewayModelId` + `CF_GATEWAY_SUBMODEL_TO_PROVIDER` and the authoritative wire schema — and `pieces/community/ai`'s `createEmbeddingModel` still reads the raw `defaultEmbeddingModel` capability with no gateway case (dead but exported). Drift scenario: a new upstream prefix (`azure-openai/…`) added to shared's map routes chat correctly while KB support silently stays off; if the two schema copies diverge by one required field, `safeParse` starts failing and every gateway silently loses embedding support with no error path.

PR #14848, opened 2026-08-16
feat(agent): add starter agents and drafting one from a sentence

hazemadelkhalel:
## Description

Two ways to start an agent, because a blank instructions box is the worst possible first screen. Continues #14846.

**`GET /v1/agents/templates`** — eight starter agents. They are static, and they deliberately carry **instructions only, no tools**. A "Support triage" starter that referenced Zendesk would look ready and fail on first run, which is exactly the failure `.claude/rules/self-hosting.md` exists to prevent. This is the review focus the roadmap called out, and the test proves it rather than asserting it: a template is fetched, created, and **published** on a platform with no AI provider and no connections configured.

**`POST /v1/agents/draft`** — turns a sentence into a reviewable definition (`displayName`, `description`, `icon`, `color`, `instructions`). This one genuinely needs an AI provider, so it fails clearly instead of half-working, and templates keep serving either way. That's the honest degradation the rule asks for: the create screen always has something to offer, even on a fresh self-host.

Drafting is gated by `WRITE_AGENT` on the project named in the body — the same permission as creating, since that's what you're about to do. It does not create anything; the client reviews the result first.

The template instructions are written to be read *by the agent*: how to decide, and what not to do, rather than a job title. "Flag likely duplicates rather than merging them yourself" and "never promise a fix date" are the parts that make a starter useful instead of decorative.

## How was this tested?

39 tests in the agent suite, all green on Postgres **and on PGlite from a fresh database**. Full runs: EE 163 passing (16 files), CE 872 passing (74 files). Lint and typecheck clean. 217 / 600 meaningful lines in `server/api`, under budget without a label.

Covering the things a user would notice:

- every template is creatable and publishable **with no AI provider configured** — the zero-setup rule, proven end to end
- every template carries non-empty instructions and a valid icon and colour, so a starter can never render broken
- template ids are distinct, since the client will key on them
- the AI draft fails honestly with no provider while templates keep working
- a viewer cannot draft, so drafting can't be used to spend credits without write access
- `/templates` does not get swallowed by `/:id`, alongside the existing `/conversations`, `/memory` and `/runs` routes

`POST /draft` spends AI credits, so it is gated by `WRITE_AGENT` on a project resolved through the standard body resource, and a viewer is refused — covered by a test. The prompt is bounded by the shared text ceiling. Templates are static server-side data with no user input path. Neither endpoint reads or writes an agent row.


hazemadelkhalel:
On the metering finding — it's correct, and I'm deliberately not fixing it in this PR. Explaining rather than silently leaving it.

`POST /draft` checks credits but never debits them, so draft spend is invisible and the admission check only fires once *other* usage (chat, flows) has exhausted the balance. That is a real gap.

I tried to close it and backed out. Neither existing credit source fits a pre-creation helper: `CreditUsageSource.AI` requires `FlowRunCreditConsumptionProperties` (there is no flow run), and `CHAT` requires `conversationId` and `turnIndex` (there is no conversation). Billing it as `CHAT` with a synthetic conversation id would pollute chat analytics with non-chat events.

Closing it properly means a fourth credit source with its own properties type and a matching meter on the billing provider — and, more importantly, a decision about **what a draft should cost, or whether it should cost anything at all.** That is a pricing call, not an engineering one, and inventing a number here risks either overcharging for "help me phrase this" or shipping a meter the provider has no matching feature for, which fails silently.

What is in place meanwhile, and why I think it is a reasonable interim:

- per-caller rate limit of 20/minute, keyed on the principal so one API key cannot exhaust another caller or the platform owner
- credit admission check, so a platform already out of credits cannot draft
- prompt bounded to 2,000 characters rather than the 51,200 chat ceiling, cutting worst-case spend ~25x
- a 30s timeout and `maxRetries` left at the SDK default, so a hung provider cannot hold a connection indefinitely

Happy to add the metering in a follow-up once the credit weight is decided. Flagging it for the pricing owner rather than guessing.

On the earlier note about `extractJsonObject` placement: this repo's CLAUDE.md specifies `Imports → Exported functions/constants → Helper functions → Types`, so helpers belong after exports here. That helper is gone now anyway, replaced by the SDK's structured output.

PR #14869, opened 2026-08-17
fix(pieces): jira cloud updated-issue trigger drops laggy updates

### LINKED ISSUE #14863: [BUG] Jira Cloud polling triggers permanently skip updates missing from the poll that advances the watermark
## Problem

Jira Cloud `Updated Issue` (custom JQL, 5-min poll) skips some qualifying updates while firing for a later one, during normal live polling well after the last publish. activepieces/activepieces#13311 republish fix already present (self-hosted 0.87.0, jira-cloud 0.3.10) — distinct defect. One observed cycle: issue A updated 5:10 PM never fired, issue B ~5:14 never fired, issue C ~5:17 fired at 5:20. Test Trigger with the same JQL returns all three; only 4 issues matched the window. Skipped events are lost permanently, silently.

Ref: Pylon 4794

## Root cause

`packages/pieces/common/src/lib/polling/index.ts:79-86` (TIMEBASED): watermark = max `updated` epoch across issues returned by that poll's fetch; emit filter is `epoch > previousWatermark`. Any qualifying issue absent from the fetch that advances the watermark past its `updated` time is lost forever — the next window excludes it and the epoch filter drops it.

Why an issue is absent from that fetch (`packages/pieces/community/jira-cloud/src/lib/triggers/updated-issue.ts:19-34` — JQL `<custom> AND updated > 'YYYY-MM-DD HH:mm'`, POST `/search/jql`, one page of 50, no ORDER BY):

1. `/search/jql` is documented eventually consistent — recently updated issues can be missing from results (worse under bulk/automation load). Fits the timeline: 5:15 poll misses A+B (index lag), 5:20 poll returns C (updated 5:17), watermark jumps to 5:17, A+B excluded forever. Atlassian's `reconcileIssues` (read-after-write consistency) not used.
2. TZ mismatch: `dayjs(lastFetchEpochMS).format('YYYY-MM-DD HH:mm')` renders in the worker's local TZ; Jira parses bare JQL datetimes in the API user's profile TZ. UTC worker + UTC+8 profile widens the window ~8h → far more than 50 matches possible → unordered 50-item page can drop qualifying updates while a later-updated one on the page advances the watermark.
3. No pagination past 50 (`searchIssuesByJql` supports `nextPageToken`; trigger fetches one page), no `ORDER BY updated ASC`.

Same class as activepieces/activepieces#14781 (mssql group-mode watermark loss). All jira-cloud polling triggers share the pattern; the watermark flaw is generic to `DedupeStrategy.TIMEBASED` consumers with eventually-consistent sources.

## Fix

* Lookback overlap (query from watermark minus a safety margin, or pass `reconcileIssues`) + dedupe by issue id + `updated` so overlap doesn't re-fire.
* `ORDER BY updated ASC` + paginate via `nextPageToken`.
* Format the JQL timestamp in the Jira account profile TZ (GET `/myself`) instead of worker TZ.
* Cap watermark advance at `now - safetyLag`.

kishanprmr:
## Description

Jira Cloud's `/search/jql` endpoint is [eventually consistent](<https://developer.atlassian.com/cloud/jira/platform/search-and-reconcile/>) — a recently updated issue can be temporarily missing from search results. The `Updated Issue` trigger's `TIMEBASED` polling checkpoint advances to the newest `updated` timestamp seen on each poll regardless of which issues were actually returned, so if a laggy-indexed issue's update wasn't visible yet while a later-updated issue was, the checkpoint jumps past the laggy issue's timestamp — and since the next poll's query is `updated > <checkpoint>`, that issue can never match again. It's silently, permanently dropped, not just delayed.

A second, compounding bug: the query's bare JQL timestamp (`updated > '2026-08-17 17:20'`) was rendered via `dayjs(...).format()` with no explicit timezone. The worker process forces `process.env.TZ = 'UTC'`, so that string is always a UTC wall-clock reading — but Jira parses a bare JQL datetime literal in the *searching account's own Jira profile timezone*, not UTC. For any connected account whose profile isn't set to UTC, this silently shifts the query window by however many hours that account is offset, which can also make it more likely to exceed the (previously unpaginated, unordered) 50-result page.

Fix, scoped to `updated-issue.ts` + shared helpers in `common/index.ts`:

* Query with a 10-minute lookback buffer past the checkpoint (`updated > checkpoint - 10min`), so a laggy issue gets re-queried across subsequent polls instead of falling permanently out of range.
* Dedupe re-fetched issues by issue id + their own `updated` timestamp, stored in the flow's own key-value store, so the lookback buffer's overlap never causes a duplicate fire — only an issue whose `updated` value actually advances again fires again.
* Format the JQL timestamp in the connected account's real Jira profile timezone (`GET /myself` → `timeZone`) instead of the worker's forced-UTC local time.
* Add `ORDER BY updated ASC` and follow `nextPageToken` (capped at 10 pages) so a wider-than-expected match can't silently truncate.

Not included in this PR (tracked as follow-ups): the same pattern exists in Jira Data Center's `updated-issue.ts` and in the other jira-cloud triggers built on this pattern (`new-issue`, `new-comment`, `issue-assigned`, `new-attachment`, `updated-issue-status`).

Fixes activepieces/activepieces#14863

## How was this tested?

Manual testing against a live Jira Cloud instance (Cloud edition path):

* Bulk-transitioned 7, 20, and 25 issues at once (including multi-step transitions) and watched the flow's Runs history across several 1-minute poll cycles.
* Confirmed the total distinct issue count always reconciled — including cases where Jira's search index visibly lagged and an issue only appeared on a later poll instead of never appearing at all.
* Confirmed re-fires of the same issue only happened when it was genuinely updated a second time (cross-checked against Jira's own issue History/changelog for several repeated cases) — no spurious duplicates from the lookback buffer's overlap re-fetching.
* `npx turbo run lint build --filter=@activepieces/piece-jira-cloud`: 0 errors, 0 new warnings (all pre-existing), clean build.
* No automated test added for this scenario — discussed and intentionally deferred.


kishanprmr:
@greptileai You're right that this specific gap exists, but I don't think it's solvable with the piece's own local state, and I chose it deliberately over the alternative.

The problem: after the fact, there's no way to distinguish "this issue's update was already correctly delivered by the old code" from "this issue's update was already silently lost by the old code" — both look identical (`epoch <= old checkpoint`, no entry in a seen-map that never existed before the upgrade). The pre-fix code only ever tracked a single scalar checkpoint, never a per-issue delivery ledger, so that information is genuinely gone by the time this code runs. The only real ground truth would be Activepieces' own flow-run history, which isn't something a piece can query from inside its own execution.

Given that, there were two options for the one transitional poll right after upgrade:

1. **(implemented)** Suppress emission for anything `<= old checkpoint`, but still record it as seen. Zero new duplicates, guaranteed. Cost: an issue that was *already* lost before the upgrade stays lost — but that's a one-time, <=10-minute-wide window, only at the exact moment of upgrade.
2. **(rejected)** Don't suppress — emit everything in the widened window. This would rescue that rare pre-upgrade loss, but at the cost of duplicate-firing every issue that had *already been correctly delivered* in that window, which per Atlassian's own docs ("most changes visible within seconds") is the common case, not the rare one.

I biased against duplicates: a duplicated side effect (double notification, double invoice, etc.) is generally a worse outcome for a downstream automation than a rare missed event confined to the moment of a version upgrade. There's no design here that gets both zero risk of missed-forever and zero risk of duplicate using only what the piece can see — it's a genuine tradeoff, not an oversight, and I'm choosing the side that avoids duplicates by default.

kishanprmr:
@greptileai Confirmed, thanks for checking — that's exactly the intent: only the first-run transition suppresses emission, every fetched item (suppressed or not) still gets merged into the store so a genuine later update on the same issue fires normally. Appreciate the thorough review on this one.

PR #14876, opened 2026-08-17
fix(auth): store one-time codes as a keyed digest and sweep the expired ones

AhmadTash:
## Description

Stacked on #14874 — both change `otp-service.ts`, so this needs that one first.

### One-time codes were stored in plain text

The `otp` table held the six-digit code as-is, and `confirm()` compared it with `===`. So read access to the database was enough to sign in as anybody — no password, no second factor — for the ten minutes the code was alive. That reaches further than the app: a read replica, a nightly dump, a support query, or any SQL read primitive inherits it. Worth noting the asymmetry it created — the password on the same identity has always gone through `passwordHasher.hash()`, while the credential that *bypasses* the password sat in cleartext.

Found while testing #14874 on a preview: reading the live code straight out of Postgres is how I fed the "correct code" to the verify endpoint.

Codes are now stored as **HMAC-SHA256 keyed with `jwtUtils.getJwtSecret()`**, compared with `timingSafeEqual`.

A bare hash would not have been enough — six digits is a million candidates, so an unkeyed digest falls to an offline sweep in milliseconds by anyone holding the table. Keying it means the database alone is insufficient, because the key never lives there. bcrypt via the existing `passwordHasher` was the alternative; HMAC was chosen because it adds no per-verify latency and the secret is already present on every instance, so there is nothing for a self-hoster to configure.

### Re-sending needed the plaintext back

The row no longer holds the code, so the re-send path recovers it from a short-lived cache keyed by identity and type, with a TTL equal to the code's lifetime.

Minting a fresh code on every request would have been simpler and I nearly did it — but it lets anyone who knows an address invalidate the owner's in-flight code at will, so the existing re-send behaviour is preserved instead. That cache lives outside the database, which is the store this finding is about.

### Expired codes were never deleted

They sat in the table until the same person happened to request another one. Two changes: `confirm()` discards a code it finds expired, and a new `OTP_CLEANUP` system job sweeps every ten minutes for whatever nobody comes back for.

### The migration clears rather than converts

A stored plaintext and a stored digest are both just strings, so nothing can tell them apart at verify time and a converted row would silently never match again. `1826000000000-ClearPlaintextOtps` deletes outstanding rows. Every code in that table expires within ten minutes, so the cost is that anyone mid-sign-in at upgrade asks for a fresh one.

## How was this tested?

- `otp-service.test.ts` — **17/17**, including three new cases pinning the properties that matter: the stored value is never what was sent and looks like a digest, the digest offered *as* a code is refused, and an expired row actually disappears. Plus two on `deleteExpired` (removes what has aged out, leaves what is in flight).
- `passwordless-authn.test.ts` — **21/21**. Its `storedOtp` helper now returns the delivered code from the cache, since the row deliberately no longer carries it, and a new assertion checks the row holds a 64-char digest.
- Full CE integration suite — **77 files / 920 tests passing**.
- `turbo run typecheck lint --filter=api` clean.

Edition paths: the OTP service and the sweep are shared code registered unconditionally in `app.ts`. Exercised on **CE**. Not yet re-run on a preview — the EE preview for #14874 is the natural place, and the check I would run there is that a code still arrives by email and still works, with the row holding only a digest.

Fixes # (issue)

The migration invalidates any code outstanding at upgrade, so someone mid-sign-in must request a fresh one. Entry added to `docs/install/reference/breaking-changes.mdx`. No new configuration — the HMAC key comes from a secret every instance already has.

### Security impact?  (required — chosen deliberately)

Touches authentication and credential storage at rest. Risk and mitigation are described above. Pre-existing: the plaintext storage dates from the original OTP entity, not from the passwordless work.


packages/server/api/src/app/authentication/otp/otp-service.ts
```diff
@@ -105,7 +130,40 @@ async function countAttempt(otpId: string): Promise<void> {
 
 async function discard({ otp, identityId, type, log }: DiscardParams): Promise<void> {
     await repo().delete({ id: otp.id })
-    log.warn({ identityId, type }, '[otpService#confirm] attempt budget exhausted, credential discarded')
+    await forgetCachedCode({ identityId, type })
+    log.warn({ identityId, type }, '[otpService#confirm] credential discarded')
+}
+
+async function digestOf(code: string): Promise<string> {
```
MrChaker: @AhmadTash we have a encryption lib that you can use
packages/server/api/src/app/helper/encryption.ts

packages/server/api/src/app/app.ts
```diff
@@ -283,6 +284,24 @@ export const setupApp = async (app: FastifyInstance): Promise<FastifyInstance> =
         },
     })
 
+    systemJobHandlers.registerJobHandler(SystemJobName.OTP_CLEANUP, async () => {
+        const removed = await otpService(app.log).deleteExpired()
+        if (removed > 0) {
+            app.log.info({ removed }, '[otpCleanup] expired one-time codes removed')
+        }
+    })
+    await systemJobsSchedule(app.log).upsertJob({
```
MrChaker: @AhmadTash move this to otp module . follow  SystemJobName.FILE_CLEANUP_TRIGGER as example

packages/server/api/src/app/authentication/otp/otp-service.ts
```diff
@@ -34,34 +36,58 @@ export const otpService = (log: FastifyBaseLogger) => ({
...
+        const removedPerType = await Promise.all(
+            Object.entries(OTP_EXPIRATION_MS).map(async ([type, lifetimeMs]) => {
+                const result = await repo()
+                    .createQueryBuilder()
+                    .delete()
+                    .where('type = :type AND updated < :cutoff', {
+                        type,
+                        cutoff: dayjs().subtract(lifetimeMs, 'millisecond').toISOString(),
+                    })
+                    .execute()
```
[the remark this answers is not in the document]
MrChaker: there won't be much rows . should be fine

MrChaker (changes requested):

packages/server/api/src/app/authentication/otp/otp-service.ts
```diff
@@ -105,7 +113,35 @@ async function countAttempt(otpId: string): Promise<void> {
 
 async function discard({ otp, identityId, type, log }: DiscardParams): Promise<void> {
     await repo().delete({ id: otp.id })
-    log.warn({ identityId, type }, '[otpService#confirm] attempt budget exhausted, credential discarded')
+    await forgetCachedCode({ identityId, type })
+    log.warn({ identityId, type }, '[otpService#confirm] credential discarded')
+}
+
+function digestsMatch(stored: string, candidate: string): boolean {
```
MrChaker: @AhmadTash move this function to encyption utils

abuaboud (approved):

MrChaker (approved):

PR #14888, opened 2026-08-18
feat(flow-runs): attribute a child run to the barrier that awaits it

othmanemad:
## Description

**Stack: 5 of 5.** Base: [#14778](https://github.com/activepieces/activepieces/pull/14778).

PRs 1–4 built a barrier that releases on committed signal rows, and a human producer that writes them. This PR adds the second producer: **a child run's own outcome**.

### Attribution

`flow_run` gains two columns. `parentWaitpointId` says which barrier counts this run; `dispatchIndex` is its ordinal within that barrier. When the runs-metadata worker persists a terminal child, it maps the run's status onto a `BarrierSignalStatus` and calls `barrierService.receive({ refId: runId })` — so the barrier never polls its children, and the ordering that makes this safe (row committed, then evaluation enqueued) is the same one decision 000015 argues for the human producer.

**This PR carries only the consumer side of attribution** — the two columns, the exclusions, the retry guard, and the terminal `receive` call. Nothing in it creates an attributed child. The writer is `flowRunService.dispatchChild`, which lands in [#14890](https://github.com/activepieces/activepieces/pull/14890) and sets `parentWaitpointId` and `dispatchIndex` in one insert, then claims the child's signal (`barrierService.claimSignal`) so `receive({ refId })` has something to match. Until that layer, the `receive` call here is inert by construction rather than broken.

An earlier revision let a caller attribute a run over the webhook path with `ap-parent-waitpoint-id` / `ap-dispatch-key`. **That was removed in 6940a00251, not hardened.** The header was only format-checked on a `securityAccess.public()` route, while `parentWaitpointId IS NULL` is exactly what the runs list, `countByStatus`, the bulk-retry filter, platform analytics and the billing usage report key on — so any caller holding a webhook URL could run flows that executed normally but appeared on no surface and in no metered count. It could never have done its job either: no signal row bound a webhook child's run id, so the terminal `receive` matched nothing and the parent waited out its deadline. The reader was a fossil of the deleted `Fan Out to Subflows` piece action, which ran in the engine sandbox and had no channel into run creation but the public webhook. A caller-facing producer, if one is ever wanted, has the per-signal confirm URL handed out at barrier creation — it completes a signal the server pre-allocated and creates no run at all.

**`parentWaitpointId` also had to go into `RUNS_METADATA_UPSERT_KEYS`.** That list is an allowlist — an unlisted key is dropped silently between the worker and the row, with no error. Same failure class as forgetting `getEntities()`.

### Children are unlisted, not invisible

Of the eight run list/count surfaces, **five exclude children and three deliberately keep them**: user-facing and billing surfaces exclude, ops and load surfaces do not. Excluding — `flowRunService.list` (one edit covering the runs endpoint, the builder list, the agent tool and the MCP tool), `countByStatus`, `filterFlowRunsAndApplyFilters`, the billing daily-execution count, and `platformAnalyticsReportService.listRuns`. Still counting — `healthMetricsService`, the telemetry daily rollup, and the admin platform retry.

Two of the five matter beyond consistency. The shared filter-and-apply helper keeps a filter-driven **bulk** retry from dragging a whole fan-out into the refusal path — it is not the lock on that door, since `bulkRetry` calls `retry()` per run and `retry()` already throws on a child, but without it a filter matching a 20k fan-out returns 20k error rows, because children share the parent's `flowId`. (An earlier revision of this description called it the *only* thing stopping that bulk retry. That was wrong — corrected in the review round below.) And the analytics rollup multiplies a user-set per-trigger constant, so counting children makes a flow that batches ten items claim eleven times the time saved.

Reaching a child by id already works. `ListFlowRunsRequestQuery.parentWaitpointId` is the opt-back-in — a filter on the existing endpoint, not a new one — ordered by `dispatchIndex`, and it **also requires `dispatchIndex IS NOT NULL`**. That is the cursor, not taste: `Paginator` slices its composite cursor at the first field missing from the decoded cursor object, so a page ending on a NULL-index child encodes no `dispatchIndex`, emits no cursor predicate at all, and returns page 1 forever. Any ordering on a nullable column has this failure mode. `archivedAt` is a separate user-driven axis and is not reused for hiding.

### Retry refuses a run that started mid-graph

**`parentWaitpointId IS NOT NULL` is the one test for "this run is a barrier child"; `dispatchIndex` is ordering data, never an identity test.** The two are written together by the single dispatcher and are set together or not at all. `retry()` refuses a run carrying *either* column — belt and braces at the entry point rather than a second definition, because gating only one strategy's update would leave the other open. Retrying a child is meaningless in both strategies: `FROM_FAILED_STEP` resumes from the trigger, so it would re-run the *parent's* whole prefix inside the child, and `ON_LATEST_VERSION` starts a fresh top-level run no barrier is counting.

A retry also clears the run's waitpoints first, so a re-executed step cannot inherit the previous attempt's pause, and `ON_LATEST_VERSION` copies `parentRunId` but **never** `parentWaitpointId` — the barrier already counted the original child, and attributing the retry to it would inflate the terminal count and risk a premature release.

### Two read-path costs the new column exposes

Both are on the runs list this PR changes, so both are fixed here rather than deferred.

**The list indexes.** Hiding children appends `parentWaitpointId IS NULL` to every default list, and none of the `created`-leading indexes carried that column — so the exclusion was a post-heap Filter, and every child newer than the page's oldest visible parent was fetched from the heap and discarded to return ten rows, on a table the UI re-polls every 15s. The three list-only indexes become **partial on `WHERE "parentWaitpointId" IS NULL`**: children leave them entirely, so cost is proportional to page size instead of fan-out size and a fan-out writes zero entries into them. `idx_run_project_id_environment_created_archived_at` is **dropped rather than replaced** — it is an exact prefix of the `created`/`status` cover index and, being narrower, the planner preferred it, which would have made the change a no-op. Net one fewer index on `flow_run` than before.

**The `status`-leading index deliberately stays full**, keeping `parentWaitpointId` as a trailing column. It is the only index left that can serve the seven `healthMetricsService` aggregates, which count children by design and so carry no such predicate; making all four partial drops those to a whole-table scan. Adding the predicate to `healthMetricsService` instead was tried and reverted — it silently reclassifies an ops surface as a user-facing one.

All five creates run before all five drops, under new names, so the hottest table in the product is never left without a list index for the length of a concurrent build.

Measured on 1.02M rows / 200 projects with a 20k-child fan-out as the newest activity: no-filter list **4.49ms → 0.11ms**, flow+status **6.22ms → 0.79ms**, `countByStatus` **20.45ms → 0.89ms**, billing daily count **10.71ms → 2.87ms** (the last two now index-only, zero heap fetches). Trailing the column instead of a partial predicate was measured too and is the weaker fix — it reaches the Index Cond but still walks every child's leaf pages, buying 2-3× where partial buys 40×.

**The `flow_version` join.** `queryBuilderForFlowRun` is the single builder behind `list`, `getOne`, `getOnePopulatedOrThrow` and `findFlowRunOrThrow`, and the only field any caller reads off the relation is `displayName`. It used `leftJoinAndSelect`, selecting all thirteen `flow_version` columns including three `jsonb` ones — `trigger` (the whole flow graph, routinely TOASTed), `backupFiles`, `notes`. The `addSelect` beside it was a raw SQL string TypeORM mapped to nothing. `Paginator.paginate` then uses `getRawAndEntities` and `mergeAdditionalColumns` copies every non-root-prefixed raw key onto the entity, so each row carried the graph twice, and `convertDatesToStrings` deep-walked both copies before zod stripped them — the wire payload stayed clean while the DB read, the heap and the per-request CPU did not.

Now `leftJoin` + `addSelect(['flowVersion.id', 'flowVersion.displayName'])`. **The PK is load-bearing, not decoration**: `RawSqlResultsToEntityTransformer.group` keys joined rows by the alias's primary columns, so a partial relation select without the id silently fails to hydrate. Measured on 1M runs / 10k flow versions with a ~6.4 KB TOASTed trigger, fully warm: a 50-row page **1.59 MB → 15 KB** returned and **4.63ms → 0.56ms**; the single-row read behind `findFlowRunOrThrow` 0.63ms → 0.30ms. `EXPLAIN (ANALYZE, BUFFERS)` **cannot** measure this and reports almost no difference — it discards output rows, so it never detoasts the wide columns.

One outbound payload shrinks as a side effect: `FLOW_RUN_RETRIED` and `FLOW_RUN_RESUMED` are fed from `findFlowRunOrThrow`, and `auditLogRepo().save` persists the event with no zod parse while `event-destinations` spreads `event.data.flowRun` into its webhook — so both had been carrying the entire flow definition under a `flowVersion` key `FlowRunEventData` never declared. They now carry `id` and `displayName`.

### Job priority

An attributed child gets `low` priority instead of `medium`/`high`. A fan-out is by definition many jobs arriving at once; without this a single wide barrier is a noisy neighbour to every interactive run on the shared queue.

### Size

> **These counts are stale and the diff above is not this PR's own.** `feat/barrier-edges` was force-pushed after this branch last merged it, so the merge base fell back to a `main` commit and the PR now renders the whole stack (64 files) instead of this layer. This branch needs a rebase onto the new base before the numbers mean anything.

As last measured against the correct base: `server/api` **683** against 600, hence `large-pr-ok` — but **only 206 of that is source**; the rest is four test files (`barrier-children-visibility`, `daily-executions`, `list-flow-runs`, `retry-flow-run`, plus mocks). Engine+worker+core-execution bucket **62/300**. The review round below adds ~24 lines of source and ~90 of tests.

## How was this tested?

- **`barrier-children-visibility.test.ts` (new)** — the five excluding surfaces and the three that keep counting, asserted per surface, plus the `?parentWaitpointId=` opt-in ordering by `dispatchIndex` and the mid-graph retry refusal.
- **`daily-executions.test.ts` (new, EE)** — the billing daily count excludes children. **This file never ran until 469e3ce3af.** It imported `billingUsageReportService` from `ee/billing-usage-report`, which `main` renamed to `licenseKeyUsageReportService` under `ee/license-key-usage-report`, and mocked `captureBillingEvent`/`flushBillingEvents`, since renamed to `captureLicenseKeyEvent`/`flushLicenseKeyPostHogEvents`. The module failed to resolve, so the assertion had never executed. Import and mocks fixed; it passes under `AP_EDITION=ee` now.
- **`list-flow-runs.test.ts`** — the join guard, asserted on the **service** result rather than the HTTP response: zod strips the heavy keys at serialization, so an HTTP-only assertion passes against the wide query too and proves nothing.
- **`retry-flow-run.test.ts`** — `parentRunId` survives an `ON_LATEST_VERSION` retry of a subflow run; both strategies refuse a barrier child and leave it FAILED. The earlier fixture set `parentWaitpointId` with no `dispatchIndex` and expected success; under the both-or-neither invariant that row cannot exist, so the case now asserts refusal.
- Full CE `flows/flow-run` + `webhooks` integration run passes, including `waitpoint.test.ts`, `barrier.test.ts` and `execute-flow-e2e.test.ts` — this PR touches the shared run-creation and retry paths, so those passing unchanged is the real check.
- `check-migrations` clean, plus a manual `up`/`down`/`up` round-trip on the new migration and `check-migration-rollback` ✅. `npm run lint-dev` 0 errors. `core-execution` unit tests 126 pass, including the new `job-priority.test.ts`.
- CE and EE paths both exercised. No UI in this PR, so no appearance/white-labelling surface.

## Review round (@amrdb)

Five comments. Four fixed across three commits, one declined with reasons.

- **`dispatchIndex` silently ignored without `parentWaitpointId`** (7707b30cdc) — the service gated the filter on `parentWaitpointId`, so `?dispatchIndex=5` alone returned the project's parent runs with a 200. Refused at the DTO now. Moving the filter out of the `if` instead would have returned zero rows, which is a different wrong answer.
- **Bulk cancel silently dropped explicitly-named child ids** (7707b30cdc) — `filterFlowRunsAndApplyFilters` backs `cancel` and `bulkArchive` as well as `bulkRetry`, so its unconditional `parentWaitpointId IS NULL` made `cancel({ flowRunIds: [childId] })` answer 200 and do nothing. The filter now applies only when the caller names no `flowRunIds`: a filter never selects a child, an explicit id is taken literally. Cancelling a child was already safe end to end — `cancelSingleRun` hands `CANCELED` to `runsMetadataQueue`, which is terminal, so the barrier gets its signal.
- **`fanIn: { barrierId }` broke the evlog field schema** (474cb48c87) — rule 1 says an entity's own id is `id` inside its group, so every correlation query keying on `waitpoint.id` missed all three sites. Renamed to `waitpoint: { id }`, which `barrierService.create` already used a few lines away. A fresh `barrier` group would have been a second path for one concept. #14890 carries two more sites.
- **`FROM_FAILED_STEP` left `parentWaitpointId` on the retried row** (7707b30cdc) — unreachable today, but this PR's own brain bullet mandated threading `parentWaitpointId` through `createQuotaExceededRun` → `runsMetadataQueue` without `dispatchIndex`, which would have built exactly that row. Guarded at the entry point and the bullet amended to require both columns. The sentence claiming attribution can legitimately be dropped is gone: nothing in the tree drops it, and it was the only justification for having two discriminators.
- **`receiveSignal` fires on every terminal upsert, not only on the transition** — *declined.* Decision 000015 records last-write-wins deliberately. `closeBarrier` deletes every signal row in the transaction that completes the barrier, so a post-release rewrite is impossible. And a compare-and-set breaks redelivery: `receiveSignal` writes and *then* enqueues, so a retry after a failed enqueue would find the signal decided, skip the enqueue, and leave the barrier to time out. Full reasoning on the thread.

Both new columns are nullable with no default, both new query params are optional, and no env var is added. The five renamed `flow_run` indexes are an internal schema detail with a working `down()`. `FLOW_RUN_RETRIED`/`FLOW_RUN_RESUMED` stop carrying a `flowVersion` key that `FlowRunEventData` never declared and no consumer could rely on by contract — called out here rather than hidden, but not a documented field being removed.

No new route and no change to any `securityAccess`. The webhook-header attribution an earlier revision added is **gone** as of 6940a00251 — it stamped `parentWaitpointId` from a format-checked header on a `securityAccess.public()` route, which let a caller hide its own run from every user-facing and metered surface. No caller-facing path can set either new column now; the only writer is the server-side dispatcher in #14890. Every new `flow_run` query stays project-scoped, and the child-listing filter runs inside the existing project scope.


packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -753,6 +781,7 @@ async function queueOrCreateInstantly(params: CreateParams, log: FastifyBaseLogg
         flowVersionId: params.flowVersionId,
         environment: params.environment,
         parentRunId: params.parentRunId,
+        parentWaitpointId: params.parentWaitpointId,
         failParentOnFailure: params.failParentOnFailure ?? true,
```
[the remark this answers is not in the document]
othmanemad: Fixed — the writer this describes no longer exists on this branch.

`6940a00251` (the tip, pushed after your review) **deletes** the `ap-parent-waitpoint-id` / `ap-dispatch-key` webhook readers rather than threading a `dispatchIndex` through them. Removing was the right call independent of this finding: the header was only format-checked on a `securityAccess.public()` route, and `parentWaitpointId IS NULL` is exactly what the runs list, `countByStatus`, the bulk-retry filter, platform analytics and the billing usage report key on — so any caller holding a webhook URL could run flows that appeared on no surface and in no metered count. It could not have worked anyway: no signal row bound a webhook child's run id.

No path in this PR sets `parentWaitpointId` at run creation now — `git grep parentWaitpointId` on the head returns only readers (the list filter, the job-data pass-through, `RUNS_METADATA_UPSERT_KEYS`, log enrichment). The single writer is `flowRunService.dispatchChild` in #14890, which sets `parentWaitpointId` and `dispatchIndex` in the same insert; #14890 also adds a unique `(parentWaitpointId, dispatchIndex)` index as a DB-level backstop, so the half-attributed state you describe is unrepresentable from that layer on.

The PR description has been corrected — it still advertised the two headers when you reviewed, which is what this finding was reading against.

packages/server/api/src/app/flows/flow-run/flow-runs-queue.ts
```diff
@@ -100,7 +102,15 @@ export const runsMetadataQueue = (log: FastifyBaseLogger) => ({
 
                             const parentRunId = savedFlowRun.parentRunId
                             const shouldMarkParentAsFailed = savedFlowRun.failParentOnFailure && !isNil(parentRunId) && ![FlowRunStatus.SUCCEEDED, FlowRunStatus.RUNNING, FlowRunStatus.PAUSED, FlowRunStatus.QUEUED].includes(savedFlowRun.status)
-                            if (shouldMarkParentAsFailed) {
+                            if (!isNil(savedFlowRun.parentWaitpointId) && isFlowRunStateTerminal({ status: savedFlowRun.status, ignoreInternalError: false })) {
+                                await barrierService(log).receive({
+                                    refId: savedFlowRun.id,
+                                    projectId: savedFlowRun.projectId,
+                                    status: toSignalStatus(savedFlowRun.status),
+                                    result: { runId: savedFlowRun.id, status: savedFlowRun.status },
```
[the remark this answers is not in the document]
othmanemad: Accurate for this PR read in isolation, and deliberate: this layer adds the **consumer**, #14890 adds the **producer**.

`barrierService.claimSignal` is what binds a signal's `refId` to a child run id, compare-and-set inside the dispatcher's transaction (`waitpoints/barrier-service.ts` + `waitpoints/fan-out-dispatcher-job.ts`, both new in #14890). It does not exist on this branch by design — the stack introduces the terminal-child hook before the thing that dispatches children.

Nothing in this PR can create an attributed child at all: after `6940a00251` removed the webhook headers, `parentWaitpointId` has no writer here (see the other thread). So `receive({ refId: runId })` on this branch is unreachable rather than failing, and the null lookup scheduling no evaluation is the intended no-op — no crash, no wrong state, no barrier to leave paused. Merged as a stack, the producer is always present before any child exists.

Worth re-raising against #14890 if you want the binding itself verified; that is where it is testable.

othmanemad:
@greptile review

Both P1s are answered in their threads and the description has been corrected — it was advertising the webhook attribution headers that the tip commit `6940a00251` deletes. Please re-review against the current head.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -61,7 +66,13 @@ export const flowRunService = (log: FastifyBaseLogger) => ({
         if (!isNil(params.environment)) {
             whereClause.environment = params.environment
         }
-        let query = queryBuilderForFlowRun(flowRunRepo()).where(whereClause)
+        let query = queryBuilderForFlowRun(flowRunRepo())
+            .where(whereClause)
+            .andWhere({ parentWaitpointId: params.parentWaitpointId ?? IsNull() })
+
+        if (!isNil(params.parentWaitpointId)) {
+            query = query.andWhere({ dispatchIndex: params.dispatchIndex ?? Not(IsNull()) })
```
amrdb, line 74: **`dispatchIndex` is silently ignored when `parentWaitpointId` isn't also passed.** The DTO accepts them independently (`list-flow-runs-request.ts`), the controller threads both through, but here the `dispatchIndex` filter is gated on `parentWaitpointId` being set. `GET /v1/flow-runs?projectId=X&dispatchIndex=5` returns the project's parent runs (all with `dispatchIndex NULL`) with no error — the caller thinks `dispatchIndex=5` matched nothing.

Fix either at the boundary (require both together in the zod schema, refuse `dispatchIndex` alone) or by moving the filter out of the `if` so it applies unconditionally when the caller passes it.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -553,6 +574,7 @@ async function filterFlowRunsAndApplyFilters(
     let query = flowRunRepo().createQueryBuilder('flow_run').where({
         projectId: params.projectId,
         environment: RunEnvironment.PRODUCTION,
+        parentWaitpointId: IsNull(),
```
amrdb, line 577: **Bulk cancel silently drops explicitly-passed child ids.** This `parentWaitpointId: IsNull()` on the shared filter helper backs both `bulkRetry` (correct — a child cannot be meaningfully retried) *and* `flowRunService.cancel` (line ~230), which is the wrong shape for cancel.

`cancel({ flowRunIds: [childId] })` silently drops the child, returns 200 OK, and the child continues running while the barrier waits. Independently canceling a stuck barrier child is a legitimate operator need (unlike retrying one).

Options: exempt the `flowRunIds`-provided path from the child filter, split the helper into a cancel-flavored and retry-flavored variant, or throw/report explicit rejection when a passed id resolves to a child.

amrdb (commented):
Five findings from a second-pass review. Two are user-visible correctness holes; two are latent under this PR's shipped writer surface but re-open with any future producer; one is an evlog schema nit worth taking while there's only one call site.

packages/server/worker/src/lib/worker.ts
```diff
@@ -382,6 +382,7 @@ async function executeJob(apiClient: WorkerToApiContract, job: ConsumeJobRequest
         ...spreadIfDefined('run', isAgentJob && 'runId' in jobData ? { id: jobData.runId } : undefined),
         ...spreadIfDefined('conversation', 'conversationId' in jobData ? { id: jobData.conversationId } : undefined),
         ...spreadIfDefined('flowVersion', 'flowVersionId' in jobData ? { id: jobData.flowVersionId } : undefined),
+        ...spreadIfDefined('fanIn', 'parentWaitpointId' in jobData && jobData.parentWaitpointId != null ? { barrierId: jobData.parentWaitpointId } : undefined),
```
amrdb, line 385: **`fanIn: { barrierId: ... }` violates the evlog grouping in `packages/server/CLAUDE.md`.** Rule 2: an entity's own id is `id` inside its group — never `<entity>Id`. Canonical spelling is `waitpoint: { id: jobData.parentWaitpointId }` (a barrier IS a waitpoint), or a fresh `barrier: { id: ... }` group if you want to reserve `waitpoint` for the run's current waitpoint.

The same misspelling already exists in `barrier-service.ts` on the base branch, so this PR broadens the exposure — worth fixing both call sites in one commit while it's still trivial. Correlation queries in the OTLP drain keying on `waitpoint.id` miss every fan-in event today.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -149,6 +168,7 @@ export const flowRunService = (log: FastifyBaseLogger) => ({
                     ? await resolveStepOutput({ step: triggerStep, flowRun: oldFlowRun, log })
                     : undefined
 
+                await waitpointService(log).deleteByFlowRunId({ flowRunId: oldFlowRun.id, projectId: oldFlowRun.projectId })
```
amrdb, line 171: **Latent: FROM_FAILED_STEP doesn't clear `parentWaitpointId` on the retried row, so the retry's terminal hook re-signals the barrier.** The update below sets `status → QUEUED / startTime / finishTime` and leaves `parentWaitpointId` in place. When the re-run finishes, `runsMetadataQueue` sees `parentWaitpointId != null + terminal` and calls `receiveSignal` again — potentially overwriting an already-committed signal status (SUCCEEDED → FAILED etc.), which can flip the release outcome under strict predicates.

Unreachable today: the only writer (#14890's dispatcher) sets `dispatchIndex` too, so the guard at line 133 blocks it. But the invariant is fragile — the design bullet in `brain/knowledge/flows-execution/flow-runs.md` explicitly allows attribution to be dropped while `dispatchIndex` is retained, and any future producer that writes `parentWaitpointId` without `dispatchIndex` re-opens this. Cheapest defense: clear `parentWaitpointId` inside the FROM_FAILED_STEP update, or gate the retry on `parentWaitpointId` too.

packages/server/api/src/app/flows/flow-run/flow-runs-queue.ts
```diff
@@ -100,7 +102,15 @@ export const runsMetadataQueue = (log: FastifyBaseLogger) => ({
 
                             const parentRunId = savedFlowRun.parentRunId
                             const shouldMarkParentAsFailed = savedFlowRun.failParentOnFailure && !isNil(parentRunId) && ![FlowRunStatus.SUCCEEDED, FlowRunStatus.RUNNING, FlowRunStatus.PAUSED, FlowRunStatus.QUEUED].includes(savedFlowRun.status)
-                            if (shouldMarkParentAsFailed) {
+                            if (!isNil(savedFlowRun.parentWaitpointId) && isFlowRunStateTerminal({ status: savedFlowRun.status, ignoreInternalError: false })) {
```
amrdb, line 105: **Latent: `receiveSignal` fires on every terminal-status upsert, not only on the first non-terminal → terminal transition.** A BullMQ job retry, or any follow-up metadata upsert on an already-terminal child (e.g., a late INTERNAL_ERROR write overriding an earlier SUCCEEDED), re-enters this branch. Duplicate `enqueueEvaluation` is bounded and manageable; a status-changing follow-up silently rewrites the barrier signal.

Gate on the transition: read the previous status off `existingFlowRun` (already fetched above), only call `receiveSignal` when the *previous* status was non-terminal. Same pattern as `shouldMarkParentAsFailed`, which correctly implies transitional intent by its column set.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
othmanemad: Fixed at the boundary in 7707b30cdc.

Diagnosis confirmed — the filter sat inside `if (!isNil(params.parentWaitpointId))`, so `?dispatchIndex=5` alone fell through to the default `parentWaitpointId IS NULL` branch and returned the project's parent runs with a 200.

Took the first of your two options. `ListFlowRunsRequestQuery` now carries a `.refine()` requiring `parentWaitpointId` whenever `dispatchIndex` is present, so the combination is a 400 rather than a wrong answer.

Worth saying why not the second: moving the filter out of the `if` would trade one wrong answer for another. The default list also applies `parentWaitpointId IS NULL`, so an unconditional `dispatchIndex` filter contradicts it and returns zero rows — the caller still learns nothing. A dispatch index is only meaningful inside the barrier that issued it, so requiring the pair is the honest contract.

The param itself stays: `barrier-children-visibility.test.ts` uses it to resolve batch N of a wide fan-out without paging to it.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
othmanemad: Fixed in 7707b30cdc, with your first option: the child filter now applies only when the caller passes no `flowRunIds`. A filter never selects a child; an explicit id is taken literally.

Your read was right and the fix turned out cheaper than the comment assumes, because the helper was never load-bearing for retry safety. `bulkRetry` calls `this.retry()` per run, and `retry()` already throws on a child — so the guard inside `retry()` is the real lock, and the helper's predicate was a redundant second one that only broke cancel. **The PR description claimed the opposite** ("the *only* thing stopping a filter-driven bulk retry"); that claim was wrong and I have corrected it.

Two things that back the cancel side being safe end to end:

- `cancelSingleRun` hands `CANCELED` to `runsMetadataQueue`, and `CANCELED` is terminal, so cancelling a child signals its barrier through the normal path rather than stranding it.
- Filter-driven cancel loses nothing by not selecting children directly. `dispatchChild` sets `parentRunId`, and `getAllChildRuns` recurses on it, so cancelling a parent already reaches its children.

Keeping the filter for the filter-driven path also keeps the bulk-retry response readable: without it, a filter matching a 20k fan-out returns 20k error rows, since children share the parent's `flowId`.

Covered by three new cases in `barrier-children-visibility.test.ts`: cancel by child id succeeds, filter-driven cancel leaves children alone, and a bulk retry naming a child by id returns one validation error instead of silence.

packages/server/worker/src/lib/worker.ts
othmanemad: Fixed in 474cb48c87, all three sites in one commit as you suggested — `worker.ts`, `barrier-service.ts` and `barrier-queue.ts`.

Took your first option, `waitpoint: { id }`, not the fresh `barrier` group. Three reasons:

- `barrierService.create` already logs a barrier as `waitpoint: { id: existing.id }`, a few lines above the `wideEvent.set` in the same file. A new group would have contradicted its own module.
- `waitpoint: { id }` is in the canonical-groups list in `packages/server/CLAUDE.md`; `barrier` is not. Adding one would be the second path for one concept that rule 1 exists to prevent.
- There is nothing to reserve `waitpoint` against. The `EXECUTE_FLOW` worker path sets no `waitpoint` group, so the parent barrier id cannot collide with a run's own pause there.

Net effect is the correlation query you wanted: a single `waitpoint.id` now traces a barrier from creation through evaluation and child job execution to release.

One correction for the record — it is rule **1** in that file, not rule 2. Rule 2 is the attributes-beside-id rule, which is why the release event became `waitpoint: { id, signalCount, releaseReason, stillRunning }` rather than keeping the extras elsewhere.

#14890 adds two more `fanIn: { barrierId }` sites in `dispatchChild`; they will be renamed on its next rebase.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
othmanemad: Fixed in 7707b30cdc, with the second of your two defences rather than the first: `retry()` now refuses a run carrying **either** `parentWaitpointId` or `dispatchIndex`.

I went that way because clearing the column inside the `FROM_FAILED_STEP` update would close only one of the two doors. `ON_LATEST_VERSION` reads the same run and starts a fresh top-level run that no barrier is counting, so a half-attributed row would still be retryable there. Gating the entry point covers both strategies and leaves nothing to clear.

Your read of the fragility was the more useful half of the comment, and it was worse than latent. This PR's own brain bullet mandates threading `parentWaitpointId` through `createQuotaExceededRun` → `runsMetadataQueue`, and `dispatchIndex` is not in `RUNS_METADATA_UPSERT_KEYS` — so that path, as documented, would have produced exactly the `parentWaitpointId` without `dispatchIndex` row you describe. No branch in the stack implements it yet. I amended the bullet to require both columns and to say the allowlist needs `dispatchIndex` first.

I also deleted the sentence that said attribution can legitimately be dropped. Nothing in the tree drops it — `git grep` finds no writer setting `parentWaitpointId` back to NULL, and `release`/`closeBarrier` never touch `flow_run`. That sentence was the only justification for having two discriminators, so it went with the split. `parentWaitpointId IS NOT NULL` is now the single identity test; `dispatchIndex` is ordering data. The retry guard reading both is belt and braces, not a second definition.

`retry-flow-run.test.ts` had to change: its fixture set `parentWaitpointId` with no `dispatchIndex` and expected the retry to succeed. Under the invariant that row cannot exist, so it now asserts refusal on both strategies, with a separate case keeping the `parentRunId` assertion for a plain subflow run.

packages/server/api/src/app/flows/flow-run/flow-runs-queue.ts
othmanemad: Not changing this one — the overwrite is deliberate, and I am fairly sure gating it would introduce a worse bug. I did build the fix before concluding that, so this is a report rather than a dismissal.

Three things, in the order they changed my mind.

**It is an accepted decision, not an oversight.** `brain/knowledge/decisions/000015` says so in as many words: "Receiving is an upsert, last write wins. It is not a state machine and does not reject a second write … the barrier must reflect the latest truth, not the first. That also makes redelivery of the receive path free."

**The post-release rewrite cannot happen.** `closeBarrier` deletes every signal row for the barrier inside the same transaction that flips it to `COMPLETED`. After a release there are no signals left, so `receiveSignal` finds nothing and returns null. The summary and the rows cannot disagree.

**Gating it breaks redelivery.** I implemented the compare-and-set — `UPDATE … WHERE status = PENDING`, return null when nothing was written — and then found the hole. `receiveSignal` writes the signal and *then* enqueues the evaluation. If the enqueue throws, the metadata job fails and retries; on the retry the CAS sees a decided signal, returns early, and never enqueues. The barrier then waits out its `timeoutAt` holding a run that was ready to resume — the exact failure 000015's coalescing bullet is written to prevent. Unconditional re-write plus re-enqueue is what makes that path self-healing.

That leaves your two concrete cases. A BullMQ redelivery writes the *same* status, so it is a no-op plus a deduplicated evaluation. A genuinely later, different terminal status is the "latest truth" case 000015 chose on purpose.

One premise of that decision has weakened and is worth noting: it justifies itself partly with "a retried child reports its terminal status again", and this PR now refuses to retry a child at all. If someone wants to revisit last-write-wins on those grounds it should be a decision doc superseding 000015, with the enqueue moved out from behind the guard — not a gate added at this call site, which would leave `resume-controller`'s own read-then-write race open anyway.

For the record, `resolveOpenSignal` in `resume-controller` already refuses a non-`PENDING` signal before calling in, so the human producer never depends on the overwrite either way.

PR #14894, opened 2026-08-18
feat(builder): render a batch step's body as a shaded region on the canvas

othmanemad:
## Description

**Stack: 9 of 12.** Base: [#14893](https://github.com/activepieces/activepieces/pull/14893).

First of two builder PRs. This one is **canvas rendering only** — how a `PROCESS_IN_BATCHES` step and its body draw. The settings form and the step-picker entry are the next PR; nothing here is reachable from the product yet.

### The region is a node, not a decoration

A batch step reads like a loop on the canvas, so it lays out like one — the container geometry was already generalised into `flowCanvasUtils.isContainerStep` in #14892, and this is the builder half. What is new is the shaded rectangle behind the body, which marks "everything in here runs in a child run".

It is a real React Flow node (`ApNodeType.BATCH_REGION`), sized from the body's own bounding box, `selectable: false` and `draggable: false` so it never intercepts a click or a drag meant for a step inside it. `batch-region.ts` computes its geometry from the laid-out graph rather than re-walking the flow, so the region cannot drift from the nodes it is drawn behind.

It is deliberately **not** in `AP_NODE_SIZE`: every other node type has a fixed size, and this one's size is derived per-instance, so it is excluded from that `Record` rather than given a meaningless default.

### INSIDE_BATCH on the add and paste paths

`ApLoopStartEdge` now carries `stepLocationRelativeToParent` in its edge data, and `loop-start-edge.tsx` reads it instead of hardcoding `INSIDE_LOOP`. One edge component now serves both containers, which is why the diff there is a deletion. The context menu gains "Paste Inside Batch" beside "Paste Inside Loop".

### Two canvas changes the region forced

Both are in `flow-canvas/index.tsx` and both are fixes, not refactors:

- **The graph key no longer includes the selected step.** It was rebuilding the entire graph on every selection change. The region's size comes from layout, not selection, so nothing needed it — and the rebuild was wasted work on a hot path before this PR too.
- **The translate extent is computed from node positions, not `getNodesBounds`.** The region node is as large as the whole body it wraps, so `getNodesBounds` counted it and inflated the pannable area by the region's own size — the canvas would let you pan a full body-height into empty space.

## How was this tested?

- `packages/web/test/app/builder/flow-canvas/utils/batch-region.test.ts` — **new, 459 lines, 12 cases.** Region geometry for an empty body, a single step, a nested container, and a body taller than the step; the region's position relative to the batch node in both orientations; and that it is excluded from selection and drag.
- `packages/web/test/app/builder/flow-canvas/utils/flow-canvas-utils.test.ts` — **+130.** Graph construction with a batch container: the region node's presence and ordering, the loop-start edge's `stepLocationRelativeToParent`, and the graph key no longer varying with the selected step.
- `npx vitest run test/app/builder/flow-canvas` — **30 passing, 5 files.**
- **Editions:** CE / EE / Cloud identical — canvas rendering, no gate, no branded surface.

### A note on `flowCanvasUtils.getStepStatus`

The branch this stack is split from deletes `getStepStatus` here, because it rewrites `step-node-status-in-run.tsx` onto the batch-aware `useStepOutputInRun` hook. That hook belongs to the run-detail layer, so **this PR keeps `getStepStatus`** and the run-detail PR removes it along with its last caller. It is not an oversight.

Fixes # (issue)


packages/web/src/app/builder/flow-canvas/index.tsx
```diff
@@ -325,5 +312,5 @@ const createGraphKey = (
   const notesGraphKey = notes
     .map((note) => `${note.id}-${note.position.x}-${note.position.y}`)
     .join('-');
-  return `${flowVersion.id}-${flowGraphKey}-${notesGraphKey}-${selectedStep}`;
+  return `${flowVersion.id}-${flowGraphKey}-${notesGraphKey}`;
```
AbdulTheActivePiecer, line 315: why was this removed ?

packages/web/src/app/builder/flow-canvas/context-menu/canvas-context-menu-content.tsx
```diff
@@ -1,7 +1,9 @@
+import { isNil } from '@activepieces/core-utils';
 import {
   FlowAction,
   FlowActionType,
   FlowOperationType,
+  flowCanvasUtils as sharedFlowCanvasUtils,
```
AbdulTheActivePiecer, line 6: please keep this named flowCanvasUtils

packages/web/src/app/builder/flow-canvas/context-menu/canvas-context-menu-content.tsx
```diff
@@ -92,11 +94,14 @@ export const CanvasContextMenuContent = ({
   );
   const showPasteAfterLastStep =
     !readonly && contextMenuType === ContextMenuType.CANVAS;
-  const showPasteAsFirstLoopAction =
+  const showPasteInsideContainer =
     selectedNodes.length === 1 &&
-    firstSelectedStep?.type === FlowActionType.LOOP_ON_ITEMS &&
+    !isNil(firstSelectedStep) &&
+    sharedFlowCanvasUtils.isContainerStep(firstSelectedStep) &&
```
AbdulTheActivePiecer, line 100: please rename isContainerStep to isLoopOrBatchStep, I had no idea what containerStep is,

packages/web/src/app/builder/flow-canvas/utils/batch-region.ts
```diff
@@ -0,0 +1,168 @@
...
+}: {
+  graph: ApGraph;
+  batchName: string;
+  childNames: string[];
+  orientation: CanvasOrientation;
+}): ApBatchRegionNode | null => {
+  const layout = flowCanvasLayoutConsts.ORIENTATION_LAYOUT[orientation];
+  const batchNode = graph.nodes.find((node) => node.id === batchName);
+  const endNode = graph.nodes.find(
+    (node) => node.id === `${batchName}-batch-subgraph-end`,
```
AbdulTheActivePiecer, line 57: please create a helper function called batchSubgraphEndId that takes the batch name and returns the id and use it as well in the place where you create those subgraph ends.

packages/web/src/app/builder/flow-canvas/utils/batch-region.ts
```diff
@@ -0,0 +1,168 @@
...
+  );
+  if (isNil(batchNode) || isNil(endNode)) {
+    return null;
+  }
+  const prefixes = [...childNames, `${batchName}-big-add-button`];
```
AbdulTheActivePiecer, line 62: same note here as before https://github.com/activepieces/activepieces/pull/14894/changes#r3841361659.
the reason being is to unify the place you have to edit when needed.

packages/web/src/app/builder/flow-canvas/utils/batch-region.ts
```diff
@@ -0,0 +1,168 @@
...
+const toLayoutSpace = ({
+  position,
+  orientation,
+}: {
+  position: ScreenPoint;
+  orientation: CanvasOrientation;
+}): LayoutPoint =>
+  orientation === 'vertical'
+    ? { along: position.y, cross: position.x }
+    : { along: position.x, cross: position.y };
```
AbdulTheActivePiecer, line 147: I think we use the same logic in other places might have to move this to flowCanvasUtils and call it in this and other files as well

packages/web/src/app/builder/flow-canvas/utils/batch-region.ts
```diff
@@ -0,0 +1,168 @@
...
+  const cross = {
+    start: Math.min(
+      spine.cross - CROSS_PADDING,
+      ...extents.map((extent) => extent.start),
+    ),
+    end: Math.max(
+      spine.cross + layout.stepCrossSize + CROSS_PADDING,
+      ...extents.map((extent) => extent.end),
+    ),
+  };
```
AbdulTheActivePiecer, line 89: can we please pick better names than than along and cross? I can't really understand what they are, I would have to trace the code and visualize to do so.

PR #14895, opened 2026-08-18
feat(builder): add the Process in Batches settings panel and put the step in the picker

othmanemad:
## Description

**Stack: 10 of 12.** Base: [#14894](https://github.com/activepieces/activepieces/pull/14894).

**This is the PR that makes the step authorable.** Four PRs have shipped the type, the executor, the dispatcher and the canvas with the step deliberately held out of `CORE_ACTIONS_METADATA`. This adds the settings panel, the picker entry and the icon, so a user can build one.

### The settings panel is not the generic piece form

Two fields — the list to process, and the items per batch — plus one derived line. That line is the point of the panel: it renders the resulting split ("Split 1,000 items, 10 each", "up to 5 · your project's limit") from the **last test run**, so the number that actually decides throughput is visible while you are choosing it, rather than something you discover on the next production run. It is labelled as coming from the last test run, because that is the only place a real item count exists at authoring time.

`batchSize` is a plain `type="number"` field, not a mention field. `items` takes expressions; `batchSize` never can, and the panel treats them differently on purpose (recorded as a gotcha in `brain/flows-execution/flows.md`).

### Nesting is refused at the search, not at each call site

A batch step inside a batch body is not supported, so the picker inside a body has to hide it. That is one `excludeProcessInBatches` flag on `piecesHooks.usePiecesSearch`, rather than a filter repeated at each consumer — `pieces-card-list` and `explore-tab-content` both go through the same search.

### Failure handling gets its own field, and here is why

The shared `ActionErrorHandlingForm` reads "Add Error Handler / Adds Success and Failure branches…". Those branches come from a `CODE | PIECE` type guard, not from the flag, so a batch step showing that form would promise branches it never grows.

So batches get `BatchFailureModeField` in their own Advanced section: a two-segment "If a batch fails" control — **Stop the step** or **Continue & report** — writing the *same* `settings.errorHandlingOptions.continueOnFailure.value`. For a batch the flag is an outcome choice, not a branch, so the copy matches what happens.

The consequence worth knowing: `settings.errorHandlingOptions.continueOnFailure.value` now has two UIs, and grepping `ActionErrorHandlingForm` no longer finds them all. Grep the settings path.

**Retry-on-failure is deliberately not offered.** A retry re-dispatches every batch, including the ones that already succeeded, so anything those batches wrote to an external system happens twice.

### One revert of branch content

The branch this stack is split from removes the `isForEmptyTrigger || openSelectorOnClick` gate in `pieces-selector/index.tsx`. That gate is [#14728](https://github.com/activepieces/activepieces/pull/14728)'s fix for a plain click on a configured step reopening its own piece selector — the branch simply predates it. **This PR keeps `main`'s version**, and does not take the branch's deletion of the matching brain bullet.

## How was this tested?

- `packages/web/test/app/builder/step-settings/process-in-batches-utils.test.ts` — **new, 87 lines.** The derived split line: batch counts, the clamp against the project limit, and the empty/absent-items cases.
- `packages/web/test/features/pieces/process-in-batches-form-schema.test.ts` — **new, 62 lines.** The form schema: `batchSize` rejects non-integers and values below 1 with the `batchSizeMustBeAWholeNumber` / `batchSizeMustBeAtLeastOne` i18n keys, and `items` is required.
- `npx vitest run test/app/builder/step-settings test/features/pieces` — **27 passing, 4 files.**
- All 23 new user-facing strings are real keys in `packages/web/public/locales/en/translation.json`, including the two zod messages.
- **Editions:** CE / EE / Cloud identical — the step is not gated by plan or flag anywhere. Worth stating explicitly since there is currently **no mechanism** to hide a core step per platform or plan; if this ever needs gating, that is new machinery in three places (picker, `UpdateActionRequest` validation, engine executor lookup).

### Docs

`brain/flows-execution/flows.md` gains nine gotchas from wiring this up — the two-kinds-of-field distinction, the five traps in core-step metadata (including that `logoUrl` points at a CDN outside this repo, so a new core step needs an upload or it renders broken), the two `continueOnFailure` UIs, and why adding a `FlowActionType` needs no `LATEST_FLOW_SCHEMA_VERSION` bump.

The user-facing `docs/flows/process-in-batches.mdx` is **not** here — it documents the batch browser on the run detail page, which does not exist yet. It lands with that PR.

Fixes # (issue)


othmanemad:
Rebased onto the fixed base. The nested-batch finding was real of this branch's tree but already fixed one PR down the stack — `4ad2180e9e` on #14892 hoists `hasNestedBatch` out of `_addAction` into `flowOperations.apply`, so `UPDATE_ACTION` (the Replace path) is now guarded too. This branch and the two below it (#14893, #14894) were still based on the pre-fix commit, which is why the old code was still visible here.

Remaining and intentional: `pieceSelectorUtils.isInsideBatch` still returns `false` for `UPDATE_ACTION`, so Replace inside a batch body offers the step and the user meets the operation-layer error rather than not seeing the option. That is affordance, not enforcement — documented under Gotchas in `brain/flows-execution/server-side-fan-out.md`.

@greptile review

PR #14898, opened 2026-08-18
feat(builder): warn before a bulk action on a flow that processes in batches

othmanemad:
## Description

**Stack: 13 of 14.** Base: [#14897](https://github.com/activepieces/activepieces/pull/14897).

Last of the batch UI. The batch step changes what **cancel** and **retry** mean for a run, and this makes the runs table say so before acting.

### Two bulk actions now carry a warning, and only where it applies

| Action | What the user does not expect |
| --- | --- |
| Cancel runs | Cancelling a parent does **not** cancel the batches it already dispatched. They run to completion — and anything those batches are themselves waiting on, an approval or a delay, is never reached, so they can sit paused indefinitely. |
| Retry from failed step | Re-dispatches **every** batch, including the ones that already succeeded. Anything they wrote to an external system happens twice. |

Both are stated in the confirmation dialog, and both appear **only when the flow actually contains a batch step** — `flowRunUtils.hasBatchStep` decides that off the flow version, so ordinary flows keep the plain dialog they have today.

### Canvas step badges read the selected batch

`step-node-status-in-run` and `step-node-run-duration-and-piece-name` move onto `useStepOutputInRun` from #14897. Two consequences:

- A step **inside** a batch body now shows the selected batch's status. Before, it had no output in the parent's run at all and showed nothing.
- A batch step whose `items` resolved to an empty array reads as **Skipped**, not as a success with no children — `batchUtils.isSkippedOnEmptyItems` from #14896, surfaced for the first time.

`flowCanvasUtils.getStepStatus` is removed here, with its last caller. #14894 deliberately kept it alive for exactly this reason.

### Three fixes the batch work surfaced

These are independent of batches and worth calling out so they are not read as refactors:

- **`loop-iteration-input` clamped nothing.** The stored index is per step and survives a re-run, so a stored index past the end of a shorter iteration list rendered an out-of-range round number in the input and in the dots. Now clamped to the actual iteration count.
- **`notes-state` mutated the notes array in place**, then did `notes[noteIndex].ownerId = …` **after** an `if (noteIndex !== -1)` guard — so dropping a note no longer present in the version threw on `undefined`. Rewritten as a `.map`, which fixes both the throw and the mutation-outside-`set` (Zustand cannot see an in-place edit, so the ownership change did not always render).
- **`useBuilderStateContext` now wraps its selector in `useShallow`.** Most callers return an array literal, which allocates a new array every render and defeats Zustand's reference equality — every store change re-rendered every one of them. This became visible with the batch UI's extra subscribers, but it was always wrong.

## How was this tested?

- `npx vitest run` in `packages/web` — **504 passing, 52 files.** `test/features/chat/lib/chunk-reducer.test.ts` fails with `window is not defined`; it fails on `origin/main` too and is unrelated.
- Every `t()` call in the changed files resolves to a real key, including the ICU-plural `batchRetryRunCount` — verified by scanning the staged files.
- **Editions:** CE / EE / Cloud identical. The dialogs are plain copy with no branding and no gate.

Fixes # (issue)


packages/web/src/app/builder/flow-canvas/nodes/step-node/step-node-status-in-run.tsx
```diff
@@ -1,24 +1,46 @@
...
+      <StepNodeBadgeContainer>
+        <div
+          className={flowRunUtils.getStatusContainerClassName('default', true)}
+        >
+          <RouteOff className="size-3" />
+          <div>{t('Skipped')}</div>
+        </div>
+      </StepNodeBadgeContainer>
+    );
+  }
```
AbdulTheActivePiecer, line 43: this logic should already be handled in packages/web/src/app/builder/flow-canvas/nodes/step-node/step-node-skipped-status.tsx
because it is a status that is shared between both in-run and draft state.

packages/web/src/app/builder/run-details/use-batch-logs.ts
AbdulTheActivePiecer, line 3: I would argue this file needs to actually be moved into packages/web/src/app/builder/state/batches-state.ts

also if other types will rely on the kind literal you have in BatchLogs, it might be better to make it an enum

packages/web/src/features/flow-runs/hooks/flow-run-hooks.ts
```diff
@@ -193,10 +213,52 @@ export const flowRunQueries = {
 
     return { categories, total, isLoading, dataUpdatedAt, refetch };
   },
+  useMayProcessInBatches: ({
+    runs,
+    hasUnknownRuns,
+    enabled,
+  }: {
+    runs: { flowId: string; flowVersionId: string }[];
+    hasUnknownRuns: boolean;
```
AbdulTheActivePiecer, line 222: hasUnknownRuns is not a clear name, please rename it

packages/web/src/features/flow-runs/hooks/flow-run-hooks.ts
```diff
@@ -130,17 +139,22 @@ export const flowRunQueries = {
             cursor: pageParam as string | undefined,
           }),
         enabled: enabled && !isNil(barrierId),
-        refetchInterval: 7000,
+        refetchInterval: (query) =>
+          isDispatchComplete && !hasPendingChild(query.state.data)
```
AbdulTheActivePiecer, line 143: hasPendingRun is a better name here, also why do we pass a isDispatchComplete flag if we are checking the query data if it has pending runs or not? same question is to the other change in this file

packages/web/src/features/flow-runs/components/runs-table/index.tsx
```diff
@@ -352,17 +394,26 @@ export const RunsTable = () => {
...
-                    onClick={() => {
+                  <ConfirmationDeleteDialog
+                    title={t('Cancel runs')}
+                    message={t(
+                      'Are you sure you want to cancel the selected runs?',
+                    )}
+                    warning={
+                      selectionMayProcessInBatches
+                        ? t(
+                            'If any of these runs processes items in batches, cancelling also cancels the batches it dispatched — queued batches never start, and any batch waiting on an approval or a delay is dropped. Batches already executing run to completion.',
```
AbdulTheActivePiecer, line 405: Nit: might need a little bit of copy enhancement, this might be a bit too much for a non power user.

PR #14899, opened 2026-08-18
fix(flows): releases and project replace now apply changed connections instead of reverting them

### LINKED ISSUE #14873: [BUG] Release apply / project replace reverts changed connection references (new placeholder left orphaned)
## Problem

Applying a release (Git Sync / Environments) or `project-replace` where a step's connection reference changed: the step silently keeps the old `{{connections['...']}}` value. The new connection IS created (status MISSING) but nothing ever references it. Surfaces as "missing connections" + stale mappings after every release for pre-existing flows.

Ref: Pylon 5614

## Root cause

Two independent mechanisms, both reproduced on main:

1. Diff layer: `normalize()` in `project-state/diff/flow-diff.service.ts` blanks `input.auth` before comparing. Auth-only change means the flow is reported unchanged (`flowsUnchanged: 1`), UPDATE_FLOW never fires.
2. Helper layer: `updateFlowInProject()` in `project-state/project-state-helper.ts:28`: when any other change triggers the update, old step auth is unconditionally restored (`if (oldStep?.settings?.input?.auth && isNotEmptyTrigger)`). No changed-reference check, no piece check. Swapping a step to a different piece grafts the old piece's connection onto the new piece (reproduced).

Shared helper, so Git Sync releases AND project-replace CLI are affected identically (`project-replace.service.ts:518,528`). Origin: f196530db6 (Jul 2024), deliberate secret preservation, predates externalId placeholder sync.

## Fix

Contract decision needed first: teams that manually re-point target-project steps to different-externalId connections rely on today's preserve behavior; naively honoring incoming auth resets their mapping on every release (breaking-change question).

If externalIds are declared environment-stable (placeholder model):

* helper: preserve old auth only when incoming auth is nil/empty; require piece match
* diff: stop blanking auth in `normalize()` (or compare it separately)

## Verification

Repro suite pinning current behavior (4 passing tests): `packages/server/api/test/integration/cloud/project-replace/release-apply-auth-clobber-claim-verify.test.ts`: auth-only skip, clobber-on-update, wrong-piece graft, blank-auth preserve control.

majewskibartosz:
Fixes [GIT-1764](https://linear.app/activepieces/issue/GIT-1764)
Closes #14873

## Problem

Applying a release (Environments / Git Sync) or running Project Replace when a step's connection changed in the source:

1. Auth-only change: `normalize()` in `flow-diff.service.ts` blanked `input.auth` before comparing, so the flow was reported unchanged (`flowsUnchanged: 1`) and `UPDATE_FLOW` never fired. The step kept the old connection; the new one was created with status `MISSING` and referenced by nothing.
2. Auth change alongside another edit: `updateFlowInProject()` unconditionally restored the destination's old auth (`if (oldStep?.settings?.input?.auth && isNotEmptyTrigger)`), reverting the incoming reference even though the rest of the release applied.
3. Piece swap: same restore ran with no piece check, grafting the old piece's connection onto a step swapped to a different piece.
4. Pre-existing, found while fixing: `normalize()` blanked auth only where it existed, so a destination step with auth versus an incoming step without one compared `{auth: ''}` against `{}` and reported the flow changed on every apply, forever.

Shared helper, so Git Sync releases and the Project Replace CLI are affected identically. Origin: `f196530db6` (Jul 2024), deliberate secret preservation that predates externalId placeholder sync.

## Fix

- `project-state-helper.ts` -> extracted `mergeDestinationAuth`: keep the destination's connection only when the incoming step carries none **and** the step is still the same piece.
- `flow-diff.service.ts` -> stopped blanking `input.auth`; the diff now compares the **merged** incoming version, so it always agrees with what apply writes (this is what kills failure mode 4).
- `docs/install/reference/breaking-changes.mdx` -> Unreleased entry.

Contract per the ticket's internal decision: externalIds are environment-stable, secrets never cross instances, dest equals source.

## Verification

- `test/integration/cloud/project-replace/release-apply-connection-refs.test.ts`, 8 new tests, all pass. Red-first proven against the base commit: 6 fail, 2 pass (the two that pass are the intentional-preserve control and the idempotency control, correct in both directions).
- `test-cloud` neighbours green: `project-replace.test.ts` 15, `project-release.test.ts` 12, plus the new 8 = 35 passed.
- `test/unit/app/ee/projects` 38 passed. `tsc --noEmit` clean. `npm run lint-dev` 32/32 tasks, 0 errors.
- Driven end to end on a real EE stack (Postgres + Redis + web + api + worker), two projects, release applied through the UI: source `slack_production` -> `slack_backup` applied, then reverted back to `slack_production` and applied again. Destination `flow_version.trigger.settings.input.auth` followed the source both ways.
- Visual: Releases -> Create Release -> From Project -> Review Changes, dark theme, before/after on the same instance.
- Other affected paths: checked - the only two callers of `updateFlowInProject`/`flowDiffService` are `project-state.service.ts` (Git Sync / Environments) and `project-replace.service.ts` (CLI); both go through the shared merge, so both are fixed. `createFlowInProject` routes through the same helper against an empty destination flow, unaffected. `app-connection.handler.ts` rewrites auth on connection rename, a separate path with no shared defect.
- Product decision embedded: honor the incoming connection reference on apply, keeping the destination's only when the incoming step has none; alternative considered: an opt-in flag to preserve destination wiring, rejected because Project Replace's stated semantics are "dest equals source" and Project Releases is being deprecated.
- Repro: real EE stack, two projects, connection re-pointed in the source and the destination's release dialog read before and after - full steps in the Reproduction block.
- Not verified: Git Sync against a real SSH repo. It shares `flowDiffService` and `projectStateHelper` with the paths above and has no code of its own on this route.

<details>

## Plan
- Two mechanisms, one contract, so fix both layers or the bug half-survives: the diff must see the change, the apply must not revert it.
- First pass patched the two sites independently (10 lines). Self-review found that made a destination-has-auth / incoming-has-none step permanently dirty, because the diff and the apply then disagreed about what "changed" means.
- Final shape puts the rule in one function both sides call, so they cannot drift. `mergeDestinationAuth` lives in `project-state-helper.ts`; `flow-diff.service.ts` imports it. No new file: `flow-diff.test.ts` already pulls in `project-state.service.ts` -> `project-state-helper.ts`, so the import graph does not get heavier.
- Footprint estimated 3 files / ~25 lines, actual 2 product files / ~30 lines plus docs.

- Incoming step has no connection -> destination wiring kept (control test), and settles to unchanged on the next apply.
- Step swapped to a different piece, incoming has a connection -> that piece's connection applied, old one not grafted.
- Step swapped to a different piece, incoming has none -> no connection, rather than the old piece's.
- Same state applied twice -> `flowsUpdated: 0`.
- Connection reverted to the previous one -> applied, not treated as noise.
- Old step is a piece with auth, new step at the same name is a CODE step -> `pieceName` mismatch, so no auth is injected into the code step's input.
- `connectionIds` on the version is recomputed by `flowVersionService` on every mutation, so it tracks the new reference rather than the old.

## Alternatives rejected
- Honor incoming auth unconditionally -> breaks the legitimate case of an incoming step with no connection, which is the behaviour the original code existed for.
- Compare auth separately in the diff instead of merging -> leaves diff and apply as two rules that can drift, which is exactly failure mode 4.
- Patch only `updateFlowInProject` -> auth-only changes stay invisible, the customer's actual report.
</details>

<details>

## Environment

- `docker-compose.dev.yml` from this repo: `postgres:14.4` on :5432, `redis:7.0.7` on :6379, both with the committed dev credentials.
- App from the worktree, not the container image: `npx turbo run serve --filter=web --filter=api --filter=@activepieces/engine --filter=worker --env-mode=loose`, with `AP_EDITION=ee AP_DB_TYPE=POSTGRES AP_POSTGRES_HOST=localhost AP_REDIS_HOST=localhost AP_JWT_SECRET=dev-secret-please-change-in-production`. Web :4200, API :3000, verified `EDITION: ee` on `/api/v1/flags`.
- One platform, two projects. Seeded login `dev@ap.com` / `12345678`.
- `platform_plan.environmentsEnabled` and `project.releasesEnabled` held on: a licence-less dev EE resets the plan to `free` on a timer and 402s the release endpoints. A dev-licence artefact, unrelated to the bug.

## Harness

Public gist, five files: https://gist.github.com/majewskibartosz/6cd003cfaca67a5c8e912eb27bf82863

    git clone https://gist.github.com/6cd003cfaca67a5c8e912eb27bf82863.git git-1764-repro
    cd git-1764-repro && REPO=/path/to/activepieces ./run.sh

- `run.sh` boots infra and the EE stack, waits for the API, holds the plan flags, runs the seed.
- `seed.py` creates the destination project, seeds one flow in the source pointing at `{{connections['slack_production']}}`, applies a first release so both projects match.
- `repoint.py <externalId>` re-points the source flow and prints the release plan the destination would render.
- `check.sh` prints the destination flow's persisted `input.auth`.

Every credential in the harness is a default already committed to this repo. No real infrastructure is touched.

## Trigger

`python3 repoint.py slack_backup`, then in the destination project: **Releases -> Create Release -> From Project -> Review Changes**.

Re-pointing is the whole trigger: nothing else about the flow changes, which is exactly the case the diff's `normalize()` erased. Adding a rename alongside it exercises the second mechanism instead, since the rename makes the diff fire and the clobber happens inside `updateFlowInProject`.

## Results

| Source step | Base commit | With the fix |
|---|---|---|
| connection changed, nothing else | plan `[]`, dialog shows "No Changes Found", destination stays on the old connection | plan `[UPDATE_FLOW]`, dialog lists "Flows Changes (1/1)", applying writes the new reference |
| connection changed + flow renamed | rename applied, connection reverted to the old one | both applied |
| swapped to a different piece, with its own connection | piece applied, old piece's connection grafted on | that piece's connection applied |
| swapped to a different piece, no connection | old piece's connection grafted on | no connection |
| no connection on the incoming step | destination wiring kept, but reported changed on every apply forever | destination wiring kept, settles to unchanged on the next apply |
| identical state applied twice | unchanged | unchanged |

The last two rows are the ones that decided the shape of the fix: patching the diff and the apply separately makes row 5 permanently dirty, which is why both now share `mergeDestinationAuth`.
</details>

### DIFF
--- a/docs/install/reference/breaking-changes.mdx
+++ b/docs/install/reference/breaking-changes.mdx
@@ -55,6 +55,19 @@ This affects the Tables piece's Find Records action and any direct API call that
 
 Nothing on upgrade. Re-check any flow or API integration that filters a Date column with `gt`, `gte`, `lt` or `lte` — it now returns the rows the filter actually describes, which may be more or fewer than before.
 
+#### Applying a release or a project replace now updates changed connection references
+
+Applying a release (Environments / Git Sync) or running Project Replace used to keep whatever connection the destination step already pointed at. A step whose connection changed in the source silently kept the old one, and the connection the release created for it stayed unreferenced with status `MISSING`. Two things change:
+
+- A changed connection reference is now a change the diff sees, so a flow whose only edit is its connection is applied instead of being reported as unchanged.
+- The destination's connection is kept only when the incoming step carries no connection of its own, and only when the step is still the same piece. Swapping a step to a different piece no longer grafts the old piece's connection onto it.
+
+Connection secrets still never cross instances. Only the reference travels, and a reference to a connection that does not exist in the destination yet lands as a `MISSING` connection for someone to authorize there.
+
+#### What you need to do
+
+Nothing on upgrade if your source project is the source of truth for connections, which is the intended model. If instead you deliberately re-point a step in the destination project at a different connection than the source uses, that re-point is now overwritten on the next apply. Either point the source at the connection you want, or clear the connection on the source step so the destination wiring is preserved.
+
 
 ## 0.87.0
 
--- a/packages/server/api/src/app/ee/projects/project-release/project-state/diff/flow-diff.service.ts
+++ b/packages/server/api/src/app/ee/projects/project-release/project-state/diff/flow-diff.service.ts
@@ -1,6 +1,7 @@
 import { assertNotNullOrUndefined, isNil, mapsAreSame } from '@activepieces/core-utils'
 import { DEFAULT_SAMPLE_DATA_SETTINGS, FlowActionType, flowPieceUtil, FlowProjectOperationType, FlowState, flowStructureUtil, FlowTriggerType, FlowVersion, ProjectOperation, ProjectState, Step } from '@activepieces/shared'
 import deepEqual from 'deep-equal'
+import { mergeDestinationAuth } from '../project-state-helper'
 
 export const flowDiffService = {
     async diff({ newState, currentState }: DiffParams): Promise<ProjectOperation[]> {
@@ -99,7 +100,7 @@ async function isFlowChanged(fromFlow: FlowState, targetFlow: FlowState): Promis
     })
 
     const normalizedFromFlow = await normalize(fromFlow.version)
-    const normalizedTargetFlow = await normalize(targetFlow.version)
+    const normalizedTargetFlow = await normalize(mergeDestinationAuth({ destinationVersion: fromFlow.version, incomingVersion: targetFlow.version }))
     return normalizedFromFlow.displayName !== normalizedTargetFlow.displayName
         || !deepEqual(normalizedFromFlow.trigger, normalizedTargetFlow.trigger) || !isMatched || !notesMatched
 }
@@ -109,13 +110,9 @@ async function normalize(flowVersion: FlowVersion): Promise<FlowVersion> {
         const clonedStep: Step = JSON.parse(JSON.stringify(step))
         clonedStep.settings.sampleData = DEFAULT_SAMPLE_DATA_SETTINGS
         clonedStep.lastUpdatedDate = ''
-        const authExists = clonedStep?.settings?.input?.auth
 
         if ([FlowActionType.PIECE, FlowTriggerType.PIECE].includes(step.type)) {
             clonedStep.settings.pieceVersion = ''
-            if (authExists) {
-                clonedStep.settings.input.auth = ''
-            }
         }
         return clonedStep
     })
--- a/packages/server/api/src/app/ee/projects/project-release/project-state/project-state-helper.ts
+++ b/packages/server/api/src/app/ee/projects/project-release/project-state/project-state-helper.ts
@@ -1,5 +1,5 @@
-import { isNil } from '@activepieces/core-utils'
-import { FlowOperationType, FlowState, FlowStatus, flowStructureUtil, FlowSyncError, PopulatedFlow } from '@activepieces/shared'
+import { isEmpty, isNil } from '@activepieces/core-utils'
+import { FlowOperationType, FlowState, FlowStatus, flowStructureUtil, FlowSyncError, FlowVersion, PopulatedFlow } from '@activepieces/shared'
 import { FastifyBaseLogger } from 'fastify'
 import { flowService } from '../../../../flows/flow/flow.service'
 import { projectService } from '../../../../project/project-service'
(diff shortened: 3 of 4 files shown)


majewskibartosz:
## Visual verification

Real EE stack (Postgres + Redis + web + api + worker), two projects on one platform. Source project holds one flow whose Slack trigger points at `{{connections['slack_production']}}`; a first release copies it into the destination. The source is then re-pointed at `{{connections['slack_backup']}}` and the destination's **Releases -> Create Release -> From Project -> Review Changes** is opened. Same instance, same data, only the two changed files differ.

| Before (base commit) | After |
|---|---|
| before | after |
| "No Changes Found — There are no differences to apply". The release cannot be created, and the destination keeps pointing at the old connection. | "Flows Changes (1/1)" listing *Notify team on new lead* as an update, ready to apply. |

States driven: releases list (empty and populated), Create Release menu, From Project selection dialog, no-changes toast, Create Project Release dialog with the flow change listed, and Apply Changes through to a written release. Dark theme; the change is not theme-sensitive and touches no `packages/web` code.

Applied in both directions on the running instance and read back from `flow_version.trigger.settings.input.auth` in the destination project:

- `slack_production` -> `slack_backup`: destination becomes `{{connections['slack_backup']}}`
- `slack_backup` -> `slack_production`: destination becomes `{{connections['slack_production']}}`

Louai-Zokerburg (approved):

PR #14825, opened 2026-08-15
feat(ai-providers): AI Center providers UI with multi-key configurations

Louai-Zokerburg:
> [!NOTE]
> **Stacked on #14738** — review only this PR's diff (base is `feat/ai-providers-redesign`, not `main`). Merge #14738 first; GitHub retargets this PR to `main` automatically when its branch is deleted.

Linear: ENG-462 · Milestone "2. Providers redesign — connect keys, pick models, scope projects"

## Description

Ships the admin UI for the multi-key providers backend (#14738): the AI providers page becomes an **AI Center** with vertical tabs (**Providers** + **Capabilities** — routing/usage tabs come with their own milestones).

**Providers tab — connect-then-configure:**
- Providers grouped by vendor as soft cards (icon-tile header, configurations count, uppercase micro-label), one row per key showing name, a models·projects meta line, project swatches, and a purple "Chat" pill on the chat-enabled key.
- **Connect dialog**: provider picker with per-provider credential fields (Azure resource, Bedrock keys/region, Cloudflare gateway, custom base URL/headers), real `POST /` verification — server errors land in the form — then navigates straight into the new key's detail view (the create endpoint now returns the created row for this).
- **Config detail**: rename, replace credentials (non-secret fields prefilled, secrets blanked), model allow-list ("All models" / "Only selected" via dropdown, names-only table in v1), project access ("All" / "Only selected" / "All except") with a searchable checkbox table listing team **and personal** projects (tagged) — the backend scopes by project id regardless of type, so a personal-only platform must be able to grant itself access. Manual-model providers (Custom, Cloudflare Gateway) edit `config.models` through the same panel. Danger-zone delete.
- **Chat provider** row selects per provider name and toggles `enabledForChat` on that provider's newest key.
- Empty state is a hero card (purple icon tile, CTA) above recommended + all-providers grids.

**Capabilities tab** reuses the existing ai-tool-config feature; the standalone `/platform/setup/ai-capabilities` page now redirects to `?tab=capabilities`. The platform sidebar consolidates the two AI entries into one "AI Center" item.

**Backend riders (small, this PR):**
- `aiProviderService.create()` returns the created row (UI navigates to it).
- Anthropic `listModels` passes `?limit=1000` — the API paginates at a small default page size, so only the first page of models ever showed (pre-existing bug on main).

[...]

PR #14904, opened 2026-08-18
feat(ai-providers): let a chat turn report the key it used

Louai-Zokerburg:
## Description

Fourth PR in the AI providers stack, on top of the key-status foundation.

The chat turn runs in the **worker**, which until now had no way to say whether the key the app handed it worked — so a key that only ever serves chat never fed its own badge. It gets one:

- The app returns the resolved key id with the chat config (`AgentConfigResponse.aiProviderId`). Before this, nothing downstream could name the key it was given.
- Every HTTP request the turn's models make reports back through `reportAiProviderOutcome` on the worker RPC contract that already exists.

### The worker deliberately classifies nothing

It forwards the raw status code and body, and the single classifier in the app decides. That matters because the worker already has its own regex-on-the-message classifier (`isCreditExhaustedError`) which disagrees with the status-based one — leaving it in charge would mean the worker reaching one verdict while the admin page shows another. One classifier, one answer, every process.

Reporting is fire-and-forget: a chat turn must never fail because reporting did.

## How was this tested?

- Full CE ai-provider + EE agent suites pass (229 tests); api, worker and web all build; worker typechecks against the widened contract.
- The foundation's own tests cover the classification and the write path — this PR only adds a caller.
- **Still to do manually:** point a key at an exhausted OpenRouter account, send a chat message, and confirm the badge reads *Out of credits*. That needs a real provider account, so it is not in CI.

Fixes # (issue)

`AgentConfigResponse` gains a required field and `WorkerToApiContract` gains a method, so a worker and an app from different releases would disagree — but that is already impossible: `apVersionUtil.versionsAreCompatible` is fail-closed and withholds jobs from a version-skewed worker. Nothing customer-facing changes, and there is no new setup step.

It forwards provider response bodies from the worker to the app over the authenticated worker RPC, where the foundation PR truncates them to 300 characters before storing. Same consideration as that PR: a provider may echo a masked key in a 401 body. No new transport and no new credential handling — the worker already receives the decrypted key in the chat config.


Louai-Zokerburg:
Parking this one. The key-status work lands after the base AI-providers functionality merges, so the stack stops at #14825 for now. Branch `feat/ai-key-status-chat` is kept as-is.

PR #14949, opened 2026-08-20
feat(auth): name a new workspace after the company in the sign-up email

AhmadTash:
## Description

Completing sign-up always named the new platform after the person, as `"Ahmad's Platform"`. That throws away the one thing a work address already tells us: where they work.

Naming now reads the email domain first, and falls back to the person:

| Sign-up email | Platform name | Project name |
| --- | --- | --- |
| `ahmad@activepieces.com` | `Activepieces` | `Activepieces's Project` |
| `ahmad@acme-widgets.com` | `Acme Widgets` | `Acme Widgets's Project` |
| `ahmad@gmail.com` | `Ahmad's Platform` | `Ahmad's Project` |

`platformNameFromPerson` keeps its pure person-to-name semantics and all of its existing tests. The new `platformNameFromSignup` composes the two paths, and is what `passwordless-auth.service.ts` calls. `personalProjectName` already handled a company-style platform name, so the project name follows for free.

### Two details that took a second pass

**The consumer denylist is keyed on the registrable label, not the full domain.** So the single entry `yahoo` also catches `yahoo.co.uk`, and we do not need a row per country.

**That label is the second-to-last domain part, stepping back one more when the part before the TLD is itself a public suffix** (`co`, `com`, `ac`, …). Without this step-back, these would have been named `Mail`, `Co` and `Eu`:

```
mail.activepieces.com      -> Activepieces
activepieces.co.uk         -> Activepieces
eu.activepieces.co.uk      -> Activepieces
activepieces.com.au        -> Activepieces
```

This is a heuristic, not a public-suffix list, and the failure mode is worth stating plainly: **a company sitting on a two-part suffix we do not list gets the suffix as its name.** That felt like the right trade against pulling in a PSL dependency and keeping it current, but it is the thing to argue with if you disagree.

No new dependency. `disposable-email-domains` is already in the tree but it lists temp-mail domains, not consumer providers, so it cannot answer "is this a work address" (gmail.com is not in it).

## How was this tested?

**Unit — `signup-names.test.ts`, 49 passing** (21 pre-existing kept green and unchanged, 28 added). Covers company extraction, subdomains, multi-part TLDs, hyphenated labels, 9 consumer providers, 5 unusable addresses, the length cap, and that the result never violates the platform-name rule (no `.` or `/`).

**Integration — `passwordless-authn.test.ts`.** The existing case asserted `"Ahmad's Platform"` from `ahmad.tash@example.com`, which this change makes wrong; it now asserts `Example` / `Example's Project`. Added a sibling case on a `gmail.com` address asserting the person-based fallback still holds, so both branches are covered end to end.

`tsc --noEmit` on `api` is clean and `eslint` reports 0 errors on both changed source files (the one remaining warning pre-exists on `main` — verified by stashing).

**Not verified:** the full `api` unit suite does not run in my worktree. It fails 54 files on `Failed to resolve @activepieces/core-utils`, which is my symlinked `node_modules` under Bun's isolated linker, not a regression: 0 assertion failures and no authentication test among them. CI runs it properly.

Edition paths: this path is Cloud in practice (`getPreferredPlatformId` returns null only on Cloud; self-hosted sign-up joins the platform that already exists). A fresh self-host with no platform takes the same path and gets the same improvement. Nothing here is edition-gated, and no new env var or setup step is introduced.

Fixes # (issue)

Ticked functional because the default name of a newly created workspace changes. Nothing to migrate and no action required on upgrade — existing platforms keep their names and renaming stays in settings. Documented in `docs/install/reference/breaking-changes.mdx`. Happy to downgrade this to "not breaking" if you read the bar differently; I would rather over-declare it than surprise a self-hoster.

Ticked by the template's mechanical definition: the change lives in `src/app/authentication/`. It makes no authentication or authorization decision — it only derives a display name. The one real surface is that user-controlled input (the email domain) reaches a persisted, rendered platform name. Mitigations: the value comes from the domain rather than the free-text local part, every non-alphanumeric character is stripped per label, the result is capped at 50 characters, and a test asserts the output always satisfies the platform-name rule (`^[^./]+$`).

---

Stacked below this: #15003, #15004, #15005 — the chat onboarding personalization work (split out of #14998, which stacked on this same PR). Merge this one first; GitHub will retarget #15003 to `main` automatically.


Louai-Zokerburg (approved):

AbdulTheActivePiecer (approved):
Nit: please check if there is a way to check via SMTP for info like this or a service that can do this for us, just so you don't have the odd case where someone's domain isn't the same as the company name.

PR #14973, opened 2026-08-21
fix(pieces): approval step no longer points to a feature that was removed

### LINKED ISSUE #14971: [BUG] Approval piece tells users to switch to Manual Task, a feature that no longer exists
## Problem

Both Approval actions render a builder warning: "Please use Manual Task feature instead from 0.48.0 and above". Manual Task (todos) was deprecated in PR 10929 and is gone from the codebase (no module, no entity, no UI), so the notice points at nothing.

Worse, Create Approval Links is currently the only way to mint a run resume link without pausing, so the warning steers users off the one working option.

The same claim is duplicated in both actions' `aiMetadata.description`, so the AI copilot repeats it when recommending steps.

Ref: Pylon 5699

## Root cause

Hardcoded strings:

* `packages/pieces/core/approval/src/lib/actions/wait-for-approval.ts:14` (MarkDown WARNING), same claim at `:10` (aiMetadata)
* `packages/pieces/core/approval/src/lib/actions/create-approval-link.ts:15`, same claim at `:11`
* translated into 10 locales under `packages/pieces/core/approval/src/i18n/`

## Fix

Replace the copy with something accurate (legacy but supported; no replacement feature exists today) or drop the warning. Update `aiMetadata.description` in both actions, `i18n/translation.json` plus the locale files, and bump the piece version.

majewskibartosz:
Fixes [GIT-1775](https://linear.app/activepieces/issue/GIT-1775)
Closes #14971

## Problem

1. Both Approval actions render a builder notice: "Please use Manual Task feature instead from 0.48.0 and above". Manual Task (todos) was deprecated and removed, so the notice points at nothing. `docs/install/reference/breaking-changes.mdx:253` sends Todos users the other way, to the Approvals tab.
2. The same claim is duplicated in both actions' `aiMetadata.description`, so the copilot repeats it when recommending steps, and it is translated into ten locales.
3. `docs/embedding/navigation.mdx` documents `todos` and `todos/{todoId}` embed routes that resolve nowhere.

## Fix

- `wait-for-approval.ts`, `create-approval-link.ts`: notice now points at the Approvals tab; per-action wording, since Create Approval Links has no equivalent there (it hands out a link without pausing). Variant WARNING to INFO, as this is no longer a "stop using this" message.
- Both `aiMetadata.description` strings: same correction, so agent-facing guidance matches.
- `i18n/translation.json`: stale key replaced with the two new source strings; the dead key removed from all ten locale files, letting Crowdin regenerate.
- `package.json`: 0.1.25 to 0.1.26.
- `docs/embedding/navigation.mdx`: two dead route rows deleted.

## Verification

- `test/legacy-notice.test.ts` added: asserts neither action mentions Manual Task, both point at the Approvals tab, and both strings exist as keys in `translation.json`. Red-first against the base commit: 4 of 6 assertions fail; all 8 tests in the piece pass with the fix.
- eslint clean on `@activepieces/piece-approval` and `web`.
- Full pre-push suite green (474 engine tests) after building the delay piece in the worktree.
- Visual: real builder at `/projects/:id/flows/:id` on a local CE instance (own Postgres/Redis, `AP_DEV_PIECES=approval`), both actions added by hand, before and after captured from the same instance. Light theme only; the change is not theme-sensitive.
- Other affected paths: checked. `Manual Task` appears nowhere else in the repo except the Woodpecker piece, where it refers to that product's own concept. Stale `todos` references also checked in the web router and embed SDK; the two doc rows were the only ones left.
- Not covered: piece packages are not in the `turbo run test` matrix in `ci.yml:152`, so this test only runs locally. Pre-existing for the whole `packages/pieces` tree, not introduced here.

Product decision embedded: the notice now recommends the Approvals tab rather than declaring the piece unreplaced; alternative considered: dropping the notice entirely, rejected because the piece stays discoverable and users deserve the newer path.

<details>

## Plan

- Root cause is copy in two files plus its translations, so the fix is string-level; no logic touched.
- Wording is per-action rather than shared: the Approvals tab actions replace Wait for Approval, but none of them replaces Create Approval Links, which is the only action handing out a resume link without pausing the run.
- Locale files get the dead key removed but no hand-written translations; `crowdin.yml` treats `translation.json` as source and generates `%two_letters_code%.json`.
- Footprint: 15 files, ~26 lines of product code. Rejected: adding a shared single string for both actions, which would have been shorter but wrong for Create Approval Links.

- Saved flows using these actions: prop key `markdown` unchanged, MARKDOWN props are skipped by input validation, so stored settings are unaffected.
- Non-English builders: see the English string until Crowdin round-trips. The stale translated warning is gone from all ten locales, so nobody sees the removed-feature claim in the meantime.
- Test giving a false green: the first version of the test read `props.markdown.value`, which is always undefined because `Property.MarkDown` stores the text in `description`. Red-first caught it; the test now asserts the field is a string before matching.

## Alternatives rejected

- Deleting the notice: loses the pointer to the better path, and the piece is still listed in the picker.
- Keeping WARNING variant: the message is informational now, not a deprecation stop sign.
- Editing locale files by hand: Crowdin owns them and machine-overwrites.

</details>

### DIFF
--- a/docs/embedding/navigation.mdx
+++ b/docs/embedding/navigation.mdx
@@ -63,8 +63,6 @@ Here is the list for routes the sdk can navigate to:
 | `/connections` | Connections table
 | `/tables` | Tables table
 | `/tables/{tableId}` | Opens up a table
-| `todos` | Todos table
-| `todos/{todoId}` | Opens up a todo
 
 
 ## Navigate to Initial Route
--- a/packages/pieces/core/approval/package.json
+++ b/packages/pieces/core/approval/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-approval",
-  "version": "0.1.25",
+  "version": "0.1.26",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
--- a/packages/pieces/core/approval/src/i18n/de.json
+++ b/packages/pieces/core/approval/src/i18n/de.json
@@ -4,6 +4,5 @@
   "Create Approval Links": "Links zur Genehmigung erstellen",
   "Pauses the flow and wait for the approval from the user": "Pausiert den Fluss und wartet auf die Genehmigung des Benutzers",
   "Create links only without pausing the flow, use wait for approval to pause": "Erstelle Links nur ohne den Fluss zu pausieren, benutze Wartezeit um zu pausieren",
-  "Markdown": "Markdown",
-  "Please use Manual Task feature instead from 0.48.0 and above": "Bitte verwenden Sie stattdessen die Manuelle Aufgabenfunktion ab 0.48.0 und höher"
+  "Markdown": "Markdown"
 }
--- a/packages/pieces/core/approval/src/i18n/es.json
+++ b/packages/pieces/core/approval/src/i18n/es.json
@@ -4,6 +4,5 @@
   "Create Approval Links": "Crear enlaces de aprobación",
   "Pauses the flow and wait for the approval from the user": "Pausa el flujo y espera la aprobación del usuario",
   "Create links only without pausing the flow, use wait for approval to pause": "Crear enlaces solo sin pausar el flujo, usar esperar aprobación para pausar",
-  "Markdown": "Markdown",
-  "Please use Manual Task feature instead from 0.48.0 and above": "Por favor, utilice la función de tarea manual en lugar de 0.48.0 o superior"
+  "Markdown": "Markdown"
 }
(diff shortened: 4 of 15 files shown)


majewskibartosz:
## Visual verification

Local CE instance (own Postgres/Redis, `AP_DEV_PIECES=approval`), real builder at `/projects/:id/flows/:id`, both actions added by hand. Before captured from the same instance by reverting the piece source and letting the dev piece watcher rebuild.

| Before | After |
| --- | --- |
| before | after |

Create Approval Links, which gets its own wording:

create approval links

States driven: Wait for Approval selected, Create Approval Links selected, both showing the notice panel and the bumped v0.1.26. Light theme only, the change is not theme-sensitive.

kishanprmr (approved):

AbdulTheActivePiecer (approved):

PR #14984, opened 2026-08-21
fix(flows): stop large flows from pegging app CPU and tripping healthchecks

abuaboud:
## Description

Every flow read went quadratic in the number of steps, pinning production app containers at their 1-CPU cap.

`removeConnectionsAndSampleDataFromFlowVersion` ran `JSON.parse(JSON.stringify(step))` inside its `transferFlow` callback. A `step` carries `nextAction` — the entire rest of the chain — plus loop/router children, so cloning step *i* copied the remaining `N-i` steps. The whole pass was **O(N²)**.

The clone was also redundant: `flowStructureUtil.transferFlow` already deep-clones the flow version *before* invoking the callback, so the callback is handed a private copy. It's replaced with a shallow spread, matching the existing callback style in `update-sample-data-info.ts`. The pass is now **O(N)**.

This runs on **every** `getFlowVersionOrThrow`, including the default `removeConnectionsName=false` / `removeSampleData=false` case — where the callback does nothing but the quadratic cloning happened anyway.

### How it was found

A CPU profile (CDP `Profiler`, 30s, 52k samples) of a production app container:

```
42.44%  (anon) @ flow-version.service.js  ← the callback; ~84% of all non-idle CPU
 4.42%  (garbage collector)
49.29%  (idle)
 <0.3%  everything else
```

Max `transferStep` recursion depth was **255**, i.e. ~32k step serialisations for a single call, plus the GC churn behind ~2.5 GB RSS.

The user-visible symptom was containers reporting **unhealthy** with nothing actually crashed: the 5s healthcheck `curl` timed out while the event loop was blocked (the leftover zombie `curl` processes in those containers are those killed healthchecks). The database was never the bottleneck — Postgres sat in `wait_event=ClientRead`, waiting on the app.

### Scope

The identical pattern exists in `ee/…/project-state/diff/flow-diff.service.ts` (`normalize`). It's deliberately **not** touched here: it's a colder path, it did not appear in the profile, and it has no test coverage — changing it risks spurious project-release diffs. It's recorded in the brain page for whoever picks it up.

## Also in this PR (docs, no code paths touched)

**`.agents/skills/profile-app-cpu/`** — the CPU counterpart to the existing `profile-worker-memory` skill, recording the procedure that found this hot spot so nobody has to rediscover it: rule out the DB via `pg_stat_activity` wait events, split `MainThread` from `V8Worker` (GC) and `libuv-worker` with `top -H`, open the V8 inspector on a live container with `SIGUSR1` (no restart), profile over CDP using Node 22+'s global `WebSocket`, and — the step that makes or breaks the read — **aggregate self-time by function first**, because V8 emits a separate node per call stack so one hot function scatters across dozens of individually-negligible entries.

**`brain/knowledge/engineering/ci-pr-review-hygiene.md`** — a gotcha found while adding the regression test above: **a unit test under `packages/server/api/test/unit/` never runs in CI.** `ci.yml` runs `turbo run test` filtered to engine/shared/sandbox/ai-providers/pieces-framework/web, plus api's `test-ce`/`test-ee`/`test-cloud`. The api package has a `test-unit` script, but nothing invokes it and the root `test-unit` filter excludes api — so the 10+ files already in `test/unit/**` gate nothing. `packages/core/execution` is in the same position. Documented rather than rewired here: turning those suites on is its own change, with its own risk of surfacing pre-existing failures.

## How was this tested?

- **New unit test** `test/unit/app/flows/flow-version/remove-connections-and-sample-data.test.ts`, verified to fail on the unfixed code and pass on the fix:
  - clone count is constant regardless of step count — **202 vs 22 before the fix**, equal after (this is the regression guard, and it's implementation-independent: it asserts the count doesn't scale with N, not a specific number)
  - connections stripped and sample data cleared at every depth in the chain
  - the input flow version is not mutated
  - The two correctness tests pass on the *old* code too, so they lock in existing behaviour rather than encoding the fix.
- `npx vitest run test/unit/app/flows/flow-version` — 16 passed (includes 3 pre-existing sibling files).
- `AP_EDITION=ce npx vitest run test/integration/ce/flows/flow` — **140 passed**.
- `tsc -p tsconfig.app.json --noEmit` — clean.
- `eslint` on the changed file — 0 errors (one pre-existing warning on the untouched factory signature at line 18).
- The two docs files touch no code paths.

Edition paths: the changed function is CE code on a shared read path, exercised by the CE integration suite above; EE/Cloud call the same function unchanged.

> Note for reviewers: CI runs api `test-ce`/`test-ee`/`test-cloud` but never api `test-unit`, so `packages/server/api/test/unit/**` (10+ existing files) isn't executed in CI today. I followed the existing convention for placement rather than rewiring CI in a bug-fix PR, and ran the test locally as shown above.

### Review notes (two things worth a second look)

**1. The `!isNil(settings.input)` guard is a deliberate, and I believe strictly safer, difference.**
The old code assigned `settings.input = removeConnectionsFromInput(settings.input)` unconditionally. `removeConnectionsFromInput` returns its argument untouched when nil, so on a step whose settings have no `input` (e.g. an `EMPTY` trigger, whose settings are `{}`) the old code *added* an `input: undefined` key. `JSON.stringify` drops undefined values, so the HTTP response was identical either way; the guard just stops us injecting a key that was never there. Nothing depended on it — the 140 CE flow integration tests pass unchanged.

**2. The shallow spread shares sub-objects with the node it replaces — this is safe, and is the existing pattern.**
`{ ...step, settings }` copies `nextAction` / `firstLoopAction` / `children` / `continueOnFailureBranches` by reference, and `transferStep` then mutates some of those (e.g. `branches.onSuccess = transferred`). That mutation is confined to the throwaway clone: `transferFlow` deep-clones the flow version *before* any callback runs, so `step` is always a node of that private clone and never of the caller's object. The parent's pointer is reassigned to the returned object, leaving the original node garbage. `update-sample-data-info.ts` already relies on exactly this. The "leaves the input flow version untouched" test pins the guarantee down.

Fixes # (issue)


abuaboud:
@greptileai Addressed the review note: the PR title now describes the user-visible outcome rather than the mechanism.

- **Before:** `fix(flows): stop cloning the whole step chain per step on every flow read`
- **After:** `fix(flows): stop large flows from pegging app CPU and tripping healthchecks`

The mechanism (the redundant per-step deep clone inside the `transferFlow` callback, and the resulting O(N²)) is covered in the description and in the commit body, so the title is free to carry the reliability outcome that self-hosters actually observe: containers sitting at their CPU cap and reporting unhealthy while nothing has crashed.

No code changed since your last review — the fix, the test, and the brain entry are unchanged at `270591d`. Could you re-review for the score?

PR #14985, opened 2026-08-21
docs: add the profile-app-cpu skill, and record that api unit tests never run in CI

abuaboud:
## Description

Adds `.agents/skills/profile-app-cpu`, the CPU counterpart to the existing `profile-worker-memory` skill. Docs only — no code changes.

It records the procedure that found the hot spot fixed in #14984, so the next person doesn't have to rediscover it:

1. Rule out the database first — sessions sitting in `wait_event=ClientRead` mean Postgres is waiting on the app, not the other way round.
2. Split `MainThread` from `V8Worker` (GC) and `libuv-worker` with `top -H`, and read `TIME+` against container age to tell a steady spin from a bursty one.
3. Open the V8 inspector on a live container with `SIGUSR1` — no restart, and it binds to localhost inside the container's netns.
4. Take a CPU profile over CDP using Node 22+'s global `WebSocket`, so the client needs no dependency.
5. **Aggregate self-time by function before reading the profile.** V8 emits a separate node per call stack, so one hot function scatters across dozens of entries that each look negligible. This is the step that makes or breaks the read.

Also captures the gotchas that cost time here: healthcheck timeouts are a blocked event loop rather than a crash, the zombie `curl` processes are killed healthchecks rather than a leak, and `docker stats` CPU% is per-core so 100% is the cap under `cpus: 1`.

## Also in this PR

A gotcha added to `brain/knowledge/engineering/ci-pr-review-hygiene.md`: **a unit test added under `packages/server/api/test/unit/` never runs in CI.** `ci.yml` runs `turbo run test` filtered to engine/shared/sandbox/ai-providers/pieces-framework/web, plus api's `test-ce`/`test-ee`/`test-cloud`. The api package has a `test-unit` script, but nothing invokes it and the root `test-unit` filter list excludes api — so the 10+ files already in `test/unit/**` gate nothing, and a new one passes review while protecting nothing. `packages/core/execution` is in the same position.

Found while adding a regression test in #14984. Documenting rather than rewiring CI here, since turning those suites on is its own change with its own risk of surfacing pre-existing failures.

## How was this tested?

Not applicable — a markdown skill file, no code paths touched. The procedure itself was executed end to end against production app containers while diagnosing #14984, and the commands and scripts in it are the ones that actually ran.

Fixes # (issue)


abuaboud:
Folded into #14984 — both docs commits (the `profile-app-cpu` skill, including the `wait_event` correction from your review, and the api-unit-tests-never-run-in-CI gotcha) now ride the same PR as the fix they came from. Closing this one.

PR #15001, opened 2026-08-23
feat(ai-providers): per-model metadata catalog from models.dev

Louai-Zokerburg:
## Description

Every AI model reached the UI as `{ id, name, type }` and nothing else. That shape was re-asserted three separate times — each provider's own `.map()`, the `modelsCache` re-map in `listModels`, and the controller's zod response schema — so every provider fetched a rich upstream payload and threw it away. `openrouter-provider.ts` literally requests OpenRouter's `pricing` object and drops it.

This adds a per-model catalog and lands it on the existing `GET /v1/ai-providers/:provider/models` response, so every consumer of that endpoint gains it at once: the web model picker, the AI piece's model dropdown, and the MCP `ap_list_ai_models` tool.

**Backend only.** Nothing in `packages/web` changes — the picker gets a proper design pass in a follow-up PR, which will need no schema work because the field is already on the response.

**Per model, where known:** context window, max output tokens, release date, input/output price per million tokens, and tool-calling / reasoning / vision support.

### How it works

A weekly workflow runs `npm run sync-model-catalog`, which fetches [models.dev](https://models.dev) (MIT), filters to the providers we support, and publishes the result to `https://cdn.activepieces.com/ai/model-catalog.json` — **795 models across 13 providers**. Nothing is committed to the repo and no process imports it.

`modelCatalog.lookup()` fetches that object at runtime: cached 24 h, concurrent callers deduped onto one in-flight request, and a 5-minute back-off after a failure so a CDN outage cannot turn "metadata absent" into "the models endpoint is slow". `AP_MODEL_CATALOG_URL` overrides the source for anyone mirroring it.

Publishing reuses the `CDN_S3_*` secrets and the `aws s3 cp` + `curl` verify already used by `publish-embed-sdk.yml`. Because there is nothing to commit, the job pushes no branch and opens no PR — which removes the `GITHUB_TOKEN`-cannot-create-PRs problem raised earlier in this review.

**Two consequences, taken knowingly (decision 000032):** an install with no egress to the CDN gets no model metadata at all — air-gapped deployments, outbound allowlists, CDN outages — falling back to today's plain `{ id, name }` row with nothing explaining the absence; and pricing data now reaches production weekly with no human reviewing a diff, so the generator refuses to publish if a provider block vanishes or the model count drops >20% against the currently published copy. The object carries `generatedAt` so staleness is diagnosable with a `curl`.

**Rollout order:** the object must exist on the CDN *before* this merges. Run the workflow once via `workflow_dispatch` and confirm the URL returns 200 — until then every install, local dev included, silently shows no metadata.

Enrichment happens in exactly one place — the `modelsCache` re-map in `fetchModels`, where every provider converges — so it costs one lookup per `(provider, key)` per day and adds ~80 B per model to the response.

`modelCatalog.lookup()` handles per-provider id shapes: `activepieces` aliases onto the `openrouter` block, and Bedrock inference-profile ids are stripped of their `us.`/`eu.`/`apac.`/`global.` region prefix. Azure returns admin-chosen deployment names and `custom` / `cloudflare-gateway` ids are hand-typed, so those return no metadata by design.

### Backward compatibility

**Not breaking, in either direction.** `metadata` is optional and every field inside it is optional.

- **The API** gains one optional object on an existing response. `fastify-type-provider-zod` serializes through `z.array(AIProviderModel)`, so unknown keys are stripped — which is exactly why the field had to be added to the shared schema to appear at all, and why an older client simply never sees it.
- **The AI piece** calls the same endpoint with `httpClient.sendRequest<AIProviderModel[]>` — a compile-time type only, no runtime validation — so a piece pinned at an older version ignores the new field entirely. It reads `id` and `name` and nothing else.
- **A new client against an older server** sees `metadata` as `undefined`, which is the same state as any model the catalog does not match — so consumers must treat it as optional, and today none of them read it yet.
- **Rolling deploys** are unaffected. `modelsCache` is a plain in-process `Map`, not `distributedStore`, so the cached-value-shape rule in `packages/server/CLAUDE.md` does not apply and no cache key version bump is needed.

### Not in this PR

- `PROVIDER_MAX_CONTEXT_TOKENS` still drives EE agent compaction off a per-provider guess. Rewiring it needs `modelId` threaded through five call sites in `agent-compaction.ts` and the worker's `run-agent-turn.ts`, and it moves agent behaviour — own PR, with `agent-evals`.
- `MANAGED_MODEL_WEIGHTS` stays as-is. It tracks real output price but is not a function of it (`claude-opus-4` $75/M → weight 45; `claude-opus-4.7-fast` $150/M → weight 200), so deriving it would silently re-price customers. **No credit value moves in this PR** — billing reads run-log telemetry times that static table and never touches `AIProviderModel`.
- **Any UI at all.** The model picker is untouched; surfacing this data is a follow-up PR with its own design pass.
- Artificial Analysis, and therefore no Speed/Intelligence scores — their free tier does not grant redistribution, and vendoring their numbers in a public repo would be redistribution.

## How was this tested?

- `npm run lint-dev`, plus web and API typechecks.
- New table-driven unit test for `modelCatalog.lookup` covering each normalization path and the deliberate misses (Azure deployment name, `custom`), and asserting an unmatched model returns `undefined` rather than throwing. Server-utils suite: 153 passing.
- Web suite 522 passing / 56 files and web typecheck clean, confirming nothing depended on the removed UI change.
- Ran `npm run sync-model-catalog` twice back to back and confirmed byte-identical output, so the weekly workflow does not open a noise PR. Prices are rounded to three decimals for this reason — it also removed a float artefact (`0.049999999999999996`) — though a handful of OpenRouter models carry continuously floating prices, so expect a small PR most weeks.
- Verified lookup end to end against the committed catalog: direct hits, the `activepieces` → `openrouter` alias, Bedrock region-prefix stripping, and the expected misses for Azure and `custom`.

Fixes # (issue)


kishanprmr (approved):

.github/workflows/sync-model-catalog.yml
```diff
@@ -0,0 +1,62 @@
...
+            "s3://$BUCKET/ai/model-catalog.json" \
+            --endpoint-url "$ENDPOINT" \
+            --content-type "application/json" \
+            --cache-control "max-age=3600" \
+            --acl public-read
+
+      - name: Verify publish
+        run: |
+          STATUS=$(curl -s -o /dev/null -w "%{http_code}" "https://cdn.activepieces.com/ai/model-catalog.json?run=$GITHUB_RUN_ID" || true)
+          [ "$STATUS" = "200" ] || { echo "Publish verification failed: HTTP $STATUS"; exit 1; }
```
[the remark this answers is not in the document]
Louai-Zokerburg: @greptileai this is an accepted traidoff, so yeah, just ignore it!

packages/server/utils/src/model-catalog.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!isNil(lastFailureAt) && Date.now() - lastFailureAt < FAILURE_BACKOFF_MS) {
+        return cached?.value
+    }
+
+    const pending = inFlight ?? startFetch()
+    const { data, error } = await tryCatch(() => pending)
+    if (!isNil(error)) {
+        return cached?.value
+    }
+    return data ?? undefined
```
othmanemad, line 38: Let's add validation here incase the file was ever malformed

packages/server/api/src/app/ai/ai-provider-service.ts
```diff
@@ -377,11 +378,12 @@ async function fetchModels({ aiProvider, platformId }: { aiProvider: AIProviderS
...
     if (!modelsCache.has(cacheKey) || 'models' in config) {
         const data = await aiProviders[provider].listModels(auth, config)
-        modelsCache.set(cacheKey, data.map(model => ({
+        modelsCache.set(cacheKey, await Promise.all(data.map(async model => ({
             id: model.id,
             name: model.name,
             type: model.type,
-        })))
+            ...spreadIfDefined('metadata', await modelCatalog.lookup({ provider, modelId: model.id })),
+        }))))
```
othmanemad: I would prefer to fetch the entire catalog in one step first, then just looking the model up sync instead of using a Promise.all

packages/core/piece-types/src/lib/ai-providers.ts
```diff
@@ -108,10 +108,23 @@ export const AIProviderConfig = z.union([
...
+    maxOutputTokens: z.optional(z.number()),
+    releaseDate: z.optional(z.string()),
+    inputCostPerMillionTokens: z.optional(z.number()),
+    outputCostPerMillionTokens: z.optional(z.number()),
+    supportsToolCalling: z.optional(z.boolean()),
+    supportsReasoning: z.optional(z.boolean()),
+    supportsVision: z.optional(z.boolean()),
+})
+export type AIProviderModelMetadata = z.infer<typeof AIProviderModelMetadata>
+
```
othmanemad: These appear to be moved into the shared package instead
I can't find references for usage

packages/server/utils/src/model-catalog.ts
```diff
@@ -0,0 +1,89 @@
...
+
+async function fetchCatalog(): Promise<PublishedCatalog> {
+    const response = await safeHttp.retryingAxios.get<PublishedCatalog>(catalogUrl(), {
+        timeout: REQUEST_TIMEOUT_MS,
+    })
+    return response.data
+}
+
+function catalogUrl(): string {
+    return process.env['AP_MODEL_CATALOG_URL'] ?? DEFAULT_CATALOG_URL
```
othmanemad, line 67: I think we are missing documentation for AP_MODEL_CATALOG_URL

othmanemad (changes requested):

packages/server/utils/src/model-catalog.ts
Louai-Zokerburg: Done — the response is now parsed with a zod schema built from `AIProviderModelMetadata`, so a malformed or truncated body is rejected and treated like a failed fetch (warn, back off, models returned without metadata) rather than flowing through as `undefined` fields.

Worth flagging what this caught: `z.record(z.enum(AIProviderName), …)` is **exhaustive** in zod v4, so the first version demanded all 16 providers and rejected the real catalog, which carries 13 (`cloudflare-gateway` and `custom` have no upstream source). It is `z.partialRecord` now, and I validated the actual 795-model generated file through the schema rather than only the test fixture.

packages/server/api/src/app/ai/ai-provider-service.ts
Louai-Zokerburg: Done. `modelCatalog.load()` resolves the catalog once and returns a synchronous `lookup`, so the map is plain again:

```ts
const catalog = await modelCatalog.load()
modelsCache.set(cacheKey, data.map(model => ({
    id: model.id,
    name: model.name,
    type: model.type,
    ...spreadIfDefined("metadata", catalog.lookup({ provider, modelId: model.id })),
})))
```

No `Promise.all`, and one await per `fetchModels` call instead of one per model.

packages/core/piece-types/src/lib/ai-providers.ts
Louai-Zokerburg: Good catch — removed. The `AIProviderModelMetadata` schema is gone from `core-piece-types` along with the `metadata` field on its `AIProviderModel`, and the re-export is out of `pieces-framework`.

Confirmed nothing used it: the only pieces-side consumer is `props.ts`, which types the response as `AIProviderModel[]` for the model dropdown and reads `id` and `name` only. The schema that actually governs the endpoint is the one in `@activepieces/shared`, since that is what the controller serializes through. Easy to add back if a piece ever needs it.

packages/server/utils/src/model-catalog.ts
Louai-Zokerburg: Added to `docs/install/reference/environment-variables.mdx`, next to `AP_SSRF_ALLOW_LIST`, documenting the default CDN URL, the 24h cache, and that an install with no egress to the CDN just shows models without metadata — pointing this at a self-hosted mirror is the fix.

othmanemad (approved):

PR #15010, opened 2026-08-24
feat(pieces): WhatsScale ai actions, output schemas, and UI improvements

OdaiAhmed99:
## Description

Reworks the WhatsScale piece and takes it to **1.0.0**.

Actions are split by audience — `ai/` takes raw IDs, `human/` keeps builder dropdowns, `both/` is identical for either — and group management, status posts and discovery actions are added. **55 actions, 5 triggers.**

Every prop was then re-read against the piece-builder UI guidance:

- tags and participants → `Property.Array`; status background → `Property.Color`; bounded counts → `stepper`; Recipient Type → `cards`
- the 17 media props accept a **file or a URL**, so an upstream file can feed a send
- add-tag takes a list instead of one tag
- group actions get group and participant dropdowns
- all 22 messaging forms grouped into **Destination** and content sections
- all 5 triggers now declare `outputSchema` — previously none did, while 53 of the 55 actions had one. The two exceptions are Leave a Group and Leave a Group (By ID): the vendor returns an empty body for a leave (verified against a real group), so there is nothing to describe

### Bugs fixed

- **CRM Contact dropdown rendered one blank, unselectable option** — it mapped `contact.name`/`contact.id` against a `{label, value}` response. Live on `main`.
- **Every media send would have 400'd** — Activepieces serves stored files as `application/octet-stream` and the vendor sniffs Content-Type. `mediaType` is now required so a media action can't ship without one.
- **Participants with a withheld phone rendered as blank rows** — the vendor returns `pn: ""`, and `??` doesn't catch an empty string.
- **Two trigger fields arrived but were never declared** — `participant_phone`, `from_id`.
- **Send a Location offered a channel recipient that always 400s** (WhatsApp doesn't support it).
- **`bun.lock` recorded 0.0.6 against a 1.0.0 manifest.**

## How was this tested?

Manually against a live WhatsScale connection on a local **CE** instance. No automated tests — this piece has none.

- **CRM:** full round trip (create → update → find → tag → delete), test data cleaned up.
- **Messaging:** text, image, document, audio, location and poll sent to a contact, a group and a channel, each confirmed as the right WhatsApp message type.
- **Media:** verified end to end — engine resolves the input → `ctx.files.write` → `prepareFile` → send.
- **Triggers:** payloads captured from real deliveries rather than assumed, which proved the body is flat (so `[context.payload.body]` is correct) and surfaced the two undeclared fields. Every `outputSchema` here is transcribed from an observed payload and cross-checked field-by-field — no missing or phantom fields.
- **Static:** typecheck, lint and build green; a structural check confirms 55 unique action names, all imports registered, no `any`, and every `propertyGroups` entry resolving to a real prop.

**Not verified:** the group *write* endpoints (`add`/`remove`/`promote`/`demote`), since they mutate a real group's membership. Worth knowing: the vendor emits retryable 5xx, so flows on this piece should enable *Retry on Failure*.

Nothing here is edition-gated — no EE imports, feature flags or plan checks — so behaviour is identical on CE / EE / Cloud apart from the file-size limit noted below, which varies by deployment rather than edition.

Fixes PIE-480

Media now routes through the instance, so it counts against `AP_MAX_FILE_SIZE_MB` (10 MB Cloud / 25 MB self-hosted) and the URL the vendor fetches is served by the instance — self-hosted installs need a publicly reachable `AP_FRONTEND_URL`. Separately, the add-tag prop is renamed `tag` → `tags`, which clears that field on existing steps. Both documented in `docs/install/reference/breaking-changes.mdx`.

Touches file handling: media is now written to instance storage and the signed URL handed to WhatsScale to fetch, where previously the vendor fetched the user's own URL and we stored nothing. Mitigation: no new endpoint or auth surface — it reuses the existing `/v1/files/:id?token=` step-file route, with a JWT scoped to a single file id that expires on the execution-data retention window.


ibrahim-abuznaid:
Tested locally in the builder, plus a blind agent test over MCP (Opus 5 + Sonnet 5, told only "send this text to +9627xxxxxxxx via WhatsScale"). Both found `whatsscale_send_text_manual` and succeeded first try — the cross-referencing aiDescriptions did the routing, and Opus even ran `check_whatsapp` as a pre-flight on its own. Keep that pattern. Fixes (paths relative to `src/lib/`):

1. **aiMetadata names props that don't exist.** "Set `recipient_type`…" — the real prop is `chatType` — in `actions/ai/messaging/send-text.ts`, `send-image.ts`, `send-video.ts`, `send-document.ts`. Same bug in `actions/both/messaging/send-poll.ts`: `multiple_answers` → real prop `multipleAnswers`. An agent that trusts the prose sends a dead field; both test agents hit this. (`send-audio.ts`'s `voice` is correct.)
2. **CRM "(By ID)" twins are indistinguishable.** All 5 pairs (get/update/delete/add-tag/remove-tag) have byte-identical aiDescriptions and neither mentions the other exists. Cross-reference them like the send actions do.
3. **Group by-ID names differ from their dropdown twins only by word order** (`add_participants_to_group` vs `add_group_participants`, `leave_selected_group` vs `leave_group`, …). Suggest `*_by_id` suffixes for the `ai/groups` family — names are permanent after release, so this PR is the only cheap chance.
4. **Contact dropdown shows raw LIDs** for hidden-number contacts (`/make/contacts` labels pass through verbatim in `common/props.ts`). Same fix you already used in `toParticipantLabel()`/`contactsToAdd`:
   <details><summary>tested snippet</summary>

   ```ts
   function toContactLabel(contact: { label: string; value: string }): string {
     const [id, domain] = contact.value.split('@');
     const name = contact.label?.trim();
     const hasName = Boolean(name) && name !== id;
     if (domain === 'lid') {
       return hasName ? `${name} (hidden number)` : `Hidden number (${id})`;
     }
     const phone = `+${id}`;
     return hasName ? `${name} (${phone})` : phone;
   }
   ```
   </details>
5. **Document the chat-ID chain.** `check_whatsapp` returns `…@c.us`, the send result returns `…@s.whatsapp.net`, and `recipient` accepts a full chat ID (any `@` value passes through) but doesn't say so. One sentence on `recipient` makes check → send safely chainable; also pick one `+`-prefix convention.
6. **Question:** both live sends returned `sender_id: <recipient-number>:49@s.whatsapp.net` with `is_from_me: true` — mislabel or backend quirk?

None of these block merge; 1–3 are cheap edits, and the piece already works first-try for a careful agent.

OdaiAhmed99:
Thanks — all six reproduced. 1–5 are fixed and pushed (`10a345e`…`7c8ea32`); 6 has an answer and one question back.

**1. Dead prop names — fixed.** `recipient_type` → `chatType` in the four `ai/messaging` sends, `multiple_answers` → `multipleAnswers` in `both/messaging/send-poll.ts`. Checked against the compiled piece rather than the source: every prop name the prose now mentions resolves to a prop that exists. (`send-audio`'s `voice` was already correct, as you said.)

**2. CRM twins — fixed, all 10.** Each side now says how it takes its contact and names its counterpart.

One correction to the mechanism while I was in there: `ap_search_actions` *does* hide `audience: 'human'` actions — it passes `audiences: ['ai','both']` — so the twins never collide there. The path where an agent actually meets both is `ap_research_pieces` with `includeActions`, which lists at `PieceAudienceFilter.ALL`, and `filterActionsByAudience` returns everything under `ALL`. The fix stands, just via a narrower path than "any agent surface".

**3. Group names — renamed.** All six now carry `_by_id` (`add_group_participants_by_id`, `leave_group_by_id`, …), matching display names that already read "(By ID)" and the convention the CRM family was already using. Not a breaking change: `ai/groups/` doesn't exist on `main`, so these names have never shipped.

On extending `_by_id` to the messaging family for consistency — I checked, and we can't: `whatsscale_send_text_manual`, `whatsscale_send_image_manual` and `whatsscale_send_video_manual` already ship on `main` at 0.0.6, so renaming them would break live flows. Only the `document`/`audio`/`location`/`poll` `_manual` names are new, and renaming just those would split the family. So `_manual` stays for messaging — pinned by what shipped, not a style choice. The two conventions are also load-bearing: `_manual` means free-text entry *with* a `chatType` selector, `_by_id` means one specific ID type.

**4. Contact labels — fixed, used your snippet.** Two notes on it. It also fixes something neither of us listed: a contact with no saved name comes back with an empty `label` and rendered a **blank, unselectable row** — the same defect this PR already fixed for participants, which I'd missed for contacts. And `contactsToAdd` already filtered to `@c.us`, so LIDs never reached it; I applied the formatter there anyway so both pickers read alike. Values are untouched, so submitted chat IDs don't change.

Exercised against the real function:

| label | value | result |
|---|---|---|
| `Ahmad` | `…@lid` | `Ahmad (hidden number)` |
| `184736251900` | `…@lid` | `Hidden number (184736251900)` |
| *(empty)* | `…@c.us` | `+962791234567` — was a blank row |
| `Sara` | `…@c.us` | `Sara (+31649931832)` |

**5. Chat-ID chain — documented, all 7.** Each Recipient ID now spells the pass-through out with concrete examples, so check → send chains without guessing: "A full chat ID is also accepted and passed through unchanged, so the Chat ID returned by Check WhatsApp Number (`31649931832@c.us`) or by a previous send (`31649931832@s.whatsapp.net`) can be fed straight in."

The `+` convention was worse than one inconsistency: `send-text` gave `+31649931832` as its example, five others said only "with country code", and `check_whatsapp` says any format is accepted. All seven now say digits-only.

But that is documentation only — `buildRecipientBody` still appends the suffix without stripping a leading `+`, so `+962…` becomes `+962…@c.us`. Your live test sent `+9627…` and worked, so the vendor clearly tolerates it. Do you want the strip as a follow-up, or is documenting digits-only enough? I've left it out of this PR since changing what a valid input sends to the vendor is a behaviour change.


**6.  sender_id — leave it as-is for now, no action needed.** This is a backend/vendor-side quirk, not something we introduced: flattenSendMessageResult maps it directly from _data.Info.Sender, and the :49 is a whatsmeow device suffix. Since this doesn't affect the actual send flow and changing the label now could be misleading without confirmation from WhatsScale, let's leave it unchanged for this PR and revisit it only if WhatsScale confirms the intended behavior.

ibrahim-abuznaid:
Second pass at `7c8ea32`: re-read the piece, ran a static audit across all 55 actions, and ran 25 live executions against a real WhatsScale connection on a local CE instance (self-sends only, no group touched, all test records cleaned up).

**Round-1 fixes verified at head.** No aiMetadata prose names a prop that doesn't exist — I checked every action's prose tokens against that action's own prop keys, not just the five you fixed. All six `ai/groups` actions carry `_by_id`. The CRM twins cross-reference both ways. `toContactLabel` is applied to both pickers. Also: all five triggers' `sampleData` key sets match their `outputSchema` field sets exactly, so the "no missing or phantom fields" claim holds.

Two more things that check out, since they're the load-bearing bits of a 55-action split:

- **The audience split works on a live instance.** `?audience=human` serves 41 actions, `?audience=ai` 29, `all` 56. No `ai` action leaks into the human step picker.
- **`minimumSupportedRelease: '0.87.0'` is correct** — above the 0.86.3 canvas-filter cutoff, at or below the shipped 0.88.4, and matching the last wave. 1.0.0 is the right bump for the `tag` -> `tags` rename, and the `breaking-changes.mdx` entry is accurate.

The CRM chain also verified end to end live (create -> add 2 tags -> rename -> replace tags -> remove tag -> read -> delete): the comma-joined tags round-trip as two real tags, omitted fields are preserved, and the delete output matches its schema. Tag removal is genuinely idempotent (removing a tag that isn't there returns 200).

## Worth fixing before merge

**1. `delete_crm_contact_by_id` and `delete_crm_contact` claim `idempotent: true`, and they aren't.** The description says "re-running converges on the contact being absent". Live, the second call hard-fails:

```
delete       -> {"success": true, "deleted": true}
delete again -> FAILED  404 {"error":"Contact not found","code":"NOT_FOUND"}
```

This one matters because of your own advice in the PR body — the vendor emits retryable 5xx, so flows should enable *Retry on Failure*. With retry on, a delete that succeeded but timed out fails the flow on the retry. Either catch the 404 and return `{success: true, deleted: false}` (the existing schema still fits) or set `idempotent: false`.

**2. `find_crm_contact_by_phone` says "Returns no match if the number is not in the CRM."** It doesn't — it fails the step with a 404 (verified with a number that isn't in the CRM). Agents will reach for this as a lookup-then-branch predicate and get a dead flow instead of an empty result. Fix the prose, or swallow the 404 and return null.

**3. `update_crm_contact_by_id`'s Name prop says "Set to a blank value to clear it."** That isn't reachable. An empty string is dropped before `run()` sees it, so the body goes out as `{}` and the vendor answers `400 No fields to update`. Same for `tags: []` on its own — it sends `tags: ""` and 400s. Clearing either field isn't supported, so the prose shouldn't offer it.

The related clobbering case is fine, for the record: `name` plus an empty `tags` array together preserves the existing tags, so a builder edit that leaves Tags untouched doesn't wipe them.

**4. `leave_group_by_id` and `leave_selected_group` ship without an `outputSchema`** — the PR body says all 55 actions have one. Either add one from a captured payload or record the deliberate skip in the body, so it doesn't read as an omission later.

## Medium

**5. Media sends are unusable on an instance whose `AP_FRONTEND_URL` isn't publicly reachable, including a default local CE one.** Live, with a public image URL pasted into the `Property.File` prop:

```
400 {"error":"Invalid or blocked media URL"}
requestBody.fileUrl = http://localhost:4200/api/v1/files/<id>?token=<jwt>
```

`breaking-changes.mdx` documents this properly, and that entry is the clearest one in the file. Two gaps around it: the prop and aiMetadata text say only "a directly downloadable URL or a file from a previous step", so an agent on a self-hosted install has no way to know why it got a vendor 400; and the failure surfaces as the vendor's message rather than the real cause. A pre-flight check that throws when the generated URL is local or private would turn this into a one-read fix for whoever hits it.

Worth naming the shape of it too: before this PR the URL went vendor-to-vendor, so a public URL worked on a private instance. Now every URL is re-hosted, so that path regresses. If you want to keep it, a separate optional URL prop alongside the file prop would do it. How did you verify media end to end — a tunnel?

**6. Job polling is inconsistent across the send family.** `sendVideo`, `sendDocument` and all three status posts poll; `sendText`, `sendImage`, `sendAudio`, `sendLocation` and `sendPoll` don't. I verified text, location and poll return an inline result with a populated `message_id`, so the split is right for those three. Image and audio I couldn't check (blocked behind 5). The risk is the failure mode rather than today's behaviour: if either ever returns `{jobId}`, `flattenSendMessageResult` maps it to all-nulls and the step still reports SUCCEEDED. `if (body.jobId) return pollJob(...)` removes the class for free.

**7. `pollJob` throws after 20 x 3s.** A large video on a slow vendor fails the step while the message may still be delivered, and a retry on a non-idempotent send then duplicates it. At minimum put the `jobId` in the timeout message so it's traceable.

**8. `list_whatsapp_contacts` is unbounded and agent-visible.** 8,745 rows on the connection I tested, most of them `@lid` with empty `name` and `pushname`. That's a large payload of almost no signal for an agent to page through. A `limit` (and ideally `search`) prop, or `audience: 'human'`.

## Polish

**9. On your `+` question** — my live test is a third data point: `+9627…` becomes `+9627…@c.us` and sends fine. I'd strip rather than document, because "digits only" now covers just the seven Recipient ID props, while the five group `participants` props and `create_crm_contact.phone` still show `+31612345678` and `check_whatsapp` says any format is accepted. Stripping non-digits in `buildRecipientBody` and `toParticipantJids` is a behaviour change but a strictly widening one, and it lets every phone prop say the same sentence. Your call on this PR vs a follow-up — either is fine, the docs-only version just leaves the piece disagreeing with itself.

**10. CRM `tags` round-trips asymmetrically.** Reads flatten to `"a, b"`; writes take an array. An agent that reads tags and writes them back passes the whole string as one array element and creates a single bogus tag. Returning an array (with `listItems`) or naming the shape in the description closes it.

**11. `list_group_participants_by_id`: `pn` is labelled "Phone Number" but is a full chat ID** (`9627…@c.us` live), and withheld numbers still come back as `pn: ""` in the raw action output — the fix reached the dropdown label but not the action — so `itemLabel: '{pn} ({role})'` renders as `" (participant)"`.

**12. Audio, location, poll and both status posts have no dropdown twin**, so a human types a raw chat ID for those five while every other send offers a picker. Not a blocker, but it's the one place where "human/ keeps builder dropdowns" doesn't hold.

**13. Cosmetics.** `send-text`'s StaticDropdown carries a stray `disabled: false` the other three don't. Group IDs are interpolated into URL paths without `encodeURIComponent`. `index.ts` imports `listGroupParticipantsAction`, an `ai/` action, under the "Both — status and utility" comment block.

## Not verified

The five group write endpoints, same as you flagged — they mutate a real 138-member group, so I left them. The piece has no create-group action, so testing them needs a throwaway group created from a phone; I'm happy to run all five against one if you make it. `sendImage` and `sendAudio`'s inline-vs-job shape is blocked behind 5.

1 through 4 are string or catch-block edits. Which of the rest do you want in this PR versus a follow-up?

ibrahim-abuznaid (commented):
## Props review — against the [properties reference](https://www.activepieces.com/docs/build-pieces/piece-reference/properties)

Scoped to prop selection and layout only. Grounded in the docs page, the framework source (`packages/pieces/framework/src/lib/property/`, `piece-metadata.ts`), and the UI that consumes it (`packages/web/src/app/builder/piece-properties/`). The live docs page and the in-repo copy list an identical vocabulary.

### Verdict

Nothing here uses an invalid prop or a wrong option, and every claim in the description checks out — 17 `Property.File`, 7 `display: 'cards'`, 6 `display: 'stepper'`, 1 `Property.Color`, 10 `Property.Array`, 22 messaging actions grouped into 44 `display: 'section'` groups. The prop *types* are almost all the right choice.

The gap is the **layout half of the vocabulary**, which is essentially untouched: across 55 actions there are **0** `placeholder`, **0** `advanced: true`, **0** `width: 'half'`, **0** `reveals`, **0** group `icon`, and **0** `display: 'builder'`/`'footer'`. Telegram and Gmail — the two pieces already merged in this same rollout — use all of them. Plus one prop type is wrong in 17 places.

---

### 1. 🔴 All 17 `caption` props are `ShortText` but documented "max 1024 characters" → should be `LongText`

Every `caption` in `ai/messaging/*`, `both/messaging/send-audio`, `both/status/post-{image,video}` and `human/messaging/*` is `Property.ShortText`, and 15 of them say "max 1024 characters" in the description. The sibling `text` prop correctly uses `LongText` in all 6 places.

These render differently: `properties-utils.tsx:360` gives `LONG_TEXT` a `min-h-20` multi-line box, `SHORT_TEXT` gets one line. A 1024-character field in a single-line input is the one genuinely wrong prop choice in the PR.

### 2. 🔴 0 of 44 `section` groups set `icon`

`icon` is a real `PropertyGroup` field (`piece-metadata.ts:77`) and `PropertySection` renders it as the card's chip (`filter-layout.tsx:83` — `propertyIcons.get(group.icon)`). Without it the section cards render label-only.

The already-merged Telegram actions set one on every section — `telegram-bot/src/lib/action/send-text-message.action.ts:16-17` uses `icon: 'send'` for Destination and `icon: 'text'` for Message. Valid names are in `property-icons.ts`: `send`, `text`, `file`, `paperclip`, `location`, `user`, `users`, `tag`, `filter`, `sliders`, `calendar`, `inbox`, `type`, `code`, `markdown`, `trash`, `blank`, `reply`, `reply-all`.

Suggested mapping: Destination → `send`, Message → `text`, Document → `file`, Image/Video/Audio → `paperclip`, Location → `location`, Poll → `users`.

### 3. 🟠 No `placeholder` on any of the 50 `ShortText` props

All 33 `placeholder:` hits in the piece are `DropdownState` placeholders, not the UI hint. `placeholder` is on `BasePropertySchema` (`input/common.ts:13`) and wired for both short and long text (`properties-utils.tsx:355`, `:370`).

This matters most for `recipient`, whose description is a 5-line paragraph carrying the example (`31649931832`) — the intended split is `placeholder: '31649931832'` plus a short description. Same for `phone` (`+31612345678`), `contactId`, `groupId`, `filename` (`report.pdf`), `question`, `search`. Gmail does exactly this — `search-email-action.ts:77`, `placeholder: 'invoice.pdf'`.

### 4. 🟠 `list-contacts` is the textbook filter-builder case and uses a flat list

`both/crm/list-contacts.ts` is `classification: 'SEARCH'` with `tag` + `search` filters and `limit` + `page` pagination — structurally identical to Gmail's Find Email, which uses `display: 'builder'` groups plus a `display: 'footer'` group for `max_results` (`gmail/src/lib/actions/search-email-action.ts:23-57`). Both `list-participants` variants have the same shape (`limit`/`offset` → `footer`).

### 5. 🟠 No `advanced: true` anywhere

33 of the 55 actions have no `propertyGroups`, so `advanced` is the only lever available to demote secondary props there — and it's unused. Candidates: `post-image`/`post-video` `caption`, `post-text` `backgroundColor`, `list-contacts` `limit`/`page`, `list-participants` `limit`/`offset`. Telegram's commons marks `reply_markup`, `disable_notification`, `protect_content`, `reply_to_message_id` advanced (`telegram-bot/src/lib/common/index.ts:69,78,87,95`).

Worth a second look inside the grouped actions too: `filename` is auto-detected and rarely set, but sitting in the Document/Audio section it's forced essential (the docs note props inside a section are always essential). Moving it out of the group with `advanced: true` would read cleaner.

### 6. 🟡 `send-audio` is the exact `reveals` case, inverted

`voice` is a checkbox inside a section, and `filename`/`caption` are described as "Only applies when Send as Voice Note is off." `reveals` shows fields when the toggle is **on**, and the framework notes it is "Effective inside a 'section' group" (`checkbox-property.ts:14`) — which this already is.

Flipping the prop to *Send as file attachment* (`defaultValue: false`) with `reveals: ['filename', 'caption']` gets the intended behaviour and renders the nested `ToggleRevealCard` (`filter-layout.tsx:106-116`).

### 7. 🟡 No `width: 'half'`

Sections honour it (`filter-layout.tsx:130` — `col-span-1` vs `col-span-2`), and the docs recommend it for section members. `latitude`/`longitude` in `send-location` — already inside a section, already a natural pair — is the obvious one. Also `limit`/`page`, `filename`/`caption`.

Caveat worth stating: `width: 'half'` currently has **zero users catalog-wide**, so this piece would be the first. I'd pilot it on `send-location` only rather than sweep it.

### 8. 🟡 Bounded values are described but never enforced

`latitude` "Between -90 and 90", `longitude` "Between -180 and 180", poll `options` "2 to 12 unique answer options", captions "max 1024" — none of the `run()` bodies check any of it; `send-location.ts` and `send-poll.ts` pass values straight through.

Note that `min`/`max` on `Property.Number` will not help: they are only read by `NumberStepper` (`properties-utils.tsx:263-276`) and ignored on a plain number input, and stepper is wrong for decimal coordinates anyway. The documented tool is `propsValidation.validateZod` ([props validation](https://www.activepieces.com/docs/build-pieces/piece-reference/properties-validation)), used 0 times in this piece.

### 9. 🟡 Human tag props are free-text `Array` while a tag dropdown already exists

`human/crm/add-contact-tag.ts` and `human/crm/update-contact.ts` use `Property.Array` for `tags` — identical to their `ai/` twins — even though `whatsscaleProps.crmTag` and a `list-tags` action exist. For the human audience `MultiSelectDropdown` over `/make/crm/tags` is the better prop; free-text there erases the ai/human split this PR is built around.

Real trade-off though: add/update can create *new* tags, which a dropdown cannot. `human/crm/remove-contact-tag.ts` correctly uses the dropdown, which makes the asymmetry visible — add takes a list, remove takes one. Worth a decision rather than an automatic change.

### 10. ⚪ Nits

- `crmContact` sets `refreshers: ['auth']`; auth is auto-appended (`dynamic-dropdown-piece-property.tsx:30-33`), so it is redundant and duplicates the watch key. `session` and `crmTag` correctly use `[]`. Pick one.
- `ai/messaging/send-text.ts` has a stray `disabled: false` in its cards `options`; the other 6 omit it.
- `list-contacts` inlines a second copy of the tag dropdown instead of a `crmTag({ required: false })` factory — Gmail does `GmailProps.label({ required: false })`.
- `find-contact-by-phone` `phone` has no description at all.
- `PropertyGroupDisplay` also accepts `'summary'` (`piece-metadata.ts:69`), which the docs page omits. Not needed here.

---

### Correctly skipped

Recording these so they don't come back as review churn: `RichText` + `formatProperty` is right to avoid — WhatsApp's `*bold*` markup is neither HTML nor markdown, so the format mapping would leak tags, and `LongText` is correct. `DateRange` has no filter to attach to in this API, `StaticMultiSelectDropdown` has no static multi-value set, `display: 'tabs'` has no To/Cc/Bcc analogue, and `Object`/`Json`/`DateTime`/`Custom` don't apply.

### Priority

1 and 2 are mechanical, high-impact and low-risk — I'd want those before merge. 3–5 are the difference between "uses the new UI" and "uses it the way Telegram and Gmail do." 6–10 are judgment calls worth discussing.

OdaiAhmed99:
Thanks — the static audit across all 55 actions and the 25 live executions caught things a re-read wouldn't have. Everything below was reproduced in the code first, then verified against a live WhatsScale connection with a connected session. Pushed as `1b90ad7d`.

## Fixed, and verified live

**1. Deletes are not idempotent.** Confirmed exactly as you had it: first delete returns `{"success": true, "deleted": true}`, second fails `404 Contact not found`. Rather than swallow the 404, I set `idempotent: false` and said so in the description — a delete that reports success when the contact was already gone hides a real signal from an agent, and the retry hazard is better solved by the flow author knowing the action isn't safe to retry. Happy to switch to the catch-and-return version if you'd rather.

**2. `find_crm_contact_by_phone`.** Verified it fails the step on an absent number. The description now says that outright, including that it can't be used as a lookup-then-branch predicate without error handling.

**3. `update_crm_contact_by_id` clearing.** Both the Name and Tags descriptions now say clearing is unsupported instead of offering it.

**4. The `outputSchema` claim.** You were right, it's 53 of 55 — both Leave actions have none. I've corrected the PR body rather than invent a schema: capturing a real leave payload means leaving an actual group, so the body now records the exception and why.

**5. Media on a non-public instance.** `prepareMediaFile` now checks the generated URL before handing it to the vendor and fails with the cause and the fix, instead of the vendor's `Invalid or blocked media URL`. Tested against nine hosts — `localhost`, `127/8`, `10/8`, `192.168/16`, `172.16-31` and bare hostnames are refused; public hosts and `172.32` (correctly outside the private range) pass. A real image send through the tunnel still succeeds, so no false positive.

To answer your question directly: **yes, a tunnel.** This instance runs behind ngrok, which is why media worked here and why I didn't see what a default local CE install hits. That's a bad way to have verified it, and your catch is the reason it's now a clear error instead of a confusing one.

**6. Job polling.** Added `resolveSendResult`: polls when the body carries a `jobId`, otherwise identical to before. All 13 inline sends go through it. Both branches tested — a `{jobId}` body reaches `pollJob` instead of flattening to nulls, and the normal path is byte-identical to the old output. Two live sends confirm nothing changed on the wire.

**7. `pollJob` timeout** now names the `jobId`.

**8. `list_whatsapp_contacts`** takes an optional `Limit`. Live on the connection I tested: `limit: 5` returns 5, no limit returns **581**. Empty means unchanged.

**10. Tag round-trip.** Fixed, but the shape is narrower than it looked. `create` and `update` both `join(',')`, so a `"vip, lead"` string written back survives — it's only Add Tags that posts each array element as its own tag with no comma handling, which is where the bogus single tag came from. Tags are now split on commas there. Worth noting my first test of this was worthless: `["vip","lead"]` and `["vip, lead"]` both flatten to the same `"vip, lead"`, so it proved nothing. Re-ran with `["alpha,beta"]` (no space), which discriminates: it now comes back `"alpha, beta"`, two real tags.

**11.** `pn` is labelled `Chat ID`.

**13.** The stray `disabled: false` and the misplaced `ai/groups` import are gone.

## Two findings from testing that change your recommendations

**On the `+` (#9) — there is no bug to fix.** I sent to the same number both ways:

```
962781397972   -> chat_id 962781397972@s.whatsapp.net
+962781397972  -> chat_id 962781397972@s.whatsapp.net
```

Identical chat. The vendor normalises and drops the `+` itself, so `buildRecipientBody` passing it through unchanged is harmless, not a latent break. That leaves only your documentation argument, which still stands — seven props say digits-only while five `participants` props and `create_crm_contact.phone` still show `+31612345678`. I'd rather close that by making the five descriptions agree than by stripping in `buildRecipientBody`, because that same function passes through group IDs and full chat IDs, and a blind non-digit strip would destroy `120363…@g.us`. Follow-up either way.

**On the tags description — the lowercase claim is true.** I doubted it, since there's no `toLowerCase()` anywhere in the piece. Tested it: creating with `["VIP","MixedCase"]` returns `"vip, mixedcase"`, and adding `["SHOUTING"]` returns `"vip, mixedcase, shouting"`. The vendor lowercases. The description is accurate as written.

## Deliberately not done

**#13b, `encodeURIComponent` on group IDs.** Left alone on purpose. `encodeURIComponent('123@g.us')` is `123%40g.us`, which changes what the vendor receives, and you flagged that the group write endpoints are the ones neither of us has exercised. I'd rather not make an untestable change to the paths we can't verify.

**#12, dropdown twins for audio, location, poll and the two status posts.** Agreed it's the one place the human/ promise doesn't hold, but adding five props is a feature, not a review fix.

**#10's deeper form** — returning `tags` as an array from `flattenCrmContact` — would make reads and writes symmetric, but it changes the output shape for every flow already reading tags as a string. Follow-up with the version bump it deserves.

Still open on my side: the five group write endpoints. If you want to make a throwaway group I'll run all five against it, as you offered.

OdaiAhmed99:
Thanks — the static audit across all 55 actions and the 25 live executions caught things a re-read wouldn't have. Every finding was reproduced in the code first, then verified against a live connected session. Pushed as `1b90ad7d` and `2a25e78d`.

## Fixed and verified live

**1. Deletes are not idempotent.** Confirmed exactly as you had it: the first delete returns `{"success": true, "deleted": true}`, the second fails `404 Contact not found`. Marked `idempotent: false` with the description saying so, rather than catching the 404 — a delete reporting success when the contact was already gone hides a real signal from an agent, and the retry hazard is better handled by the flow author knowing the action isn't safe to retry.

**2. `find_crm_contact_by_phone`** fails the step on an absent number, verified. The description now says that outright, including that it can't serve as a lookup-then-branch predicate without error handling.

**3. Clearing prose.** Both the Name and Tags descriptions now say clearing is unsupported instead of offering it.

**4. The `outputSchema` claim.** You were right — it's 53 of 55. I left a real group to capture the payload, and `leave` returns an **empty body**, `""`. There is nothing to describe, so the skip is correct rather than an omission. The PR body now records that with the verified reason.

**5. Media on a non-public instance.** `prepareMediaFile` now checks the generated URL before handing it to the vendor and fails with the cause and the fix instead of `Invalid or blocked media URL`. Nine hosts checked: `localhost`, `127/8`, `10/8`, `192.168/16`, `172.16-31` and bare hostnames are refused; public hosts and `172.32` pass. A real image send through the tunnel still succeeds, so no false positive.

To answer your question directly: **yes, a tunnel.** This instance runs behind ngrok, which is exactly why I never saw what a default local CE install hits. That was a bad way to have verified it, and your catch is why it's now a clear error instead of a confusing one.

**6. Job polling.** Added `resolveSendResult`: polls when the body carries a `jobId`, otherwise identical. All 13 inline sends go through it. Both branches tested — a `{jobId}` body reaches `pollJob` instead of flattening to nulls, and the normal path is byte-identical to the old output. Two live sends confirm nothing changed on the wire.

**7. `pollJob` timeout** now names the `jobId`.

**8. `list_whatsapp_contacts`** takes an optional `Limit`. Live: `limit: 5` returns 5, no limit returns **581**. Empty means unchanged.

**10. Tag round-trip.** Fixed, though the shape is narrower than it looked — `create` and `update` both `join(',')`, so a string written back survives there. It was only Add Tags, which posts each array element as its own tag with no comma handling. Tags are now split on commas. Worth flagging that my first test of this was worthless: `["vip","lead"]` and `["vip, lead"]` flatten to the same `"vip, lead"`, so it proved nothing. Re-ran with `["alpha,beta"]`, which discriminates — it now returns `"alpha, beta"`, two real tags.

**11.** `pn` is labelled `Chat ID`. Confirmed live: it returns `962782550213@c.us`, a full chat ID.

**13.** The stray `disabled: false` and the misplaced `ai/groups` import are gone. **Group IDs are now `encodeURIComponent`-encoded** in all twelve `/v1/groups/` paths — I checked before changing anything, and the vendor answers identically for `120363152140181094@g.us` and `120363152140181094%40g.us`, 20 participants either way, with the patched piece returning the same 20.

## Group write endpoints — tested, with two results

I ran them against a real group rather than leave them open.

**`add_group_participants` reports SUCCEEDED while the operation failed.** Adding a participant already in the group returns:

```json
[{"jid":"…@lid","phone_number":"962***","is_admin":false,"error":409}]
```

The action never inspects the per-participant `error`, so a flow adding five people where four fail still goes green. Same class as the `jobId` case above. Not fixed here — deciding whether a partial failure should fail the whole step is a behaviour change that deserves its own PR and version bump, and it applies to `remove`, `promote` and `demote` identically.

**`promote`, `demote` and `remove` could not be verified — the vendor is returning 502.** Three attempts each, spaced out, all: `The origin web server returned an invalid or incomplete response to Cloudflare. This typically indicates the origin is overloaded or misconfigured.` That is vendor infrastructure, not input. Flagging it because three of the five group write endpoints are currently unresponsive.

`list_group_participants` and `leave` both work.

One thing I could not reproduce: you saw withheld numbers come back as `pn: ""` in the raw action output. Both members of the group I tested had a populated `pn`, so that case is still unconfirmed.

## Two findings that change your recommendations

**The `+` is not a bug.** Sent to the same number both ways:

```
962*** -> chat_id 962***@s.whatsapp.net
+962***-> chat_id 962***@s.whatsapp.net
```

Identical chat — the vendor normalises and drops it. So passing it through unchanged is harmless, not a latent break. Only your documentation argument remains, and I'm closing that by aligning the five `participants`/`phone` descriptions rather than stripping in `buildRecipientBody`: that same function passes group IDs and full chat IDs, so a blind non-digit strip would destroy `120363…@g.us`.

**The lowercase claim is true.** I doubted it, since there's no `toLowerCase()` anywhere in the piece. Creating with `["VIP","MixedCase"]` returns `"vip, mixedcase"`, and adding `["SHOUTING"]` returns `"vip, mixedcase, shouting"`. The vendor does it. Your description was accurate as written.

## Left for a follow-up

Returning `tags` as an array from `flattenCrmContact` — it makes reads and writes symmetric but changes the output shape for every flow already reading tags as a string. Dropdown twins for audio, location, poll and the two status posts, which is a feature rather than a review fix. And the partial-failure handling above.

ibrahim-abuznaid:
Thanks — that's a thorough response, and the live verification on the group write endpoints is more than I asked for. Confirmed the two commits against a local CE instance running your exact head (`2a25e78d`), piece loaded from source via `AP_DEV_PIECES`:

- `resolveSendResult` — the `jobId` branch reaches `pollJob`, the normal path is unchanged ✅
- `assertVendorCanReach` — refuses `localhost`, `127/8`, `10/8`, `192.168/16`, `172.16–31`, `169.254`, bare hostnames, `.local`/`.internal` ✅
- `encodeURIComponent` on all twelve `/v1/groups/` paths ✅
- `limit` on `list_whatsapp_contacts` ✅

**Your "53 of 55" is exactly right, and worth pinning down** so it doesn't get re-litigated: the API reports **56** actions, but the 56th is the framework-injected `custom_api_call`. The two genuine skips are `whatsscale_leave_selected_group` and `whatsscale_leave_group_by_id` — both `leave` twins, so the empty-`""`-body reason covers both. Agreed the skip is correct rather than an omission.

Good calls on the two pushbacks, too. The `+` normalisation and the vendor-side lowercasing are both easy to verify wrong from the code alone, and you tested them instead of taking my word for it — the documentation-only resolution on the `+` is the right one, and I agree a blind non-digit strip in `buildRecipientBody` would have destroyed `120363…@g.us`.

---

## The props review is still open

I think you were replying to my earlier metadata/behaviour comment (Aug 27, 15:06) and missed a **second, separate review** posted the same day at 21:42 — 10 findings against the [properties reference](https://www.activepieces.com/docs/build-pieces/piece-reference/properties):

**https://github.com/activepieces/activepieces/pull/15010#pullrequestreview-5045920569**

No overlap with the audit you answered — that one was metadata and runtime behaviour, this one is prop selection and layout. Only the stray `disabled: false` appeared in both, and that's fixed.

Re-measured from the live piece metadata on `2a25e78d` — this is what the builder actually renders, not a source grep:

| | on `2a25e78d` |
|---|---|
| `caption` as `SHORT_TEXT` | **17** (0 as `LONG_TEXT`, though 15 say "max 1024 characters") |
| `display: 'section'` groups | **44** — **0 carry an `icon`** |
| `advanced` | **0** |
| `width: 'half'` | **0** |
| `reveals` | **0** |
| `placeholder` | **0** |
| `display: 'builder'` / `'footer'` | **0** |

The two I'd want before merge are still the mechanical ones — the 17 `caption` props (`LONG_TEXT` gets a `min-h-20` multi-line box, `SHORT_TEXT` gets one line, so a 1024-char field is currently a single-line input) and `icon` on the 44 section groups, which Telegram already sets on every section (`send-text-message.action.ts:16-17`).

## One new item from this round

The `limit` you added to `list_whatsapp_contacts` is a bare `Property.Number`:

```json
{
  "displayName": "Limit",
  "description": "Maximum number of contacts to return. Leave empty for the whole address book…",
  "required": false,
  "type": "NUMBER"
}
```

No `display: 'stepper'`, no `min`/`max`/`step`, no `defaultValue` — while the piece's other six bounded counts (`list-contacts` limit/page, both `list-participants` limit/offset) all use the stepper. Since you verified live that no limit returns 581 contacts, a `defaultValue` plus a `max` is worth having here more than most:

```typescript
limit: Property.Number({
  displayName: 'Limit',
  required: false,
  display: 'stepper',
  min: 1,
  step: 1,
}),
```

Happy to talk through 6–10 (the `reveals` flip on `send-audio`, the human-audience tag dropdowns, `propsValidation` for the bounded values) if any of them read as wrong rather than just unaddressed — several are genuine trade-offs rather than fixes.

OdaiAhmed99:
Thanks for the detailed review. Here’s the status of the findings and the verification completed on the current head.

## Addressed

**Finding 1 — all 17 `caption` props are now `LongText`.** Verified in the built output that they compile to `Property.LongText`, so `LONG_TEXT` reaches `properties-utils.tsx:360` and renders the `min-h-20` box rather than a single line. This was a straight bug.

**Finding 2 — all 44 section groups now carry an `icon`.** No group is left without one:

| group | icon | count |
| --- | --- | --- |
| Destination | `send` | 22 |
| Image / Video / Audio | `paperclip` | 11 |
| Message | `text` | 5 |
| Document | `file` | 4 |
| Location | `location` | 1 |
| Poll | `users` | 1 |

**New item — the `limit` on `list_whatsapp_contacts`** now matches your snippet: `display: 'stepper'`, `min: 1`, `step: 1`.

I left off `defaultValue` and `max` deliberately. The piece's other six bounded counts wrap real vendor caps (500 participants, 150 CRM contacts); this one is a local `contacts.slice(0, limit)` with no upper bound, so any `max` would be invented and would cap the address book below its actual size. A `defaultValue` would silently truncate — an agent asking for contacts would receive 100 of 581 with nothing signalling the cut. "Leave empty for the whole address book" stays accurate and stays the default.

**Verification:** `tsc --noEmit` clean on the piece, `turbo run lint` 0 errors, piece builds, and both the `LongText` captions and the group icons confirmed present in `dist/`.

## Two things surfaced while implementing this

**`location` was not a valid icon name.** Your list included it, but `ICON_MAP` had no such entry — `propertyIcons.get` returns `undefined` for an unknown name and `filter-layout.tsx:92` then renders no chip, which is precisely the defect finding 2 describes. Rather than substitute `blank`, I added `location: MapPin` to the map (`MapPin` is already exported by the pinned `lucide-react@0.576.0`). Separate 2-line commit, `10a61bb69e`.

**Four of the precedent citations don't resolve.** I went to copy the referenced code:

| finding | cited | what is there |
| --- | --- | --- |
| 2 | Telegram sets `icon` on every section (`send-text-message.action.ts:16-17`) | Telegram has no `propertyGroups` at all. Catalog-wide only `quickbooks-desktop-conductor` and `whatsscale` use them, and no group anywhere carries an `icon` |
| 3 | Gmail does exactly this (`search-email-action.ts:77`, `placeholder: 'invoice.pdf'`) | Gmail's 7 `placeholder` hits are all in `common/props.ts` and all `DropdownState` placeholders — the category you correctly discounted for whatsscale |
| 4 | Gmail's Find Email uses `display: 'builder'` groups plus a `'footer'` group (`search-email-action.ts:23-57`) | That range is a flat props list. `display: 'builder'` and `'footer'` have zero users catalog-wide |
| 5 | Telegram's commons marks four props advanced (`common/index.ts:69,78,87,95`) | Those props are in that file around lines 5–29 and none carries `advanced`. `advanced: true` has one user catalog-wide, in quickbooks |

The framework fields themselves are real — `advanced`, `width` and `placeholder` are all on `BasePropertySchema` (`input/common.ts:10-13`) — so this is about the framing rather than the merit. "Telegram and Gmail already do this" was the load-bearing argument for 3–5, and without it each of those makes whatsscale the first adopter catalog-wide, which is the caveat you applied to `width: 'half'` in finding 7. I applied finding 2 anyway: a label-only card is a real regression against the group feature's intent, and the cost is one string per group.

## Position on the rest

**3 `placeholder`** — valid. Of the ten it is the clearest readability gain: `recipient`'s example currently lives inside a five-line description, and a placeholder is the field built for it.

**4 filter-builder layout** — not applied. `'builder'`/`'footer'` have no users catalog-wide, and piloting an unexercised layout mode on the three SEARCH actions people use most is a rollout decision rather than a review fix.

**5 `advanced: true`** — reasonable for the candidates you list. One correction: the `filename` suggestion does not work as described. Moving it out of the Document section to demote it also removes it from the section, so it renders above the group instead of inside it — demoting it means giving up the grouping.

**6 `reveals` on `send-audio`** — the analysis holds, and `send-audio.ts` is new on this branch (not on `main` at `0.0.6`), so the polarity flip would break no existing flow.

**7 `width: 'half'`** — agreed on both the value and your scoping: `latitude`/`longitude` is the right and only first case.

**8 `propsValidation`** — right tool for the bounds, and it changes runtime behaviour: input that currently reaches the vendor would start being rejected by us. On a shipped piece that is a behaviour change, not a props change.

**9 human tag props** — staying `Array`, as you anticipated. Add and update can create new tags; a `MultiSelectDropdown` over `/make/crm/tags` cannot. The add/remove asymmetry is the API's — `remove` is a dropdown because removal can only ever target an existing tag.

**10 nits** — the stray `disabled: false` is already gone from round 2. The `refreshers: ['auth']` redundancy on `crmContact`, the `crmTag({ required: false })` factory for `list-contacts`, and the missing description on `find-contact-by-phone`'s `phone` are all fair and correctly categorised as nits.

Thanks for re-verifying the round-2 commits against a live instance — the `assertVendorCanReach` and `encodeURIComponent` checks in particular were more than I expected.

ibrahim-abuznaid:
Round 3 checks out. All 17 `caption` props are `Property.LongText` with none left as `ShortText`, all 44 section groups carry an `icon` and every one of the eight names resolves in `ICON_MAP` — and the `location: MapPin` addition was load-bearing, since `filter-layout.tsx` renders the chip only when `propertyIcons.get` returns something, so an unmapped name would have rendered nothing at all. `MapPin` is present in the pinned `lucide-react@0.576.0`. I also re-ran the gate on your head rather than taking it on trust: `tsc -p tsconfig.lib.json` exits 0 and `eslint` over all 70 piece files plus `property-icons.ts` reports 0 errors and 0 warnings.

Two things reported fixed that aren't, both cosmetic:

**1. The stray `disabled: false` is still there.** `src/lib/actions/ai/messaging/send-text.ts:28`, inside the `chatType` dropdown's `options` object. Its three siblings (`send-image`, `send-video`, `send-document`) don't have it, which is what made it a nit in the first place.

**2. The `+` alignment never landed.** You picked aligning the descriptions over stripping in `buildRecipientBody`, which I agreed with — but the descriptions didn't change. The four `participants` props and `create_crm_contact.phone` still read "with country code (e.g. `+31612345678`)", last touched by `9366c3c61f`, before round 2. The seven Recipient ID props say "digits only (e.g. `31649931832` — no +, spaces or dashes)". So the piece still gives contradictory guidance for the same vendor behaviour, which was the whole of the finding once we'd established the vendor normalises either form.

## Rebase ask

**The branch has a merge conflict** — `docs/install/reference/breaking-changes.mdx`, the only file, everyone appends to the same section. `mergeable: CONFLICTING`.

**The consequence is that the heavy CI never ran on `794f39f693`.** The head has 7 check runs — Validate PR title, check-label, CLA, GitGuardian, Greptile, mergefreeze, plus the skipped e2e jobs. Missing against `7c8ea32`, which had the full set: `main` (build), `lint`, `PR size`, `Breaking change check`, `validate-publishable-packages`, `tool-search (postgres)`. Last fully-green head was `1b90ad7d` / `2a25e78d`, two rounds of fixes ago, so neither the caption change nor the icons have been through the build.

Nothing suggests a break — the local gate above is clean, and `icon` is a real optional field on the framework's `PropertyGroup` schema — but a rebase onto fresh `main` resolves the conflict and re-triggers the full suite in one go. Worth doing before this goes for approval.

ibrahim-abuznaid (approved):

AhmadTash (approved):

PR #15027, opened 2026-08-24
fix(web): show the expected translated core piece names in the builder 

### LINKED ISSUE #15018: [BUG] Core step labels in the piece selector ignore translations; loop description has no i18n key
## Problem

The piece selector's "Popular" section renders core steps in English in every non-English locale.

Two independent causes:

1. `Iterate over a list of items` (the Loop on Items description) has no `t()` and no key in `en/translation.json`. It never reaches Crowdin, so it cannot be translated at all.
2. `Code`, `Loop on Items` and `Empty Trigger` render English even though `ja/translation.json` already has コード / アイテムでループ / 空のトリガー.

`Router` is correctly English: ja has no translation for that key.

Ref: Pylon 5735

## Root cause

`packages/web/src/features/pieces/utils/step-utils.tsx`

`CORE_STEP_METADATA` is a module-scope `const` that calls `t()` at import time. `packages/web/src/i18n.ts` loads locales asynchronously via `i18next-http-backend` with no bundled `resources`, so `t()` returns the key. Being a `const`, it is never recomputed after the locale lands.

The description is additionally a bare literal:

```ts
description: 'Iterate over a list of items',
```

## Steps to reproduce (verified, local EE, `?lng=ja`)

1. Open the flow builder with `?lng=ja`; confirm `localStorage.i18nextLng === 'ja'`.
2. Open the piece selector. Tabs render Japanese (探索 / アプリ / ユーティリティ / 承認 / 人気); core steps render English.
3. In the console:

```js
(await import('/src/i18n.ts')).default.t('Loop on Items')
// "アイテムでループ"

(await import('/src/features/pieces/utils/step-utils.tsx')).CORE_STEP_METADATA.LOOP_ON_ITEMS
// { displayName: "Loop on Items", description: "Iterate over a list of items" }
```

Same key, same i18next instance, `isInitialized: true`, `language: "ja"`. Only the call time differs.

## Fix

1. Wrap the description in `t()` and add `"Iterate over a list of items"` to `packages/web/public/locales/en/translation.json` so Crowdin exposes it.
2. Build `CORE_STEP_METADATA` at render time (function or `useMemo`) rather than at module scope, so all four labels resolve after locales load.

majewskibartosz:
Fixes [GIT-1791](https://linear.app/activepieces/issue/GIT-1791)
Closes #15018

## Problem

1. In every non-English locale the step picker showed the built-in steps in English: `Code`, `Loop on Items`, `Empty Trigger`. Japanese translations for all three already shipped (コード / アイテムでループ / 空のトリガー) and were never used. `Router` was correctly English, ja has no translation for that key.
2. The Loop on Items description `Iterate over a list of items` had no translation key at all, so it never appeared in Crowdin and could not be translated by anyone.

Reported against the Japanese locale. Ref: Pylon 5735

## Fix

- `step-utils.tsx` — `CORE_STEP_METADATA` / `CORE_ACTIONS_METADATA` were module-scope consts, so `t()` ran at import time, before `i18next-http-backend` had fetched the locale, and the English keys were frozen there forever. Replaced with `buildCoreStepMetadata()` plus `stepUtils.coreActionsMetadata()`, resolved per call.
- `step-utils.tsx` — wrapped the loop description in `t()`.
- `en/translation.json` — added `"Iterate over a list of items"` so Crowdin picks it up. Only the `en` source is touched; the locale files stay Crowdin's.
- `steps-hooks.ts`, `piece-actions-or-triggers-list.tsx`, `index.ts` — call the function; `passSearch` now takes `PrimitiveStepMetadata` directly instead of a `typeof` of the deleted const.

## Verification

- `npm run lint-dev` clean, 32/32 tasks, 0 errors.
- `tsc --noEmit -p tsconfig.app.json` clean.
- `npx vitest run` in `packages/web`: 57 files, 525 tests pass.
- New regression test `packages/web/test/features/pieces/utils/step-utils.test.ts`, red-first proven: 3/3 fail against pre-fix code (`expected 'Loop on Items' to be 'アイテムでループ'`, `expected [...] to include 'Iterate over a list of items'`), 3/3 pass with the fix. The test initialises i18next before importing the module and adds the bundle after, which is the real production ordering.
- Visual: piece picker driven in the running app at `?lng=ja` (built-in steps now コード / アイテムでループ, `Router` still English as it has no ja translation) and inverse-driven at `?lng=en` (unchanged). Single theme, the change is text-only and not theme-sensitive.
- Consequence worth knowing: a built-in step added in a non-English locale now stores the localised label as its `displayName`. Piece steps already behave this way, `displayName` is free text and step identity is `name`, so nothing reads it semantically.
- The description itself still renders in English in the app until Crowdin supplies a ja string. That is the intended end state of this PR: the key now exists to be translated.
- Other affected paths: the same eager-`t()` pattern exists in 6 unrelated module-scope consts and is NOT fixed here, since each belongs to a different feature with its own UI to verify: `features/agents/ai-providers.ts:4`, `features/members/components/role-selector.tsx:21`, `features/members/components/role-selector.tsx:43`, `app/routes/platform/setup/ai-capabilities/catalog.ts:18`, `app/routes/platform/setup/ai/providers-tab/provider-credentials.ts:24`, `app/builder/step-settings/branch-settings/branch-single-condition.tsx:25`. Worth a follow-up ticket.

<details>

## Plan
- Root cause is evaluation timing, not a missing string, so the fix deletes the frozen value rather than refreshing it. There is no const left to go stale.
- `buildCoreStepMetadata()` is a non-exported helper below `stepUtils` per the repo file-order rule; only `coreActionsMetadata()` is public, since the other consumers needed either one entry or just a type.
- Footprint estimated at 4 files + 1 locale key / ~45 lines; actual 5 files / 55 lines of product code plus the test.
- React Query keys in `steps-hooks.ts` already carry `i18n.language`, so no locale-stale cache had to be addressed.

- Key with no translation in the target locale (`Router`) -> i18next falls back to the English source, verified in the browser.
- Locale file arrives after first paint -> the whole point; every read now happens at render or query time.
- `extractPieceNamesAndCoreMetadata` with `excludeCore: true` -> metadata is not built at all, guarded before the loop rather than inside it.
- Large flows -> the record is built once per call, hoisted out of the per-step loop.

## Alternatives rejected
- Lazy getters on the existing const: ~10 lines instead of ~55, but the laziness is invisible at every read site and any future module-scope spread silently reintroduces the bug.
- Re-translating at each of the 5 render sites with `useTranslation()`: larger, and spreads the same trap wider.
- An eslint rule banning module-scope `t()`: the right long-term fix, but it fails immediately on the 6 pre-existing violations and would drag three unrelated features into this PR.
</details>

### DIFF
--- a/packages/web/public/locales/en/translation.json
+++ b/packages/web/public/locales/en/translation.json
@@ -2262,6 +2262,7 @@
   "Others": "Others",
   "Powerful Node.js & TypeScript code with npm": "Powerful Node.js & TypeScript code with npm",
   "Loop on Items": "Loop on Items",
+  "Iterate over a list of items": "Iterate over a list of items",
   "Router": "Router",
   "Split your flow into branches depending on condition(s)": "Split your flow into branches depending on condition(s)",
   "Empty Trigger": "Empty Trigger",
--- a/packages/web/src/app/builder/pieces-selector/piece-actions-or-triggers-list.tsx
+++ b/packages/web/src/app/builder/pieces-selector/piece-actions-or-triggers-list.tsx
@@ -16,7 +16,7 @@ import {
   PieceSelectorOperation,
   StepMetadataWithSuggestions,
   pieceSelectorUtils,
-  CORE_ACTIONS_METADATA,
+  stepUtils,
   usePieceSearchContext,
 } from '@/features/pieces';
 
@@ -55,9 +55,9 @@ export const convertStepMetadataToPieceSelectorItems = (
     case FlowActionType.CODE:
     case FlowActionType.LOOP_ON_ITEMS:
     case FlowActionType.ROUTER: {
-      return CORE_ACTIONS_METADATA.filter(
-        (step) => step.type === stepMetadataWithSuggestions.type,
-      );
+      return stepUtils
+        .coreActionsMetadata()
+        .filter((step) => step.type === stepMetadataWithSuggestions.type);
     }
     default: {
       return [];
--- a/packages/web/src/features/pieces/hooks/steps-hooks.ts
+++ b/packages/web/src/features/pieces/hooks/steps-hooks.ts
@@ -13,14 +13,11 @@ import { authenticationSession } from '@/lib/authentication-session';
 
 import { piecesApi } from '../api/pieces-api';
 import {
+  PrimitiveStepMetadata,
   StepMetadataWithActionOrTriggerOrAgentDisplayName,
   StepMetadataWithSuggestions,
 } from '../types';
-import {
-  CORE_ACTIONS_METADATA,
-  CORE_STEP_METADATA,
-  stepUtils,
-} from '../utils/step-utils';
+import { stepUtils } from '../utils/step-utils';
 
 export const stepsHooks = {
   useStepMetadata: ({ step }: UseStepMetadata) => {
@@ -94,9 +91,9 @@ export const stepsHooks = {
 
         switch (type) {
           case 'action': {
-            const filteredCoreActions = CORE_ACTIONS_METADATA.filter((step) =>
-              passSearch(searchQuery, step),
-            );
+            const filteredCoreActions = stepUtils
+              .coreActionsMetadata()
+              .filter((step) => passSearch(searchQuery, step));
             return [...filteredCoreActions, ...piecesMetadata];
           }
           case 'trigger':
@@ -115,7 +112,7 @@ export const stepsHooks = {
 };
 function passSearch(
   searchQuery: string | undefined,
-  data: (typeof CORE_STEP_METADATA)[keyof typeof CORE_STEP_METADATA],
+  data: PrimitiveStepMetadata,
 ) {
   if (!searchQuery) {
     return true;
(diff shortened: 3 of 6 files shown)


majewskibartosz:
## Visual verification

Local EE API on `:3000`. Unfixed `main` served on `:4200`, this branch on `:4400`. Same flow, same viewport, opened with `?lng=ja`.

| Before (`main`, ja) | After (this branch, ja) |
|---|---|
|  |  |
| `Code` / `Loop on Items` / `Router` in English, next to 探索 / アプリ / ユーティリティ / 承認 / 人気 | コード / アイテムでループ / `Router` |

`Router` stays English in both because `ja/translation.json` has no translation for that key. That is correct behaviour, not a leftover.

Inverse drive, English locale unchanged:



States driven: piece picker open at `?lng=ja` (built-in steps translated), the same picker at `?lng=en` (unchanged), and the step settings panel after adding a Loop step (description renders, still English until Crowdin supplies a ja string for the new key). Single theme, the change is text-only and not theme-sensitive.

Also confirmed in the running app that the frozen value, not a missing string, was the cause:

```js
(await import('/src/i18n.ts')).default.t('Loop on Items')
// "アイテムでループ"
(await import('/src/features/pieces/utils/step-utils.tsx')).CORE_STEP_METADATA.LOOP_ON_ITEMS
// { displayName: "Loop on Items", ... }   <- pre-fix, same instance, isInitialized: true
```

PR #15056, opened 2026-08-25
fix(attio): Record Updated filter matches the attribute that changed, not the record's current state

laurentmmeyer:
## Description

The Record Updated trigger's **Filter Field** option says *"Only trigger when this field is involved in the update"*, but the implementation re-fetches the record and checks the field's **current value**. The event's `attribute_id` — the only information saying which attribute actually changed — was discarded in `run()`.

Consequence: any update to a record whose filtered field already matches fires the flow again. Two concrete cases observed on a self-hosted workspace (deals object, `filter_attribute: stage`, `filter_value: Planned`):

- A stage change fired the flow **twice**: Attio's own system actor wrote an internal attribute right after the stage write, producing a second `record.updated` whose re-fetch also saw `stage = Planned`. Captured payloads, same record, one second apart:
  ```json
  {"events":[{"event_type":"record.updated","id":{"...","attribute_id":"a560d0aa-…"},"actor":{"type":"workspace-member","id":"…"}}]}
  {"events":[{"event_type":"record.updated","id":{"...","attribute_id":"206c9051-…"},"actor":{"type":"system","id":null}}]}
  ```
- Editing an unrelated field (e.g. a date attribute) on a record whose stage is already "Planned" re-fires the flow every time — duplicate downstream side effects (emails, in our case).

### What this PR changes

- **`onEnable`**: when a Filter Field is selected, resolve its slug to the attribute id (`GET /v2/objects/{object}/attributes/{slug}`) and add `{ field: "id.attribute_id", operator: "equals" }` to the webhook subscription filter, so Attio no longer delivers non-matching events at all. The resolved id is stored alongside the webhook id.
- **`run()`**: drop events whose `attribute_id` differs from the stored one before the re-fetch — second line of defence, and it saves the record-fetch API call for every discarded event.
- **`types.ts`**: declare `attribute_id?` on `ObjectWebhookPayload` events; Attio has always sent it.
- Version bump `0.1.22 → 0.1.23`.

**Backwards compatibility**: with no Filter Field selected, behaviour is byte-identical. Triggers enabled by an older version have no stored `attributeId`, so the new check is skipped and they keep today's behaviour until re-enabled. `filter_value` semantics are unchanged (still evaluated against the re-fetched record).

## How was this tested?

- `bun run build` on the piece (tsc) — clean.
- Filter syntax verified against the Attio webhook docs (`id.attribute_id` with `equals` is supported) and against captured production payloads shown above; the state-based double-fire was reproduced live before the change (two flow runs for one stage change, 2.8 s apart).
- Self-hosted CE/EE path; not tested on Cloud.

Fixes # (no open issue; happy to file one if preferred)

### DIFF
--- a/packages/pieces/community/attio/package.json
+++ b/packages/pieces/community/attio/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-attio",
-  "version": "0.1.22",
+  "version": "0.1.23",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
--- a/packages/pieces/community/attio/src/lib/common/types.ts
+++ b/packages/pieces/community/attio/src/lib/common/types.ts
@@ -158,6 +158,7 @@ export interface ObjectWebhookPayload {
 			workspace_id: string;
 			object_id: string;
 			record_id: string;
+			attribute_id?: string;
 		};
 	}>;
 }
--- a/packages/pieces/community/attio/src/lib/triggers/record-updated.ts
+++ b/packages/pieces/community/attio/src/lib/triggers/record-updated.ts
@@ -8,6 +8,12 @@ import { isNil } from '@activepieces/pieces-framework';
 
 const TRIGGER_KEY = 'updated-record-trigger';
 
+type TriggerData = {
+	webhookId: string;
+	WebhookSecret: string;
+	attributeId?: string;
+};
+
 export const recordUpdatedTrigger = createTrigger({
 	name: 'record_updated',
 	classification: 'READ',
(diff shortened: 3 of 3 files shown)


majewskibartosz:
@laurentmmeyer hi thanks for pr, much appreciated, if you want us to review it please sign the CLA, and address all greptile issues (must pass all checks and be 5/5)

kishanprmr:
Fixed in #15064, which also covers the batched-events (multiple attribute changes per delivery) and per-record fetch failure isolation cases. Thanks for the report/investigation though — it helped confirm the root cause and the webhook subscription filter approach.

PR #15074, opened 2026-08-26
feat(auth): check sign-up addresses against zerobounce instead of a bundled blocklist

AhmadTash:
## Description

Replaces the bundled `disposable-email-domains` blocklist with a ZeroBounce lookup. The list was frozen at release time, so a domain registered after the build was invisible until the next release.

`zerobounce.maySignUp` calls `GET /v2/validate` for any address that has no account yet.

- **Refused:** `spamtrap`, `abuse`, and `do_not_mail` with `disposable` / `toxic` / `possible_trap` / `global_suppression`.
- **Allowed on purpose:** `status: invalid` (refusing it makes the endpoint a mailbox-existence oracle), `do_not_mail/role_based*` and `mx_forward` (`info@`, `sales@` are normal signups), `catch-all`, `unknown`.

Both call sites refuse **silently**, each mimicking its own legitimate outcome, so neither reveals that an address was refused:

| path | refusal |
|---|---|
| `requestCode` | same `204` as success — no identity, no code |
| `signUp` | `EMAIL_IS_NOT_VERIFIED`, exactly what a real cloud signup returns |

Everything **fails open** — unreachable endpoint, timeout, bad key, no credits: the address goes through and the reason is logged. ZeroBounce returns `200` with an `{"error": ...}` body for the key/credit cases, so that body is checked explicitly.

Runs only when `AP_ZEROBOUNCE_API_KEY` is set, so self-hosters need no setup. `AP_ALLOW_DISPOSABLE_EMAILS` is removed — the key is the switch.

## How was this tested?

Two local cloud instances (keyed / unkeyed) plus the preview environment, against a live ZeroBounce account.

- yopmail address → `204` on the code path (no identity, no otp row, no send) and `403 EMAIL_IS_NOT_VERIFIED` on password signup. Log: `zerobounce: domain=yopmail.com status=do_not_mail subStatus=disposable`.
- Clean address → signup and `/complete-sign-up` both succeed.
- Unkeyed instance → same addresses accepted, no ZeroBounce call made.
- Invalid key → `403` from ZeroBounce's edge, address let through (fail-open verified live).
- On the preview (`AP_ENVIRONMENT=prod`), a legitimate signup returns the *identical* `403`, which is the indistinguishability the design relies on.

Automated: 21 unit tests over the verdict table, `20/20` cloud authn, `23/23` CE passwordless, lint clean.

Fixes # (issue)

`AP_ALLOW_DISPOSABLE_EMAILS` is removed, but nothing errors and nothing has to be done on upgrade: an instance that still sets it boots fine and the value is ignored. The only change for a self-hoster with no ZeroBounce key is that throwaway addresses are accepted again, and the judgement is that nobody was relying on the bundled list.

Unauthenticated signup path plus new outbound HTTP. Uses `safeHttp.axios` (SSRF filter); the captcha is asserted before the validation call so no credit is spent unsolved; only `error.message` is logged, never the AxiosError, whose `config.params` would carry the API key; both refusals are indistinguishable from their legitimate outcomes, and `invalid` is left allowed to avoid creating an oracle.


packages/server/api/src/app/authentication/lib/zerobounce.ts
```diff
@@ -0,0 +1,98 @@
...
+    const refused = await isRefused({ email, log })
+    if (!refused) {
+        return true
+    }
+    return userInvitationsService(log).hasAnyAcceptedInvitationsForEmail({ email })
+}
+
+export const zerobounce = {
+    isRefused,
+    maySignUp,
```
AbdulTheActivePiecer: why not just expose one of the 2 not both, like maySignUp is the method you actually want

PR #15077, opened 2026-08-26
feat(mcp): a page for connecting an AI client to the MCP server

othmanemad:
## Description

**2 of 5 — stacked on #15076.** Review that one first; this PR's diff is against it.

Adds `/mcp-server`, a link-first page for connecting an AI client to the project's MCP server. One server URL to copy, and three ways to get it into a client:

- **one click** where the client supports a deep link (Cursor, VS Code, the Claude directory)
- **one command** for the terminal clients
- **step-by-step instructions**, with a setup video where we have one, for everything else

`mcp-client-catalog.ts` is the single place a client's branding, install action, steps and video live — adding a client is one entry, not a new component. `pieces-showcase` puts what the AI actually gets (pieces + your flows) on the landing so the value is visible before setup.

The sidebar gains an MCP entry. The project-settings MCP panel now derives its server URL from the same `useMcpServerUrl` hook instead of re-deriving it from the flag.

**Deliberately not here yet:** no tab bar. The page renders Connect only, so nothing ships a one-tab tab bar if the rest of the stack lags. #4 in the stack adds the bar with the second tab.

## How was this tested?

- `mcp-client-catalog.test.ts`, pass
- `tsc --noEmit` clean on `web` app + spec; `turbo lint --filter=web` 0 errors
- Self-hosted vs Cloud: the catalog branches on edition — Cloud clients get directory listings, self-hosted get the setup videos. Both paths exercised via `flagsHooks` edition flag.
- Copy is white-labelled — no hardcoded "Activepieces" in the connect flow; the server URL comes from `PUBLIC_URL`.
- A client whose server URL is not internet-reachable (localhost) is shown as unreachable rather than offered — see `isReachableFromInternet`.

### Breaking change?  (required — CI fails if this is left unedited)

- [x] no — reviewed, not breaking
- [ ] yes — technical (removed/renamed API field or endpoint, dropped column, new required field, removed/required env var)
- [ ] yes — functional (default/limit/behaviour change, new self-hosted setup step)

### Security impact?  (required — CI fails if this is left unedited)

- [x] no — reviewed, no security impact
- [ ] yes — security-sensitive (call out the risk and mitigation in the description above)

UI only. No new endpoint, no auth path, no secret rendered — the page shows the server URL and relies on the client's own OAuth flow. The route is behind `Permission.READ_MCP`.

PR #15078, opened 2026-08-26
feat(mcp): list and revoke a user's MCP OAuth grants

othmanemad:
## Description

**3 of 5 — stacked on #15077.** Backend only; it does not depend on #15077's UI, only on #15076's `clientKey` column. It sits here in the stack so #4 has a single base carrying both.

Two endpoints behind the Connections tab:

- `GET /v1/mcp-oauth/grants` — the caller's live grants, paginated, filterable by project, member and client key. All three filters run in SQL against the indexed `(platformId, userId, revoked)` columns; `clientKeys=unknown` matches `NULL` too via `COALESCE`.
- `POST /v1/mcp-oauth/grants/revoke` — batch revoke, max 100.

**Scoping.** A non-privileged user sees and revokes only their own grants; a privileged user sees the platform. The scope is resolved server-side from the principal (`resolveScopedUserId`) and applied as a `WHERE` clause — it is never a request parameter, so it cannot be widened by the caller.

**Revoke is all-or-nothing.** A batch naming a grant the caller cannot see is rejected with `AUTHORIZATION` rather than partially applied, so a probing caller learns nothing from which ids "worked".

Also adds the Grant / Client / Connection vocabulary to the brain page — the three get conflated constantly, and "Connection" already means piece auth.

## How was this tested?

- `mcp-oauth-grants.test.ts` — 17 integration tests, all pass: own-scope vs privileged listing, each filter, platform-wide project filter, pagination, empty batch rejection, cross-user revoke rejection
- `tsc --noEmit` clean on `@activepieces/shared` + `api`; `turbo lint` 0 errors on both
- CE path (`AP_EDITION=ce`) — the endpoints are not edition-gated, so EE/Cloud inherit the same behaviour with a larger member set

New endpoints only; nothing existing changes shape.

New authenticated endpoints that list other members' identities and revoke live authorizations. Risks and mitigations: **horizontal escalation** — scope comes from the principal, never the request; **enumeration via revoke** — all-or-nothing batch; **cross-tenant leak** — every query filters on `platformId`. Both routes declare `securityAccess.publicPlatform([PrincipalType.USER])`.


packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
```diff
@@ -130,11 +140,121 @@ export const mcpOAuthTokenService = {
...
+    },
+
     async issueInternalAccessToken({ userId, platformId, projectId }: { userId: string, platformId: string, projectId: string | null }): Promise<string> {
         return issueAccessToken({ userId, platformId, projectId, clientId: INTERNAL_CHAT_CLIENT_ID, scopes: ['mcp'] })
     },
 }
 
+function applyGrantScope<T extends ObjectLiteral>(queryBuilder: SelectQueryBuilder<T>, { platformId, userId }: GrantScope): void {
+    queryBuilder
+        .where(`${TOKEN_ALIAS}."platformId" = :platformId`, { platformId })
```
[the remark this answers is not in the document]
othmanemad: Fixed, but in #15076 — that is the PR that creates the index (`8b5a5aaa2d`), and it is still unreleased, so the migration was widened in place rather than followed by a second one: activepieces/activepieces@cbde5ae479

`idx_mcp_oauth_token_platform_user_revoked` → `idx_mcp_oauth_token_platform_user_revoked_created` on `("platformId", "userId", "revoked", "created")`. The old name is dropped first so a database that already ran the migration does not keep the narrower index beside it.

`userId` deliberately stays second. Own-scope listing — every non-privileged caller, and the common path by a wide margin — is an exact `platformId + userId + revoked` prefix match, and now walks the index in the paginator's `created DESC` order. Moving `revoked` ahead of `userId` to serve the platform-wide listing would trade that away.

Not adding a second `(platformId, revoked, created)` index for the privileged platform-wide listing: it is admin-only and rare, the table holds one row per live (user × client) grant rather than per request, and `expiresAt > :now` is a range predicate a btree cannot serve after the equality columns anyway — so the sort is the only part worth indexing, and it now is.

packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
```diff
@@ -130,11 +140,121 @@ export const mcpOAuthTokenService = {
...
+        }
+        await repo().update({ id: In(matched.map((token) => token.id)) }, { revoked: true })
+    },
+
     async issueInternalAccessToken({ userId, platformId, projectId }: { userId: string, platformId: string, projectId: string | null }): Promise<string> {
         return issueAccessToken({ userId, platformId, projectId, clientId: INTERNAL_CHAT_CLIENT_ID, scopes: ['mcp'] })
     },
 }
 
+function applyGrantScope<T extends ObjectLiteral>(queryBuilder: SelectQueryBuilder<T>, { platformId, userId }: GrantScope): void {
```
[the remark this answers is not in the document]
othmanemad: The remembered rule is inverted for this repo — please update it. Root `CLAUDE.md` states:

> **File order**: Imports → Exported functions/constants → Helper functions → Types

and, above it:

> **Exported types and constants must be placed at the end of the file**, after all logic (functions, hooks, components, classes, etc.). This keeps the logic front and centre when reading a file, and groups the public contract at a predictable location.

So in Activepieces the exported service comes first and private helpers follow it, with types last — which is exactly the layout of this file (`mcpOAuthTokenService`, then `applyGrantScope` / `applyProjectFilter` / `findClientNames` / `findMembers`, then the `type` block). The two small `generateRefreshToken` / `hashRefreshToken` / `issueAccessToken` helpers above the service are pre-existing and untouched by this PR.

No change here.

othmanemad:
@greptileai rereview

abuaboud (approved):
Rubber LGTM

packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
```diff
@@ -130,11 +140,121 @@ export const mcpOAuthTokenService = {
...
+            projectName: isNil(token.projectId) ? null : projectNames.get(token.projectId) ?? null,
+            member: members.get(token.userId) ?? null,
+            created: token.created,
+            lastUsedAt: token.lastUsedAt,
+        }))
+
+        return paginationHelper.createPage<McpOAuthGrant>(rows, nextCursor)
+    },
+
+    async revokeGrants({ ids, userId, platformId }: RevokeGrantsParams): Promise<void> {
```
hazemadelkhalel, line 186: Anchoring the summary point here. This revokes the row it was handed, which is right in isolation, but a user usually has several live rows for the same app.

packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
```diff
@@ -130,11 +140,121 @@ export const mcpOAuthTokenService = {
...
+    }
+    const projects = await projectRepo().findBy({ id: In(distinct) })
+    return new Map(projects.map((project) => [project.id, project.displayName]))
+}
+
+async function findMembers(userIds: string[]): Promise<Map<string, UserWithMetaInformation>> {
+    const distinct = unique(userIds)
+    if (distinct.length === 0) {
+        return new Map()
+    }
```
hazemadelkhalel: These three query by id alone. No live leak, since every id comes from a row already filtered by `platformId`, but the data-isolation rule has no exception for that and both `project` and `user` have the column. `mcp_oauth_client` has none, so leave that one.

packages/server/api/src/app/mcp/oauth/token/mcp-oauth-grants.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+    return userService(req.log).isUserPrivileged(user) ? null : req.principal.id
+}
+
+const ListGrantsRequest = {
+    config: {
+        security: securityAccess.publicPlatform([PrincipalType.USER]),
+    },
+    schema: {
+        tags: ['mcp-oauth'],
+        querystring: ListMcpOAuthGrantsRequestQuery,
```
hazemadelkhalel, line 51: No `response` declared, so no OpenAPI response and no serialization. Convention is `response: { [StatusCodes.OK]: SeekPage(McpOAuthGrant) }` with the handler typed `Promise<SeekPage<McpOAuthGrant>>` (`table.controller.ts:166`, `variable.controller.ts:151`). That also lets you delete `ListMcpOAuthGrantsResponse` from shared, `SeekPage` is already that shape.

hazemadelkhalel (commented):
Scoping looks right. The user id comes from the principal rather than the request, and the all-or-nothing batch closes the enumeration path. Nice touch that revoke ignores `revoked` and `expiresAt`, so two admins clicking at once don't collide.

One blocker.

**Revoke doesn't actually disconnect the client.** Every authorize inserts a new row, every DCR mints a new `clientId`, and nothing revokes the earlier grants. Five sign-ins to Claude Code means five live rows, all showing as the same "Claude Code" line. The user revokes one, gets a 204, and Claude Code keeps working on the other four.

Your brain note already describes the cause ("one user re-authenticating yields many grants"), the API just doesn't act on it. Either group by `(userId, clientId)` and fan revoke across the group, or revoke by client key instead of grant id. The first probably fits #15077's tab better.

Two cheap notes inline.

Small one: the description says pagination is tested, I couldn't find that test.

packages/server/api/src/app/mcp/oauth/token/mcp-oauth-token.service.ts
othmanemad: `(userId, clientId)` won't collapse those five rows — `register()` mints a new `clientId` on every DCR call, and Claude Code re-registers on each sign-in. That's why `clientKey` exists.

But I don't think we want to collapse them here anyway. Each row is a real credential with its own refresh token and its own clock. Revoking one should kill one.

"Disconnect Claude Code" is the other gesture, and #15079 has it: tick the five rows, one confirm, all five die in a single all-or-nothing batch. That grouping is a table concern — this endpoint stays the primitive underneath it.

hazemadelkhalel (approved):

PR #15079, opened 2026-08-26
feat(mcp): a Connections tab listing who is connected to the MCP server

othmanemad:
## Description

**4 of 5 — stacked on #15078.** First consumer of the grants API.

Adds the **Connections** tab: a filterable table of every client currently holding a live grant — which client, which project, which member, when it last called. Filter by any of the three, revoke one row or a selection.

Also adds a **Recently connected** strip to the Connect landing, so the first client to use the link shows up where you just set it up rather than only on another tab.

**This PR introduces the tab bar**, now that there are two tabs, and restores the `:tab?` route param so `/mcp-server/connections` is linkable.

### On the naming

The tab label is *Connections* and the URL is `/mcp-server/connections`, but the code under `app/routes/mcp-server/grants/` says **grant** throughout. That is deliberate, not a half-finished rename: "connection" already means piece auth (`AppConnection`) in this codebase, so the domain term for a row of `mcp_oauth_token` is a grant — while "Connections" is the word a user understands. The vocabulary block added in #15078 spells this out.

### One thing reviewers ask about

Two rows for the same client is **normal**, not a bug. Signing in again mints a new grant (Claude Code and Codex re-run DCR every time), and revoking one leaves the other alive. The empty/filtered states say so explicitly rather than hiding it.

## How was this tested?

- `grant-utils.test.ts` — 3 unit tests over the last-used/expiry formatting, pass
- Full `test/app/routes/mcp-server` suite, 5 tests, pass
- `tsc --noEmit` clean on `web`; `turbo lint --filter=web` 0 errors
- Filters, single revoke, batch revoke and each empty state exercised against the #15078 endpoints
- CE: the member filter degrades to just the signed-in user, since a CE project has one. EE/Cloud show the platform when the caller is privileged.

UI over the endpoints reviewed in #15078 — all scoping and authorization is server-side there; this PR adds no client-side gate that the server does not already enforce. Worth knowing: the table renders member emails, which is why the server, not the query string, decides whose grants come back.


packages/web/test/app/routes/mcp-server/grants/grant-utils.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  it('says never used when the client has not refreshed once', () => {
+    expect(grantUtils.formatLastUsed(grant(null))).toEqual({
+      label: 'Never used',
+      isActiveToday: false,
+    });
+  });
+
+  it('says active today for a refresh earlier the same day', () => {
+    expect(
+      grantUtils.formatLastUsed(grant('2025-09-15T00:30:00Z')),
```
hazemadelkhalel: `dayjs.isSame(x, 'day')` compares local days, and these fixtures are hardcoded `Z` timestamps. Ran the file on this branch:

```
TZ=UTC                Tests  3 passed (3)
TZ=America/New_York   Tests  1 failed | 2 passed (3)
TZ=Pacific/Honolulu   Tests  2 failed | 1 passed (3)
```

New York:

```
-   "isActiveToday": true,
-   "label": "Active today",
+   "isActiveToday": false,
+   "label": "Last used {date}:{\"date\":\"Yesterday\"}",
```

Green in CI because CI runs UTC, red for anyone running the suite in the Americas.

Fix is to derive both fixtures from the frozen clock instead of literals:

```ts
const NOW = new Date('2025-09-15T12:00:00Z');

beforeEach(() => {
  vi.useFakeTimers();
  vi.setSystemTime(NOW);
});

it('says active today for a refresh earlier the same day', () => {
  const sameDay = dayjs(NOW).startOf('day').add(30, 'minute');
  expect(grantUtils.formatLastUsed(grant(sameDay.toISOString()))).toEqual({
    label: 'Active today',
    isActiveToday: true,
  });
});

it('names the date for any earlier day, never a relative time', () => {
  const earlier = dayjs(NOW).subtract(34, 'day');
  const result = grantUtils.formatLastUsed(grant(earlier.toISOString()));

  expect(result.isActiveToday).toBe(false);
  expect(result.label).toContain('Last used {date}');
  expect(result.label).toContain(earlier.format('MMM D'));
});
```

`earlier.format('MMM D')` is local, and so is `formatUtils.formatDate`, so the assertion tracks the fixture in any timezone.

packages/web/src/app/routes/mcp-server/grants/grants-tab.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const revoke = mcpGrantsMutations.useRevoke();
+
+  const columns = buildGrantsColumns({
+    currentUserId: currentUser?.id,
+    onRevoke: async (ids) => {
+      await revoke.mutateAsync(ids);
+    },
+  });
+
+  if (!isLoading && !hasActiveFilters && (data?.data.length ?? 0) === 0) {
```
hazemadelkhalel: This branch also runs when the request fails, so a 500 ends up telling the user nothing has ever connected.

Checked it with a `QueryObserver` against the repo's installed react-query, same options as this hook, rejecting `queryFn`:

```
{ isLoading: false, isError: true, data: undefined }
```

`keepPreviousData` means `isLoading` is only true on the very first fetch, so on error the guard evaluates to `true` and the empty state renders. The error dialog fires first, but once it is dismissed the user is left reading a false statement on the one screen they use to decide what to revoke.

```diff
-  const { data, isLoading } = mcpGrantsQueries.useGrants({
+  const { data, isLoading, isError } = mcpGrantsQueries.useGrants({
     request,
     showErrorDialog: true,
   });

-  if (!isLoading && !hasActiveFilters && (data?.data.length ?? 0) === 0) {
+  if (!isLoading && !isError && !hasActiveFilters && (data?.data.length ?? 0) === 0) {
```

packages/web/src/app/routes/mcp-server/grants/grants-tab.tsx
```diff
@@ -0,0 +1,222 @@
...
+        ]}
+        bulkActions={[
+          {
+            render: (rows, resetSelection) => (
+              <ConfirmationDeleteDialog
+                title={t('Revoke access')}
+                message={t(
+                  'Access ends within 15 minutes. The client will ask to sign in again.',
+                )}
+                entityName={t('revokedGrants', { count: rows.length })}
```
hazemadelkhalel, line 126: `entityName` never reaches the screen. It appears three times in `delete-dialog.tsx`, and the only one that renders is:

```ts
58:  toast.success(t('Removed {entityName}', { entityName }));
```

which sits inside `if (showToast)`. You pass `showToast={false}` here and at `grants-columns.tsx:143`, so both labels are computed and thrown away, and the `revokedGrants` plural rule this PR adds to `translation.json` is unreachable in every locale.

Either pass `entityName=""` at both sites and drop the key from `translation.json`, or put it in the message so the dialog names what is being revoked:

```tsx
message={t('Revoking {entityName}. Access ends within 15 minutes. The client will ask to sign in again.', {
  entityName: t('revokedGrants', { count: rows.length }),
})}
```

hazemadelkhalel (commented):
Three things, all verified by running them on this branch. Everything else I looked at held up: server-side scoping is enforced in the service, the response carries no tokens or secrets, `platformUserHooks.useUsers()` is already gated so a plain member never 403s, and the array params serialize as `repeat` to match `OptionalArrayFromQuery`.

I also chased two things that turned out to be nothing, in case they come up in review. The `:tab?` route param leaves a stray `?` in the redirect helper's path rewrite, but `createPath` drops a bare `?` and `matchRoutes` resolves every variant, so there is no symptom. And `client_name` is capped at 255 with a control-character refinement at registration, so the missing `truncate` on the ClientChip is cosmetic at worst.

hazemadelkhalel (approved):

PR #15080, opened 2026-08-26
feat(mcp): Pieces tab in the MCP page showing available pieces

othmanemad:
## Description

**5 of 5 — stacked on #15079.** Last of the MCP server page stack.

Adds the **Pieces** tab: every piece a connected client can reach in this project, and every action inside it, each classified read / search / write / destructive. "What can a client that holds my grant actually touch" becomes something you read off a page instead of infer from the piece list.

- Search across pieces *and* action names
- Warns when piece actions are switched off for the project — clients still see the list, but every call fails, which is confusing without being told
- Points a platform admin at the piece set that decides the list, and shows the enterprise upsell where restricting the list is gated

**Route note:** this PR *inserts* the Pieces trigger between Connect and Connections rather than appending it, so the final tab order reads Connect · Pieces · Connections. The diff on `index.tsx` therefore both adds a trigger and shifts one — that shift is intentional, not a stray reorder.

## How was this tested?

- `pieces-utils.test.ts` — 8 unit tests over the piece/action rollup and classification counts, pass
- Full `test/app/routes/mcp-server` suite, 13 tests across all three tabs, pass
- `tsc --noEmit` clean on `web`; `turbo lint --filter=web` 0 errors
- Edition paths — this is the one tab with gated surface:
  - **CE**: all pieces reachable, `LockedAlert` explains that restricting the set is enterprise
  - **EE / Cloud enterprise**: piece set drives the list, admin gets the "Review piece set" link
  - **Cloud freemium/paid**: same as CE plus the `RequestTrial` path
- Project-settings dialog opens correctly from the "actions are off" warning

The page itself is a read-only mirror of what the project's piece set already allows — it grants nothing and changes no permission, and the route is behind `Permission.READ_MCP`.

But the project picker made `projectId` user-controlled on `GET /v1/pieces` for the first time. That route (and `/v1/pieces/:name`, `/v1/pieces/:scope/:name`) is declared `securityAccess.unscoped(ALL_PRINCIPAL_TYPES)` yet takes `projectId` off the query string and feeds it to `resolvePieceSetForProject`, which looks the project up by id alone. Before this PR every caller passed the session project id, so nothing exercised it; with `?project=` in the URL, a hand-typed id would have revealed which pieces and actions a sibling project in the same platform has enabled.

**Mitigation:** all three routes now assert membership via `rbacService.assertPrinicpalAccessToProject` — the same helper the security middleware runs for every `securityAccess.project` route. Membership only, no permission, since `/v1/pieces` is the general catalogue the flow builder depends on. Skipped for principals with no platformId (WORKER / UNKNOWN / ONBOARDING), for whom `resolveVisibility` already returns null and `projectId` was inert. Platform admins and operators keep cross-project access through the implicit role `getRole` grants them, which is the same rule that decides what the picker lists.


packages/web/src/app/routes/mcp-server/reach/reach-hooks.ts
```diff
@@ -0,0 +1,25 @@
...
+  useReachablePieces: (projectId: string | null) => {
+    const { i18n } = useTranslation();
+    const query = useQuery<PieceMetadataModelSummary[], Error>({
+      queryKey: ['mcp-reach-pieces', projectId, i18n.language],
+      queryFn: () =>
+        piecesApi.list({
+          projectId: projectId ?? undefined,
+          suggestionType: SuggestionType.ACTION,
+          locale: i18n.language as LocalesEnum,
+        }),
```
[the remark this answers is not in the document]
othmanemad: Confirmed and fixed in 20b3d890ef.

The three `GET /v1/pieces*` routes (`/`, `/:name`, `/:scope/:name`) all accept `projectId` off the query string and feed it to `resolveVisibility` → `resolvePieceSetForProject`, which looks the project up by id alone. Fixed all three rather than just the list route the hook calls — the detail routes leak more (full action/trigger maps for a piece, not just summaries).

The handlers now assert membership via `rbacService.assertPrinicpalAccessToProject`, the same helper the security middleware runs for every `securityAccess.project` route. Membership only, no permission: `/v1/pieces` is the general catalogue the flow builder lives on, so requiring `READ_MCP` would break the piece selector for editors without MCP access. Skipped when the principal has no platformId (WORKER / UNKNOWN / ONBOARDING) — `resolveVisibility` already bails for those, so `projectId` was inert for them and their behaviour is unchanged.

The picker keeps working: `projectMemberService.getRole` grants platform admins an implicit ADMIN role on every project in their platform, which is also exactly what `/v1/projects` uses to populate the picker.

Covered by three new cases in `piece-component-filtering.test.ts` — a project member gets 200 for their own project, 403 for a sibling project on both the list and detail routes.

packages/web/src/app/routes/mcp-server/reach/reach-hooks.ts
```diff
@@ -0,0 +1,25 @@
...
+export const reachQueries = {
+  useReachablePieces: (projectId: string | null) => {
+    const { i18n } = useTranslation();
+    const query = useQuery<PieceMetadataModelSummary[], Error>({
+      queryKey: ['mcp-reach-pieces', projectId, i18n.language],
+      queryFn: () =>
+        piecesApi.list({
+          projectId: projectId ?? undefined,
+          suggestionType: SuggestionType.ACTION,
+          locale: i18n.language as LocalesEnum,
```
AbdulTheActivePiecer: this can be moved to the pieces hooks, you can create a usePiecesWIthActions hook there and use it.

packages/web/src/app/routes/mcp-server/reach/reach-piece-row.tsx
```diff
@@ -0,0 +1,102 @@
...
+        </div>
+      </CollapsibleTrigger>
+      <CollapsibleContent>
+        <div className="grid grid-cols-1 gap-6 bg-muted/40 px-4 py-4 sm:grid-cols-2 lg:grid-cols-4">
+          {row.groups.map((group) => (
+            <div key={group.classification} className="flex flex-col gap-2">
+              <div className="flex items-center gap-1.5">
+                <span
+                  className={cn(
+                    'text-[11px] font-semibold uppercase tracking-wide',
```
AbdulTheActivePiecer: again with the text sizes

packages/web/src/app/routes/mcp-server/reach/reach-piece-row.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    group.classification === 'DESTRUCTIVE'
+                      ? 'text-destructive'
+                      : 'text-muted-foreground',
+                  )}
+                >
+                  {ACTION_CLASSIFICATION_BADGES[group.classification].label()}
+                </span>
+                <span className="text-[11px] text-muted-foreground">
```
AbdulTheActivePiecer: here too

packages/web/src/app/routes/mcp-server/pieces/pieces-utils.ts
```diff
@@ -0,0 +1,133 @@
...
+  return {
+    piece,
+    groups,
+    actionCount: actions.length,
+    destructiveActionCount: actions.filter(
+      (action) => action.classification === 'DESTRUCTIVE',
+    ).length,
+    forceExpanded,
+  };
+}
```
AbdulTheActivePiecer, line 58: forceExpanded is taken as a param and returned as is, so what's the point of it as a param ?

packages/web/src/app/routes/mcp-server/reach/reach-utils.ts
```diff
@@ -0,0 +1,133 @@
...
+      return [toReachablePiece({ piece, actions, forceExpanded: false })];
+    }
+    const matchingActions = actions.filter(
+      (action) =>
+        containsQuery({ text: action.displayName, lowercaseQuery }) ||
+        containsQuery({ text: action.description, lowercaseQuery }),
+    );
+    if (matchingActions.length === 0) {
+      return [];
+    }
```
AbdulTheActivePiecer: Shouldn't be filtering on display name or description since the backend already would do that for you in the list endpoint already if you use it passing the query param.

packages/server/api/src/app/pieces/metadata/piece-metadata-controller.ts
```diff
@@ -151,6 +157,20 @@ const basePiecesController: FastifyPluginAsyncZod = async (app) => {
...
+        return
+    }
+    if (isNil(projectId) || isEmpty(projectId) || isNil(getPlatformId(principal))) {
+        return
+    }
+    await rbacService(log).assertPrinicpalAccessToProject({
+        principal,
+        permission: undefined,
+        projectId,
+    })
```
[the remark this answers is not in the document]
othmanemad: Not a real issue — dismissing this one.

Project IDs are 21-character base62 `apId`s (~125 bits of entropy, `packages/core/utils/src/lib/id-generator.ts`). "Enumerating valid project IDs" by distinguishing 404 from 403 requires guessing an unguessable identifier first; at that point the oracle tells the attacker nothing they didn't already have. There is no enumerable ID space here — these aren't sequential integers.

Collapsing 403 into 404 would only make real access errors harder to debug for legitimate users, for zero security gain. Leaving the 403/404 distinction as-is.

packages/web/src/app/routes/mcp-server/pieces/pieces-tab.tsx
```diff
@@ -0,0 +1,218 @@
...
+    keepPreviousResults: true,
+    showErrorDialog: true,
+  });
+  const { data: mcpServer } = mcpHooks.useMcpServer(projectId ?? '');
+
+  const rows = piecesUtils.toReachablePieces({
+    pieces: pieces ?? [],
+    isSearching,
+  });
+  const visibleRows =
```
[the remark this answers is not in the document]
othmanemad: Fixed in 7cf1b5e.

`usePieces` now returns `isError`/`error`, and the tab renders the failure instead of falling through to the empty state: an access-specific message (`AUTHORIZATION` / `PERMISSION_DENIED` / `ENTITY_NOT_FOUND`) telling the user to pick another project or ask for access, and a retryable "The pieces failed to load" otherwise. `pieces ?? []` stays, but it is only reached on a successful response now.

Dropped `showErrorDialog` on this query — the inline alert explains the failure in place and offers a retry, so the global "Failed to load data" modal was just noise on top of it.

AbdulTheActivePiecer (approved):

PR #15094, opened 2026-08-27
fix(agents): agent setup problems are no longer reported as system errors

hazemadelkhalel:
## Description

When an agent run failed because of a setup problem — no AI provider configured, a bad API key, a model the provider no longer serves — we reported it as an internal system error. The job moved to the failed-jobs queue, where nothing a retry can do would ever clear it. One 30-day retention window had built up roughly **5,900 such jobs**, burying the failures that are genuinely ours to fix.

Those runs now complete as `USER_FAILURE`. Nothing changes for the person using the agent: they see the same error in chat and the same failed step in the builder. What changes is that the failed-jobs queue goes back to meaning "something is wrong with Activepieces".

Anything unrecognised still fails loudly. `classifyAgentRunError` encodes four judgements, each of which is a way to get this wrong:

- **The user-fault statuses are an allow-list (401/403/404), not `!APICallError.isRetryable`.** The AI SDK calls every 4xx non-retryable, so the tempting one-liner blames the user for a `413` prompt-too-long after *our* compaction, or a `400` from an illegal tool name *we* generated.
- **An auth failure on the managed `activepieces` provider stays internal.** It runs on our own OpenRouter key and would fail every platform at once, so it has to page.
- **Credit is read from a status or the specific `insufficient_quota` marker, never loose text in a response body.** Scanning a body for "credits" made a provider `500` whose HTML error page happened to mention credits complete as a billing failure — hiding a real outage. It also fixes the inverse: OpenAI signals billing exhaustion as a *retryable* `429`, which we previously retried pointlessly.
- **`ENTITY_NOT_FOUND` counts only for an AI-provider `entityType`, and `VALIDATION` for nothing.** A bare not-found is our bug. The `VALIDATION` that reaches this surface is the conversation concurrency lock, where a conversation stuck mid-stream is a state worth seeing.

### Reading the failed job you do get

The worker↔app RPC envelope carried only `error.message`, so three unrelated causes all arrived as an indistinguishable bare `ENTITY_NOT_FOUND`. It now also carries the error's `code` and `entityType`.

Deliberately **not** the whole `params`: it is typed `unknown`, and socket.io JSON-encodes this ack from inside a `catch` where nothing handles a throw — one cyclic or BigInt-bearing value would send no ack at all and stall the caller for the full 60s RPC timeout. A regression test pins this with a cyclic-params case and a JSON-round-tripping fake socket.

### CI was not running any of this

`worker` and `core-execution` were in neither the CI test filter nor `npm run test-unit`, so none of the above was enforced on a PR. Both are added, which brought **226 previously-unrun worker tests** into CI.

Getting there needed one skip: `worker-settings-override.test.ts` has 9 cases that all time out on the socket handshake, rotted precisely because the package was never in that filter. Those tests cover the cloud sandbox-isolation gate, so the skip comment names what it guards — **that production check is intact and untouched by this PR**. Repairing the file is worth its own issue.

### Not in this PR

- The curated-model-list cleanup (deprecated Gemini/Grok ids). Removing them re-points every Google platform on the default tier from a GA model to a preview one via the `curatedModels[0]` fallback, with no migration — that belongs in its own change, next to a decision about the replacement id.
- Draining the existing ~5,900 backlog. Separate, needs its own approval; it also self-expires at `REDIS_FAILED_JOB_RETENTION_DAYS` (default 30).
- A failed agent step still leaves `flow_run` SUCCEEDED, because the Run Agent piece returns the failed result rather than throwing. Out of scope: a naive fix would also fail runs where the turn succeeded but one tool errored, so it needs the turn-level failure carried separately.

## How was this tested?

- `npm run test-unit` — 28/28 tasks green. `npm run lint-dev` — 33/33, zero errors.
- 17 classifier cases covering the full status table (400/401/402/403/404/408/409/413/422/429/500/503), the managed-provider carve-out, `RetryError` unwrapping, and the 5xx-mentioning-credits case that used to misclassify.
- 3 RPC envelope cases, including the cyclic-params one. Verified each new guard actually fails when the logic is broken: reverting the narrow projection makes the cyclic test throw `Converting circular structure to JSON` and hang for the full timeout; forcing the classifier to always return `user` reds 5 tests.
- Verified independently that the 9 `worker-settings-override` failures pre-date this branch (stash, re-run, restore).
- Edition paths: the agent job is EE/Cloud only (`src/app/ee/agent`); CE does not register it, so CE is unaffected. No new env var, secret, or setup step — zero-setup on self-host.

Fixes # (issue)

The RPC envelope change is additive and degrades cleanly in both directions during a rolling deploy: an old worker ignores the new field, and a new worker reads `undefined` from an old app and classifies as before. No API field, column, env var, or setup step changes.

Reviewed on the disclosure question specifically, since an error's payload now crosses a process boundary: only `code` and a string-checked `entityType` are carried, so nothing in `params` (engine stdout/stderr, license keys, API keys, JWTs, `userId`/`projectId`) can reach the worker, a user-visible string, or the log drain. The user-visible message still derives solely from `err.message`, exactly as before. No authorization code is in the user-class set, and the ack is a per-packet callback on one authenticated worker socket.


hazemadelkhalel:
Adding `large-pr-ok`: 336/300 in engine/worker/execution, 36 over, and 159 of those lines are the three test files. Splitting would put the classifier in a separate PR from the RPC envelope it reads, so neither half would be reviewable on its own — worse for review than one PR this size. Nothing here was covered in CI before (`worker` was in neither the CI filter nor `test-unit`), so the tests are the point rather than padding.

hazemadelkhalel:
Simplification pass pushed. Removed the duplication the earlier shape invited: `apErrorOf` had two branches building the same projection, the quota marker was written twice, `apErrorOf` ran twice per classify, and three single-caller predicates wrapped what reads better as one top-to-bottom decision. The classifier tests are now one status table — shorter than the repeated assertions, and covering four statuses they missed (408/409/413/422): 25 cases, up from 17.

engine/worker/execution is 291/300, so I've removed `large-pr-ok` — the size gate passes on its own now.

hazemadelkhalel:
## Validated against 30 days of production logs

Queried `default.otel_logs` (2.91B rows, 30-day retention — the whole window that exists). The log line this PR adds doesn't exist in prod yet, so the proxy is the worker's wide event: `outcome:'failed'` is set exactly where the handler re-throws, so **one such row == one BullMQ failed-set entry**.

**12,044 agent-run failures over 30 days. 11,934 (99.09%) would now complete; 110 (0.91%) still fail.** In the ticket's own Aug 25–27 window: 5,935 of 5,952 (99.71%) — their arithmetic checks out exactly.

All five status-code claims confirmed from real `error.statusCode` values:

| Cluster | Count | Status | Now |
|---|---|---|---|
| `ENTITY_NOT_FOUND` provider not configured (+ bare, pre-Aug-17) | 11,672 | — | user |
| `No endpoints found for <model>` | 158 | **404** | user |
| `Unauthorized` on custom provider | 67 | **401** | user |
| `error: {}` — thrown value serialized to nothing | 47 | — | internal |
| Anthropic tool-name pattern | 24 | **400** | internal ✓ correct |
| `is deprecated` / `no longer available` | 16 | **404** | user |
| everything else | 60 | mixed | mostly correct |

`No endpoints found` is a **404, not a 400** — so the allow-list covers it. The 24 pattern errors stay internal, which is the intent: we generate tool names Anthropic rejects.

### One real gap the data found, and why I have not patched it

5 rows of a Google **429** reading `You exceeded your current quota`. No status and no `insufficient_quota` marker catches it. A `/quota/i` pattern would — but Google returns 429 `RESOURCE_EXHAUSTED` for **per-minute rate limiting as well as billing**, so that pattern risks telling a merely throttled user to top up. I have pulled the untruncated message to settle it rather than guess. The last commit does fix the adjacent inconsistency (the loose patterns now always run on the provider's own message, never on a response body).

### Two things outside this PR, worth their own tickets

1. **All 158 `No endpoints found` rows are `provider: "activepieces"`** — 100% on our managed OpenRouter key, where the model catalog is *ours*. After this PR they complete silently as "user config", so a stale curated model list goes quiet. Happy to make a managed-provider 404 internal instead if you'd rather keep that visible — say the word.
2. **A second path for the same misconfiguration, larger than this one.** `POST /api/v1/agents/runs` rejects before `jobQueue.add`, so it never becomes a job. Aug 25–27 alone: **10,985 `QUOTA_EXCEEDED` + 1,340 `ENTITY_NOT_FOUND`, all logged `httpStatus: 500`** while `error.status` is 402/404. That is 12,325 spurious api-side 500s in two days from pure user billing/config — more than twice the volume of the failed-set problem this PR fixes, and untouched by it.

hazemadelkhalel:
## Simplification pass

Rewritten as one clean commit. **299 → 261** meaningful lines in engine/worker/execution.

**Collapsed the classifier from four functions into one.** It had two overlapping credit branches (the message check ran before the `isNil(apiError)` split, then a second body check after it) plus three single-caller helpers — `classifyUserConfig`, `isCreditText`, `messageOf`. The four billing signals are one question, so they are now one condition, and the two class decisions are one ternary each. Behaviour is identical: `apiError?.statusCode` and `apiError?.responseBody` are undefined-safe, so merging the branches changes nothing.

**All added comments removed**, with one deliberate exception: the marker on `describe.skip` in `worker-settings-override.test.ts`. A bare skip with no reason is worse than a comment, and that one names #15096 so it gets deleted rather than living forever. Every other *why* is in the commit message and in the "Four judgements" section above.

I also restored the pre-existing `// "User not found" is OpenRouter refusing a key` comment to its original single line — I had extended it, which was diff noise.

**Gates re-proven after the rewrite:** substituting the naive `!isRetryable` rule back in reds 4 tests (400, 413, 422, and the managed-key case), which is exactly the laundering the allow-list exists to prevent.

`npm run test-unit` 28/28, `npm run lint-dev` 33/33 zero errors, `bun install --frozen-lockfile` clean.

PR #15095, opened 2026-08-27
fix(agents): a failed agent step now fails its flow run

hazemadelkhalel:
## Description

A Run Agent step whose turn died reported success. `run-agent.ts` **returned** the failed `AgentResult` on RESUME instead of raising it, so the step read SUCCEEDED, the flow run completed SUCCEEDED, and the following steps acted on a payload that carried the failure inside it. A platform with no AI provider configured, or a revoked key, produced runs that looked green.

**This completes the invariant the original ticket asked for.** Its table wanted `flow_run` FAILED for a user-config error and only *asserted* that was already the case — it wasn't.

### #15094 made this more pressing, not less

That PR deliberately stopped these jobs landing in the failed set. So as of today:

| Signal | Before #15094 | After #15094 (now) |
|---|---|---|
| BullMQ job | failed | **completed** |
| flow_run status | SUCCEEDED | **SUCCEEDED** |
| Durable trace | failed-set entry | a `warn` log |

**6,150 flow-step agent runs failed in 16 days on Cloud**, every one of them reading SUCCEEDED. #15094 removed the last durable signal, so without this the invariant is half-implemented in the wrong direction.

### Scoped to a turn that produced nothing usable

`AgentResult` gains an optional `failure`, set **only** where the handler catches a dead turn, and the RESUME branch throws only when it is present. `stepResultFrom` is untouched, so the three other paths that pass a `failure` string keep returning their result and the flow keeps going:

- a turn cut short by the **output limit or usage budget**, which still returns the work it completed
- a turn where **one tool call errored** but the agent carried on
- a run **cancelled by the user**

An earlier revision of this PR keyed off `stepResultFrom` and therefore failed the step in all of those too. That was much broader than the problem, and it is the main thing that changed here — the blast radius is now the genuinely-dead turns only.

`isNil` rather than a truthiness check is deliberate: an `Error` with an empty message would give a blank reason, and `isNil` still fails the step for it. Failing with an empty message beats completing as a success.

**Both `AgentResult` declarations carry the field, and that is not redundant.** They are deliberate twins — `core-execution` on `zod` for the server, `core-piece-types` on `zod/mini` for pieces and the engine. Pieces resolve theirs through `pieces-framework`, so patching one leaves the `piece-ai` build unable to see it. I checked whether one could re-export the other: it cannot, because `AgentStepBlock` differs between the two zod flavours.

## How was this tested?

- 2 tests, both verified to fail in the right direction:
  - Reverting the handler line so the dead-turn path stops setting the signal reds the config-failure test (`expected { prompt: 'do it', … } to match object { failure: StringContaining{…} }`).
  - Restoring the over-broad version, where `stepResultFrom` sets it, reds the passthrough test — so the narrowing itself is pinned, not just the fix.
- `npm run test-unit` — 28/28 tasks. `npm run lint-dev` — 33/33, zero errors. All affected packages rebuild clean.
- Edition paths: the agent step is EE/Cloud only (`agentModule` is registered in the CLOUD and ENTERPRISE branches of `app.ts`), so a Community self-hoster is unaffected. No new env var, secret, or setup step.
- **Not verified live:** I have not run a real agent step end to end. Worth one manual check — a Run Agent step on a platform with no provider for the chosen vendor should now produce a FAILED run.

## On the breaking-change classification

Raised in review as possibly not breaking, on the grounds that agents aren't live. The data says the step is:

**10,682 Run Agent flow-step executions across 397 distinct platforms** in 16 days (Cloud only, so that is a floor). 201 platforms used it on 2+ days, 70 on 5+ days, and distinct platforms per day roughly doubled week over week. The action dates to 2025-05-06, is ungated (`auth: PieceAuth.None()`, no plan flag, no `LockedFeatureGuard`), and six flow-version migrations exist purely to rewrite persisted `run_agent` steps inside existing customer flows.

The real distinction: the **new Agents surface** *is* gated — `agentController` sits behind `platformMustHaveFeatureEnabled((p) => p.plan.agentsEnabled)`. But `agentRunController`, which the flow step calls, is registered deliberately **outside** that wrapper, so the step works for every Cloud and EE platform regardless of the flag.

The concrete dependency that makes it breaking: a flow can branch on the agent result today — `Run Agent → Router: if status is FAILED, notify`. That router no longer runs when the step raises. The docs entry tells people to move that handling onto the step's failure path.

Fixes # (issue)

A flow whose agent step dies now has a FAILED run where it previously had a SUCCEEDED one, and its downstream steps no longer run. No API field, column, env var or setup step changes, and nothing needs migrating. Entry added to `docs/install/reference/breaking-changes.mdx`.

Adds one optional field to an internal result type and one throw. No authentication, authorization, secret handling, crypto, outbound HTTP or file handling is touched. The `failure` string is the same message already shown to the user in the step output today.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -116,7 +116,7 @@
     },
     "packages/core/execution": {
       "name": "@activepieces/core-execution",
-      "version": "0.16.0",
+      "version": "0.17.0",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -148,7 +148,7 @@
     },
     "packages/core/piece-types": {
       "name": "@activepieces/core-piece-types",
-      "version": "0.6.0",
+      "version": "0.7.0",
       "dependencies": {
         "@activepieces/core-utils": "workspace:*",
         "tslib": "2.6.2",
@@ -334,7 +334,7 @@
     },
     "packages/pieces/community/ai": {
       "name": "@activepieces/piece-ai",
-      "version": "0.8.0",
+      "version": "0.9.0",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
--- a/docs/install/reference/breaking-changes.mdx
+++ b/docs/install/reference/breaking-changes.mdx
@@ -23,6 +23,21 @@ The `SMTP Blacklist Sender` checkbox is replaced by a `Blocked Sender Addresses`
 #### What you need to do
 
 Nothing to configure or migrate, and no existing step stops working. Review any flow whose Create or Update Contact step relied on the old behaviour: a step that was silently wiping attributes will now leave them intact, and a step you were relying on to blacklist contacts must have the checkbox explicitly ticked. To block a sender for a contact, list that sender's address in `Blocked Sender Addresses` — it must be an active sender in your Brevo account, or Brevo answers *"One of the sender is invalid or inactive"*.
+#### A Run Agent step whose turn dies now fails its flow run
+
+A Run Agent step whose turn died — no AI provider configured for the chosen vendor, a revoked API key, a model the provider no longer serves — used to report success. The step returned the failed agent result rather than raising it, so the step read SUCCEEDED, the run completed SUCCEEDED, and later steps acted on a payload that carried the failure inside it. Such a run is now FAILED, and the agent's error message is the step's error.
+
+Two cases are deliberately unaffected, because the agent did produce usable output:
+
+- A turn that finished but had a single tool call error. The agent may have recovered from it and reported on it, so those runs keep succeeding exactly as before.
+- A turn that was cut short by the output limit or the run's usage budget. It still returns the work it completed, and the flow still continues.
+
+#### What you need to do
+
+Nothing to configure or migrate, and no new environment variable. Two things to check if either applies to you. If a flow relied on continuing past a dead agent step, tick **Continue on failure** on that step — it now takes effect, where previously the step never failed for it to apply to. If a flow branches on the agent result (a Router reading the step's `status`), that branch no longer runs when the turn dies, because the step raises instead of returning; move that handling onto the step's failure path.
+
+If you alert on run status, expect agent flows that were silently completing on a misconfigured provider to start showing as failed. That is the misconfiguration surfacing, not new breakage.
+
 
 #### Workers no longer pre-warm the flow cache on startup by default
 
--- a/docs/install/reference/breaking-changes.mdx
+++ b/docs/install/reference/breaking-changes.mdx
@@ -23,6 +23,7 @@ The `SMTP Blacklist Sender` checkbox is replaced by a `Blocked Sender Addresses`
 #### What you need to do
 
 Nothing to configure or migrate, and no existing step stops working. Review any flow whose Create or Update Contact step relied on the old behaviour: a step that was silently wiping attributes will now leave them intact, and a step you were relying on to blacklist contacts must have the checkbox explicitly ticked. To block a sender for a contact, list that sender's address in `Blocked Sender Addresses` — it must be an active sender in your Brevo account, or Brevo answers *"One of the sender is invalid or inactive"*.
+
 #### A Run Agent step whose turn dies now fails its flow run
 
 A Run Agent step whose turn died — no AI provider configured for the chosen vendor, a revoked API key, a model the provider no longer serves — used to report success. The step returned the failed agent result rather than raising it, so the step read SUCCEEDED, the run completed SUCCEEDED, and later steps acted on a payload that carried the failure inside it. Such a run is now FAILED, and the agent's error message is the step's error.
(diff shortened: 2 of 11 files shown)


hazemadelkhalel:
CI caught something worth recording: `AgentResult` is declared **twice** on purpose — `core-execution` on `zod` for the server, `core-piece-types` on `zod/mini` for pieces and the engine, which is a deliberate bundle-size split. Pieces resolve theirs through `pieces-framework`, so adding `failure` to only the core-execution copy left `run-agent.ts` unable to see it and the `piece-ai` build failed with `Property 'failure' does not exist on type 'AgentResult'`.

The brain page for this area already warns that a field added to one of these schemas has to be added to its twin as well. I missed it. Both copies now carry the field, and `piece-ai` is bumped to 0.9.0, which `validate-publishable-packages` requires for any change to a published piece.

hazemadelkhalel:
On the stale-piece-version finding: that was already addressed before this comment landed — `@activepieces/piece-ai` is at **0.9.0** as of `0408445578`, which also added the `failure` field to the `zod/mini` twin of `AgentResult` in `core-piece-types` (the piece build was failing without it). Greptile reviewed the first commit on the branch.

The finding did point at something real nearby, though, which I have now fixed: the three version bumps on this branch (`core-execution`, `core-piece-types`, `piece-ai`) were not reflected in `bun.lock`, so `bun install --frozen-lockfile` in Docker and the release workflows would have failed. Same issue was flagged on #15100 and applied here and on #15094 too. Verified `--frozen-lockfile` now passes with no changes.

hazemadelkhalel:
## Simplification pass

Rewritten as one clean commit. **43 meaningful lines**, all added comments removed — the *why* is in the commit message and in the description above.

**One correctness decision worth stating**, since it is no longer visible in the code: the RESUME branch uses `isNil(result.failure)` rather than a truthiness check, deliberately. An `Error` with an empty message would produce a blank reason, and `isNil` still fails the step for it. Failing with an empty message beats completing as a success, which is what a truthiness check would do.

**Both `AgentResult` declarations carry the field, and that is not redundant.** They are deliberate twins — `core-execution` on `zod` for the server, `core-piece-types` on `zod/mini` for pieces and the engine, which is a bundle-size split the codebase documents. Pieces resolve theirs through `pieces-framework`, so patching one left the `piece-ai` build unable to see it. I checked whether one could re-export the other: it cannot, because `AgentStepBlock` differs between the two zod flavours.

`npm run test-unit` 25/25, `npm run lint-dev` 33/33 zero errors, all four core packages rebuild clean.

hazemadelkhalel:
## Checked whether this is really breaking — it is

The question raised was that agents aren't live yet, so a behaviour change here affects nobody. Thirty days of Cloud logs say otherwise.

**Run Agent flow-step executions, 2026-08-12 → 08-27 (16 days):**

| | |
|---|---|
| Executions | **10,682** |
| Distinct platforms | **397** |
| Distinct projects | **398** |
| Distinct flow runs | 10,206 |
| Platforms using it on 2+ days | **201** |
| Platforms using it on 5+ days | **70** |
| Distinct platforms per day | 28 → 122 (roughly doubled week over week) |

Excluding one platform's single-day incident (5,769 failures on 08-26, worth its own look), flow-step runs are **4,913 with 92.2% success** — so thousands of runs currently complete SUCCEEDED, and the ones carrying a dead turn are exactly what this PR flips to FAILED.

**The discriminator is code-verified, not guessed.** There is no `source` field in the wide event (`JSONHas(Body,'source')` is 0 on every row), so presence of `flowRun`/`waitpoint` was used instead. `run-agent.ts` posts `flowRunId: context.run.id` plus a waitpoint id, and `agent-run-controller.ts` is the only enqueue site passing them — the chat path passes neither. Separately: **`source` being absent from the wide event is worth fixing**, since the discriminator is inferential today.

**It is also not new and not gated.** The action dates to 2025-05-06 (`d78481dfc8`), with `auth: PieceAuth.None()`, `audience: 'both'`, no plan flag and no `LockedFeatureGuard`. The strongest evidence is six flow-version migrations (`migrate-v14/15/16/20/22`) whose only job is rewriting persisted `run_agent` steps inside existing customer flow versions — nobody writes migrations for a step that exists in no saved flow.

**Where the confusion likely comes from, and it is a real distinction:** the new Agents surface *is* gated — `agentController` sits behind `platformMustHaveFeatureEnabled((p) => p.plan.agentsEnabled)`. But `agentRunController`, which the flow step calls, is registered **deliberately outside** that wrapper in `agent.module.ts`, so the step works for every Cloud and EE platform regardless of the agents flag.

Keeping the breaking-change label and the `breaking-changes.mdx` entry. Happy to reclassify if you read it differently, but not on this evidence.

### Caveats on the numbers

- Cloud only. Self-hosted Enterprise usage is invisible in this stream, so 397 platforms is a floor.
- No pre-08-12 baseline: `EXECUTE_AGENT_RUN` as a job type only appears from 08-12, matching the thin-client rewrite (#14699). The step is older than its current job type, so this measures when the job type started, not when usage started.
- Log ids only, no plan tiers, so I cannot say how many of the 397 are paying versus trialing versus internal.

PR #15096, opened 2026-08-27
fix(worker): prevent settings override tests from timing out

hazemadelkhalel:
## Description

Repairs the 9 tests in `packages/server/worker/test/lib/worker-settings-override.test.ts`, which all failed on a 5-second timeout waiting for a `FETCH_WORKER_SETTINGS` callback that never fired. No user-facing behaviour changes — this is a test-only repair.

Root cause: the worker builds its machine info *before* it emits `FETCH_WORKER_SETTINGS`, and `collectMachineInfo` probes `<AP_FRONTEND_URL>/api/v1/health`, racing that fetch against a 5-second sleep. The test's fake API was a bare `createServer()` with no request listener of its own. Socket.IO's `attach()` replaces the server's `request` listeners with its own and forwards non-socket.io paths to the listeners that existed when it attached — of which there were none — so the health request was accepted and never answered. The probe burned its full 5-second budget, and settings arrived just after the test's own 5-second deadline. Giving the fake API a request listener, exactly as the sibling `worker.test.ts` already does, fixes all 9.

Nothing was weakened to get there: no timeout was raised, no assertion removed, no case skipped, and no production code touched. The cloud sandbox-isolation gate these tests guard — the one that refuses a worker group running anything but `SANDBOX_PROCESS` / `SANDBOX_CODE_AND_PROCESS` on cloud edition — stays covered exactly as before.

These rotted unnoticed because the `worker` package was in neither the CI test filter nor `npm run test-unit`. #15094 adds it to both; with this merged, the `describe.skip` that PR needed can be dropped.

A gotcha bullet is added to `brain/knowledge/execution-runtime/workers.md` so the next person writing a fake API for the worker meets this in place.

## How was this tested?

- `npx vitest run test/lib/worker-settings-override.test.ts` — 9 passed in 1.8s total, well clear of the 5s deadline (they previously failed at 5.02–5.07s each), so this is a real fix and not a timing squeeze.
- `npx turbo run test --filter=worker` — whole package green: 221 passed, 1 skipped, 0 failed.
- `npx tsc -p tsconfig.lib.json --noEmit` in `packages/server/worker` — clean.
- `npx turbo run lint --filter=worker` — 0 errors, 29 pre-existing warnings.

Edition paths: the suite itself exercises community, enterprise, and cloud edition branches of the settings-override gate; all pass.


hazemadelkhalel:
Applied — retitled to `fix(worker): stop the worker settings tests timing out`, which names the outcome rather than the activity. Keeps the Conventional Commit prefix.

For the record, the cause was that the test's fake API server had no request listener of its own, so socket.io's `attach()` left `/api/v1/health` accepted and never answered; the worker's startup probe then burned its full 5s budget and settings arrived just after the test's own 5s deadline. Every failure clustered at 5022–5072ms, which is what gave it away. Suite now finishes in 1.8s rather than 4.9s.

hazemadelkhalel:
Adopted your exact wording: **`fix(worker): prevent settings override tests from timing out`**.

I had already retitled to "stop the worker settings tests timing out" before seeing this, but yours is better on two counts — `(worker)` is already the scope so repeating it was redundant, and "settings override" matches the file name (`worker-settings-override.test.ts`) rather than paraphrasing it. Switched to yours and resolved the thread.

No coding-style objection to record: naming the corrected failure rather than the activity is the right rule for this repo, since PR titles feed the changelog.

PR #15099, opened 2026-08-27
fix(engine): create the output dir before writing the build metafile

hazemadelkhalel:
## Description

The engine's esbuild `rebuildLogger` plugin writes `<outfile>.meta.json` next to the bundle without ensuring the directory exists. Normally esbuild has already created `dist/packages/engine/` by the time `onEnd` fires, so this is invisible. When something removes or recreates that directory concurrently, the write fails:

```
ERROR: [plugin: engine-rebuild-logger] ENOENT: no such file or directory,
open '/home/runner/.../dist/packages/engine/piece-child.js.meta.json'
```

That failed the `main` job on #15098 twice in a row. The likely trigger is concurrency rather than anything in that PR: the CI step runs two turbo invocations in parallel, remote caching is on, and five PRs were building at once — two engine builds racing on one `dist/` path. `main` passed on #15096 and #15097, which build the engine too but ran when less was in flight.

One `mkdirSync(..., { recursive: true })` makes the write order-independent. `path` and `fs` are already imported.

### Honest limitation

**I could not reproduce the failure locally, so I cannot prove this fixes it.** With the fix reverted and `dist/packages/engine` deleted, the build still succeeds single-threaded — esbuild creates the directory itself, which is exactly why the bug only shows under concurrency. I reached the race by elimination, not by observing it.

What I can say for certain: `mkdirSync` with `recursive: true` on a directory that already exists is a no-op, so this cannot regress anything, and it removes the plugin's dependence on another process not touching the directory. If CI keeps failing after this, the cause is elsewhere and this line is still correct.

## How was this tested?

- `rm -rf dist/packages/engine && npx turbo run build --filter=@activepieces/engine --force` — clean, and both `main.js.meta.json` and `piece-child.js.meta.json` are produced.
- Same command with the change reverted — also clean, which is the limitation described above.
- No test added: this is a build-script guard with no runtime surface, and a test would have to simulate a concurrent directory removal to mean anything.

Fixes # (issue)

Build-time only. The emitted bundles are byte-identical; only the sibling metafile write is made robust. No API, column, env var, or setup step is touched.

Creates a build output directory if absent. The path is derived from the build config, not from any input — no user or network data reaches it. No authentication, secret, crypto, outbound HTTP, or runtime file handling is involved.

### DIFF
--- a/packages/server/engine/esbuild.config.mjs
+++ b/packages/server/engine/esbuild.config.mjs
@@ -10,7 +10,7 @@ const pieceChildOutfile = path.join(outdir, 'piece-child.js');
 
 const watch = process.argv.includes('--watch');
 
-fs.rmSync(outdir, { recursive: true, force: true });
+fs.mkdirSync(outdir, { recursive: true });
 
 const zodLocaleTrim = {
   // Drop zod's 46 unused locale packs (~184KB). The app surfaces validation
--- a/packages/server/engine/esbuild.config.mjs
+++ b/packages/server/engine/esbuild.config.mjs
@@ -39,6 +39,7 @@ function rebuildLogger(outfile) {
       });
       build.onEnd((result) => {
         if (result.metafile) {
+          fs.mkdirSync(path.dirname(outfile), { recursive: true });
           fs.writeFileSync(
             outfile + '.meta.json',
             JSON.stringify(result.metafile)


hazemadelkhalel:
Good catch, and it changes the fix — you were right that the race remained.

`esbuild.config.mjs` also called `fs.rmSync(outdir, { recursive: true, force: true })` at module load, so a second invocation could wipe the directory between the first one's `mkdirSync` and its `writeFileSync`. My change narrowed the window without closing it, and the PR description overclaimed by saying it made the write order-independent.

The clean turned out to be buying nothing: there is no code splitting and no content hashing, so both targets write a fixed `outfile` and every file `rmSync` deleted was immediately rewritten under the same name. Its only effect was creating a destructive window for concurrent builds sharing `dist/`. It is now a `mkdirSync` of the same directory, so nothing in this config removes it at all.

I kept the plugin's `mkdirSync` as well, deliberately: during a rollout an older checkout of this file may still run the `rmSync`, and the guard covers that until it is gone everywhere.

Verified from a clean dist and again over a populated one — 6 outputs both times, all deterministic names, so nothing accumulates without the clean. The one residual cost is that renaming an outfile later would leave the old one behind as a dead file, which is cheap next to breaking parallel builds.

PR #15101, opened 2026-08-27
feat(billing): given a specific link, automatically activate the trial to customers

AbdulTheActivePiecer:
## Description

Signing in through a trial activation link (`/sign-in?licenseKey=…`) used to reopen the **Activate Trial Key** dialog — prefilled and auto-submitted — on top of whatever page the user happened to land on. The dialog was never the right surface for this: there is nothing to fill in, and a user who isn't a platform admin got no feedback at all, because the activation was skipped silently.

The stashed key now hands off to a dedicated full-screen activation surface with four states:

| State | What the user sees |
| --- | --- |
| **Activating** | Progress ramp with rotating status copy while the request runs |
| **Success** | Confetti, then a 5s countdown that drops them at their default route |
| **Not a platform admin** | The activation link to copy and forward to an admin |
| **Failed** | Retry, plus a Contact support mailto |

It's an overlay rendered from `AllowOnlyLoggedInUserOnlyGuard` rather than a new route, so the user's original destination isn't lost — and, as before, **no new endpoint and no migration**. Activation still posts to the existing `POST /v1/platform-billing/activate`.

### How it's wired

- One mount-only `useEffect` fires the activation and starts a 250ms clock. The view, the progress ramp, the 75s client-side timeout and the redirect countdown are all *derived* from that clock plus the mutation's own `isPending`/`isSuccess`/`isError` — `now` is the only state.
- `AutomaticTrialActivation` decides whether to show the screen in a lazy `useState` initialiser; every input (stash, platform, edition, platform role) is a suspense query or a synchronous read, so there's nothing to wait for.
- `alreadyLicensed` is frozen at mount. Activation invalidates the platform query, so reading it live would flip the guard and unmount the success panel mid-confetti.
- The stash is cleared on **successful** activation rather than on first read, so a failed attempt survives a reload and retries. `sessionStorage` bounds that to the tab.

### Also in here

- The Activate Trial Key dialog reverts to its pre-trial shape (`initialLicenseKey` / `autoSubmit` / the in-dialog success view are gone).
- `useUpdateLisenceKey` accepts a nullable error message, so the screen can render the outcome itself instead of toasting over it.
- `brain/knowledge/platform-editions-ee/license-keys.md` updated.

### Deviations from the design handoff

Worth a reviewer's eye:

- The URL row uses the existing `CopyToClipboardInput` instead of the spec'd field + "Copy link"/"Copied" text button — `packages/web/CLAUDE.md` forbids hand-rolling that pair.
- Confetti goes through `canvas-confetti` (already a declared dep, previously unused) with our own canvas, which preserves the spec'd layering behind the card. `disableForReducedMotion` covers the reduced-motion case.
- `--ap-neutral-400` maps to `text-muted-foreground/70` so it flips in dark mode.
- No per-attempt reference code in the support mailto — there's no such id, which the handoff said to drop in that case.

## How was this tested?

- `tsc -p packages/web/tsconfig.app.json` clean.
- `npm run lint-dev` — 0 errors (71 pre-existing warnings, count unchanged).
- Edition paths: **Community** and already-licensed platforms skip the screen entirely and render nothing (unchanged gating). **Cloud** and **Enterprise** get the full flow. The header uses `FullLogo`, so the surface follows platform branding rather than hardcoding Activepieces.

Fixes # (issue)

Web-only. No endpoint, schema, migration or env var changes; the `licenseKey` query-param contract the console builds links against is unchanged.

The platform-admin check it branches on is presentation only — `POST /v1/platform-billing/activate` enforces its own authorization server-side, unchanged. The `not_admin` state renders the licence key back to the same user who arrived holding it in their own URL, so it exposes nothing new. The one behavioural change is that the key now lingers in `sessionStorage` until activation succeeds instead of being cleared on first read — same origin, same tab, same user, and gone when the tab closes.


AbdulTheActivePiecer:
@greptileai SAML users would never activate a license through signing in with SAML.

AbdulTheActivePiecer:
@greptileai it is allowed for gmail sign ins SSO and email, no SAML, also re review

PR #15109, opened 2026-08-27
feat(web): make the rail handle many projects, and order them by recent flow activity

AhmadTash:
## Description

Makes the project rail usable once a platform has more than a handful of projects, and changes what "Recently used" means.

**The rail squished.** Nav buttons (Chat / Agents / Explore / Impact) and the project rows lived in one `flex flex-col overflow-y-auto` column with no `shrink-0`, so flexbox compressed the `h-10` buttons to their min-content height before letting the container scroll — measured at **40px → 20px** with 21 projects. The nav block is now pinned, the project list is its own scroll area, and every row carries `shrink-0`. `scrollbar-none` is gone, so the app's global scrollbar styling applies to the list.

**"Recently used" now means the project whose flows changed most recently**, not the project you clicked last. The old ordering read a client-side access history in `localStorage`, so it knew nothing about work done from another browser, by a teammate, or by a schedule. It is now backed by a new `analytics.lastFlowUpdated` on `ProjectWithLimits` — a grouped `MAX(flow.updated)` computed alongside the existing per-project aggregates in `enrichProjects`, so no extra round trip.

**The order stays live without a refresh.** `projectCollection` runs on its own `QueryClient` and never refetches, so the new field would have been frozen at page load. Flow mutations patch the row locally through `flowsApi` rather than refetching every project on each builder autosave.

**Rows animate to their new position** with a `motion` layout transition, disabled under `prefers-reduced-motion`. The local write is deferred two frames: landing it in the same commit as the mutation's re-render made motion fast-forward through most of the tween (measured **228px → 103px in one 6ms frame**), which read as a jump rather than a move.

Also adds a separator above **Platform Admin**, matching the one above **Projects**.

### Note for reviewers

`analytics.lastFlowUpdated` is additive on a response schema. Anything constructing a `ProjectWithLimits` in tests needs the new field — one fixture updated here.

`packages/core/shared` is bumped 0.147.0 → 0.148.0 (minor: new exported field).

## How was this tested?

- `npm run test-unit` equivalent for web: **537 tests / 58 files pass**, including the updated `projects-collection` fixture.
- `turbo run build` green for both `api` (tsc) and `web`; `turbo run lint` reports **0 errors** on both.
- Manually on **CE** (`AP_EDITION=ce`) against a platform seeded with 21 projects:
  - nav buttons hold their full 40px height with the list overflowing; project list scrolls independently.
  - creating a flow in a project sitting 8th moved it to 1st **without a reload** (`performance.getEntriesByType('navigation').length === 1`).
  - renaming a flow from the automations list moved its project 3rd → 1st and animated over 384ms, easing `76px → 70 → 48 → 29 → 16 → 9 → 5 → 2.7 → 1.4 → 0.6 → 0.25 → 0.08 → 0`.
  - collapsed rail verified too: nav icons pin, project badges scroll.
- Not exercised on EE / Cloud. The changed code is edition-agnostic — `enrichProjects` and the rail are shared by all editions — but a reviewer on a Cloud build is worth having.

Fixes # (issue)


AhmadTash:
@greptile review

AhmadTash:
@greptile review

AhmadTash:
@greptile review

AhmadTash:
@greptile review

AhmadTash:
@greptile review

Louai-Zokerburg (approved):

PR #15116, opened 2026-08-27
fix(front): attachments are sent as multipart, so they actually arrive

### LINKED ISSUE #15119: [Bug]:[Pieces] - Front piece silently drops attachments (JSON body instead of multipart)
Supporting ticket for community PR activepieces/activepieces#15116 (supersedes #15115, bot-closed).

## Bug

The Front piece's `attachments` field is a `Property.Array` of plain strings ("List of attachment URLs") posted inside a JSON body. Front only accepts attachments on `multipart/form-data`, as file bytes, so it takes the message and drops `attachments`. No error, no warning, HTTP 202. Affects `sendMessage`, `sendReply`, `createDraft`, `createDraftReply`.

## What the PR changes (3 commits)

1. **engine** — `handleUrlFile` buffered the response without checking status, so a 4xx/5xx produced an `ApFile` containing the error page. The streaming branch (`handleStreamingFile`) already rejected non-ok; the buffered branch now does too.
2. **pieces-common** — `BaseHttpClient.getHeaders` stamps `Content-Type: application/json` on every request with a body, so a global `FormData` body goes out labelled JSON and the fetch-generated boundary is lost. Header is now dropped when the body is a global `FormData`. Author claims zero blast radius: `discord`, `instabase`, `image-router` use the `form-data` package (different `serializeBody` branch, own boundary), `docusign` is not multipart.
3. **front** — `attachments` becomes `Property.Array` of `Property.File`, same shape as `gmail`/`discord`/`docusign`. URLs still work (props-processor recurses into ARRAY-with-`properties`, `fileProcessor` resolves each to an `ApFile`). No-attachment messages keep the JSON path. Indexed-bracket naming (`to[0]`, `attachments[0]`).

Tests: 12 new in `packages/pieces/community/front/test/attachments.test.ts`, 1 new in `packages/server/engine/test/variables/file-processor.test.ts`.

## Review questions for pieces

* **Version bump is 0.1.6 → 0.1.7 (patch).** Steps pinned `~0.1.6` resolve any 0.1.x (`findNextExcludedVersion` in `piece-metadata-service.ts`), so existing flows would pick up the prop-shape change automatically and lose their attachment config. Should this be a minor bump (0.2.0) instead?
* Commit 2 touches shared HTTP client behaviour for every piece. Worth confirming the blast-radius claim independently.
* Commit 1 changes engine file-resolution semantics: a failed download now errors instead of yielding a garbage file. Any flow relying on the old lenient behaviour changes.
* Author offered to split into three PRs, or to drop commits 1 and 2 and route around them inside the piece.

## Impact

Reported from a production flow forwarding uploaded documents to external recipients through Front. The field looks configured and does nothing.

alexandronic:
## The bug

Put a file URL in the Front piece's **Attachments** field, run the flow, and the message sends with no attachment. No error, no warning, HTTP 202.

The field is a `Property.Array` of plain strings described as "List of attachment URLs", and those strings are posted inside a JSON body. Front only accepts attachments on a `multipart/form-data` request, as the bytes of the file, so it takes the message and drops the `attachments` field on the floor. The same thing happens on `sendMessage`, `sendReply`, `createDraft` and `createDraftReply`.

Fixing it turned out to need three layers, one commit each. They are independent enough to split if you'd rather review them separately.

### 1. `fix(engine): a failed download is not a file`

`handleUrlFile` buffered whatever came back from the URL without checking the status, so a 4xx or 5xx produced an `ApFile` named after the URL and filled with the server's error page. A piece can't tell that apart from the real thing, so it travels on as the attachment: an expired signed link becomes an email with the storage provider's XML error stapled to it.

The streaming branch of the same processor already refuses a non-ok response (`handleStreamingFile`). The buffered branch now does too.

### 2. `fix(pieces-common): let fetch set the multipart boundary`

`BaseHttpClient.getHeaders` stamps `Content-Type: application/json` onto every request that carries a body, so a `FormData` body goes out labelled as JSON. Writing `multipart/form-data` by hand doesn't help either: the boundary that separates the parts is chosen by fetch while it encodes the body, and a Content-Type decided before that point names something the receiver can't split on.

The header is now dropped when the body is a global `FormData`, leaving fetch to write the whole thing.

To be clear about the blast radius: **no existing piece is affected.** `discord`, `instabase` and `image-router` all build their multipart body with the `form-data` package, which supplies its own boundary through `getHeaders()` and takes a different branch of `serializeBody`; I checked each against the unpatched client and all three come out correct. `docusign` doesn't use multipart at all. The Front piece here is the first caller to use a global `FormData`, which `serializeBody` already passes through untouched and which `getHeaders` then mislabels.

### 3. `fix(front): actually attach the files`

Attachments become `Property.File` entries, the same shape `gmail`, `discord` and `docusign` use:

```ts
attachments: Property.Array({
  properties: { file: Property.File({ ... }) },
})
```

A URL is still all the user supplies, because `props-processor` recurses into an ARRAY-with-`properties` and `fileProcessor` resolves each entry into an `ApFile` before the action runs. A step that already holds a file can be wired straight in.

A message with no attachments keeps its JSON path, so nothing changes for a flow that never used the field, and an entry whose file couldn't be resolved is skipped rather than failing the send. Array fields use Front's documented indexed-bracket naming (`to[0]`, `attachments[0]`).

## How was this tested?

- 12 new tests in `packages/pieces/community/front/test/attachments.test.ts`. They build a real `Request` from the init object so the assertions read the actual serialised wire bytes: multipart boundary present, `to[0]` / `attachments[0]` naming, the PDF bytes in the body, the filename on the part, and the JSON path still taken when there are no attachments.
- 1 new test in `packages/server/engine/test/variables/file-processor.test.ts` (9 pass in that file).
- `test/variables` goes 42 → 43 passing with no new failures.
- Also checked end to end against the real `propsProcessor`: a plain URL in the field is fetched by the engine and those exact bytes come out in Front's multipart body.

## Note

Alvys is an existing customer; this came out of a production flow where drivers upload documents that need forwarding to brokers through Front. Happy to split the three commits into separate PRs, or to drop commits 1 and 2 and route around them inside the piece, if that's easier to land.

Supersedes #15115, which the bot closed and which we can't reopen from our side; this one carries the same three commits with the comments tightened up.

- [x] yes — technical (removed/renamed API field or endpoint, dropped column, new required field, removed/required env var)

`attachments` is a list of files rather than a list of strings, so an existing flow has to re-pick the file. That's a change from a setting which silently did nothing to one that works. Piece version bumped 0.1.6 → 0.2.0.

Commit 1 also changes behaviour for every piece property that takes a file: a URL that answers 4xx or 5xx now fails the step instead of yielding a file full of the error page. A flow that was silently attaching error pages starts failing at that step.

Both are written up in `docs/install/reference/breaking-changes.mdx` under Unreleased.

Security-sensitive by the template's definition on two counts: it touches file handling (`variables/processors/file.ts`) and outbound HTTP headers (`BaseHttpClient.getHeaders`).

The risk in commit 2 is mislabelling a request body, so the mitigation is scope: the header is dropped only when the body is a global `FormData`, a branch no shipped piece takes today — `discord`, `instabase` and `image-router` use the `form-data` package and go down a different path in `serializeBody`, each verified unchanged against the unpatched client. Commit 1 removes an exposure rather than adding one: an unauthorized or expired URL's error body is no longer packaged as a file and forwarded to a third party.


OdaiAhmed99:
Thanks for this — the diagnosis is exactly right, and I reproduced the bug independently before reading your write-up. I've pushed one commit on top of your three rather than opening a competing PR, so your work stays yours. Here's what changed and why.

## The shared HTTP change isn't needed

This is the main finding, and it's good news for the PR.

`serializeBody` already has a dedicated branch for the **form-data package** (`isNodeFormData`): it pipes the stream and returns `extraHeaders` from `getHeaders()`, boundary included. `finalHeaders` merges `extraHeaders` **last**, so it overrides the JSON default. 75 files in this repo already send multipart that way.

You reached for a global `FormData`, which `serializeBody` passes through as a plain `BodyInit` — and only *that* choice needs `getHeaders` patched. Switching the piece to `form-data` makes commit 2 unnecessary, so I've reverted it here.

The proof is in the PR: **the 18 tests in this piece pass against the unpatched client.**

## About the blast radius

I checked the claim independently, and it's wider than it looked. Scanning all 11,927 tracked `.ts` files for a global `FormData` reaching `httpClient.sendRequest` turns up **29 files across 17 pieces**: cloudconvert, documentpro, extracta-ai, flow-parser, gptzero-detect-ai, groq, housecall-pro, matomo, pandadoc, parseur, personal-ai, photoroom, smartsheet, straico, timelines-ai, zoho-bookings, zoo.

Your reading of discord, instabase and image-router was correct — those use the form-data package and are genuinely unaffected, along with 72 other files. And docusign has no multipart at all.

But the direction matters: `getHeaders` spreads `request.headers` last, so today 11 of those 29 send a hand-written `multipart/form-data` with **no boundary** and the rest send multipart **labelled as JSON**. All 29 are broken, and your header change fixes every one of them.

So commit 2 is worth *more* than you claimed, not less — which is exactly why it deserves its own PR with its own release note and spot-checks, rather than riding along here. Please do open it; there's a home for the test in `packages/pieces/common/test/`.

## The engine commit is right — I was wrong about it

I initially thought it swapped one silent failure for another. It doesn't, and I want to correct that on the record.

I reproduced it against live Front. On `main`, a URL returning **404 with a body** (an expired signed link) is buffered without a status check, and the draft reached Front carrying:

```
filename: expired.pdf   content_type: application/pdf   size: 99
```

99 bytes of `<Error><Code>AccessDenied</Code><Message>Request has expired</Message></Error>`, delivered as a bill of lading. Step status: SUCCEEDED.

With your commit applied, the same run **fails** with `Expected file url or base64 with mimeType, received: http://.../expired.pdf`, and a healthy URL still succeeds. It doesn't drop silently — `fileProcessor` returns `null` and the props validator rejects the step before the action runs. (Small note: the commit returns `null` rather than throwing, unlike the PR description.)

It's a real fix. My only suggestion is that it also belongs in its own PR, purely because it changes file resolution for every piece that accepts a file.

## Other changes in my commit

**An unreadable attachment now fails the step.** Previously an entry that couldn't be resolved was skipped and the message went out without it — which is this bug reached by another route.

**Version is 0.2.0, not 0.1.7.** The property changes shape, and `update-piece-version-dialog` filters the version list to patch releases by default, hiding minor/major behind a toggle. A patch bump would offer a config-breaking change as a one-click upgrade.

**`translation.json` regenerated** for the changed description. (For what it's worth, the nested `File` / `The file to attach.` strings can't be translated by anything today — `pathsToValuesToTranslate` stops at `actions.*.props.*.displayName` with no path for nested `properties.*`. Gmail has the same gap.)

**Content type per part** now derives from the filename via form-data's mime lookup, so a PDF arrives as `application/pdf` rather than `application/octet-stream`.

**Tests cover all four actions** instead of two — `createDraft` and `createDraftReply` had none — plus a property-contract test, and they read the serialised multipart bytes.

## Verified against live Front

All four actions deliver. Beyond the happy path: a 256-byte file covering every byte value arrives byte for byte, a 2 MB file, an empty file, five attachments at once, duplicate filenames, a URL-sourced file, and unicode subjects and bodies.

Two limitations found, neither caused by this PR:

- A **non-ASCII filename** shows as mojibake in Front. We put correct UTF-8 on the wire — I dumped the bytes and they match `Buffer.from(name,'utf8')` exactly — so Front is decoding `filename=` as latin1. Your approach would emit the same bytes.
- A filename containing `(` or `)` is **lost entirely**, arriving as `unknown`. The engine's `getFileName` matches `filename*=UTF-8''([\w%\-.]+)`, and that class excludes characters `encodeURIComponent` leaves literal. Pre-existing, affects every piece resolving a file from a URL.

## One follow-up for whoever merges

Front's four attachment actions should get `unsafeSteps` entries in `piece-upgrade-register.json` before it's next regenerated, or the v23 migration will roll live flows onto the new shape and silently empty their attachment config.

Happy to adjust anything here if you'd rather take a different route on any of it.

OdaiAhmed99 (approved):

MrChaker (approved):

PR #15121, opened 2026-08-27
fix(flows): the Versions option no longer disappears when editing a flow from a run

### LINKED ISSUE #13556: [BUG] Versions menu item stays hidden after switching from run view to edit mode (embedded builder)
## Problem

In the builder, the `⋯` menu's "Versions" entry is hidden whenever the router path contains `/runs`, and in an embedded builder it never comes back after switching to edit mode. Reported twice by the same paid embed customer.

Ref: Pylon 5029, Pylon 5794

## Root cause

Two different `useLocation` hooks disagree about where the user is.

* `packages/web/src/app/components/flow-actions-menu.tsx:82` — `isRunsPage = useLocation().pathname.includes('/runs')` (from `react-router-dom`), gating the item at `:294` with `insideBuilder && !isRunsPage`.
* `builder-header/flow-status/view-draft-or-edit-flow-button.tsx:6,29` — "Edit flow" branches on `useLocation` from `react-use`, i.e. `window.location`.

Embeds run on a memory router (`app/guards/index.tsx:140`), so `window.location` is the host app's URL and never contains `/runs`. Edit flow therefore takes the `switchToDraft()` branch (`builder/flow-canvas/hooks.tsx:99`), which mutates builder state only. Router path stays `/runs/<runId>`, `isRunsPage` stays true, Versions stays hidden for the rest of the session.

Non-embed: `window.location` does contain `/runs`, so the button navigates to `/flows/<id>` and the item returns. Embed-only symptom.

## Fix

Gate on builder state, not the URL. `BuilderHeader` (`builder-header.tsx:147`) is the only caller with `insideBuilder: true`: read `state.run` there, pass it down, render on `insideBuilder && isNil(run)`, then delete `isRunsPage` and its `react-router-dom` import. `clearRun` nulls `run` (`builder/state/run-state.ts:118`), so the item returns on leaving the run in embed and non-embed alike.

Accepted delta: `setRun` sets `readonly: true` (`run-state.ts:98`), so Versions is also hidden while watching a test run inside the builder. Today it shows there.

## Out of scope

* Showing Versions *during* run view (this ticket's original title) — separate call.
* Making "Edit flow" navigate under the memory router (import `useLocation` from `react-router-dom` there). Separate defect: embedded users edit a draft while the URL still says `/runs/<runId>`, so refresh and deep links reopen the run and the host app's route sync reports the wrong page.

## Steps to reproduce

1. Embed the builder via the SDK, open `/runs`.
2. Click any run, then **Edit flow**.
3. Open the flow-name `⋯` menu: Rename and Import are present (draft is editable), Versions is absent.
4. Same steps outside an embed: Versions is present.

majewskibartosz:
Fixes [GIT-1531](https://linear.app/activepieces/issue/GIT-1531)
Closes #13556

## Problem

1. In the flow menu, "Versions" is hidden whenever the router path contains `/runs`, and in an embedded builder it never comes back after switching to edit mode: the user is editing the draft with no way to reach version history.
2. The gate and the Edit-flow button read two different locations: `flow-actions-menu.tsx` used `useLocation()` from `react-router-dom` (the in-app router), while `view-draft-or-edit-flow-button.tsx` reads `useLocation` from `react-use` (`window.location`). Embeds mount a memory router, so `window.location` never contains `/runs`, Edit flow takes the in-place `switchToDraft()` branch, and the router path stays on the run for the rest of the session.

## Fix

- `flow-actions-menu.tsx`: gate "Versions" on a `viewingRun` prop instead of `useLocation().pathname`; the `react-router-dom` import is gone from this component.
- `builder-header.tsx`: derive `viewingRun={!isNil(run)}` from builder state. It is the only caller with `insideBuilder: true`, and `clearRun` nulls `run`, so the entry returns the moment the user leaves the run, embed or not.

Product decision embedded: "Versions" is now hidden whenever a run is displayed on the canvas, including in-builder test runs where it used to show; alternative considered: restrict the gate to PRODUCTION runs, rejected because picking a version mid-run desyncs the canvas from the run overlay in either case (`flow-versions-card.tsx` calls `setVersion` without clearing the run).

## Verification

- `packages/web`: lint 0 errors, `tsc --noEmit` clean, 541 tests pass (60 files).
- Regression tests, red-first proven against the pre-fix code: menu case "shows Versions in edit mode while the route still points at a run" fails pre-fix (`expected 'menuRenamePush to Git…' to contain 'Versions'`); both builder-header wiring cases fail pre-fix (`expected undefined to be false/true`). The unchanged case (run being viewed → hidden) passes in both.
- Driven on a local instance (real API, Postgres, seeded flow with production runs): run view → menu has no Versions; Edit flow → menu has Versions.
- Visual: run view (Versions hidden), edit mode after leaving the run (Versions shown), light theme, local instance. Before/after is not observable outside an embed: pre-fix, non-embed edit mode already showed the entry because the navigation changed the URL. The embed condition is covered by the red-first tests.
- Other affected paths: the same URL-derived-state pattern lives in `flow-canvas/hooks.tsx` `useListenToExistingRun` (live production-run polling never enables in an embed, so the run view goes stale) and in `view-draft-or-edit-flow-button.tsx` (embedded users edit a draft while the URL still says `/runs/<runId>`, so refresh and deep links reopen the run). Both are left unfixed here per the ticket's Out of scope, and both are recorded on GIT-1531.
- Pre-existing failures unrelated to this change: `@activepieces/engine` delay-piece tests and `@activepieces/sandbox` (heap OOM) fail on untouched code in a clean worktree.

<details>

## Plan
- Root cause is a state/URL mismatch, so the fix belongs where the state lives: one prop from the builder header, no new abstraction.
- `BuilderHeader` is the only consumer of `FlowActionMenu`, so the prop plumbing is a 2-file change; the component stays free of builder-store coupling because it lives in `app/components/`, not `app/builder/`.
- Footprint: 2 product files, 12 insertions / 6 deletions; rejected the larger alternatives: plumbing a route-aware flag from `FlowRunPage`, and switching the Edit-flow button's `react-use` import (wider blast radius, out of ticket scope).
- Two test files rather than one: the menu test covers the gate, the builder-header test covers the derivation from `state.run` so the wiring cannot silently break.

- Test run inside the builder: run is set with `readonly: true`, so Versions hides. Deliberate, see the product-decision line.
- Run selected from the builder's Runs panel: navigates to `/runs/<runId>`, so it is the ordinary run-view case.
- `insideBuilder: false` arm: currently unreachable (no caller); left in place with `viewingRun: null` rather than deleted, to keep this diff scoped to the fix.
- Version panel already open when a run loads: unchanged behavior, the sidebar keeps its own state; only the menu entry is gated.

## Alternatives rejected
- Drop the guard entirely and allow Versions during run view: picking a version while a run is displayed desyncs canvas and run overlay; that needs its own product call.
- Have `FlowActionMenu` read the builder store itself: couples a shared component to the builder context.
- Keep the pathname check and special-case embeds: leaves two sources of truth, which is the bug.
</details>

### DIFF
--- a/packages/web/src/app/builder/builder-header/builder-header.tsx
+++ b/packages/web/src/app/builder/builder-header/builder-header.tsx
@@ -1,4 +1,4 @@
-import { Permission } from '@activepieces/core-utils';
+import { isNil, Permission } from '@activepieces/core-utils';
 import {
   ApFlagId,
   FlowOperationType,
@@ -66,12 +66,14 @@ export const BuilderHeader = () => {
     moveToFolderClientSide,
     applyOperation,
     setRightSidebar,
+    run,
   ] = useBuilderStateContext((state) => [
     state.flow,
     state.flowVersion,
     state.moveToFolderClientSide,
     state.applyOperation,
     state.setRightSidebar,
+    state.run,
   ]);
 
   const { embedState } = useEmbedding();
@@ -149,6 +151,7 @@ export const BuilderHeader = () => {
                       setRightSidebar(RightSideBarType.VERSIONS);
                     }}
                     insideBuilder={true}
+                    viewingRun={!isNil(run)}
                     flow={flow}
                     flowVersion={flowVersion}
                     readonly={!isLatestVersion}
--- a/packages/web/src/app/components/flow-actions-menu.tsx
+++ b/packages/web/src/app/components/flow-actions-menu.tsx
@@ -21,7 +21,6 @@ import {
   User,
 } from 'lucide-react';
 import React, { useState } from 'react';
-import { useLocation } from 'react-router-dom';
 import { toast } from 'sonner';
 
 import { ConfirmationDeleteDialog } from '@/components/custom/delete-dialog';
@@ -62,8 +61,12 @@ type FlowActionMenuProps = {
   onDelete: () => void;
   onOwnerChange?: () => void;
 } & (
-  | { insideBuilder: true; onVersionsListClick: () => void }
-  | { insideBuilder: false; onVersionsListClick: null }
+  | {
+      insideBuilder: true;
+      onVersionsListClick: () => void;
+      viewingRun: boolean;
+    }
+  | { insideBuilder: false; onVersionsListClick: null; viewingRun: null }
 );
 
 const FlowActionMenu: React.FC<FlowActionMenuProps> = ({
@@ -78,8 +81,8 @@ const FlowActionMenu: React.FC<FlowActionMenuProps> = ({
   onOwnerChange,
   onVersionsListClick,
   insideBuilder,
+  viewingRun,
 }) => {
-  const isRunsPage = useLocation().pathname.includes('/runs');
   const { platform } = platformHooks.useCurrentPlatform();
   const openNewWindow = useNewWindow();
   const { gitSync } = gitSyncHooks.useGitSync(
@@ -301,7 +304,7 @@ const FlowActionMenu: React.FC<FlowActionMenuProps> = ({
             </PermissionNeededTooltip>
           )}
 
-          {insideBuilder && !isRunsPage && (
+          {insideBuilder && !viewingRun && (
             <DropdownMenuItem onClick={onVersionsListClick}>
               <div className="flex cursor-pointer  flex-row gap-2 items-center">
                 <GalleryVerticalEnd className="h-4 w-4" />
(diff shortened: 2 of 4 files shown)


majewskibartosz:
## Visual verification

Driven on a local instance (real API + Postgres, seeded flow with production runs), light theme.

| Viewing a run | Edit mode after leaving the run |
| --- | --- |
| run view | edit mode |
| Menu: Rename, Move To, Duplicate, Import, Export, Share, Delete. No Versions, as intended while a run is on the canvas. | Menu: Rename, Move To, Duplicate, **Versions**, Import, Export, Share, Delete. |

States driven: run view (`/runs/<runId>`), Edit flow from the run view, draft builder, flow menu opened in each. Before/after is not observable outside an embed: pre-fix, non-embed edit mode already showed the entry because leaving the run changed the URL. The embed condition (router on `/runs`, no run in state) is covered by the red-first regression tests.

PR #15131, opened 2026-08-28
fix(fillout-forms): new-form-response trigger silently drops all webhooks (double JSON.parse)

### LINKED ISSUE #15149: Fillout Forms new-form-response trigger silently drops every webhook (JSON.parse on already-parsed body)
Reported by customer, Pylon #5806. Customer opened PR #15131 with the fix; PR auto-closed by the unsolicited-PR bot.

## Verified

`packages/pieces/community/fillout-forms/src/lib/triggers/new-form-response.ts`:

```ts
async run(context) {
  const payload = JSON.parse(context.payload.body as string) as { submission: Record<string, any> };
  return [payload.submission];
}
```

`convertBody()` in `packages/server/api/src/app/webhooks/webhook-request-converter.ts` returns `request.body` for non-binary / non-multipart requests. Fastify has already parsed `application/json`, so `context.payload.body` is an **object**. `JSON.parse(object)` stringifies to `"[object Object]"` and throws `SyntaxError: "[object Object]" is not valid JSON`. The `as string` cast hides it from tsc.

Failure is silent end to end:

* engine wraps the throw as `INTERNAL_ERROR` (`packages/server/engine/src/lib/operations/index.ts`)
* `execute-webhook.ts` only submits payloads on `EngineResponseStatus.OK`, so no run is created
* `recordTriggerRun` writes a Redis stat counter only. No run row, no issue, no alert
* job returns `FIRE_AND_FORGET / OK` — worker logs look healthy

`test()` polls `GET /forms/:id/submissions` instead, so the trigger looks fine in the builder while dropping 100% of live submissions.

## Broken since

`ebe302caf2` (2025-07-07, first tag `0.65.0`) replaced the working `return [context.payload.body]` from `78d700cf86` with the `JSON.parse`. Broken in every release since. Not a recent regression.

## Fix

Same as PR #15131:

```ts
const body = context.payload.body;
const payload = (typeof body === 'string' ? JSON.parse(body) : body) as {
  submission: Record<string, any>;
};
```

Bump `@activepieces/piece-fillout-forms` to 0.1.9.

## Scope

Fillout only. Grep for `payload.body as string` across `packages/pieces` returns one other hit, `salesforce/src/lib/trigger/new-outbound-message.ts`, which is `text/xml` — Fastify does not parse it, body is genuinely a string there. Not affected. `respaid` already guards with `typeof`.

## Repro test

Passing locally. Intended repo path `packages/pieces/community/fillout-forms/src/lib/triggers/new-form-response-body-parse.test.ts`, needs a `vitest.config.ts` in the piece dir mirroring the salesforce one:

```ts
import { describe, expect, it } from 'vitest'
import { newFormResponse } from './new-form-response'

const FILLOUT_WEBHOOK_BODY = {
  formId: 'vs1PXaHmRfus',
  formName: 'Contact form',
  submission: {
    submissionId: 'abc123',
    submissionTime: '2026-08-31T10:00:00.000Z',
    questions: [
      { id: '5AtgG35AAZVcrSVfRubvp1', name: 'What is your name?', type: 'ShortAnswer', value: 'John Doe' },
    ],
    calculations: [],
    urlParameters: [],
  },
}

const buildContext = (body: unknown) => ({
  payload: {
    method: 'POST',
    headers: { 'content-type': 'application/json' },
    body,
    queryParams: {},
  },
})

describe('fillout new-form-response run()', () => {
  it('throws on a parsed JSON body, which is what the webhook controller hands it', async () => {
    await expect(newFormResponse.run(buildContext(FILLOUT_WEBHOOK_BODY) as never)).rejects.toThrow(SyntaxError)
  })

  it('only works when the body is a raw string, which application/json never is', async () => {
    const output = await newFormResponse.run(buildContext(JSON.stringify(FILLOUT_WEBHOOK_BODY)) as never)
    expect(output).toEqual([FILLOUT_WEBHOOK_BODY.submission])
  })
})
```

laurentmmeyer:
## The bug

The `new-form-response` trigger never fires on real webhook deliveries:

```ts
async run(context) {
  const payload = JSON.parse(context.payload.body as string) as { ... };
  return [payload.submission];
}
```

Fillout sends its webhooks with `Content-Type: application/json`, so by the time the engine hands the payload to the trigger, `context.payload.body` is already a **parsed object**, not a raw string. `JSON.parse(object)` stringifies it to `"[object Object]"` and throws `SyntaxError: Unexpected token 'o'`. The engine swallows the error as "payload filtered out", so the job reports `outcome: success` and **no run is ever created — silently**. The `as string` cast hides the mismatch from the compiler.

The `test()` function is unaffected (it polls `GET /forms/:id/submissions`), which is why the trigger looks healthy in the builder while dropping every live submission.

## Repro

1. Create a flow with the Fillout **New Form Response** trigger, publish + enable (webhook is registered correctly at Fillout).
2. Submit the form.
3. Webhook is delivered, `EXECUTE_WEBHOOK` job succeeds in the worker logs — but no flow run appears.

Observed on self-hosted 0.88.1; the parsed-body behavior comes from the shared webhook controller, so this should affect Cloud identically.

## The fix

Accept both shapes — parse only when the body is actually a string:

```ts
const body = context.payload.body;
const payload = (typeof body === 'string' ? JSON.parse(body) : body) as {
  submission: Record<string, any>;
};
```

Verified on our instance (patched bundle of 0.1.8): the same webhook payloads that were silently dropped now create runs with the submission as trigger output.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -3141,7 +3141,7 @@
     },
     "packages/pieces/community/fillout-forms": {
       "name": "@activepieces/piece-fillout-forms",
-      "version": "0.1.8",
+      "version": "0.1.9",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -3150,6 +3150,7 @@
       },
       "devDependencies": {
         "tslib": "2.6.2",
+        "vitest": "3.2.6",
       },
     },
     "packages/pieces/community/fireberry": {
--- a/packages/pieces/community/fillout-forms/package.json
+++ b/packages/pieces/community/fillout-forms/package.json
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
     "bundle": "node ../../../../dist/packages/cli/src/index.js pieces bundle",
-    "lint": "eslint 'src/**/*.ts'"
+    "lint": "eslint 'src/**/*.ts'",
+    "test": "vitest run"
   },
   "dependencies": {
     "@activepieces/pieces-common": "workspace:*",
@@ -15,6 +16,7 @@
     "@activepieces/core-utils": "workspace:*"
   },
   "devDependencies": {
-    "tslib": "2.6.2"
+    "tslib": "2.6.2",
+    "vitest": "3.2.6"
   }
 }
--- a/packages/pieces/community/fillout-forms/package.json
+++ b/packages/pieces/community/fillout-forms/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-fillout-forms",
-  "version": "0.1.8",
+  "version": "0.1.9",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
--- a/packages/pieces/community/fillout-forms/src/lib/triggers/new-form-response.test.ts
+++ b/packages/pieces/community/fillout-forms/src/lib/triggers/new-form-response.test.ts
@@ -1,0 +1,69 @@
+/// <reference types="vitest/globals" />
+
+import { newFormResponse } from './new-form-response';
+
+const SUBMISSION = {
+  submissionId: 'abc123',
+  submissionTime: '2026-08-31T10:00:00.000Z',
+  questions: [
+    {
+      id: '5AtgG35AAZVcrSVfRubvp1',
+      name: 'What is your name?',
+      type: 'ShortAnswer',
+      value: 'John Doe',
+    },
+  ],
+  calculations: [],
+  urlParameters: [],
+};
+
+const WEBHOOK_BODY = {
+  formId: 'vs1PXaHmRfus',
+  formName: 'Contact form',
+  submission: SUBMISSION,
+};
+
+const buildContext = (body: unknown) =>
+  ({
+    payload: {
+      method: 'POST',
+      headers: { 'content-type': 'application/json' },
+      body,
+      queryParams: {},
+    },
+  } as never);
+
+describe('fillout new-form-response run()', () => {
+  test('returns the submission when the body is an already-parsed object', async () => {
+    const output = await newFormResponse.run(buildContext(WEBHOOK_BODY));
+
+    expect(output).toEqual([SUBMISSION]);
+  });
+
+  test('still returns the submission when the body is a raw JSON string', async () => {
+    const output = await newFormResponse.run(
+      buildContext(JSON.stringify(WEBHOOK_BODY))
+    );
+
+    expect(output).toEqual([SUBMISSION]);
+  });
+
+  test('returns an array holding exactly one item', async () => {
+    const output = await newFormResponse.run(buildContext(WEBHOOK_BODY));
+
+    expect(Array.isArray(output)).toBe(true);
+    expect(output).toHaveLength(1);
+  });
+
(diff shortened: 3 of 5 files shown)


OdaiAhmed99:
Thanks for this @laurentmmeyer — the diagnosis is exactly right, and I verified the fix end to end on a local instance before reviewing:

- Reproduced the bug live with the unpatched build: webhook delivered, `HTTP 200` returned, **no run created and no error surfaced** — matching your report precisely.
- With your fix: the run is created and the trigger output is the `submission` object.
- Confirmed the blast radius is Fillout only. `salesforce/new-outbound-message` is `text/xml`, which Fastify does not parse, so its `as string` is genuine. `respaid` already guards with `typeof`.

Tracked internally as #15149.

### I pushed tests to your branch

Hope you don't mind — I added a commit (`b26cee6d6c`) with a regression test, a `vitest.config.ts` and the `vitest` devDependency, since the fix had no test and this silently dropped every webhook since `0.65.0`. Five cases: parsed object, raw JSON string, array shape, `{}` without a submission, and a non-JSON string.

They are mutation-proven — reverting the fix fails the object cases with `SyntaxError: "[object Object]" is not valid JSON`, while the raw-string case keeps passing, which is exactly why this went unnoticed for so long.

Please `git pull` before pushing again so you don't lose it.

### Two things needed from you before this can merge

**1. The CLA.** `license/cla` is still pending — *"Contributor License Agreement is not signed yet"*. Nothing can merge until you sign it here:

https://cla-assistant.io/activepieces/activepieces?pullRequest=15131

**2. Two required sections in the PR description.** Our `Breaking change check` job parses the PR body and fails if either section is missing or has no box ticked. Your description does not have them yet — the check has not run on this PR (fork PRs wait for a maintainer to approve workflow runs), so it will fail as soon as it does.

Please paste these at the end of the description and tick exactly one box in each:

```markdown

```

For this change I would tick **no** on both: it restores the trigger's documented behaviour rather than changing a contract, and webhook body parsing is not in the security-sensitive list (auth, secrets, permissions, cryptography, SSRF, file handling). Your call though — the checkbox is yours to tick, and the job only requires that exactly one is selected in each section.


### One thing worth your opinion

The fix handles both real body shapes, but a **bodyless** POST to the webhook URL still throws — `typeof null === 'object'`, so `payload.submission` raises `TypeError: Cannot read properties of null`. That lands on the same silent-drop path as the original bug, and any probe or health-check POST to a public webhook URL would hit it.

Arguably not creating a run when there is no submission is correct, and the real problem is the platform swallowing the throw rather than surfacing it — which is bigger than this piece. So I have deliberately not widened your fix. Happy to hear if you would rather guard it here.

OdaiAhmed99:
Thanks again for this one, @laurentmmeyer — the diagnosis was exact, and it caught a trigger that had been silently dropping every webhook since `0.65.0`.

Closing this PR only because of the CLA: `license/cla` needs every committer on the branch to sign, and without that signature nothing on this branch can merge no matter how green the rest of the checks are. Rather than leave it sitting, we've re-landed your fix from an internal branch so it ships in the next release — same `typeof` guard, plus the regression tests, and you're credited in the PR description and the commit.

Replacement PR: #15247
Tracking issue: #15149

On the bodyless-POST question I raised above: we left it out. A webhook with no submission creating no run is arguably right, and the real problem is the platform swallowing the throw rather than surfacing it — that's bigger than this piece and gets its own ticket.

Appreciate you taking the time to open this. 💜

PR #15166, opened 2026-09-01
feat(gmail): let an approval email carry attachments

alexandronic:
## Description

**Request Approval in Email** is the only action in this piece that composes a message and can't carry a file with it.

Send Email, Create Draft and Update Draft each take an `attachments` array; Reply to Email and Create Draft Reply take a single `attachment` file; Forward Message carries the original message's attachments across. Request Approval in Email builds its mail options the same way as the rest, through `MailComposer`, but exposes no field at all and hardcodes an empty list:

```ts
const mailOptions: Mail.Options = {
  to: receiverEmail,
  ...
  html: htmlBody,
  attachments: [],   // nothing can ever go in here
};
```

So a flow that wants a human to look at a document *before* deciding can't send them the document. The approver gets a bare "Review & Respond" link, and the file has to reach them some other way, which rather defeats the point of a human-in-the-loop gate on a document.

### The change

`attachments` becomes a prop of exactly the shape `send_email`, `create_draft` and `update_draft` already use, so it configures identically to the rest of the piece:

```ts
attachments: Property.Array({
  displayName: 'Attachments',
  required: false,
  properties: {
    file: Property.File({ displayName: 'File', required: true }),
    name: Property.ShortText({ displayName: 'Attachment Name', required: false }),
  },
}),
```

The engine's `props-processor` recurses into an ARRAY-with-`properties` and `fileProcessor` resolves each entry into an `ApFile` before the action runs, so the user can supply a URL, a base64 string, or wire in a file from an earlier step. Building the nodemailer attachments is the same block as `send-email-action.ts`: `name ?? file.filename`, content type from `mime.lookup(file.extension)`, `encoding: 'base64'`.

A message with no attachments (or an empty array) takes the identical path it took before, so nothing changes for a flow that never used the field.

No new i18n keys: `Attachments`, `File` and `Attachment Name` are already the display names `send_email` uses.

## How was this tested?

7 new tests in `packages/pieces/community/gmail/test/request-approval-attachments.test.ts`, following `docs/build-pieces/misc/testing-pieces.mdx`: `createMockActionContext` from the framework, and the same `vitest.config.ts` (alias block included) that `packages/pieces/core/text-helper` uses. The only addition to the shared mock context is the waitpoint pair the approval action needs.

They mock the Gmail client, read the base64url `raw` field the action actually hands to `users.messages.send`, and parse it back with `mailparser` (already a dependency of this piece), so the assertions are against the real sent message rather than the intermediate mail options:

- the attached bytes come out byte-identical (`content.equals(...)`)
- `filename` is the file's name, and the optional **Attachment Name** overrides it
- content type is derived from the extension (`application/pdf`, `image/png`)
- several attachments all travel, in order
- the confirmation link, subject and recipient are unchanged when an attachment is present
- no attachments field, and an empty array, both send a message with zero attachments

Checked the tests actually bite: replacing the action with its `main` version fails 4 of the 7, and the 3 that still pass are exactly the two no-attachment regression guards plus the link check.

Also run locally: `bun install --frozen-lockfile --dry-run` clean against the updated `bun.lock`, `tools/scripts/breaking-change-check.ts` passes with this description, `tsc -p tsconfig.lib.json` clean, and `eslint 'src/**/*.ts'` has 0 errors (40 pre-existing warnings, none in the touched file).

Note that `ci.yml` doesn't currently run `turbo test` for community pieces, so these won't run on CI as things stand, same as the existing `github` and `text-helper` piece tests.

Fixes GIT-1832

The prop is new and optional, and an absent or empty value takes the pre-existing code path. Piece version bumped 0.13.0 → 0.14.0.

Security-sensitive only by the template's definition that it touches file handling. There is no new file handling here: the action consumes an `ApFile` the engine has already resolved, through the same `mime.lookup` / base64 mapping `send_email` has shipped for a long time, and no engine or HTTP-client code is touched. The one behavioural note worth stating plainly is that an approval email can now carry flow files out to an external recipient, which is the same exposure `send_email` already has and is what the field is for.

## Note

Alvys is an existing customer; this came out of a production flow where a document has to be approved by a human before it moves on, and the approver couldn't see the document in the approval mail.

For what it's worth, `microsoft-outlook`'s `request_approval_in_mail` has the exact same gap against its own `send-email` action. Happy to add it to this PR if you'd like the pair fixed together; left out here to keep the change to one piece.


OdaiAhmed99:
Reviewed this against a real Gmail account rather than only the mocks, and it holds up. Every claim in the description checked out, including the one I most wanted to verify: swapping the action back to its `main` version fails **exactly** 4 of the 7 tests, and the 3 survivors are precisely the two no-attachment regression guards plus the link check. The mapping block really is byte-identical to `send-email-action.ts`, the prop shape matches `send_email` field for field, and the vitest config is byte-identical to `text-helper`'s.

**End to end on a live account.** Sent the approval email carrying a real 429-byte PDF served over HTTP, then read the message back out of the mailbox through the Gmail API:

| | |
|---|---|
| delivered filename | `Load 10042 rate con.pdf` — the **Attachment Name** override works |
| content type | `application/pdf`, from the extension |
| size | 429 bytes, byte-identical to what was served |
| bytes fetched back from Gmail | 429, starting `%PDF-` — a genuinely openable file |
| run | `PAUSED`, then `SUCCEEDED` on approve, with the approval link intact beside the attachment |

**Pushed two small additions** to the branch:

- Nine edge cases the original seven do not reach. The one with a security consequence: a newline inside an attachment name **cannot** inject a mail header — nodemailer strips the CR/LF, no `Bcc:` and no `X-Injected:` appear in the built MIME. Also pinned down that an empty **Attachment Name** still leaves the file named (the `??` keeps an empty string, and nodemailer then generates one), that a missing or unrecognised extension becomes `application/octet-stream`, that a non-ASCII name round-trips, that ten attachments keep their order, that threading headers are untouched, and that resuming does not send a second email.
- The `File` translation key. `Attachments` and `Attachment Name` were already present, but `File` was missing — for `send_email` too, so this is a pre-existing gap rather than anything this PR introduced.

---

Two things worth filing separately. **Neither blocks this PR**, and neither is a defect in the code here.

**1. Props are re-resolved on resume, so a vanished attachment kills an already-approved run.**

This action is now the only one in the repo that puts a `Property.File` on a *pausing* action, so it is the first to meet this engine behaviour. `processProps` runs in `piece-context-builder` before `executionType` is even consulted, so every prop is resolved again when the approver clicks, including re-downloading the file the resume branch never touches.

Measured it with a file server that logs each fetch: one hit when the mail is sent, and **a second hit at resume time**. Then repeated it with the server stopped before approving, and the run failed *after* the human approved:

```
{ "attachments": { "properties": [ { "file": [
  "Expected file url or base64 with mimeType, received: http://localhost:8899/rate-confirmation.pdf" ] } ] } }
  at processProps
```

The click is consumed by the POST, so recovery means re-running the step, which emails the approver again. The realistic trigger is an attachment wired from an earlier step, since those are served as signed URLs that expire — an approval left pending longer than the URL's life then fails on approve. An inline base64 file is unaffected, being re-validated rather than re-fetched.

**2. `In Reply To` crashes when the Message-ID cannot be found.**

Unrelated to attachments and already on `main`, in this action and in `send-email-action.ts`: `messages.data.messages?.[0].threadId` is one link short. When the search returns no messages, `messages?.[0]` is `undefined` and reading `threadId` throws `Cannot read properties of undefined`. My first edge-case run tripped over it by accident before I mocked a found thread.

Nice piece of work — the gap it closes is real, and the tests were honest about biting.

OdaiAhmed99 (approved):

PR #15181, opened 2026-09-01
fix(mcp): revoking a grant cuts the client off now, not in 15 minutes

othmanemad:
## Description

Revoking an MCP OAuth grant from the Connections tab set `mcp_oauth_token.revoked = true` and stopped there. That kills the *refresh* flow, which reads the column — but `resolveIdentity` on `POST /mcp` only verified the JWT signature and never consulted the row, so **an already-issued access token kept calling tools for up to fifteen more minutes** while the UI reported the grant revoked the instant the request returned. The API was asserting something it had not made true.

Access tokens now carry the grant's row id as a `grantId` claim, revoking writes a short-lived **Revocation List** in Redis keyed on it, and every `POST /mcp` reads that list before serving.

**Why `grantId` and not `clientId`:** a DCR registration is shared by every user who signed in through that client, so keying the list on `clientId` would have revoked strangers' live tokens. `jti` tracking would have meant a Redis write on every token issue to enable a rare read.

**Both revoke paths write it:** `revokeGrants` (the Connections tab) and the RFC 7009 `/revoke` a client calls on disconnect. Postgres is written first and a Redis failure fails the request — the reverse order would leave a grant that looks revoked for fifteen minutes and then quietly works again, and swallowing the error would return `204` while the token still functions.

**TTL is derived, not typed:** `MCP_OAUTH_REVOKED_GRANT_TTL_SECONDS = MCP_OAUTH_ACCESS_TOKEN_TTL_SECONDS + 60`, so tuning the token lifetime cannot leave the list covering less than the tail of it. The extra minute absorbs clock skew between the node that signed the JWT and Redis.

**Tokens with no `grantId` claim skip the check** — deliberately. Tokens minted before this ships self-heal within fifteen minutes (`refreshAccessToken` stamps the claim, so 30-day refresh tokens are covered too), and `issueInternalAccessToken` tokens for EE chat have no grant row at all.

Also fixes stale copy the revoke dialog and toast were showing — *"Access ends within 15 minutes."* — which is the exact claim this PR makes false.

Reasoning is recorded in [decision 000033](brain/knowledge/decisions/000033-a-revoked-mcp-grant-is-blocked-by-a-redis-revocation-list.md), with a gotcha on the MCP Server page for the three populations that skip the check.

### Behaviour change worth knowing (not a required action)

`POST /mcp` now returns **503 + `Retry-After`** when the revocation list cannot be read, where it previously returned 200. It fails closed on purpose — failing open would silently reinstate the bug — but it answers 503 rather than 401 because a `WWW-Authenticate: error="invalid_token"` challenge makes several MCP clients drop the refresh token and push the user through consent again over a 200 ms hiccup. Redis is already mandatory for the API (BullMQ), so this adds no new dependency and needs no setup.

Platform teardown is a known remaining gap: it bulk-`DELETE`s the rows rather than revoking them, so no keys are written and those tokens stay valid until they expire. Left out deliberately — the platform's data is being erased anyway, and covering it would mean either enumerating every id or a second platform-scoped key that turns the hot-path `GET` into an `MGET`.

## How was this tested?

**Automated** — new integration suite `mcp-oauth-revocation.test.ts` (5 tests, real in-memory Redis):
- revoke → the same bearer immediately gets `401` with `error="invalid_token"` (the regression test for the whole PR)
- revoking one member's grant leaves another member's token on the same `clientId` alive
- an `issueInternalAccessToken` token keeps working after an unrelated revoke
- RFC 7009 `POST /revoke` also blocks the live access token
- a revocation-list read failure answers `503` with `Retry-After` and no `WWW-Authenticate`, not `401`

Unit test pins the key shape and that the TTL exceeds the access-token TTL. All 14 CE MCP suites pass (253 tests). `npm run lint-dev` clean (0 errors, 33 packages).

**Manual, CE (`AP_EDITION=ce`), against a local dev server with real Redis + Postgres** — two grants created through the real OAuth consent page, then:

| Check | Result |
|---|---|
| `grantId` claim in the issued JWT matches the DB row | ✅ |
| Both live grants call `POST /mcp` | ✅ 200 / 200 |
| Revoke one via the Connections tab UI | ✅ row disappears |
| Revoked token's very next call | ✅ **401** `error="invalid_token"` |
| The other grant on the same client | ✅ still 200 |
| Redis key after revoke | ✅ `mcp_oauth:revoked_grant:<id>` = `true`, TTL 917 s of 960 |
| Key for the untouched grant | ✅ absent, exactly 1 key total |
| RFC 7009 `POST /revoke` | ✅ 200, then that token → 401 |
| Internal chat token (no claim) | ✅ 200, unaffected |
| Refresh with a revoked refresh token | ✅ 400 `invalid_grant` |

EE / Cloud paths are unexercised by this change: the endpoint, the entity and the Connections tab are all CE, and the only edition-specific caller is `issueInternalAccessToken` (EE chat), which is covered by the third test above and the manual internal-token check.

Fixes # (issue)

No env var, no schema change, no removed or renamed field, no new setup step. The 503 noted above is a new failure response on a path whose Redis dependency is already mandatory — nothing a self-hoster must act on. Flagging it here rather than burying it, in case a reviewer reads it the other way.

Touches authentication: it adds a claim to the MCP access token and a new authorization check on `POST /mcp`.

- **Risk: keying the list wrong would over-revoke.** Mitigated by keying on the grant row id, never `clientId` — pinned by the cross-member isolation test.
- **Risk: failing open would silently reinstate the vulnerability.** The check fails closed; a lookup error is 503, never a pass-through — pinned by the 503 test.
- **Risk: claim-less tokens are trusted.** Bounded and deliberate (see above): the pre-deploy population drains in 15 minutes, and internal chat tokens have no grant to revoke. Anything that mints an MCP access token without a grant row inherits this exemption, which is written down as a gotcha on the MCP Server page.
- **Risk: a skew-sized hole at the end of the TTL.** Mitigated by the +60 s buffer over the access-token lifetime.


othmanemad:
Fixed the P1 — it was a real bug, and it defeated the exact guarantee this PR is built on.

`multi().exec()` in ioredis resolves with a `[error, result]` tuple per command instead of rejecting, so a single `SETEX` refused by Redis (OOM, a replica in read-only mode) left `putBatch` reporting success. Both revoke endpoints would then have returned `204` with the row marked revoked and the revocation key missing — silently reinstating the 15-minute window.

`assertEveryCommandApplied` now throws on any command-level error. It also handles a case the report didn't mention: `exec()` resolves **`null`** when the transaction is discarded, which is likewise not success. Applied to `putBooleanBatch` too — same latent bug, and it currently has no callers, so making it safe now costs nothing.

Pinned by tests at both levels:
- `distributed-store-batch.test.ts` — a failing command in the tuple array throws, a discarded (`null`) transaction throws, an all-OK batch resolves, an empty batch opens no transaction.
- `mcp-oauth-revocation.test.ts` — `POST /v1/mcp-oauth/grants/revoke` does **not** answer `204` when the revocation-list write fails, which is the user-visible consequence the review called out.

Revocation + grants + token suites and all MCP/database unit suites green (13 files, 120 tests). `npm run lint-dev` 0 errors.

Louai-Zokerburg (approved):

PR #15182, opened 2026-09-01
feat(pieces): outputSchema for Moxie CRM, Cryptolens, Hugging Face and Crypto

OdaiAhmed99:
## Description

Four pieces had no `outputSchema`, so their step output rendered as raw JSON in the data selector and gave agents nothing to reason about. This adds one per describable step — **34 schemas in total** — each authored from a payload captured live against the real API and validated field by field, never from the vendor's reference or the in-source `sampleData`.

Linear: https://linear.app/activepieces/issue/PIE-484/outputschema-for-moxie-crm-crypto-cryptolens-and-hugging-face

| Piece | Version | Schemas | Field paths validated |
|---|---|---|---|
| `moxie-crm` | 0.1.8 → **0.2.0** | 26 — 10 actions + 16 triggers | 489 |
| `cryptolens` | 0.0.8 → **0.1.0** | 5 — 4 actions + 1 trigger | 29 |
| `hugging-face` | 0.1.7 → **0.1.8** | 7 of 8 actions | 241 |
| `core/crypto` | 0.0.27 → **0.0.28** | 1 of 7 actions | — |

Moxie CRM and Cryptolens take a **minor** bump because each also registers actions that already existed in the codebase with no call sites: eight for Moxie (Create Contact, List Invoice Templates, plus List Clients, Search Clients, Search Contacts, Search Projects, List Pipeline Stages and List Workspace Users) and Extend License for Cryptolens.

### Six actions were entirely non-functional, and are fixed here

Neither Cryptolens nor Hugging Face had ever been run against a live API. No payload could be captured until these were fixed, and every fix is proven by a before/after call against the real service.

**Cryptolens — nothing in the piece worked at all:**

- `makeRequest` appended `?token=` unconditionally, but all five callers pass a path that already carries a query string. The URL came out as `…?Name=X&Email=Y?token=XXX`, so the token became part of the previous parameter's *value* and no token was ever sent. Every call answered `401 "Unable to authenticate."`
- Parameters were sent in the query string on `POST` with a JSON content type and an empty body. Cryptolens reads POST parameters from a form body, so it answered `"The input parameters were incorrect."` Non-GET requests now send the query string as an `application/x-www-form-urlencoded` body, with the token kept in the query, so mutations stay POSTs rather than being downgraded to GET.
- The New API Event trigger read `responseBody.Events`, but `ai/GetWebAPILog` returns `{ logs, result, message }`. The guard was therefore always true and the trigger returned an empty list forever, with a `NaN` dedupe timestamp from a field that does not exist. The log record is camelCase and unrelated to the PascalCase interface the piece declared, so the interface, the timestamp field and the sample data now match what the API returns.

**Hugging Face — three of eight actions:**

- Object Detection and Image Classification collected a required model and never sent it to the inference call, using it only to decorate their own output. Every call reached the provider with no model and failed with `"Specifying a model is required when provider is 'auto'"`. Fixed for the standard and zero-shot classification branches alike.
- Text Summarization always set a `generate_parameters` object inside its request parameters, which the API rejects: `"The following model_kwargs are not used by the model: ['generate_parameters']"`. Its two useful values are documented summarization parameters, so `min_length` and `max_length` now sit directly on `parameters`.

### Three shapes that capture corrected against the documentation

- **Cryptolens Create Key** does **not** return `customerId`, even when a customer is associated — a drafted field that was dead and dropped. Its `NoOfKeys > 1` branch, on the other hand, *is* reachable from any flow and returns `{ keys: [{ key, result, message }] }`, so both branches are described.
- **Hugging Face Object Detection** adds a `technical` or `analytics` key only for the matching Output Format, so both are described and each is annotated as conditional.
- **Hugging Face Text Classification** returns `customCategories` in zero-shot mode and omits it for a pre-trained model; one schema covers both branches, with no field absent from every branch.

Object Detection's analytics label distribution is keyed by whichever labels the model detected, so it is described as a dynamic-key map rather than an invented fixed field set.

### Deliberate skips, each with a stated reason

- **Hugging Face Document Question Answering** — every model in its dropdown is unreachable: two return "No Inference Provider available" and the third fails its provider lookup. There was no payload to describe, and the action is unusable as shipped regardless of schemas. Worth its own issue.
- **Crypto, six of seven actions** — hash, hmac, base64 encode/decode, generate password and RSA signature all return a bare string. `OutputSchema` is a list of fields and the empty-string wrapper covers only a top-level array, so a scalar has nothing to describe. Only OpenPGP Encrypt returns an object, and its success/error union is described in full.
- **Moxie, six of 22 triggers**, for three different reasons: the trial plan blocks invoice *emails* (invoice sent, payment received), no API endpoint books a scheduler meeting (3 meeting triggers), and Moxie does not emit the webhook for API-created form submissions (1).
- **Moxie List Invoice Templates** — returns `Promise<string[]>`, a top-level array of bare strings, so there are no fields to list.
- Every piece's inline `custom_api_call`.

## How was this tested?

Per piece, in this order:

1. **Live capture.** Every schema authored from a real payload, not from docs and not from the in-source `sampleData`. Moxie's 22 webhook triggers were fired by hand through ngrok; Cryptolens keys and customers were created against a real product; Hugging Face ran real inference on seven models.
2. **Path validation.** Every field's `value ?? key` resolved against the captured JSON at the correct scope — top level against the root, `children` against the parent object, `listItems` against one array item. 489 + 241 + 29 paths resolve; the only non-resolving paths are the documented either/or branches above, and no field is absent from every captured branch.
3. **Both consumers.** The builder auto-unwraps `body` while the MCP flattener does not, so the flattened paths an agent is handed were checked separately rather than assumed from a green UI.
4. **Lint and unit tests.** `turbo run lint` clean on all four pieces (0 errors). 18 new Moxie client tests plus the existing 20 Crypto tests pass; repo unit suites green.
5. **Builder UI**, after a full stack restart: Moxie and Hugging Face confirmed rendering curated labels, list items labelled by name rather than "Item N", the dynamic-key map, and three levels of nesting, with `description` surfacing as a tooltip.

Edition paths: these are piece-local changes with no server, database, EE or appearance surface, so behaviour is identical on CE, EE and Cloud.

Fixes PIE-484

Nothing here requires action from a self-hoster or API consumer. No field is removed or renamed, no env var or migration is involved, and steps pin an exact piece version, so no existing flow changes behaviour. The six repaired actions could only ever fail before this PR, so no flow can have depended on their previous behaviour.

Two things reviewers should look at deliberately:

- **The Cryptolens fix changes how the API token is transmitted.** It stays in the query string as before, while the other parameters move to a form body on POST. The token was previously mangled into another parameter's value and never authenticated, so this is the first time the piece sends it correctly. No token is logged, and the error path still surfaces only the API's own message.
- **Cryptolens Add Customer returns a `secret`**, the customer's own auth secret, and this PR describes it as a field. It is the endpoint's product rather than a credential of ours, so it is surfaced rather than hidden — flagged here so that choice is reviewed rather than assumed. No `config`, `headers`, token or credential field is described anywhere in the 34 schemas.


OdaiAhmed99:
Good catch on `parameters.generate_parameters.max_length` — following it up turned out to expose a bug rather than a schema omission.

That path only exists because Language Translation nests Max Translation Length inside a `generate_parameters` object, which is the same shape the API rejects for summarization. Because the parameter is optional the action worked fine until someone filled that field, and then it failed every time:

```
Failed to perform inference: The following `model_kwargs` are not used by
the model: ['generate_parameters']
```

Reproduced against the live API — identical call, only `maxLength` differing:

| Input | Result |
|---|---|
| no `maxLength` | SUCCEEDED, `parameters={clean_up_tokenization_spaces, src_lang, tgt_lang}` |
| `maxLength: 40` (before fix) | FAILED, `model_kwargs` error above |
| `maxLength: 40` (after fix) | SUCCEEDED, `parameters={…, max_length: 40}` |

So rather than describe the nested path, `max_length` now sits directly on the request parameters, matching the fix already applied to Text Summarization in this PR — and the schema describes `parameters.max_length`, annotated as present only when the field is set. That brings the piece to **four** repaired actions and the validated path count to 242.

Pushed as c14054fbea.

kishanprmr (approved):

PR #15188, opened 2026-09-01
feat(http): improve the Send HTTP request step form

Talaljaber:
The Send HTTP request form shipped 13 properties in a single flat column, using none of the layout options the builder has gained since. This groups them into Request / Authentication / Body sections, moves six secondary options into Advanced, and renders the two mode selectors as cards.
It also fixes three misleading required markers: headers, queryParams and authType were all required: true, but an empty dictionary is a valid request and "None" is a valid auth choice.
Two notes for reviewers: propertyGroups labels are not translated (pathsToValuesToTranslate doesn't cover them), and timeout gains max: 300 — a UI-only clamp, so existing larger values stay valid.

How was this tested?
Manually in the builder on CE with AP_DEV_PIECES=http: every section, both card selectors, the stepper, and the Advanced section. turbo run build lint --filter=@activepieces/piece-http green. No runtime behaviour change — run() is unchanged apart from a removed comment.

### Breaking change?
- [x] no
- [ ] yes-technical
- [ ] yes-functional

### Security impact?
- [x] no
- [ ] yes-security-sensitive


ibrahim-abuznaid:
Thanks Talal, I pulled this branch locally together with your other PRs and clicked through the form. The direction is right: grouping the fields, fixing the wrong `required` flags and moving the rare options into Advanced all make sense. A few things need another pass before it's good for users.

**Timeout: remove the limit.** Drop `max: 300`. There's no reason for that number, and the platform's own limit is `AP_FLOW_TIMEOUT_SECONDS` (600 by default, self-hosters can raise it). A user who needs 400 can't type it anymore. While you're there, a plain number field fits better than a stepper for a value that ranges into the hundreds.

**Double labels.** The section header says "Authentication" and the first field under it says "Authentication". Same with "Body" and "Body Type". The section box adds space but no information. Either drop the Authentication and Body sections and keep only Request, or rename the inner fields (e.g. "Type") so nothing repeats.

**Cards truncate at the default panel width.** Auth shows "Bearer T...", "Token in t...", "No creden...", "Username ...". Body shows "Structured...", "Fields and ...", and the fourth option "Raw" wraps alone onto a full-width row. Cards you can't read are worse than the dropdown they replaced. For auth type, three obvious options, go back to a dropdown. For body type, either a dropdown too, or keep the cards but with short labels and no descriptions so nothing gets cut.

**Descriptions.** The builder cuts descriptions at 70 characters and adds "show more". Almost every field now hits that, so the form is full of "show more" links. Keep descriptions under 70 characters and put the examples in placeholders, like you already did for URL.

**Overall check.** After the changes, open the step at the default panel width and compare with the current form on main. The new one should be shorter and easier to scan than the old one, not longer. Please attach before/after screenshots to the PR when you push.

Also: the CLA check is still pending on all your PRs, please sign it so we can merge once this is ready.

Talaljaber:
All five addressed, pushed as a second commit on top of the original.

Timeout — uncapped, back to a plain number input with min: 1. The stepper was display: 'stepper', removed.

Double labels — the section header was repeating the field label right below it. Dropped every property group except Request, so Authentication, Body Type and the rest render flat.

Cards truncating — display: 'cards' cut off the longer option text. Both dropdowns are normal selects again.

Descriptions — the builder folds anything over 70 characters behind "read more", which is what you were seeing. All eight rewritten to fit, nothing truncates now.

Overall — before/after attached. Honest read: shorter, but by ~120–170px at the default panel width, not dramatically. The real win is that six rarely-touched options moved to Advanced and Headers/Query params are no longer marked required, so a plain GET is four fields instead of six.

Two deliberate calls: Headers and Query params lost their examples rather than moving them into placeholders — Property.Object renders a DictionaryInput, which takes no placeholder prop. And minimumSupportedRelease moves to 0.88.2 because propertyGroups and advanced need a newer engine.
Screenshot 2026-09-03 145638
Screenshot 2026-09-03 145646
Screenshot 2026-09-03 145703
Screenshot 2026-09-03 145717
Screenshot 2026-09-03 145727

packages/pieces/core/http/src/lib/actions/send-http-request-action.ts
```diff
@@ -172,79 +149,83 @@ export const httpSendRequestAction = createAction({
...
+            description: 'Stored in the flow and visible in exports.',
+            required: false,
+          }),
+        };
       },
     }),
     timeout: Property.Number({
-      displayName: 'Timeout(in seconds)',
+      displayName: 'Timeout',
+      description: 'Seconds to wait before giving up. Empty means no timeout.',
```
ibrahim-abuznaid: Thanks, all five points are addressed and the form reads well now. Verified locally: typed 400 into Timeout, accepted, and no field shows "show more" anymore.

One last copy fix before I approve. "Empty means no timeout" is true for the request itself, but the run still stops at the flow limit (`AP_FLOW_TIMEOUT_SECONDS`, 600 s by default), so a user reading this will expect the request to wait forever. Suggest:

```suggestion
      description: 'Seconds to wait for a response. Empty: up to the flow limit (10 min).',
```

69 characters, so it stays under the fold.

ibrahim-abuznaid (approved):

packages/pieces/core/http/src/lib/actions/send-http-request-action.ts
ibrahim-abuznaid: Pushed this myself in 26295b94ea so we don't hold the merge on a one-liner. Nothing else changed.

PR #15195, opened 2026-09-01
fix(slack): make Message optional on Update / Send-Direct-Message actions (allow blocks-only)

AdamSelene:
## Problem

**Update Message** (`updateMessage`, AI `slack_update_message`) and **Send Message To A User** (`send_direct_message`, AI `slack_send_direct_message`) mark **Message** (`text`) as **required** and always render it as a section block. So you can't send a **blocks-only** message (e.g. a single context block), and a required text field forces a section even when your content is already in `blocks`.

`Post Message` / `Send Message To A Channel` already handle this: `text` is optional and only rendered as a section when present (also serving as the notification fallback). This PR brings the Update / Direct-Message actions in line.

## Change

- Make **Message** optional on the four actions above.
- Build a section from `text` **only when it's provided** (so blocks-only works); `text` is still passed as the notification fallback.
- Require **text or blocks** (can't update/send truly empty).
- `piece-slack` patch bump 0.17.9 → 0.17.10.

Behavior is otherwise unchanged and consistent with Post/Send Message: if you provide **both** `text` and `blocks`, `text` renders as a section above the blocks (that's opt-in, same as the channel actions). To send blocks-only, leave **Message** empty.

Out of scope: `request_action_*` / `request_approval_*` (they build their own buttons and take no user `blocks`).

### Breaking change?
- [ ] Yes
- [x] No

### Security impact?
- [ ] Yes
- [x] No

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "activepieces",
@@ -8343,7 +8344,7 @@
     },
     "packages/pieces/community/slack": {
       "name": "@activepieces/piece-slack",
-      "version": "0.17.9",
+      "version": "0.17.10",
       "dependencies": {
         "@activepieces/core-piece-types": "workspace:*",
         "@activepieces/core-utils": "workspace:*",
@@ -8355,6 +8356,7 @@
       },
       "devDependencies": {
         "tslib": "2.6.2",
+        "vitest": "3.2.6",
       },
     },
     "packages/pieces/community/slashed": {
--- a/packages/pieces/community/slack/package.json
+++ b/packages/pieces/community/slack/package.json
@@ -13,11 +13,13 @@
     "@activepieces/core-utils": "workspace:*"
   },
   "devDependencies": {
-    "tslib": "2.6.2"
+    "tslib": "2.6.2",
+    "vitest": "3.2.6"
   },
   "scripts": {
     "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
     "bundle": "node ../../../../dist/packages/cli/src/index.js pieces bundle",
-    "lint": "eslint 'src/**/*.ts'"
+    "lint": "eslint 'src/**/*.ts'",
+    "test": "vitest run"
   }
 }
--- a/packages/pieces/community/slack/package.json
+++ b/packages/pieces/community/slack/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-slack",
-  "version": "0.17.9",
+  "version": "0.17.10",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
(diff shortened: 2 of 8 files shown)


AdamSelene:
Thanks @greptile-apps — addressed both P1s:

1. **Fallback text still rendered when both `text` and `blocks` are supplied** — this is intentional and consistent with the existing `post-message` / `send-message` actions, which likewise render `text` as a section when it's present. The goal of this PR is to make `text` **optional** so a blocks-only message is possible (leave Message empty); supplying *both* fields and getting both rendered is opt-in, matching the channel actions. I've reworded the prop descriptions so they no longer imply `text` is fallback-only — they now state it renders as a section when provided. (Making it a pure fallback would diverge from `post-message`; happy to align all send/update actions in a follow-up if maintainers prefer that semantic.)

2. **Generated footer bypasses validation** (update-message) — good catch, fixed. The "text or blocks required" check now runs **before** the optional Mention-flow-of-origin footer is appended, so the footer can no longer satisfy the requirement on its own.

AdamSelene:
@greptile-apps why is the confidence score still at 4/5 since you indicate all findings are solved

OdaiAhmed99:
Reviewed this and the change is correct — thanks @AdamSelene. Two things I want to call out as right: you didn't loosen the shared `text` prop in `common/props.ts`, so the four `request_approval_*` / `request_action_*` actions stay untouched (verified at runtime — they still serve `text.required: true`), and the guard-before-footer ordering in `updateMessage` is the right call.

I have three small follow-ups plus the piece's first test suite ready, but I can't push to your branch — GitHub's maintainer-edit doesn't apply to organization-owned forks, so my permission on `alan-eu/activepieces` is `push: false` regardless of the `maintainer_can_modify` flag. If you'd like them on this PR:

```
git remote add odai https://github.com/OdaiAhmed99/activepieces.git
git fetch odai review/slack-15195-followups
git cherry-pick 66804821aa
```

Nothing in it changes behaviour — it's the guard message unified onto the channel actions' wording (your two Update actions threw a different string from the two DM actions), the explanatory comments dropped per CLAUDE.md, the Message descriptions shortened, and `test/blocks-only.test.ts` with 28 cases driving the real `run()` of all four actions against a mocked `WebClient`.

On the descriptions: I first shortened them to the channel action's one-liner and then realised that re-introduces the fallback-only implication you deliberately removed for Greptile's P1. They're now short **and** accurate — "Renders as a section above any Block Kit blocks, and is used as the notification fallback. Leave empty to send blocks only." Entirely optional; skip it if you prefer your wording.

**One thing needed here before this can merge:** the PR description is missing the `### Breaking change?` and `### Security impact?` template sections. `breaking-change-check` errors on a missing section and the workflow fires on `edited`, so please add **both** in a single edit and tick `no` on each — this isn't breaking, since `required: true` → `false` can't invalidate an existing flow.

AdamSelene:
Done, thanks @OdaiAhmed99 !

OdaiAhmed99 (approved):

PR #15197, opened 2026-09-01
feat(flows): sort the automations list by name

### LINKED ISSUE #14946: [Feature]: Sort flows list by name (alphabetical)
## Problem

Flows dashboard has no way to sort by name. Hard to find a specific flow once a project has many. The list is force-sorted by last-modified at two layers and no column header is interactive.

## Current behavior (verified in code)

* Backend order is hard-coded `status DESC, updated DESC` in `flow.service.ts` `list()`. `ListFlowsRequest` (`packages/core/execution/src/lib/flows/dto/list-flows-request.ts`) has no sort/order field.
* Frontend re-sorts client-side by `updated` desc, pins first (`packages/web/src/features/automations/lib/utils.ts`).
* Table headers are static divs; no sort control on any column (`packages/web/src/features/automations/components/automations-table.tsx`).

## Proposed

Add alphabetical (A to Z / Z to A) sort by flow name via a clickable "Name" column header. Default order unchanged.

* Add `sort`/`order` to `ListFlowsRequest`.
* Wire the "Name" header control; stop the unconditional updated-desc client re-sort when a sort is chosen.

## Implementation note

`displayName` lives on the joined `flow_version` table, not on `FlowEntity`. The cursor paginator only orders on base-entity columns, so name sort needs either denormalizing the name onto `FlowEntity` or reworking the paginator to order on a joined column. This is the real work, not just a new enum value.

AhmadTash:
## Description

The flows dashboard had no way to sort by name, which makes finding a specific flow hard once a project has a few hundred of them. The list was force-sorted by last-modified and no column header was interactive.

The **Name** column header is now a three-state control: default order → A to Z → Z to A → back to default. It applies to the whole merged tree, so folders, flows and tables reorder together exactly as they already interleave under the default recency order. When a name sort is active it is fully alphabetical and the enabled-first grouping is dropped, which is what "sort by name" is normally taken to mean. The chosen sort lives in a `sort` URL param, so a shared link and a refresh both preserve it, and "Clear all filters" deliberately leaves it alone (sorting is not filtering).

**Ordering happens in Postgres, before the `LIMIT`.** This is the part that matters: the page fetches a capped window of rows and previously did all sorting client-side, so re-sorting in the browser would only have reordered whichever rows happened to be fetched — a rarely-touched flow sitting past the cap by `updated` could never appear under A–Z no matter what you clicked. Sorting server-side means the database orders the project's entire set and *then* slices, so the ordering is globally correct.

Two details worth flagging for review:

- **Case-insensitivity is deliberate, not incidental.** A raw `ORDER BY` is case-sensitive under PGlite's `C` collation, so Community Edition would have sorted `Zebra` before `apple` while managed Postgres did the opposite. Both flows and tables order on a lowercased expression so the two editions agree.
- **A name-sorted response returns no cursor.** The shared cursor paginator builds its keyset boundary from `status`/`updated`/`id`, so it cannot page correctly under a name order. Rather than hand out a `next` that would produce wrong pages, the response omits it, and `sortBy` combined with a `cursor` is rejected. Nothing in the web app passes a cursor here. Lifting this would mean teaching `Paginator` to carry an expression through both its `ORDER BY` and its cursor comparison — a change to code shared by ~22 services, deliberately out of scope.

No migration, no new column, and no new index. An earlier draft of this denormalised a sort key onto `flow`, but that turned out to buy nothing: `flow.service.list` hits TypeORM's DISTINCT-subquery path, which clears the inner `ORDER BY`, so no index on `flow` could serve the sort anyway. Registering the order on the query builder before `paginate()` works because the paginator only ever calls `addOrderBy` after, leaving the caller's criterion leading and `status`/`updated`/`id` as tiebreaks.

### Known limitation

Rows past the fetch cap remain unreachable, and which rows land in the window now changes with the sort — a late-alphabet flow visible in default order disappears under A–Z while the page count looks identical. A notice appears while sorted to disclose this. Flows inside folders also still sort within a budget shared across all folders; server-ordering that single shared query would starve late-alphabet folders entirely, so it stays client-sorted and is worth its own issue.

## How was this tested?

Automated, on **CE**:

- 16 integration tests in `flow.test.ts` and 2 in `table.test.ts` (the latter run twice via `describeWithAuth`). They cover A–Z, Z–A, the absent cursor, the `sortBy`+`cursor` rejection, and that a sorted row carries no internal sort key in the response.
- One assertion pins the **default** order (`status DESC`, then `updated DESC`). Nothing guarded it before, so "default order unchanged" was previously unenforceable.
- The case-folding test uses names (`Zeta`, `alpha`, `mid`) chosen so it actually fails if the lowercasing is removed — the obvious `Alpha/mid/zeta` fixture passes either way under a `C` collation and proves nothing.
- A vitest over the pure client comparator: mixed case, numeric ordering (`Table 2` before `Table 10`), and pinned-first preserved in both directions.

Manual, against a local instance seeded with **1200 flows and 40 tables** (deliberately over the cap):

- A–Z and Z–A verified against the live API. 461 capitalised names appear *after* the first lowercase one, so the cases genuinely interleave; the returned order matches a case-folded sort exactly.
- Confirmed 170 flows are outside the default window but inside the A–Z window (and 165 for Z–A) — e.g. `Ainvoice 0000` is absent by default and is the first row under A–Z, while search still finds it either way since search also filters before the `LIMIT`.
- Default order confirmed unchanged and still returning a cursor; `sortBy`+`cursor` confirmed 400.

Not verified: EE and Cloud were not exercised. The change is edition-agnostic (no plan flags, no `ee/` code touched), but the PGlite-vs-managed-Postgres collation difference is precisely why the lowercasing is there, so a reviewer on Cloud double-checking the ordering would be welcome.

Fixes #14946

Both new query params are optional and the response is byte-identical when they are absent. `@activepieces/shared` is bumped for the `ListTablesRequest` addition.

The params are a zod `z.enum` on both ends, so the sort direction can never reach SQL as free text, and the ordering expression is a fixed string. Project scoping is untouched.


AhmadTash:
Valid finding, fixed in 22c37da361.

I reproduced it against a live instance rather than reasoning about it. With four flows named `zzTable 1/2/10/100`:

```
SERVER  (LOWER(), lexical):    zzTable 1, zzTable 10, zzTable 100, zzTable 2
CLIENT  (numeric: true):       zzTable 1, zzTable 2,  zzTable 10,  zzTable 100
```

The display mismatch inside a page is the cosmetic half. The half that actually matters is the one you identified: Postgres decides *which* rows fall inside the row limit, so under the client's numeric ordering a row belonging early in the visible sort could be excluded from the window altogether — exactly the class of bug that moving the sort into SQL was meant to eliminate.

Natural sort isn't portable in SQL here (no extensions on managed Postgres, and Community Edition runs on PGlite), so the client was the side to change. The comparator now drops `numeric: true` and uses `sensitivity: 'accent'` rather than `'base'` — `'base'` also folds accents, which `LOWER()` does not, so it was a second, smaller instance of the same divergence. Re-ran the same probe after the change and the two orderings match exactly.

Two unit tests cover it: embedded numbers now assert the lexical order (`Table 10` before `Table 2`), and a new one asserts `Eclair` before `éclair`.

@greptile review

AhmadTash:
The actionable half of this is real and is fixed in 0bd4dd543a. The other half can't be fixed, and I'd rather say why than leave it looking open.

**Real:** `localeCompare` with no locale uses each browser's default, and that genuinely changes the order:

```
en / de   äpple < zebra      chleb < hora
sv        äpple > zebra      chleb < hora
cs        äpple < zebra      chleb > hora
```

Two people viewing the same project could see the same rows in a different order. The comparator now pins the locale, so every browser agrees, and it agrees with the `en_US` collation Cloud's database uses.

**Not fixable:** exact client/server parity, because the two editions' servers don't agree with each other. Cloud runs `en_US.UTF-8`; Community Edition runs on PGlite, whose `C` collation is byte order. Those give opposite answers for the very first pair above — `C` puts `äpple` after `zebra` (0xE4 > 0x7A), matching Swedish rather than English. No single client locale reproduces both, so pinning to `en` narrows the gap to CE-with-non-ASCII-names instead of closing it. Changing the server to `COLLATE "C"` for parity would trade this away for worse ordering on Cloud, where é and CJK currently sort sensibly — a bad trade for an effect that is invisible below the row cap.

So the residual is: on Community Edition, a project over the row cap with non-ASCII names can have window membership that disagrees with the visible order. That is a pre-existing property of the two collations, not something this PR introduces, and it is called out under "Known limitation" in the description.

On the cross-locale test: I added one asserting `äpple` before `zebra`, which pins the intended behaviour and would fail if the explicit locale were removed while CI ran under a Swedish locale. I can't make it fail under an `en` runtime, since the default and the pinned locale agree there — worth being clear that it documents intent more than it guards.

@greptile review

AhmadTash:
I ran this against a real PGlite instance (`datcollate: C`) using the exact expression the server emits, rather than reasoning about it. The specific claim here is incorrect; the general one is right and already documented.

```
NAMES: Eclair, éclair, zebra, äpple, Table 2, Table 10

PGlite  ORDER BY LOWER(COALESCE(name,'')):  Eclair, Table 10, Table 2, zebra, äpple, éclair
client  localeCompare('en', accent):        äpple, Eclair, éclair, Table 10, Table 2, zebra

  Eclair vs éclair -> PGlite: Eclair < éclair | client: Eclair < éclair  AGREE
  zebra  vs äpple  -> PGlite: zebra < äpple   | client: äpple < zebra    DISAGREE
```

**On the flagged test:** the accented expectation (`Eclair` before `éclair`) is exactly what PGlite's C collation produces, so it is not a mismatch — it is the one accent case where the two agree. `LOWER('Eclair')` starts `0x65`, `LOWER('éclair')` starts `0xC3`, so byte order puts `Eclair` first, and English collation independently agrees. `Table 10` before `Table 2` likewise matches on both sides, which is the numeric fix from two rounds ago holding.

**On the general point:** correct, and it is the residual already recorded under "Known limitation". `äpple`/`zebra` is the real shape of it — C collation is byte order, so `ä` sorts after `z`, while every human locale puts it near `a`.

I'm not changing the comparator again, because there is no correct value to change it to. The three options are: match Cloud's `en_US` and mismatch CE's `C` (current), match `C` and mismatch Cloud, or force `COLLATE "C"` server-side and make é and CJK sort after `z` for every Cloud customer. The current choice mismatches the smaller population, and only for non-ASCII names in projects above the row cap. That is a deliberate trade, not an oversight.

Not re-triggering a review on this one, since it is the third pass raising the same inherent limitation and there is no change that would resolve it.

Louai-Zokerburg (approved):

amrdb (approved):

PR #15211, opened 2026-09-02
feat(pieces): add Form.io piece

OdaiAhmed99:
## Description

GIT-1833

Form.io is an open-source form and API platform that public-sector teams commonly self-host. A customer running it on their own infrastructure asked for this, and no piece existed.

**Connection** — a Project URL and an API key. Form.io addresses everything relative to a project rather than a host: a self-hosted server usually serves forms straight off the host (`https://forms.example.gov`) while a hosted project nests them under a project name (`https://xyzabc.form.io`). One URL field covers both. The key travels in the `x-token` header; user-login JWTs are out of scope.

The connection is validated against `GET {projectUrl}/role`, deliberately **not** `/form` — form definitions are readable anonymously, so validating there would accept a wrong key and leave every later step to fail at runtime. `/role` refuses an unauthenticated caller, so a bad key is caught while the dialog is still open.

**Actions** — Create, Get, Find, Update and Delete Submission, List Forms, Get Form, and a custom API call.

Find Submissions exposes Form.io's own query language rather than inventing one: each filter row is a field path, an operator and a value, mapping onto the suffixes the API accepts (equals, not equal, greater/less than, in, regex, exists). It returns the rows alongside the count returned and the total matching, which differ under a limit — Form.io reports the total only in a `Content-Range` header.

Update Submission **merges by default**. Form.io's `PUT` replaces the submission outright, so an update sending one field clears every other one; on a form holding somebody's submitted details that is unrecoverable, while a field left uncleared is not. The raw behaviour is one unticked box away.

**Triggers** — New Submission and Updated Submission, both webhooks.

The webhook triggers register their own Form.io webhook action on enable and delete it on disable, so nobody wires a webhook by hand. The project API key can create and delete an action but cannot list them, so the action id is kept in the trigger store and read back on disable. Each trigger registers its own action scoped to a single event, so a flow only wakes for what it asked for.

The ticket also asked for a polling fallback, and this PR deliberately does not ship one: per @kishanprmr, where a webhook is supported the piece should keep only that, since both do the same work. It is also the right call technically — a webhook needs Form.io to reach Activepieces, while polling needs Activepieces to reach Form.io, which is the same direction all seven actions already need, so polling could never cover the case a webhook cannot.

What that leaves is sample data, and the test button on both triggers fetches the newest submissions on the chosen form rather than making somebody fire a live event to see a payload shape. Updated Submission skips rows whose `modified` still equals their `created`, since Form.io sets the two equal on insert and a never-edited row is a poor sample of an edit. There is deliberately no cursor in that path: with no polling trigger there is nothing to resume from, so carrying one would have been an unreachable code path in a new piece.

**outputSchema** on all seven actions and both triggers, each authored from a payload captured against a live self-hosted instance: 123 field paths, none dead. `data` stays a single field because it holds whatever the form defines, and `metadata.headers` is described down to the four headers Form.io actually keeps — safe because Form.io filters the request headers before storing them, which I confirmed by submitting with `Authorization`, `Cookie` and a custom secret header and finding none of them retained.

## How was this tested?

Against a self-hosted Form.io (`formio/formio:rc` + mongo) with a seeded citizen-intake form, through the real engine rather than mocks.

- **34 action edge cases**: a wrong API key rejected at connect time, a scheme-less URL rejected, a form URL pasted as the project URL rejected, missing required fields, unknown field keys silently dropped by Form.io, malformed ids, all six filter operators, paging with the total read from `Content-Range`, blank filter rows ignored, and delete proving idempotent.
- **Triggers, both directions**: the webhook reaches Activepieces and the payload is the saved submission with none of Form.io's internal request/scope machinery; an update does not fire New Submission and a create does not fire Updated Submission; `onDisable` leaves nothing behind on the form. The test button returns real submissions from the form as sample data.
- **Merge**: with it on, a one-field update keeps the rest; with it off, the omitted fields are cleared, matching the raw API.
- **31 unit tests** covering the client contract and the sample-data path. They were mutation-checked: sorting oldest-first instead of newest fails two of them, and dropping the never-edited filter fails another.
- **Builder UI**: every action and trigger tested from the panel, the form picker offering forms by title, and the output tree reading its curated labels rather than raw keys.

Edition paths: piece-local only, with no server, database, EE or appearance surface, so behaviour is identical on CE, EE and Cloud.

Fixes GIT-1833

A new piece. Nothing existing changes, no env var or migration is involved, and no self-hoster has to do anything.

Ticked deliberately, since three things here deserve a reviewer's attention rather than a wave-through:

- **The connection holds an API key** and sends it in the `x-token` header. It is never described in an `outputSchema`, never logged, and the error path surfaces only Form.io's own message.
- **The project URL is user-supplied and drives every outbound request.** It is normalised once in the client rather than at each call site, and a URL without an `http`/`https` scheme is refused rather than guessed at. Requests run through the piece `httpClient` in the engine sandbox, as with every other piece.
- **`metadata.headers` is described in the output**, which is only safe because Form.io filters the request headers before storing them. Confirmed by submitting with `Authorization`, `Cookie` and a custom secret header: only `host`, `user-agent`, `accept`, `content-type` and `content-length` were retained.


kishanprmr (approved):

OdaiAhmed99:
Both findings were worth raising. The first is a real bug and is fixed; the second is real but cannot be fixed through this API, and I tested that rather than assuming it.

**1. Updated-submission polling could stall behind untouched rows — correct, and worse than a stall.**

The helper advances its cursor to the highest timestamp among the items a poll *returns*, and the updated poll drops never-edited rows. So a page consisting entirely of never-edited submissions returned nothing, the cursor stayed put, and the next poll fetched the same page again. Once a hundred new submissions accumulated past the cursor, every later edit sat behind them and was **never** delivered — on a form that takes many submissions and is occasionally corrected, which is exactly this piece's use case, the trigger would quietly stop working.

Fixed in 688ad252f5: the poll keeps its own high-water mark in the trigger store, advanced to the newest row of the page it *fetched* rather than of the rows it emitted, so progress no longer depends on anything surviving the filter. The helper's cursor still wins when it is ahead, so re-enabling a flow does not replay old rows, and a sample run ignores the mark and leaves it untouched.

You were also right that the tests did not cover it. Six new ones do, including an edit sitting behind a page of untouched rows arriving on the following poll, and each form keeping a separate mark. Checked that they bite: dropping the high-water-mark write fails three of them.

**2. The merged update has a lost-update window — correct, and not closeable here.**

Merge reads the submission and writes it back, so a change committed in between is overwritten. I looked for a way to make the write conditional and tested each option against a live self-hosted instance:

| Attempt | Result |
|---|---|
| Read `__v` and send it back | the API never returns `__v` in a submission at all |
| Send `__v` anyway as a precondition | accepted and ignored — the write succeeds |
| Send a stale `modified` as a precondition | accepted, and the server replaces it with its own timestamp |

There is no version field exposed, no conditional-write header, and no field that acts as a precondition, so nothing the piece does can close the window.

That leaves a choice between two failure modes rather than a fix. With merge **off** there is no read and therefore no race, but every field not supplied is cleared — a *guaranteed* loss on every update. With merge **on** the loss is *possible*, and only when two writers touch the same submission in the same instant. For a form holding somebody's submitted details I would rather risk the rare race than wipe fields on every ordinary update, so merge stays the default — but the window is now stated plainly in the field description and in the AI metadata, with the advice to write disjoint fields with merge off where several flows update one submission (6f90e0fe64).

Happy to flip the default if you would rather the piece never read-modify-write implicitly; it is a one-line change and the behaviour is covered either way.

OdaiAhmed99:
Dropped the two polling triggers in 492d3f743f, and the reasoning behind them was wrong twice over.

On convention: a piece should not ship the same event as both a webhook and a polling trigger, and the repo bears that out — of the community pieces only azure-devops does it. The ones I cited in the original description (copper, campaign-monitor) carry both *kinds* of trigger, but for different events, so quoting them as precedent was misleading.

On the substance, which matters more: a webhook needs Form.io to reach Activepieces, while polling needs Activepieces to reach Form.io — the same direction all seven actions already need. The case I invoked to justify polling was an on-premise Form.io that Activepieces cannot reach, and in that case polling fails too, along with every action. It was never a fallback for that situation.

Polling stays where it belongs, behind the test button for sample data, and it keeps the paging and high-water-mark behaviour the tests cover. The piece is back to two triggers, and the description is corrected.

kishanprmr (approved):

PR #15213, opened 2026-09-02
perf(pieces): stop shipping zod in every piece bundle

abuaboud:
## Description

Every one of our 757 piece bundles shipped **316 KB of zod** that nothing on the piece path ever called.

`@activepieces/pieces-framework` declared 54 top-level zod schemas (`ShortTextProperty`, `PieceMetadata`, `WebhookRenewConfiguration`, …), each sitting next to a hand-written TypeScript twin of the same name. The twins are what the codebase uses; the schemas were left over from the TypeBox → zod migration. esbuild cannot drop a top-level `z.object(...)` call — it has to assume side effects — so every piece that imported `Property` or `PieceAuth` pulled the whole property-schema tree, and the zod runtime with it, and rebuilt it on every piece load.

**Measured, `schedule` piece, cold `require()` of the real bundle:**

| | bundle | heap | rss | load |
|---|---|---|---|---|
| before | 386.7 KB | 2.47 MB | 23.0 MB | 23 ms |
| after | **41.2 KB** | **0.31 MB** | **2.2 MB** | **2.1 ms** |

The engine keeps up to 5 pieces loaded at once (`MAX_LOADED_PIECES`), so that cost was paid five times over in a worker.

More bundles (before → after, and the zod inside them):

| piece | bundle | zod |
|---|---|---|
| discord | 503.7 → 157.9 KB | 316.6 → 0 KB |
| notion | 1484.2 → 1137.9 KB | 316.7 → 0 KB |
| airtable | 1309.2 → 963.3 KB | 316.6 → 0 KB |
| schedule | 386.7 → 41.2 KB | 316.5 → 0 KB |
| slack | 818.6 → 483.4 KB | 316.7 → 10.9 KB |

slack keeps a tree-shaken zod because it calls `propsValidation.validateZod` itself — that is a deliberate piece-author choice and stays.

### What changed

1. **Deleted the 54 dead schema consts** across 31 framework files. The TS twins stay; the 13 types that were `z.infer<typeof X>` are now written out.
2. **The two places that really composed with those schemas moved to the API.** `admin-platform.controller.ts` defines the six small ones (`PropertyGroup`, `ErrorHandlingOptionsParam`, `AiMetadata`, `Audience`, `ActionClassification`, `WebhookRenewConfiguration`) locally, beside the `Action`/`Trigger` schemas it already declares. `app.ts` drops `globalRegistry.add(PieceMetadata, { id: 'piece' })` — that component is an orphan in `docs/openapi.json`, no route schema `$ref`s it.
3. **The engine's one runtime use** of a deleted schema, `DateRangeValue.safeParse` in `props-processor.ts`, is now `dateRangeUtils.isDateRangeValue` — a plain type guard in the framework, covered by the existing engine test. (The engine builds with esbuild, so tsc never flagged this; the test did.)
4. **A gate so it cannot come back.** `bundlePiece` now fails when a bundle contains zod *through first-party code* and the piece's own `package.json` does not declare it. Third-party SDKs that ship their own zod are not the piece's business and are ignored. Two unit tests cover both directions.
5. **Manifests made honest:** the 6 pieces that reach first-party contract modules which build schemas at load (`ai`, `gmail`, `mcp`, `mcp-client`, `oracle-database`, `forms`, `subflows`) declare zod; `elastic-email` declared it without importing it and no longer does. `ai` and `semver` are dropped from the framework manifest — neither is imported anywhere in its source.

### Notes for the reviewer

- Steps I planned but did **not** do, because measurement said they were unnecessary: moving `piecePropertiesUtils.buildSchema` off the framework barrel (it tree-shakes out on its own once nothing else forces zod in — the minimal-piece bundle is zod-free with it in place), and subpath-splitting `core-utils` / `core-piece-types` (their zod modules stop being emitted once the framework stops holding the graph open, except for the 6 pieces above).
- Replacing `propsValidation.validateZod` in the 64 pieces that use it would mean re-implementing zod's `transform` / `pipe` / `refine` / `union` / `record` surface across ~200 call sites, trading 11 KB per piece for real behavioural risk in integrations. Left alone deliberately.

## How was this tested?

- `npx turbo run build --filter='@activepieces/piece-*'` — 763/763 pass (CI builds all pieces because the framework changed).
- `npm run test-unit` — 28/28 tasks, including the engine test that caught the `DateRangeValue` regression.
- `npm run lint-core`, plus lint on `pieces-framework` and `cli` — 0 errors.
- `api`, `web`, `worker`, `engine` builds pass; engine `tsc` error count unchanged from `main` (23, all pre-existing).
- Bundled all 755 pieces through the new gate: 755 pass, 0 fail.
- Before/after bundle and runtime numbers above were taken with the real `bundle-piece-utils` esbuild config.

Fixes # (issue)

`@activepieces/pieces-framework` is not published to npm (pieces inline it at build time — see `validate-publishable-packages.ts`), so removing exported values from it is internal. The only externally visible change is the orphan `piece` schema leaving the OpenAPI components, which nothing referenced.

https://claude.ai/code/session_012DTUh7J4orkUYdNmrqAjeT


abuaboud:
Addressed the review's one finding: `assertZodDeclared` now runs on every forked-entry bundle as well as the main one, with a test that fails if a forked entry ships undeclared zod. Verified against `oracle-database`, the only piece that declares `bundleForkedEntries` — it builds clean (40 KB, forked entry emitted).

AhmadTash (approved):

PR #15217, opened 2026-09-02
fix(builder): give piece property editors an accessible name

### LINKED ISSUE #15055: [BUG]: step-input editors have no accessible name — the field label's for attribute points at nothing
**Describe the bug**

In the builder's step-settings panel, a piece property's input is a rich-text editor rendered as a bare `<div contenteditable="true">`. It has no `id`, no `role`, no `aria-label`, no `aria-labelledby` and no placeholder. Its visible label *is* a real `<label>`, but the `for` attribute names an id that no element on the page carries, so the label labels nothing.

The result is that the most frequently used control in the product has no accessible name. A screen reader lands on an unlabelled editable region; anything that resolves controls by their label — assistive tech, browser autofill, end-to-end tests — cannot find it at all.

The wiring is visible in the source. `packages/web/src/components/ui/form.tsx` mints one id per field and hands it to both halves:

```tsx
// FormItem
const id = React.useId();
// useFormField
formItemId: `${id}-form-item`,

// FormLabel
<Label ... htmlFor={formItemId} {...props} />

// FormControl
<Slot.Root data-slot="form-control" id={formItemId} ... />
```

`Slot.Root` forwards `id` to its single child, and for these fields that child is `TextInputWithMentions`. But its props type accepts no `id`:

```tsx
// packages/web/src/app/builder/piece-properties/text-input-with-mentions/index.tsx
type TextInputWithMentionsProps = {
  className?: string;
  wrapperClassName?: string;
  initialValue?: unknown;
  onChange: (value: string) => void;
  placeholder?: string;
  disabled?: boolean;
  enableMarkdown?: boolean;
  autoFocus?: boolean;
  outputFormat?: 'text' | 'html';
};
```

so the id is dropped on the floor, and `TiptapEditor` sets only a `class` on the editable element (`tiptap-editor.tsx:457-465`, `editorProps.attributes`). Nothing ever puts the id, or any label, on the `contenteditable` div.

**To Reproduce**

1. Sign in, open a flow in the builder, and select a step with text properties — the Storage → `Put` step is a clear case, with `Key`, `Value` and `Store Scope`.

The builder's step settings panel for a Storage Put step, showing the Key, Value and Store Scope fields

2. Read the labels and the fields off the live DOM:

```
LABELS:
  "Key *"          for="_r_3o_-form-item"  document.getElementById(...) -> null
  "Value *"        for="_r_3p_-form-item"  document.getElementById(...) -> null
  "Store Scope *"  for="_r_3q_-form-item"  document.getElementById(...) -> null

[contenteditable="true"] elements: 2
  <DIV> id=null role=null aria-label=null aria-labelledby=null placeholder=null data-placeholder=null
  <DIV> id=null role=null aria-label=null aria-labelledby=null placeholder=null data-placeholder=null

real <input>/<textarea> on the panel: 1
  <input placeholder="Search">      <- the piece search box, not a property
```

All three `for` attributes resolve to `null` via `document.getElementById`.

3. Try to reach a field by its label:

```
getByLabel("Value") -> 1 match; first = <button aria-label="Dynamic value">
fill("Value") FAILED: Element is not an <input>, <textarea>, <select> or [contenteditable]
                      and does not have a role allowing [aria-readonly]

getByLabel("Key")   -> 0 matches
fill("Key")   FAILED: Timeout 4000ms exceeded.
```

`Value` resolves to the neighbouring `Dynamic value` button, whose accessible name merely *contains* the word, so the click lands on the wrong control. `Key` resolves to nothing at all.

4. The control is available by a raw CSS ordinal, which is the check that this is a naming problem and not a broken field:

```
typing into [contenteditable] nth(1) SUCCEEDED — the field accepts input normally
```

**Expected behavior**

The property editor should carry the id the form already minted for it, so that the existing `<label for=...>` resolves:

- add `id` (and ideally `aria-label`) to `TextInputWithMentionsProps` and pass it through to `TiptapEditor`, which can put it on the editable element via `editorProps.attributes` alongside the `class` it already sets there; and
- give the editable element `role="textbox"` and `aria-multiline`, since a `contenteditable` div with no role is not announced as a text field.

That is enough to make `Key`, `Value` and every other piece property announce their name, without changing any behaviour.

**Screenshots**

Included above.

**Additional context**

Version 0.88.3, Community edition, Docker Compose with Postgres and Redis, at `d4cc3eaff77209b773f31e947b3b71306b02059d`. Chromium.

The generated ids in the transcript above (`_r_3o_`, `_r_3p_`, `_r_3q_`) are React `useId` output and will differ per render; the `resolves-to-an-element = null` result is what matters and reproduces every time.

majewskibartosz:
Fixes [GIT-1803](https://linear.app/activepieces/issue/GIT-1803)
Closes #15055

## Problem
1. In the builder step-settings panel, every piece property editor renders as a bare `contenteditable` div with no `id`, `role`, or `aria`. Its visible `<label>` uses the react-hook-form `formItemId` in `for`, but no element on the page carries that id, so the label labels nothing.
2. The most frequently used control in the product has no accessible name: screen readers land on an unlabelled editable region, and anything resolving controls by label (assistive tech, browser autofill, `getByLabel` in e2e tests) either finds nothing or hits the wrong control.

## Fix
- `text-input-with-mentions/index.tsx` — new `FormFieldMentionInput` reads `useFormField()` and passes `id={formItemId}` into the editor; `TextInputWithMentions`/`TiptapEditor` gain an optional `id`.
- `tiptap-editor.tsx` — set `id`, `role="textbox"`, `aria-multiline` on the contenteditable via ProseMirror `editorProps.attributes` (a `Slot`/`FormControl` can't reach it: it targets the editor's wrapper div, not the editable).
- `properties-utils.tsx`, `auto-form-field-wrapper.tsx`, `rich-text-property.tsx` — use `FormFieldMentionInput` at the labelled field sites (static text/number, dynamic-value toggle, rich text).

## Verification
- lint: `turbo run lint --filter=web` clean (0 errors).
- Live (local API + built frontend, Storage → Put): before — `Key`/`Value` label `for` resolves to null, editors `id`/`role` null. After — both labels resolve to their editor; editors carry `id=formItemId`, `role="textbox"`, `aria-multiline="true"`. Editor still accepts input; no new console errors; panel visually unchanged.
- Inverse drive: dynamic-value toggle (`Store Scope` → dynamic mode) also resolves with `role="textbox"`.
- Independent code review: no defects.
- Repro: browser a11y-probe on a live local instance, before=null / after=resolves — full steps in the Reproduction block.
- Visual: before/after DOM label-resolution probe on Storage Put (Key, Value) + dynamic-toggle path — see the Visual verification comment.
- Other affected paths: checked — dictionary/object value inputs and array-item inputs left as plain `TextInputWithMentions` by design (many inputs share one FormItem, so a single `formItemId` would create duplicate ids); `Store Scope` and other Radix `Select` dropdown labels still don't resolve, a separate pre-existing issue on a different control, out of this ticket's contenteditable scope.

<details>

## Plan
- Root cause: `formItemId` (minted by `FormItem`, used by `FormLabel` `htmlFor`) never reaches the mention editor. The text-property path has no `FormControl`, and a `FormControl`/Radix `Slot` would forward the id to `TiptapEditor`'s wrapper div, not the contenteditable — so the id must be threaded explicitly into `editorProps.attributes`.
- Shape: a tiny context-reading wrapper (`FormFieldMentionInput`) supplies `formItemId`; it renders inside the same `<FormItem>` as the label, so the id matches the label's `for`.
- Footprint: 5 files, ~29 lines — rejected: adding an `id` prop with per-caller id plumbing, and wrapping the shared children in `FormControl`.
- Risk radar: (a) no flow-schema impact — pure frontend DOM attributes; (b) not a breaking change — additive optional prop + a11y attributes; (c) no product decision embedded.

- Editor used outside a form (code step, loops, dictionary values) → keeps plain `TextInputWithMentions`, no `useFormField` dependency, cannot throw.
- Dynamic-value toggle mode → swapped to `FormFieldMentionInput`, label resolution verified live.
- Duplicate ids → dictionary/object/array inputs (many per FormItem) deliberately not swapped.

## Alternatives rejected
- Add an `id` prop only (the ticket's prescription) → nothing supplies the value on the text path; the label still resolves to null.
- Wrap children in `FormControl` inside `AutoFormFieldWrapper` → `Slot` targets the editor wrapper div, not the contenteditable, and risks other property types (Slot single-child rule, nested `FormControl`s on the checkbox path).
</details>

<details>

## Environment
- Activepieces at this branch, built frontend + local API against Postgres/Redis. Any flow with a step that has text properties (repro used Storage → `Put`: `Key`, `Value`, `Store Scope`).

## Harness
Paste this into the browser devtools console with a step's settings panel open. It reports, for each field label, whether its `for` resolves and what `role`/`aria-multiline` the target carries.

```js
(() => {
  const labels = Array.from(document.querySelectorAll('label[for]'))
    .filter((l) => /Key|Value|Store Scope/.test(l.innerText))
    .map((l) => {
      const t = document.getElementById(l.getAttribute('for'));
      return {
        label: l.innerText.trim().replace(/\s+/g, ' '),
        forId: l.getAttribute('for'),
        resolves: !!t,
        role: t ? t.getAttribute('role') : null,
        ariaMultiline: t ? t.getAttribute('aria-multiline') : null,
      };
    });
  const editables = Array.from(document.querySelectorAll('[contenteditable="true"]'))
    .map((d) => ({ id: d.id || null, role: d.getAttribute('role'), ariaMultiline: d.getAttribute('aria-multiline') }));
  return { labels, editables };
})();
```

## Trigger
Open a flow in the builder, select a step with text properties (Storage → `Put`), run the snippet.

## Results
- Before this fix: `Key`/`Value` labels → `resolves: false`; both `contenteditable` editors → `id: null, role: null`.
- After this fix: `Key`/`Value` labels → `resolves: true`, target `role: "textbox"`, `aria-multiline: "true"`; editors carry `id` = the label's `for`. `Store Scope` (a Select, not a contenteditable) is unchanged and out of scope.
</details>


majewskibartosz:
## Visual verification

Live on a local instance (built frontend + real API), Storage → `Put` step, with an in-page a11y probe overlay that reports, per field label, whether its `for` resolves and the target's `role`/`aria-multiline`.

| Before | After |
|---|---|
| before | after |

**Before** — `Key`/`Value`/`Store Scope` labels all resolve to `null`; both `contenteditable` editors have `id=null role=null aria-multiline=null`.

**After** — `Key` and `Value` labels **resolve** to their editor; editors carry `id` = the label's `for`, `role="textbox"`, `aria-multiline="true"`. The panel is visually unchanged.

Driven states: default panel; typed input into `Value` (still accepts input); dynamic-value toggle on `Store Scope` (switches to a mention editor that also resolves with `role="textbox"`). `Store Scope`'s static dropdown (a Radix `Select`, not a contenteditable) stays `null` — a separate pre-existing issue, out of this ticket's scope.

othmanemad (approved):

PR #15222, opened 2026-09-02
feat(clay): send rows to a table and trigger on rows Clay sends back

OdaiAhmed99:
## Description

GIT-1369

Adds a **Send Row to Clay Table** action and a **Row Received from Clay** trigger, closing the GTM loop the ticket describes: push a lead into Clay, let Clay enrich it, receive the enriched row back.

**Two corrections to the ticket, since it would otherwise mislead a reviewer.**

It is filed as "New piece — Clay", but `@activepieces/piece-clay` already exists: @kishanprmr shipped it in #15176 on the same day this was picked up, with Search People, Search Companies and a custom API call, and no triggers. So this extends an existing piece rather than creating one, and bumps it `0.0.1 → 0.1.0`.

It also states Clay has "no public REST API". Clay launched one on 9 July 2026 at `api.clay.com/public/v0`, authenticated with a `clay-api-key` header, and this repo's own piece already calls it. The ticket's conclusion still holds, but for a better reason: Clay's docs say plainly *"Can I build or write to Clay tables via CLI or API? No"*, and the `Tables` primitive is read-only and Enterprise-only. A table's webhook source really is the only way to write a row on any plan.

**Send Row to Clay Table** posts the row to that source with the token in `x-clay-webhook-auth`. The token is optional, because a source can exist without one, but Clay answers `401` when a source *has* a token and the request omits it, so the two cases get separate messages naming what to fix — one points at the field to fill in, the other mentions that a refreshed token takes up to a minute to become active.

**Row Received from Clay** is a manual-setup webhook trigger, because Clay exposes no endpoint for registering webhooks: thirteen candidate paths all `404` against a live API key, while `POST /search/query-mode` and `POST /tables/query` answer `400` with field-level validation on the same probe, so the negative result is trustworthy rather than a naming guess.

### The signature, and why a conventional implementation would break

Clay signs deliveries with `x-clay-signature: sha256=<hex>`, an HMAC-SHA256 over the raw body. **The key is the entire secret string including the `whsec_` prefix.** That prefix is the Standard Webhooks / Stripe convention, where implementations strip it and base64-decode the remainder — do the conventional thing here and every genuine Clay request is rejected, which would read as Clay's fault. This was settled by capturing a real signed request and computing all eight plausible key-and-encoding combinations against it; exactly one reproduced the signature. A test pins it with a synthetic secret and a precomputed expected value, and reverting to the conventional derivation fails three tests.

Verification is opt-in for a real reason, not as a nicety: Clay reaches a flow by two routes, and only one of them can sign. A webhook created in Clay's settings signs every delivery; a table's **HTTP API column** is a general-purpose request builder that cannot compute an HMAC at all. A required secret would make that second route — the one that actually sends table rows — impossible to use. The trigger's setup instructions and the field's own description both say so.

Clay also sends a signed verification request the moment a webhook is created, carrying an empty row. That ping starts no run, otherwise enabling the trigger would fire the flow on empty data and, on a flow that writes to a CRM, create an empty record. The check requires the enveloped shape, so a flat payload from an HTTP API column is never mistaken for a ping — an earlier version keyed only on an empty `data`, which silently discarded every delivery from the most common route.

### outputSchema on the action only, deliberately

The action's output is normalised to `{ success, response }` so it is stable, which is what lets it have a schema at all: a source set to "Send response as: Plaintext" replies with the bare string `OK` rather than `{"success": true}`, so `response` passes Clay's own acknowledgement through untouched while `success` stays dependable.

The trigger has none, and adding one would be a mistake. It receives the enveloped `{webhookId, createdAt, data}` from a settings webhook or a flat body from an HTTP API column, and either way the field names are the user's own table columns. Any fixed schema describes somebody's payload wrongly, and the visible symptom is a "friendly view" full of empty values — observed during testing before the schema was removed. With no schema the builder shows the payload that actually arrived. `sampleData` is a flat row, matching the common route.

## How was this tested?

Against a live Clay workspace, not mocks.

- **Action, on two real tables**: correct token accepted; missing token and wrong token each refused with their own message; a scheme-less URL and a blank URL refused; a whitespace-padded URL trimmed and accepted. Rows confirmed present in the table by the account owner, and the rejected ones confirmed absent.
- **Trigger, end-to-end through an ngrok tunnel into a published flow**: signed deliveries produced runs with the row available to later steps; a tampered body and an unsigned request were refused; the creation ping produced no run. The flat HTTP-API-column route was driven from a real Clay table and confirmed by the account owner.
- **36 unit tests**, mutation-checked five ways: switching the HMAC key to the Standard Webhooks derivation fails four, removing the trim on a pasted secret fails four, removing the destination check fails six, loosening it to a substring match fails three, and reverting the flat-payload ping fix fails one.
- **Neither component needs a Clay connection.** Raised by review: neither reads `context.auth`, yet both forced the user to create an API-key connection — a key that is still beta in Clay — before Clay could be pointed at the flow. Both are now `requireAuth: false`, following `discord`'s `send-message-webhook` (a webhook URL as a step prop) and `kallabot-ai`'s `call-events` (a manual-setup webhook trigger on an authed piece). The piece's two existing search actions still require it, since they call the API.

Edition paths: piece-local only, with no server, database, EE or appearance surface, so behaviour is identical on CE, EE and Cloud.

**Left unproven rather than guessed:** `aiMetadata.idempotent` is `false`. Whether Clay upserts a repeated identical row or appends another one could not be established from the API — the acknowledgement is the same either way — so the conservative value stands. It matters because an agent told a write is idempotent will retry it, and a wrong `true` would quietly duplicate rows in somebody's GTM data.

Fixes GIT-1369

Two additive steps on an existing piece. The existing actions, its auth and any existing connection are untouched, and no env var or migration is involved.

Ticked deliberately, since three things here deserve a reviewer's attention:

- **Webhook signature verification** is implemented here, with the non-obvious key derivation described above. Comparison is `timingSafeEqual` over equal-length buffers, and a signature that is absent, empty, wrong-length or unparseable is refused rather than throwing. Verifying against `payload.rawBody` is load-bearing: a re-serialised body changes the bytes and fails verification, so a test asserts that a parsed body is rejected.
- **It proves origin, not freshness.** Clay sends no timestamp and signs the body alone, so a captured request stays valid indefinitely and can be replayed. Worth knowing rather than assuming otherwise.
- **The send destination is constrained to Clay.** The row and the source token both travel to whatever URL the step carries, so validating only the scheme would have let a mistyped host, or one resolved from an upstream expression, receive them. An https address on `clay.com` or a subdomain of it is required, and anything else is rejected before a request is attempted. Raised by review; the permissive version had a test asserting it, on the reasoning that Clay might change its URL shape later, which is the wrong trade against a silent credential leak.
- **The source's auth token is a step property, so it lives in the flow's settings rather than an encrypted connection.** The framework has no masked step property, and the piece's existing auth holds a workspace API key while this token is per-table, so folding it into the connection would mean one connection per table. Reviewed and agreed to merge as is, since step-prop webhook URLs are how the Slack and Discord webhook actions work too. Note also that the webhook URL alone is *not* sufficient to write to a source, since a token-bearing source refuses an unauthenticated POST. **Correcting an earlier claim in this description:** it said both routes were "recorded in the ticket for a follow-up decision". @ibrahim-abuznaid checked, and neither GIT-1369 nor the synced GitHub issue carries that note — so the follow-up needs a PIE ticket rather than being already captured. It should cover the connection-based route (the framework accepts an array of auth types, so a "Table webhook source" `CustomAuth` variant could sit beside the API-key auth without a breaking change), plus the two existing search actions' missing `outputSchema` and their 25 optional filters with no `advanced: true`.

### DIFF
--- a/packages/pieces/community/clay/package.json
+++ b/packages/pieces/community/clay/package.json
@@ -1,16 +1,20 @@
 {
     "name": "@activepieces/piece-clay",
-    "version": "0.0.1",
+    "version": "0.1.0",
     "main": "./dist/src/index.js",
     "types": "./dist/src/index.d.ts",
     "scripts": {
         "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
-        "lint": "eslint 'src/**/*.ts'"
+        "lint": "eslint 'src/**/*.ts'",
+        "test": "vitest run"
     },
     "dependencies": {
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
         "@activepieces/shared": "workspace:*",
         "tslib": "2.6.2"
+    },
+    "devDependencies": {
+        "vitest": "3.2.6"
     }
 }
--- a/packages/pieces/community/clay/package.json
+++ b/packages/pieces/community/clay/package.json
@@ -11,7 +11,6 @@
     "dependencies": {
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
-        "@activepieces/shared": "workspace:*",
         "tslib": "2.6.2"
     },
     "devDependencies": {
--- a/packages/pieces/community/clay/src/index.ts
+++ b/packages/pieces/community/clay/src/index.ts
@@ -4,18 +4,21 @@ import { PieceCategory } from '@activepieces/shared';
 import { clayAuth } from './lib/auth';
 import { searchCompaniesAction } from './lib/actions/search-companies';
 import { searchPeopleAction } from './lib/actions/search-people';
+import { sendRowToTableAction } from './lib/actions/send-row-to-table';
+import { rowReceivedTrigger } from './lib/triggers/row-received';
 
 export const clay = createPiece({
     displayName: 'Clay',
-    description: 'Search Clay\'s GTM database for people and companies.',
+    description: 'Search Clay\'s GTM database, and move table rows in and out of Clay.',
     minimumSupportedRelease: '0.36.1',
     logoUrl: 'https://cdn.activepieces.com/pieces/clay.png',
     categories: [PieceCategory.SALES_AND_CRM],
     auth: clayAuth,
-    authors: ['kishanprmr'],
+    authors: ['kishanprmr', 'odaithalji'],
     actions: [
         searchCompaniesAction,
         searchPeopleAction,
+        sendRowToTableAction,
         createCustomApiCallAction({
             baseUrl: () => 'https://api.clay.com/public/v0',
             auth: clayAuth,
@@ -24,5 +27,5 @@ export const clay = createPiece({
             }),
         }),
     ],
-    triggers: [],
+    triggers: [rowReceivedTrigger],
 });
--- a/packages/pieces/community/clay/src/index.ts
+++ b/packages/pieces/community/clay/src/index.ts
@@ -1,6 +1,5 @@
-import { createPiece } from '@activepieces/pieces-framework';
+import { createPiece, PieceCategory } from '@activepieces/pieces-framework';
 import { createCustomApiCallAction } from '@activepieces/pieces-common';
-import { PieceCategory } from '@activepieces/shared';
 import { clayAuth } from './lib/auth';
 import { searchCompaniesAction } from './lib/actions/search-companies';
 import { searchPeopleAction } from './lib/actions/search-people';
@@ -14,7 +13,7 @@ export const clay = createPiece({
     logoUrl: 'https://cdn.activepieces.com/pieces/clay.png',
     categories: [PieceCategory.SALES_AND_CRM],
     auth: clayAuth,
-    authors: ['kishanprmr', 'odaithalji'],
+    authors: ['kishanprmr', 'OdaiAhmed99'],
     actions: [
         searchCompaniesAction,
         searchPeopleAction,
--- a/packages/pieces/community/clay/src/lib/triggers/row-received.ts
+++ b/packages/pieces/community/clay/src/lib/triggers/row-received.ts
@@ -1,0 +1,89 @@
+import {
+    createTrigger,
+    Property,
+    TriggerStrategy,
+} from '@activepieces/pieces-framework';
+import { clayAuth } from '../auth';
+import { clayWebhook } from '../common/webhook';
+
+export const rowReceivedTrigger = createTrigger({
+    auth: clayAuth,
+    name: 'row_received',
+    displayName: 'Row Received from Clay',
+    description: 'Triggers when Clay sends a row to this flow.',
+    aiMetadata: {
+        description:
+            'Fires when Clay posts a row to this flow, carrying whatever fields the Clay side was configured to send. Clay has no API for registering webhooks, so somebody has to point a Clay webhook or an HTTP API enrichment column at this trigger\'s URL by hand before it receives anything.',
+    },
+    type: TriggerStrategy.WEBHOOK,
+    props: {
+        setupInstructions: Property.MarkDown({
+            value: `
+Clay has no API for registering webhooks, so point Clay at this flow yourself. Copy this URL:
+
+\`\`\`text
+{{webhookUrl}}
+\`\`\`
+
+Then use whichever route suits you. They behave differently, so the choice matters.
+
+**To send rows from a table - add an HTTP API column.** This is the usual choice.
+
+- Set the method to \`POST\` and paste the URL into **Endpoint**. Put nothing else in that field: typing \`/\` there opens Clay's column picker, and an inserted column silently corrupts the URL.
+- Set the **Body** to the columns you want, for example \`{"Domain": "/Domain"}\`. With an empty body nothing useful arrives.
+- Leave **Signing Secret** below **empty**. Clay's HTTP API column cannot sign a request, so a secret there rejects every delivery.
+
+**To receive workspace events - create a webhook in Clay's settings.**
+
+- Paste the URL into Clay's **Webhook URL** field.
+- Clay immediately sends a signed verification request carrying an empty row. This trigger ignores it, so it will not start a run.
+- Optionally paste that webhook's \`whsec_...\` secret into **Signing Secret** below, and every delivery's signature is then checked.
+
+The fields you receive are whatever Clay sends, so they follow your own table's columns rather than a fixed shape.
+            `,
+        }),
+        signingSecret: Property.ShortText({
+            displayName: 'Signing Secret',
+            description:
+                'Only for a webhook created in Clay\'s settings, which signs each delivery with an x-clay-signature header. Paste that webhook\'s whsec_... secret to have every delivery verified and anything unsigned rejected.\n\nLeave this empty when Clay reaches this flow through a table\'s HTTP API column. That column is a general-purpose HTTP request builder and cannot compute a signature, so a secret here would reject every delivery it sends.',
+            required: false,
+        }),
+    },
+    sampleData: {
+        Domain: 'activepieces.com',
+        Company: 'Activepieces',
+    },
+
+    async onEnable() {
+        return;
+    },
+
+    async onDisable() {
+        return;
+    },
+
+    async run(context) {
+        const { signingSecret } = context.propsValue;
+
+        if (signingSecret) {
+            const verified = clayWebhook.verifySignature({
+                signingSecret,
+                rawBody: context.payload.rawBody,
+                signatureHeader: clayWebhook.signatureHeaderOf(
+                    context.payload.headers,
+                ),
+            });
+            if (!verified) {
+                throw new Error(
+                    'The x-clay-signature header did not match the signing secret, so this request was not accepted. Check that Signing Secret matches the secret Clay showed for this webhook, including its whsec_ prefix.',
+                );
+            }
+        }
+
+        if (clayWebhook.isVerificationPing(context.payload.body)) {
+            return [];
+        }
+
+        return [context.payload.body];
+    },
+});
--- a/packages/pieces/community/clay/src/lib/triggers/row-received.ts
+++ b/packages/pieces/community/clay/src/lib/triggers/row-received.ts
@@ -3,11 +3,9 @@ import {
     Property,
     TriggerStrategy,
 } from '@activepieces/pieces-framework';
-import { clayAuth } from '../auth';
 import { clayWebhook } from '../common/webhook';
 
 export const rowReceivedTrigger = createTrigger({
-    auth: clayAuth,
     name: 'row_received',
     displayName: 'Row Received from Clay',
     description: 'Triggers when Clay sends a row to this flow.',
@@ -16,6 +14,7 @@ export const rowReceivedTrigger = createTrigger({
             'Fires when Clay posts a row to this flow, carrying whatever fields the Clay side was configured to send. Clay has no API for registering webhooks, so somebody has to point a Clay webhook or an HTTP API enrichment column at this trigger\'s URL by hand before it receives anything.',
     },
     type: TriggerStrategy.WEBHOOK,
+    requireAuth: false,
     props: {
         setupInstructions: Property.MarkDown({
             value: `
@@ -63,7 +62,9 @@ The fields you receive are whatever Clay sends, so they follow your own table's 
     },
 
     async run(context) {
-        const { signingSecret } = context.propsValue;
+        const signingSecret = clayWebhook.trimmedSecretOf(
+            context.propsValue.signingSecret,
+        );
 
         if (signingSecret) {
             const verified = clayWebhook.verifySignature({
--- a/packages/pieces/community/clay/src/lib/triggers/row-received.ts
+++ b/packages/pieces/community/clay/src/lib/triggers/row-received.ts
@@ -18,33 +18,25 @@ export const rowReceivedTrigger = createTrigger({
     props: {
         setupInstructions: Property.MarkDown({
             value: `
-Clay has no API for registering webhooks, so point Clay at this flow yourself. Copy this URL:
+Clay cannot register a webhook for you, so paste this URL into Clay yourself:
 
 \`\`\`text
 {{webhookUrl}}
 \`\`\`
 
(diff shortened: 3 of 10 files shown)


OdaiAhmed99:
Fixed in 388cdd47e4 — the finding was right, and my original reasoning for the permissive version was the wrong trade.

The row and the source token both travel to whatever URL the step carries, so validating only the scheme meant a mistyped host, or one resolved from an upstream expression, would receive them. The destination is now required to be an https address on `clay.com` or a subdomain, and it is rejected before any request is attempted.

Worth admitting why it was loose: I deliberately declined to check the host, on the reasoning that Clay might change its URL shape and a hardcoded host would break the piece. That weighs a clear, one-line failure against a silent credential leak, which is backwards. A rejected URL tells the user exactly what to fix; a token posted to the wrong host tells nobody anything.

The permissive behaviour even had a test asserting it ("http is accepted, since a source may be proxied"), which is a reminder that a test can pin a bad decision as easily as a good one. That test is gone, replaced by seven covering lookalike domains (`notclay.com`, `evil-clay.com`, `clay.com.evil.example`), loopback, plain http on a Clay host, and that no request is attempted when the destination is refused. Removing the check fails six of them; loosening it to a substring match fails three.

Verified live afterwards: the real Clay source still accepts rows, and a non-Clay URL is refused with the host named in the message. 32 tests green.

Still open and flagged in the description rather than fixed here: the token lives in the step's settings rather than an encrypted connection, because the framework has no masked step property and the piece's existing auth holds a workspace-wide API key while this token is per-table. That one needs a decision from @kishanprmr, whose piece this extends.

packages/pieces/community/clay/src/index.ts
```diff
@@ -4,18 +4,21 @@ import { PieceCategory } from '@activepieces/shared';
...
 export const clay = createPiece({
     displayName: 'Clay',
-    description: 'Search Clay\'s GTM database for people and companies.',
+    description: 'Search Clay\'s GTM database, and move table rows in and out of Clay.',
     minimumSupportedRelease: '0.36.1',
     logoUrl: 'https://cdn.activepieces.com/pieces/clay.png',
     categories: [PieceCategory.SALES_AND_CRM],
     auth: clayAuth,
-    authors: ['kishanprmr'],
+    authors: ['kishanprmr', 'odaithalji'],
```
ibrahim-abuznaid: `odaithalji` is not a GitHub account (404). `authors` is the GitHub login by convention, and your QuickBooks Desktop piece already uses `OdaiAhmed99`. Form.io has the same slip and needs a follow-up.

packages/pieces/community/clay/src/lib/triggers/row-received.ts
```diff
@@ -0,0 +1,89 @@
+import {
+    createTrigger,
+    Property,
+    TriggerStrategy,
+} from '@activepieces/pieces-framework';
+import { clayAuth } from '../auth';
+import { clayWebhook } from '../common/webhook';
+
+export const rowReceivedTrigger = createTrigger({
+    auth: clayAuth,
```
ibrahim-abuznaid: This trigger never reads `context.auth`, but as written it forces the user to create a Clay API-key connection (still "beta" in Clay) before they can point a table at the flow. `requireAuth: false` fixes that; the YouTube and Kallabot triggers are precedent. The action doesn't use the key either, so the same could apply there, but that one is more of a judgment call.

packages/pieces/community/clay/src/lib/triggers/row-received.ts
```diff
@@ -0,0 +1,89 @@
...
+    async onEnable() {
+        return;
+    },
+
+    async onDisable() {
+        return;
+    },
+
+    async run(context) {
+        const { signingSecret } = context.propsValue;
```
ibrahim-abuznaid: Worth a `.trim()` here and on `authToken` in `sendRow`. The URL is trimmed but the secrets are not, and the engine's text processor does not trim. A pasted trailing space then surfaces as "signature did not match" or "token does not match", which points the user at the wrong cause.

packages/pieces/community/clay/src/lib/triggers/row-received.ts
```diff
@@ -0,0 +1,89 @@
...
+        if (signingSecret) {
+            const verified = clayWebhook.verifySignature({
+                signingSecret,
+                rawBody: context.payload.rawBody,
+                signatureHeader: clayWebhook.signatureHeaderOf(
+                    context.payload.headers,
+                ),
+            });
+            if (!verified) {
+                throw new Error(
```
ibrahim-abuznaid, line 70: Non-blocking, just so it's a known trade: throwing here records a FAILED trigger run in the piece's stats for every rejected request. Cursor's trigger does the same, so it is consistent. It will get noisy if someone sets a secret and then points an HTTP API column at the flow, which the setup text already warns about.

packages/pieces/community/clay/package.json
```diff
@@ -1,16 +1,20 @@
...
     "scripts": {
         "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
-        "lint": "eslint 'src/**/*.ts'"
+        "lint": "eslint 'src/**/*.ts'",
+        "test": "vitest run"
     },
     "dependencies": {
         "@activepieces/pieces-common": "workspace:*",
         "@activepieces/pieces-framework": "workspace:*",
         "@activepieces/shared": "workspace:*",
```
ibrahim-abuznaid: Pre-existing from #15176, but since this file and `index.ts` are both open: `PieceCategory` is re-exported from `@activepieces/pieces-framework`, and the piece lint rule bans `@activepieces/shared` everywhere else (this piece's `.eslintrc.json` has the generator's empty rules, which is why lint passes). Dropping the import and this dependency also removes the need for the `@activepieces/shared` alias in `vitest.config.ts`.

packages/pieces/community/clay/test/webhook.test.ts
```diff
@@ -0,0 +1,337 @@
+/// <reference types="vitest/globals" />
+
+const sendRequest = vi.fn();
+
+vi.mock('@activepieces/pieces-common', () => ({
```
ibrahim-abuznaid: Nit: mocking the whole module means the first time `webhook.ts` imports anything else from `pieces-common`, the suite fails with an unrelated-looking error. Coupa and Workday do `vi.spyOn(httpClient, 'sendRequest')` instead, which survives new imports.

ibrahim-abuznaid (approved):
Approving. This is well built and the security reasoning holds up. A few small items below that I'd like in before merge, plus one design question I'm deliberately parking.

**What I checked**
- CI built the piece on this head (`piece-clay:build` in the `main` job). Lint and all 32 tests pass locally. CI runs neither for pieces, so worth saying that they ran.
- Everything the code relies on exists on `main`: `classification`, trigger `aiMetadata`, `rawBody` on the trigger payload, `tryCatch`. The server only sets `rawBody` for string-parsed bodies, so HMAC over raw bytes works for Clay's JSON. The engine turns an empty optional text prop into `undefined`, so the two 401 messages branch correctly.
- Greptile's destination finding is properly closed in 388cdd4.

**Token in a step prop vs a connection**
Merge as is. Step-prop webhook URLs are how the Slack and Discord webhook actions work too. The description says the two routes are recorded in the ticket, but neither GIT-1369 nor the synced GitHub issue has that note, so let's open a PIE ticket for the connection-based route. The framework accepts an array of auth types (Intercom does OAuth-or-token this way), so a "Table webhook source" `CustomAuth` variant could sit beside the API-key auth without a breaking change. That is also what the original ticket scoped.

**Not for this PR, noting for the piece:** the two search actions have no `outputSchema` and 25 optional filters with no `advanced: true`. Same follow-up ticket.

packages/pieces/community/clay/src/index.ts
OdaiAhmed99: Fixed in 39512dc. `OdaiAhmed99`, matching the QuickBooks Desktop and MSSQL pieces. Form.io has the same slip and I'll fix it on its own branch rather than reaching into it from here.

packages/pieces/community/clay/src/lib/triggers/row-received.ts
OdaiAhmed99: Fixed in 39512dc, and applied to the action too.

Trigger: `auth` dropped and `requireAuth: false`, following `kallabot-ai`'s `call-events` — same shape, a manual-setup webhook trigger on a piece that has auth. Confirmed it does what you describe: `showAuthForTrigger` in `piece-settings/index.tsx` gates the connection field on it, and `buildSchema` stops requiring the auth prop.

On the action, which you left as a judgment call — I took it, because `discord`'s `send-message-webhook` is the exact precedent: a webhook URL as a step prop, `auth` omitted, `requireAuth: false`, on a piece that has auth. Same reasoning as yours for the trigger, and one more: the credentials it actually uses are the webhook URL and the per-table source token, so a workspace API key it never sends was pure friction. Keeping `auth: clayAuth` alongside `requireAuth: false` would have been worse than either — `context.auth` typed as present while the UI never collects it.

packages/pieces/community/clay/src/lib/triggers/row-received.ts
OdaiAhmed99: Fixed in 39512dc. You were right that the wrong-cause message was the real cost.

Rather than two `.trim()` calls I put one `trimmedSecretOf` in `webhook.ts`, used by the trigger for the secret and by `sendRow` for the token, so it is testable and a future caller cannot forget it.

It also picks up a second bug I had not seen: whitespace alone now normalises to `undefined`, not `'   '`. Before, a stray space in Signing Secret was truthy, so verification switched on with an effectively-empty key and rejected **every** delivery. Same on the token — a whitespace-only value used to send a junk header and get the "token does not match" 401, when the right message is the one naming the field to fill in.

Four tests cover it, and removing the trim fails all four.

packages/pieces/community/clay/src/lib/triggers/row-received.ts
OdaiAhmed99: Agreed, and left as is — noted as a known trade rather than fixed. Cursor's trigger behaves the same way, and a rejected request that leaves no trace is worse than a noisy one.

The trim above does take the sharpest edge off it: a stray space in Signing Secret used to fail every delivery, which would have produced exactly that flood of FAILED runs from a cause the message pointed away from.

packages/pieces/community/clay/package.json
OdaiAhmed99: Fixed in 39512dc, including the part that let it through.

`PieceCategory` now comes from `@activepieces/pieces-framework`, and the `@activepieces/shared` dependency and the `vitest.config.ts` alias are both gone.

Dropping the import alone would not have held, though — you had already identified why lint passed. The piece's `.eslintrc.json` now carries the same restricted-imports block as `formio`, `coupa` and `microsoft-sql-server` instead of the generator's empty rules. Verified it bites: re-adding a `@activepieces/shared` import errors with `'@activepieces/shared' import is restricted from being used by a pattern`, where before it was silent.

packages/pieces/community/clay/test/webhook.test.ts
OdaiAhmed99: Fixed in 39512dc, and it was more than a nit — the suite is now importing the real `pieces-common`, which the module mock had been standing in for.

`vi.spyOn(httpClient, 'sendRequest')` per Coupa and Workday. Assertions read the request through a typed `HttpRequest` recorded by `mockImplementation` rather than `mock.calls[0][0] as HttpRequest`, which keeps it off the `as` cast the repo bans, and `.method` now asserts against the real `HttpMethod.POST` instead of the mock's string.

36 tests green, and every mutation check still holds: the HMAC-key mutation now fails four rather than three, since the pasted-secret test depends on the same derivation.

OdaiAhmed99:
All six items from the review are in **39512dc**, plus the one you left as a judgment call. Thanks — two of them turned out to be covering real bugs rather than tidiness.

| Note | What landed |
|---|---|
| `odaithalji` is not a GitHub account | `OdaiAhmed99` |
| Trigger forces a connection it never reads | `auth` dropped, `requireAuth: false` — and the same on the send action |
| Secrets not trimmed | one `trimmedSecretOf` in `webhook.ts`, used by both sides |
| Throwing records a FAILED run | left as is, as a known trade |
| `PieceCategory` from `@activepieces/shared` | from `pieces-framework`; dep, vitest alias **and the empty eslint rules** all fixed |
| Whole-module mock | `vi.spyOn(httpClient, 'sendRequest')` |

**Two of these were hiding a bug.** The trim: whitespace alone was truthy, so a single stray space in Signing Secret switched verification on with an effectively-empty key and rejected *every* delivery — the flood of FAILED trigger runs from your fourth note, from a cause its message pointed away from. Whitespace now normalises to `undefined`. And the eslint note: dropping the import alone would not have held, since you had already found why lint passed. The piece now carries the standard restricted-imports block, and I verified it errors on a re-added `@activepieces/shared` import where it used to be silent.

36 tests (was 32). Every mutation check still holds, and the HMAC one now fails four rather than three because the pasted-secret test depends on the same derivation. Lint and build green via turbo.

**One correction to my own description, which you caught.** It claimed both auth routes were "recorded in the ticket for a follow-up decision". You checked; they are not. I have rewritten that paragraph rather than leaving it standing, and it now lists what the PIE ticket should carry: the `CustomAuth`-variant route for the per-table token, and the two search actions' missing `outputSchema` and 25 unflagged optional filters. I have not opened the ticket yet — say the word and I will, or it is yours if you would rather scope it.

Form.io has the same `authors` slip; I will fix it on its own branch rather than reaching into it from here.

ibrahim-abuznaid (approved):
Re-approving on 39512dc. Checked each item on this head rather than from the summary: lint clean, 36/36 tests pass locally, CI green with `piece-clay:build` in the log, and I confirmed the new eslint rule actually errors on a re-added `@activepieces/shared` import. Good catch on the whitespace-only secret, that was a real bug.

Please open the PIE ticket yourself, scoped as you wrote it in the description: the `CustomAuth` variant for the per-table token, plus `outputSchema` and `advanced: true` on the two search actions. Link this PR from it. The Form.io `authors` fix on its own branch is fine.

@kishanprmr this extends your piece, a look from you would be good before merge.

PR #15239, opened 2026-09-03
refactor(mcp): share the retention sweep and the listing helpers

othmanemad:
## Description

Pure refactor, no behaviour change. Lifts two helpers out of their single caller so a second caller can use them, and nothing else. **1 of 3** in the MCP Activity stack — split out so the feature PRs behind it read as new code rather than a mix of moved and new.

**The project-aware retention sweep** moves from `agentRetention` to `helper/retention/execution-data-retention.ts`. It was already generic: group projects by their `executionDataRetentionDays` override, then delete older-than in bounded passes so a backlog drains across the hourly schedule instead of one statement holding locks and WAL. Now it takes `repo` / `alias` / `scopeCondition` and `agentRetention` calls it.

**The listing helpers** move from the OAuth grants listing to `mcp/mcp-listing-utils.ts` — the platform-wide-or-scoped project filter, the batched project-name and member lookups, and `resolveUserIdFilter` (which narrows a non-privileged caller to their own rows). `mcp-oauth-token.service.ts` and `mcp-oauth-grants.controller.ts` call them instead.

Net +64 lines: the two new files are almost entirely offset by what leaves the two callers.

## How was this tested?

Both helpers are already covered by the tests of the callers they came from, and those pass unchanged — that is the point of the split:

- `mcp-oauth-grants.test.ts` — 19 tests
- `agent-retention.test.ts` — 3 tests
- `tsc --noEmit` and `turbo lint --filter=api` clean

Edition paths: the retention helper is called from CE (`file.module.ts`, via `agentRetention`) and the listing helpers from the CE grants controller. No EE-only path is involved and no import crosses into `src/app/ee/` from CE.

Fixes # (issue)

Internal code movement only. No schema, no endpoint, no env var, no response shape changes.

`resolveUserIdFilter` is an authorization boundary — it returns `null` for a privileged user (see everything on the platform) and the caller's own id otherwise. Moving it to a shared module means a future bug there widens data exposure for every caller at once instead of one. Flagged for that reason, not because the logic changed: it is byte-identical to what the grants controller ran, its two call sites are unchanged, and `mcp-oauth-grants.test.ts` covers both the privileged and unprivileged paths.


lina-alomari (approved):
Approving — the refactor holds up. I went looking for breakage in the retention move specifically and couldn't find any.

**How I tested it**

Existing suites first: `agent-retention` (3) and `mcp-oauth-grants` (19) pass unchanged, and the full CE MCP integration folder is green at 284. `tsc` and lint clean on all five files, no import cycle from the new `helper/retention/` module.

Then ~29 extra tests against a real DB, since the existing retention unit test mocks `projectRepo.find` to `[]` and so never executes the project-scoped pass — the one branch whose SQL you actually changed:

- **Retention:** deletes past the default boundary, never touches a `CHAT` conversation at any age, honours a project-scoped shorter retention, keeps one project's short retention away from another project's rows, and sweeps project-less rows on the default boundary only. The `alias."projectId"` form resolves correctly against the real schema.
- **Listing:** `projectIds` undefined / empty / platform-wide sentinel / sentinel mixed with a real id / foreign-platform id / unknown id. `findProjectNames` and `findMembers` both refuse rows from another platform.
- **`resolveUserIdFilter`** (the bit you flagged): member narrowed to their own grants, operator privileged and sees everyone, and a member revoking someone else's grant gets a 403 with the row left unrevoked.

One thing worth knowing about the code you moved: the project-scoped pass is unreachable under default config. `getEffectiveExecutionDataRetentionDays` computes `min(30, max(x, 30))`, which is always 30 and therefore always `>= EXECUTION_DATA_RETENTION_DAYS`, so every project gets filtered out of the grouping. I had to set `AP_PAUSED_FLOW_TIMEOUT_DAYS=1` to reach it at all. Pre-existing and not yours to fix here — just flagging it since you now own the helper.

**One small thing, totally fine by me**

`clientKey` on the access token JWT (`mcp-oauth-token.service.ts` 42, 202, 239, 299) is new code rather than moved code. Nothing reads it, so it's inert — I confirmed tokens minted by current `main` still authenticate on both `/mcp` and `/mcp/platform`, with a bad-token 401 as a control. It's just the one part of this PR that isn't a lift, and PR 2 is where it gets consumed. Move it or leave it, I don't mind — not blocking the approval.

PR #15240, opened 2026-09-03
feat(mcp): record what a client ran against the MCP server

othmanemad:
## Description

**2 of 3** in the MCP Activity stack — stacked on #15239, review that first. This one writes the rows; the API that serves them is #15241.

Nothing has persisted MCP tool calls since `mcp_run` was dropped in `DropLegacyTables1766015156683`. `MCP_TOOL_CALLED` telemetry fires on flow tools only and isn't queryable. This adds `mcp_activity`, shaped around the two ways the V1 table failed.

**Exactly one tool is recorded: `ap_run_action`.** Not a hint-derived predicate over 26 tools — a literal title match. An agent session is mostly `ap_create_flow` / `ap_add_step` / `ap_update_step`, and a row per edit answers nothing a user asked. `ap_run_action` is the only tool that runs a real connector step against a real connected account, which is the question the feed exists to answer. Widening it later needs no migration, since `toolName` is still stored.

**The row is thin and carries no payload.** V1 stored `input` and `output` as two non-null JSONB columns written synchronously per call. Here the row is denormalized (`pieceName`, `actionName`, `connectionExternalId`) so the future list query does zero joins, and the payload goes to the `file` table under a new `MCP_CALL_PAYLOAD` type, capped at 128KB and ZSTD-compressed.

**The write is off the hot path.** `withActivityRecording` returns the tool result *before* the row and file are written, via `rejectedPromiseHandler` — the same idiom as the `MCP_TOOL_CALLED` telemetry beside it. The activity context is a thunk rather than an awaited value specifically because on the platform server the project is only known after a Redis read (`mcpProjectSelection.get`); resolving it eagerly would put that read back on the request path while looking like it didn't. The trade is a row lost to a crash mid-write, which is acceptable for a feed and would not be for an audit log — see the decision page in `brain/knowledge/decisions/`.

**Retention exists on day one.** `audit_event` at 362M rows / 475GB unpruned is what "we'll prune it later" looks like. Rows and payload files both ride the hourly `FILE_CLEANUP_TRIGGER` already scheduled, under the existing `AP_EXECUTION_DATA_RETENTION_DAYS` and its per-project override, via the sweep extracted in #15239. No new env var, no new cron, nothing for a self-hoster to configure.

### Three things worth a reviewer's attention

**Recording cannot live in `PermissionChecker.wrapExecute`.** That's the obvious place — it already wraps every static tool — and it's wrong: CE uses `ALLOW_ALL`, whose `wrapExecute` is the identity function. A recorder hung there records nothing in Community Edition and everything in Cloud, which reads as a data bug rather than an edition bug. `withActivityRecording` composes *around* the already-permission-wrapped `execute` at the registration sites instead.

**`platformId` comes from the token, never from `mcp.platformId`.** `mcp_server.platformId` is NULL on every `PROJECT` row and can never be set — `idx_mcp_server_platform_id` is UNIQUE, so two project servers in one platform would collide. Reading it to build the activity context is why an earlier revision recorded nothing at all on the project path. The access token already carries `platformId` for both server types, so `resolveIdentity` threads it through `buildServer`.

**The selected project is read once per call, not twice.** On the platform path the selection is mutable Redis state — an uncached read, scoped per *user* rather than per session, on a stateless transport where concurrent requests share one key. An earlier revision resolved it inside `execute` and then again from the recorder after the response had gone out, so a selection that moved during the seconds an `ap_run_action` engine run takes put the row and its payload file in a project the action never touched, invisibly. The handler now resolves it once and builds both the executor and the activity context from that one value; `withActivityRecording` keeps its contract, and the platform path has a single read site, so the invariant is structural rather than a convention. Consequence worth noting: a call made with **no project selected** now records nothing, where it used to write a `FAILED` row with a null `projectId` plus a stored payload that no project-scoped query could surface.

**`status` is only as good as the tool's `isError`, and the `❌` glyph is not the flag.** Any failure path returning error text without `isError: true` records as `SUCCEEDED`. `mcpUtils.validateAuth` had exactly that bug — its return type was an inline content shape with no such field — and now returns `McpToolResult | null` with the flag set. All five call sites already returned it as an early error, so nothing else changed.

## How was this tested?

- `mcp-activity-recording.test.ts` — 12 CE integration tests driving **real tool calls** through an in-memory MCP client, not inserted rows. That matters: the context is nullable and `record()` returns silently on null, so asserting on rows you inserted yourself would have passed while the feature recorded nothing. Covers the project-server path, client attribution, the throw-still-records path, a rejected `auth` argument landing as `FAILED`, hostile model-written arguments (NUL bytes, 400-char names) surviving the `varchar(256)` insert, and the read-only and formerly-recorded mutating tools writing nothing.
- `mcp-activity-recorder.test.ts` — 17 unit tests: the record/skip predicate, canonical piece-name storage, and every branch of the payload cap staying under 128KB — including an oversized *input*, which an output-only truncation would have let through.
- `mcp-activity-recording.test.ts` also carries the **first integration coverage of the platform MCP path** — there was none. It spies the selection read to return the right project once and the wrong one thereafter, then asserts the row names the project the action ran in and that the read happened exactly once. Verified to fail on the previous commit for that exact assertion.
- No regressions: the whole CE MCP suite (266 across 14 files) and the MCP unit suite (86).
- `tsc --noEmit` and `turbo lint --filter=api` clean.

Edition paths: CE and EE both exercised. CE is the one that matters here — it uses `ALLOW_ALL`, exactly where a recorder hung off the permission checker would have silently no-opped.

Not verified locally: the payload path against a real S3 backend. With `AP_FILE_STORAGE_LOCATION=DB` (the local default) it takes the bytea path; `MCP_CALL_PAYLOAD` is registered in `isExecutionDataFileThatExpires`, so it routes to S3 wherever that's configured.

Fixes # (issue)

Additive throughout: a new table, a new `FileType` member. The migration creates a table and runs no destructive DDL, so the class carries `breaking = false`. `buildServer` gains two optional parameters.

**The risk:** this persists piece-action inputs and outputs to the file store. An `ap_run_action` input is fully-resolved values and an output is whatever the third-party API returned, so this is new data-at-rest that can contain sensitive content.

**Mitigations:** nothing reads it in this PR — the endpoint arrives in #15241, where the access rules are reviewable in one place. Payloads expire on the existing execution-data retention rather than living forever. And this is not a new *class* of exposure: `FLOW_RUN_LOG` already persists step inputs and outputs under exactly this retention and storage path. Auth is never among the stored bytes — `ap_run_action` takes a `connectionExternalId`, a reference the engine resolves later, not a credential. Flagged rather than waved through because the 128KB cap and the retention window are the only things bounding it.


packages/server/api/src/app/mcp/tools/flow-run-utils.ts
```diff
@@ -141,8 +141,8 @@ export async function executePieceActionRun({
     offload?: ActionRunOffload
     log: FastifyBaseLogger
 }): Promise<McpToolResult> {
-    const { auth: inlineAuth, ...inputWithoutAuth } = input ?? {}
-    const effectiveExternalId = connectionExternalId ?? (typeof inlineAuth === 'string' ? inlineAuth : undefined)
+    const inputWithoutAuth = omit(input ?? {}, ['auth'])
+    const effectiveExternalId = mcpUtils.resolveConnectionExternalId({ connectionExternalId, input })
 
     const authError = mcpUtils.validateAuth(effectiveExternalId)
```
lina-alomari, line 142: You fixed `validateAuth`'s missing `isError` here — but the six failure paths below this line still return bare `content`, so every one of them records as `SUCCEEDED` with `errorMessage: null`:

- `:191` / `:194` — `diagnosis.unknownKeys` / `missing` / `invalidEnums` (the most common model-authored failures)
- `:226` — `parsedAction` validation failure
- `:241` / `:249` — both `TIMEOUT` branches
- `:259` — `INTERNAL_ERROR`
- `:277` — **the failed-run path**, `formatPieceActionRunResult`'s `❌ … failed (run …)` + `structuredContent.errorSummary`

`grep -c isError` on this file returns 0 after this PR. `:277` is the one that matters: the vendor 401'd, the Slack message wasn't sent, and the feed shows it green. `runFlowAsTool` already has the idiom — `...(isOkay ? {} : { isError: true })`.

Both tests that assert `FAILED` hit the two helper paths you fixed, and nothing drives a run to a terminal outcome, so CI stays green and `SUCCEEDED` is never asserted against the DB at all. A unit test on `withActivityRecording` with `execute` stubbed to return `{ content, structuredContent: { errorSummary } }` and no `isError` pins this in ~10 lines, no engine needed.

packages/server/api/src/app/database/migration/postgres/1841000000000-AddMcpActivity.ts
```diff
@@ -0,0 +1,70 @@
+import { QueryRunner } from 'typeorm'
+import { Migration } from '../../migration'
+
+export class AddMcpActivity1841000000000 implements Migration {
```
lina-alomari, line 4: `1841000000000` is already taken on `main` by `AddPlatformConfiguration1841000000000`. Needs `1842…` in the filename, the class name, the `name` field, the import in `postgres-connection.ts` and the `getMigrations()` entry.

This is half of what `mergeable: CONFLICTING` is reporting.

packages/core/shared/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/shared",
-  "version": "0.159.0",
+  "version": "0.160.0",
```
lina-alomari, line 3: `main` is already at `0.161.0`, so this wants `0.162.0`. Other half of the conflict — same rebase.

lina-alomari (changes requested):
Two blocking things, both left inline.

Everything else I found is non-blocking and I've kept it off the PR to stay reviewable — the bigger ones are worth a conversation rather than a thread: `status` collapsing both `TIMEOUT` branches into one value (the "may have partially completed, don't re-run blindly" case is the row a user would actually act on), `projectId` being nullable-but-unreachable while #15241 ships a `platform-wide` filter and a test that depend on the NULL state, and `capPayload` reimplementing `largeResultUtils.fitToBudget` — which `flow-run-utils.ts:2` already imports. Ping me and I'll walk through them.

One more for the rebase pass: the decision page is numbered `000033`, which `main` already uses (it's at `000034`).

lina-alomari (approved):
Approving — the correctness blocker is properly fixed.

`92dc73a48e` went further than the ask: rather than sprinkling `isError` at the call site, `formatPieceActionRunResult` now returns `McpToolResult` directly, so the flag is set where the outcome is known. Verified all seven returns in `executePieceActionRun` carry it, the refactor has a single production caller with `PieceActionRunResult` cleanly deleted, and `createMockPieceMetadata` accepts every field the new fixture passes. The unit tests pin failed / failed-without-message / successful without needing an engine, and the integration test drives a real `ap_run_action` into `diagnosis.missing` and asserts `FAILED`. The brain gotcha is on the right page under `### Gotchas`. I did not run the suites — CI is the check there.

**Still to do before merge, not review objections:** the branch needs re-stacking onto the rebased #15239 (its base was force-pushed, so the PR sits on the old `6e4780d92c` and the diff currently reads across #15239's rebase — that's the `CONFLICTING` state), and the three numbers renumbered against whatever `main` looks like at that moment: the migration timestamp, `@activepieces/shared`, and the decision page. Worth re-running the size check after, since no number is meaningful until then.

Two small things I'd take if you're touching the file anyway, neither blocking:

- `executeFlowTest` still has five unflagged error returns (`:33`, `:37`, `:46`, `:95`, `:104`). Correctly out of scope — different tool, not recorded — but the new brain bullet says to grep the file for `isError` before trusting a derived status, and the file isn't clean. One clause noting `executeFlowTest` is unconverted saves the next reader the trip.
- Nothing writes a `SUCCEEDED` row to the database in any test. Much lower risk now the formatter is pinned both ways, so: nice-to-have.

Carried over and still open, for the record rather than as conditions: `capPayload` duplicating `largeResultUtils.fitToBudget` and its 128KB constant (`flow-run-utils.ts:2` already imports both), the recorder's single `log.warn({ error }, …)` carrying no entity groups on a path designed to lose rows, and the recorder exporting four loose functions instead of a grouped const.

The one I'd genuinely like settled in the stack rather than after it: `projectId` is nullable in the migration and the schema but unreachable by any writer, while #15241 ships a `platform-wide` filter and a test that depend on the NULL state. Either the column is `NOT NULL` here or that filter is unreachable UI there — worth picking one before both land.

PR #15241, opened 2026-09-03
feat(mcp): serve the recorded calls as an Activity feed

othmanemad:
## Description

**3 of 3** in the MCP Activity stack — stacked on #15240, review that first. #15240 writes the rows; this serves them. Frontend follows separately.

Adds `GET /v1/mcp-activity` and `GET /v1/mcp-activity/:id/payload` over the `mcp_activity` rows the recorder already writes — the list of what a connected AI client actually ran, filtered by project, member, client, status and date range.

**Access follows the grants endpoint exactly.** Platform-scoped, and `resolveUserIdFilter` (shared in #15239) narrows a non-privileged caller to their own rows. A platform ADMIN or OPERATOR sees the platform; everyone else sees themselves.

**No count endpoint.** A `COUNT(*)` over a table designed to grow is the shape that produced the `audit_event` statement timeout (GIT-1705). The design's "1,284 runs" is dropped rather than served — the feed is cursor-paginated newest-first.

**The connection name is hydrated at read time, keyed by `(projectId, externalId)`.** The row stores the `connectionExternalId` the client asked for, unresolved — `ap_run_action` never holds an `AppConnection`, it embeds `{{connections['...']}}` and the engine resolves it later — so a hallucinated id is recorded honestly with no lookup on the write path. It can't be shown as-is, since a UI-created connection's externalId defaults to `apId()` and the feed would print a random 21-char string. The lookup is batched after pagination, like `findProjectNames`. **Keying that map by externalId alone would name the wrong account**: `idx_app_connection_platform_id_and_external_id` is not unique and `app_connection` has no `projectId` column, so two projects in one platform can hold the same externalId.

**Every filter is rejected at the edge, not by PostgreSQL.** `createdAfter`/`createdBefore` use a new shared `OptionalIsoDateTimeQuery` (`@activepieces/core-utils`): `z.iso.datetime({ offset: true })` plus a year bound, because the ISO grammar's `\d{4}` year accepts `0000`, which PostgreSQL has no representation for — `?createdAfter=0000-01-01T00:00:00Z` would otherwise reach the database and return a 500. Binding a JS `Date` instead looks like the cleaner fix and is not one: the entity-bound query builder ISO-stringifies it again before the driver sees it, so the range has to be rejected in the schema. `projectIds` is typed as an `ApId` or the `platform-wide` literal rather than a bare string, so a cleared `?projectIds=` returns 400 like its sibling filters instead of parsing to `['']` and matching nothing — and the sentinel is now part of the request contract rather than a runtime check.

## How was this tested?

- `mcp-activity.test.ts` — 23 CE integration tests: platform isolation, unprivileged members narrowed to their own rows, every filter, cursor paging newest-first, the `platform-wide` project filter, retention deleting past the boundary while sparing what's inside it, and the payload route scoped by both platform and member — removing either filter fails a test.
- `mcp-activity-recording.test.ts` — grows from 14 to 17, adding the endpoint-backed cases: the payload served back, the connection named on a listed entry, a hallucinated connection listing with a null name, and the two filter indexes existing (those indexes ship in #15240; this diff has no entity or migration change).
- No regressions: `mcp-tools` (100), `mcp-oauth-grants` + `mcp-oauth-token` (22), MCP unit suite (89).
- `tsc --noEmit` and `turbo lint --filter=api` clean.

Edition paths: CE and EE both exercised — the controller is CE and the privilege check is the same `isUserPrivileged` gate the grants endpoint uses on both.

Not verified locally: the payload read against a real S3 backend (local default is the bytea path).

Fixes # (issue)

Two new endpoints and two new indexes. Nothing removed, renamed or made required.

**The risk:** this is the read side of the payload store #15240 introduced. `GET /v1/mcp-activity/:id/payload` returns fully-resolved piece-action inputs and third-party API outputs — sensitive business data, now reachable over HTTP.

**Mitigations:** both routes are `securityAccess.publicPlatform([PrincipalType.USER])` and filter on `req.principal.platform.id`, so a row is unreachable across platforms. `resolveUserIdFilter` is applied to the payload endpoint too, not just the list — an unprivileged user passing another member's activity id gets a 404, not their payload. The same rule as the grants endpoint, and covered by the isolation tests above.


packages/core/shared/src/lib/automation/mcp/mcp-activity.ts
```diff
@@ -25,4 +26,51 @@ export const McpActivity = z.object({
...
+
+export const ListMcpActivityRequestQuery = z.object({
+    cursor: z.string().optional(),
+    limit: z.coerce.number().int().min(1).max(100).optional(),
+    projectIds: OptionalArrayFromQuery(z.string()),
+    memberIds: OptionalArrayFromQuery(ApId),
+    clientKeys: OptionalArrayFromQuery(McpOAuthClientKey),
+    statuses: OptionalArrayFromQuery(McpActivityStatus),
+    createdAfter: z.iso.datetime({ offset: true }).optional(),
+    createdBefore: z.iso.datetime({ offset: true }).optional(),
```
hazemadelkhalel: Blocking. zod accepts year `0000` but Postgres rejects it, so `?createdAfter=0000-01-01T00:00:00Z` binds straight into `created >= :createdAfter` and throws `QueryFailedError`. That is a 500 plus a Sentry event on every such request. Both fields need a year bound.

packages/core/shared/src/lib/automation/mcp/mcp-activity.ts
```diff
@@ -25,4 +26,51 @@ export const McpActivity = z.object({
...
+    durationMs: z.number(),
+    hasPayload: z.boolean(),
+})
+
+export type PopulatedMcpActivity = z.infer<typeof PopulatedMcpActivity>
+
+export const ListMcpActivityRequestQuery = z.object({
+    cursor: z.string().optional(),
+    limit: z.coerce.number().int().min(1).max(100).optional(),
+    projectIds: OptionalArrayFromQuery(z.string()),
```
hazemadelkhalel: `?projectIds=` is how a UI usually sends a cleared filter, and it parses to `['']`, which becomes `projectId IN ('')` and returns nothing instead of skipping the filter. `memberIds`, `statuses` and `clientKeys` return 400 instead, because their schemas reject an empty string. Worth settling before the frontend is built against it.

packages/server/api/test/integration/ce/mcp/mcp-activity.test.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        it('404s on another members activity for an unprivileged member', async () => {
+            const member = await createMemberContext(app!, ctx, { projectRole: DefaultProjectRole.ADMIN })
+            const mine = await insertActivityRow({ userId: ctx.user.id, projectId: ctx.project.id })
+
+            const response = await member.get(`/v1/mcp-activity/${mine}/payload`)
+
+            expect(response.statusCode).toBe(404)
+        })
+    })
```
hazemadelkhalel, line 307: Blocking. `createMemberContext` puts the member on `parentCtx.platform.id`, so this passes on the `userId` filter alone. Remove `platformId` from `getPayload` and nothing here fails.

That matters because `resolveUserIdFilter` returns null for a privileged caller, which leaves `platformId` as the only tenant guard on this route. `list` has `never leaks activity from another platform`; the payload route needs the same.

packages/server/api/src/app/mcp/activity/mcp-activity-service.ts
```diff
@@ -1,12 +1,167 @@
...
+            throw new ActivepiecesError({
+                code: ErrorCode.ENTITY_NOT_FOUND,
+                params: { entityType: 'mcp_activity', entityId: id, message: 'No payload stored for this activity' },
+            })
+        }
+        const file = await fileService(log).getDataOrThrow({
+            fileId: activity.payloadFileId,
+            type: FileType.MCP_CALL_PAYLOAD,
+            ...(isNil(activity.projectId) ? {} : { projectId: activity.projectId }),
+        })
```
hazemadelkhalel, line 88: `spreadIfDefined('projectId', activity.projectId)` replaces this ternary, and the same helper fits the `userId` spread in the `findOneBy` above. It already handles null internally, and `listGrants` uses it exactly this way in `mcp-oauth-token.service.ts:187`.

Separate point on the same line: when `projectId` is null the key drops out, and TypeORM ignores undefined where-keys, so the query becomes `WHERE id AND type`. `getDataOrThrow` never filters on `platformId` in any branch. Nothing reaches this today, but the column is nullable and the tests already insert such rows. Passing `platformId: activity.platformId` unconditionally closes it.

packages/server/api/src/app/mcp/activity/mcp-activity-service.ts
```diff
@@ -1,12 +1,167 @@
...
+        const { input, output } = StoredPayload.parse(JSON.parse(file.data.toString('utf-8')))
```
hazemadelkhalel: A corrupt payload file throws `SyntaxError` or `ZodError` here and surfaces as a 500 rather than a handled error. `tryCatch` per the repo convention; nine other files in this module already use it.

`StoredPayload` is two `z.unknown()` fields, so it validates nothing. `McpActivityPayload.parse({ ...JSON.parse(...), truncated: activity.payloadTruncated })` gives you one schema instead of two, and it checks the shape you actually return.

packages/core/shared/src/lib/automation/mcp/mcp-activity.ts
```diff
@@ -25,4 +26,51 @@ export const McpActivity = z.object({
...
+
+export const McpActivityPayload = z.object({
+    input: z.unknown(),
+    output: z.unknown(),
+    truncated: z.boolean(),
+})
+
+export type McpActivityPayload = z.infer<typeof McpActivityPayload>
+
+export const GetMcpActivityPayloadParams = z.object({
```
hazemadelkhalel: Nit. Over 45 routes inline `params: z.object({ id: ApId })`. A cross-package export for a single route costs more than it saves.

packages/server/api/test/integration/ce/mcp/mcp-activity-recording.test.ts
```diff
@@ -1,6 +1,7 @@
 import { Client } from '@modelcontextprotocol/sdk/client/index.js'
 import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
 import { McpActivity, McpOAuthClientKey } from '@activepieces/shared'
+import { createMockConnection } from '../../../helpers/mocks'
```
hazemadelkhalel: Line 12 already imports from `../../../helpers/mocks`, so this is a duplicate. `import-x/no-duplicates` would catch it, except `packages/server/api` lints `src/**/*.ts` only, so nothing under `test/` is ever linted. The glob is the more useful finding here.

hazemadelkhalel (changes requested):
Good work, and the hard calls here are right. No COUNT endpoint, cursor pagination, connection names resolved at read time and keyed by `(projectId, externalId)`, and an explicit response mapper. That last one is load-bearing, by the way: `server.ts:172` falls back to raw `JSON.stringify(data)` when `safeParse` fails, so replacing the mapper with a spread would put `platformId`, `userId` and `payloadFileId` on the wire the first time the schema drifts. Worth keeping as-is.

Two blockers, three things worth fixing, the rest are nits. Details are inline.

The blockers:

1. `?createdAfter=0000-01-01T00:00:00Z` returns a 500 and logs a Sentry event. zod accepts year 0000, Postgres rejects it. I verified both halves separately: the schema parses it, and `SELECT '0000-01-01T00:00:00Z'::timestamptz` errors with `date/time field value out of range`. I did not reproduce it over HTTP, so treat the seam between those two facts as read from the code rather than observed.

2. The payload route's `platformId` filter has no test that can fail. Both payload tests build the member on the same platform, so they pass on the `userId` filter alone. For a privileged caller `resolveUserIdFilter` returns null, which leaves `platformId` as the only tenant guard on a route that serves resolved piece inputs and third-party outputs.

Two fixes for the description: the two composite indexes are already on the base branch, and this diff has no entity or migration change, so the stack reviewer will go looking for them. And it is 18 new tests, not 16.

One thing that is not yours to fix in this PR but is worth a decision. Both routes use `securityAccess.publicPlatform`, and `nonEmbedUsersOnly` is the tighter guard for a console surface. Embedded users are not reliably MEMBER either: `ee/scim/scim-user-service.ts:22` takes `platformRole` straight from the IdP payload. Related, `resolveUserIdFilter` reads the user through `getOneOrFail({ id })` with no platform scoping while the row filter uses the token's platform, and `getOneByIdAndPlatformIdOrThrow` sits three lines below it. Both apply equally to the grants endpoint.

othmanemad:
Thanks — both blockers were real and both are fixed in 0177159b68, with the brain pages in 1f9b00989f. Details are inline; three things belong here.

**The year-zero fix is not the one I expected.** Binding a JS `Date` instead of the string genuinely works in a scratch reproduction: `node-postgres` serialises a year-zero `Date` to `0001-01-01 … BC`, and both a bare `pg` query and an entity-less TypeORM query builder return rows for it. Through the entity-bound query builder the value is ISO-stringified again before the driver sees it, and the query fails with `date/time field value out of range: "0000-01-01T00:00:00.000Z"` exactly as before. Only the integration test caught it. The bound has to be in the schema. It is on the brain page now, because the scratch reproduction is convincing and wrong.

**Both payload tests were dead, not one.** `insertActivityRow` hardcoded `payloadFileId: null`, so every payload test 404s on the `isNil(activity.payloadFileId)` branch before any filter runs — `404s on another members activity for an unprivileged member` passed with `userId` removed too. The fixture now writes a real `MCP_CALL_PAYLOAD` file, and both guards are verified by mutation: removing `platformId` fails only the new cross-platform test, removing `userId` fails only the member test.

**On `publicPlatform` and `resolveUserIdFilter`, I would leave both.** `nonEmbedUsersOnly` is on exactly 1 route and `publicPlatform` on 51, so adopting it here makes this the second user of a guard the codebase has not standardised on — that reads as a platform-wide decision rather than this PR's. The `getOneByIdAndPlatformIdOrThrow` swap also costs more than it looks: it runs `findOne({ id, platformId })` and *then* `getMetaInformation({ id })`, so a second query on every list request, and it returns `UserWithMetaInformation` where `isUserPrivileged` takes `User`. The invariant it would assert holds by construction anyway — `authentication-utils.ts` mints the USER token with the same `platformId` the user was looked up by, and `user.platformId` is a single column — so the check would be dead code. Happy to be talked out of this one.

Description fixed: the composite-indexes paragraph is dropped (they ship in #15240, and this diff has no entity or migration change), and the counts now read 23 and 14 → 17.

Deliberately not done here: the date bound lives in a shared `OptionalIsoDateTimeQuery` but is only adopted by this endpoint. `flow_run`, `audit_event`, health metrics and admin-platform all still bind `z.string()` into a timestamp comparison and 500 on `?createdAfter=invalid` today; that wants its own PR.

othmanemad:
Replies to the remaining threads. GitHub's review-comment reply endpoint is returning a persistent HTTP 500 for this PR — only the `createdAfter` thread accepted one — so they are collected here and keyed to each thread.


### `mcp-activity.ts` — `projectIds` and the empty filter

Settled — and thanks for raising it before the frontend was built against it.

`projectIds` is now `OptionalArrayFromQuery(z.union([ApId, z.literal(PLATFORM_WIDE_PROJECT_FILTER_VALUE)]))`, so `?projectIds=` returns 400 like its three siblings instead of parsing to `['']` and matching nothing. It also promotes the `platform-wide` sentinel out of `applyProjectFilter`'s runtime check and into the request contract. Test added.

I left the shared `OptionalArrayFromQuery` alone. The same `''` behaviour is at roughly 18 call sites — flows, tables, connections, templates, audit-events, platform — and making it drop empty strings would change list behaviour on all of them, which the repo's own rules class as a functional breaking change. `mcp-oauth.ts:73` has the identical shape, so the grants endpoint deserves the same decision rather than a drive-by here.

### `mcp-activity.test.ts` — the payload route's `platformId` filter

Right, and it was worse than one dead test — both were.

`insertActivityRow` hardcoded `payloadFileId: null`, so every payload test 404s on the `isNil(activity.payloadFileId)` branch before any filter runs. `404s on another members activity for an unprivileged member` therefore passed with `userId` removed too.

Fixed: the row fixture can now write a real `MCP_CALL_PAYLOAD` file, the member test asserts against a row that has one, and the cross-platform test you asked for is added. Both verified by mutation — removing `platformId` fails only `404s on another platforms activity for a privileged caller`, and removing `userId` fails only the member test.

One trap for anyone reusing the fixture: `createMockFile` defaults `compression` and `type` to `faker.helpers.enumValue(...)`, so it pins `FileCompression.NONE` and `FileType.MCP_CALL_PAYLOAD` explicitly. Unpinned, `decompress` would throw on roughly half of runs.

### `mcp-activity-service.ts:89` — `spreadIfDefined`, and the file lookup's tenant scope

`spreadIfDefined` applied at both sites — thanks, and the `listGrants` precedent was the right pointer.

On the second half: passing `platformId: activity.platformId` does not compile. `fileService.getDataOrThrow` takes `GetOneParams = { fileId?, projectId?, type? }` (`file.service.ts:407`) and has no `platformId` parameter, even though the `file` table does have the column. Closing it means widening a signature that every file read in the product goes through.

I left it, because nothing can reach the gap: `payloadFileId` is read off a row already scoped by `(id, platformId, userId)`, so a caller supplies an activity id, never a file id, and there is no path by which an activity on one platform holds a file id from another. The reasoning is on the brain page so it does not get reopened. If you think `fileService` should carry a platform scope for all callers, that seems worth its own PR rather than one parameter used by a single call site.

### `mcp-activity-service.ts:90` — `tryCatch` and `StoredPayload`

Both done. `StoredPayload` is gone, the read and parse run inside `tryCatch`, and a failure logs with the activity id and file id before throwing `ENTITY_NOT_FOUND` with `Stored payload could not be read` — so a corrupt payload stays distinguishable in logs from a genuinely absent one, rather than surfacing as a 500.

One caveat on "one schema instead of two": `McpActivityPayload` validates `truncated` and nothing else, since `input` and `output` are `z.unknown()`. It is still the right call — the shape lives in one place, and parsing is what keeps `input`/`output` typed as `unknown` without an `as` cast — but it is not extra validation.

### `mcp-activity.ts:70` — `GetMcpActivityPayloadParams`

Agreed, inlined. `GetMcpActivityPayloadParams` had exactly one consumer, and 45 routes already write the shape inline — 13 on a single line, 32 across two.

### `mcp-activity-recording.test.ts:4` — duplicate import, and the lint glob

Duplicate merged into the line 12 import.

The glob is the better finding, and I have not touched it here. Measured: `eslint 'test/**/*.ts'` today reports 122 errors and 1,061 warnings across 187 of 305 files. 714 of the warnings are `no-non-null-assertion` — the `app!` idiom every integration test uses — and 347 are `explicit-function-return-type`. A `test/**` override turning those two off clears 1,061 of the 1,183 problems and leaves 122 errors, 68 of them auto-fixable. That is a mechanical but wide PR, so it wants to be its own. For what it is worth, this file is one of 4 `import-x/no-duplicates` errors already in the tree.

packages/core/shared/src/lib/automation/mcp/mcp-activity.ts
othmanemad: Fixed in 0177159b68. The bound now lives in a shared `OptionalIsoDateTimeQuery` in `core-utils`, so `flow_run`, `audit_event` and health metrics can pick it up — they all still 500 on `?createdAfter=invalid`.

One warning. Binding a JS `Date` instead of the string looks like the better fix: `node-postgres` turns a year-zero `Date` into `0001-01-01 … BC`, and both bare `pg` and an entity-less TypeORM builder accept it. The entity-bound query builder stringifies it again, so Postgres rejects it exactly as before. Only the integration test caught that. The bound has to be in the schema.

packages/server/api/test/integration/ce/mcp/mcp-activity.test.ts
othmanemad: Fixed in 0177159b68. Both payload tests were dead, not one: `insertActivityRow` hardcoded `payloadFileId: null`, so every payload test 404s on the no-payload branch before any filter runs. The unprivileged-member test passed with `userId` removed too.

The fixture now writes a real payload file. Removing `platformId` fails the new cross-platform test, and removing `userId` fails the member test.

packages/core/shared/src/lib/automation/mcp/mcp-activity.ts
othmanemad: Settled. `projectIds` is now `ApId | 'platform-wide'`, so `?projectIds=` returns 400 like the other three filters, and the sentinel moves into the request contract. Test added.

I left the shared `OptionalArrayFromQuery` alone. The same `''` behaviour is at ~18 call sites, so changing it there is a separate decision. The grants schema has the identical shape.

packages/server/api/src/app/mcp/activity/mcp-activity-service.ts
othmanemad: `spreadIfDefined` applied at both sites.

On the second half: `platformId: activity.platformId` does not compile. `getDataOrThrow` takes `{ fileId, projectId, type }` and has no `platformId` parameter, so closing it means changing a service every file read in the product goes through.

I left it. `payloadFileId` is read off a row already scoped by `(id, platformId, userId)`, so a caller passes an activity id and can never steer the file id. Happy to do the `fileService` change as its own PR if you think it earns one.

hazemadelkhalel (approved):

packages/core/utils/src/lib/base-model.ts
```diff
@@ -29,6 +29,10 @@ export const OptionalBooleanFromQuery = z.pipe(
     z.optional(z.boolean()),
 )
 
+export const OptionalIsoDateTimeQuery = z.optional(
+    z.iso.datetime({ offset: true }).check(z.refine((val) => new Date(val).getUTCFullYear() >= 1)),
```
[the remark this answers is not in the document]
othmanemad: Correct, fixed in ce640f2cce. `0001-01-01T00:00:00+01:00` normalized to UTC year 0000 and returned 400, and PostgreSQL accepts that timestamp — the service binds the raw string, not a `Date`, so the guard was checking a value the driver never sees.

Took the suggested `slice(0, 4)` check. The zod ISO grammar fixes the year at `\d{4}` with no sign, so the slice is exactly what reaches PostgreSQL, and `0000` is the only unrepresentable value.

The existing "earliest window" test used `0001-01-01T00:00:00Z`, which is UTC year 1 and passed either way. Added the positive-offset case that does not.

PR #15246, opened 2026-09-03
fix(sandbox): pin the isolated linker for piece installs

### LINKED ISSUE #15196: [BUG]: 0.89.0: newly installed pieces fail to resolve after upgrade (bun 1.4 hoisted layout)
**Describe the bug**
On 0.89.0, installing any new piece fails with `PieceNotFoundError` / `ENGINE_OPERATION_FAILURE` (500) on `POST /api/v1/pieces`. The worker logs `[pieceInstaller] Installed registry pieces using bun` and reports success — but the engine then can't resolve the piece.

0.89.0 bumped bun 1.3.x → 1.4.0 (2596d6c5). Per bun's 1.4 breaking changes (oven-sh/bun#24236), a `bun.lock` carrying `"configVersion": 0` (written by bun ≤1.3) keeps the legacy **hoisted** layout. Combined with the `--filter` flag `bunRunner.install` always passes, the piece is linked only at the workspace root:

```
cache/v14/common/node_modules/@scope/piece-x          <- where bun 1.4 puts it
cache/v14/common/pieces/@scope/piece-x-1.0.0/node_modules/@scope/piece-x   <- where the engine looks
```

The engine resolver only ever checks the nested path, so the piece is unresolvable. Pre-existing pieces keep working (their nested dirs survive from the bun 1.3 era), so this only affects **newly installed** pieces — which makes it look intermittent.

**To Reproduce**
1. Run 0.88.x so the shared workspace gets a `bun.lock` with `"configVersion": 0`.
2. Upgrade to 0.89.0.
3. Settings → My Pieces → Install Piece, add any piece not already cached.
4. Install "succeeds" in the worker logs, then the request 500s with `ENGINE_OPERATION_FAILURE`.
5. `ls cache/v14/common/pieces/<scope>/<name>-<version>/node_modules` → missing; the package is at `cache/v14/common/node_modules/<scope>/<name>` instead.

Minimal repro without the UI, in the workspace root:
```bash
bun install --ignore-scripts --filter "./pieces/@scope/piece-x-1.0.0"
```
Nested layout is created on bun 1.3.1 and missing on bun 1.4.0.

**Expected behavior**
A newly installed piece is linked into `pieces/<name>-<version>/node_modules/<realname>` (isolated layout) and resolves, as it did on 0.88.x.

**Screenshots**
N/A — logs above.

**Additional context**
`bunfig.toml` does ship `linker = "isolated"`, but it lives at `/usr/src/app/bunfig.toml` while bun runs with `cwd=/usr/src/app/cache/<v>/common`. Bun reads bunfig from **cwd and `$HOME` only, never ancestors**, so the setting never applies. Independent confirmation: the same file's `minimumReleaseAge = 259200` is also inert — a package published 15h earlier installed fine.

Verified matrix (same workspace, same tarball, one variable changed):

| bun | lock `configVersion` | `--filter` | nested layout |
|---|---|---|---|
| 1.3.1 | none | yes | ✅ |
| 1.4.0 | fresh | no | ✅ (writes `configVersion: 1`) |
| 1.4.0 | `0` | no | ✅ |
| 1.4.0 | `0` | yes | ❌ |
| 1.4.0 | `0` + bunfig in cwd or `$HOME` | yes | ✅ |

All three conditions are required, which is why a clean install passes and an upgraded instance fails.

Also note installing a second version of the same piece overwrites the root copy and leaves *both* versions without a nested dir, so a new install can break a previously working one.

Suggested fix — either:
- pass `--linker=isolated` explicitly in `bunRunner.install` (`packages/server/sandbox/src/lib/utils/bun-runner.ts`), or
- write `bunfig.toml` into the workspace root alongside the generated `package.json` in `createRootPackageJson`.

Workaround: mount a `bunfig.toml` containing only `[install]` + `linker = "isolated"` at `$HOME/.bunfig.toml`. Do **not** copy the shipped bunfig verbatim — that would activate `minimumReleaseAge` and block piece versions published <72h ago.

Affects 0.89.0. `main` still pins `bun@1.4.0` with no explicit linker, so releases cut from it carry this. `release-candidate` is still on `bun@1.3.3`.

Environment: self-hosted, CE 0.89.0, `AP_EXECUTION_MODE=UNSANDBOXED` (all pieces share the `common` workspace, so custom/archive pieces are affected too).

AhmadTash:
## Description

On 0.89.0, installing any new piece fails with `PieceNotFoundError` / `ENGINE_OPERATION_FAILURE` (500) on `POST /api/v1/pieces`. The worker logs `[pieceInstaller] Installed registry pieces using bun` and reports success, then the engine can't resolve the piece.

The engine resolves a piece only at `pieces/<name>-<version>/node_modules/<realname>` (`resolveInstalledPieceEntry`, `piece-loader.ts`), which the **isolated** linker produces. Three things have to line up, and on an upgraded instance all three do:

1. **bun 1.4.0** (`Dockerfile`, bumped in 2596d6c5) defaults workspaces to isolated — but honors a legacy-hoisted default when the lockfile asks for it.
2. **A pre-1.4 lockfile on the persisted cache volume**, written while the box ran 0.88.x, carries `"configVersion": 0` and pins the workspace to hoisted. A *clean* 0.89.0 install writes `configVersion: 1` and works, which is why this only reproduces after an upgrade and looks intermittent.
3. **`--filter`**, which `bunRunner.install` always passes. Narrowed to one member there is no version conflict to force nesting, so the package hoists to the workspace root — where the resolver never looks (`traverseAllParentFoldersToFindPiece` only probes `*/pieces/<pkg>`).

Pre-existing pieces keep working because their nested dirs survive from the bun 1.3 era. Hoisting also collapses several installed versions of a piece onto one, so a new install can break a previously working one.

The fix pins `--linker=isolated` on the `bun install` command line, making the layout independent of whatever `configVersion` the inherited lockfile carries.

**Why not the shipped `bunfig.toml`.** It already sets `linker = "isolated"`, but it never applies: it lives at `/usr/src/app/bunfig.toml` while bun runs with `cwd=/usr/src/app/cache/<v>/common`, and **bun reads bunfig from cwd and `$HOME` only, never ancestors**. Writing a bunfig into the generated workspace root would fix the linker but also switch on that file's `minimumReleaseAge = 259200`, blocking piece versions published less than 72h ago. The CLI flag avoids that. (Independent confirmation the file is inert today: a package published 15h earlier installs fine.)

**Why not teach the resolver to accept a hoisted layout.** A hoisted root copy is a single version shared by every workspace member, so accepting it would silently serve the wrong piece version. Isolated linking is a correctness requirement, not a path convention.

**Scoped to piece installs.** `bunRunner.install` is shared with the code-step installer, and isolating that too is not free: bun 1.4's isolated default applies only to *workspaces*, so a lone code-step package still gets hoisted today, and isolating it stops **undeclared transitive imports** from resolving. Measured on bun 1.4.0 — a code step declaring only `axios` and importing `form-data` builds today (404,708 bytes) and fails under isolated with `Could not resolve "form-data"`. Code steps gain nothing from isolation (esbuild bundles everything, then `node_modules` is deleted), so the flag is `true` for pieces and `false` for code steps.

Existing broken instances self-heal: `pieceCheckIfAlreadyInstalled` already drops the `ready` marker when the nested `node_modules` is missing, so affected pieces reinstall correctly on next use.

## How was this tested?

**Reproduced the bug and verified the fix on a real bun 1.4.0 binary.** Built a workspace exactly as `createRootPackageJson` / `createPiecePackageJson` write it, with real npm tarballs (`piece-google-sheets@0.16.1`, `piece-google-sheets@0.16.3`, `piece-google-drive@0.7.8` — two versions of one piece, plus a second piece sharing `googleapis@129.0.0`), installed under bun 1.3.9, forced the lock to `configVersion: 0`, then added a member with bun 1.4.0 and `--filter`:

| install | result |
| --- | --- |
| no `--linker` | `piece-google-sheets-0.16.3` has no nested dir → unresolvable; `@activepieces/piece-google-sheets` appears hoisted at the workspace root |
| `--linker=isolated` | all three resolve, nothing hoisted at the root, all three `require()` successfully |

Side effects checked rather than assumed, since isolated linking uses **symlinks** where bun 1.3's nested layout used real directories:

- **Resolver** — `utils.folderExists` uses `fs.access` (follows symlinks) with no `isDirectory()` check, so the symlinked package dir resolves. Confirmed by simulating `resolveEntryFromPackageDir` + `require` on all three pieces.
- **`assertPathInside`** resolves realpath and would reject a store symlink leaving the piece folder — it is only called from EE git-sync, never on the piece path.
- **`deleteStaleCache`** reads only top-level version dirs and `rm -r`s whole trees; `rm` unlinks symlinks rather than following them.
- **Memory** — `googleapis@129.0.0` resolves to a single store realpath shared by all three pieces, so Node's `require.cache` holds one copy. No regression versus hoisted, better than per-member nested copies.
- **Disk** — the fixed layout is smaller: 182 MB / 14,788 paths versus 187 MB / 15,836 for the broken hoisted state.
- **Piece strictness** — bun 1.3.9's default for this workspace shape is already isolated, so pieces have been isolated since 0.88; no phantom-dependency change for piece bundles.

**Automated:** `packages/server/sandbox` — 21 files / 254 tests pass. New `test/lib/utils/bun-runner.test.ts` pins the exact arg vector (removing the flag fails it), and `piece-installer.test.ts` now asserts every piece install call, batch and individual-fallback, requests the isolated linker. `eslint 'src/**/*.ts'` 0 errors; `tsc --noEmit` clean.

**Editions:** the piece install path is edition-independent (`packages/server/sandbox`, no `ee/` involvement, no feature flag, no new env var or setup step) and runs identically on CE / EE / Cloud. Reported on self-hosted CE 0.89.0 with `AP_EXECUTION_MODE=UNSANDBOXED`, where all pieces share the `common` workspace, so custom and archive pieces are affected too.

Fixes #15196


majewskibartosz:
resolved in activepieces/activepieces#15262

AhmadTash:
resolved in #15262

PR #15253, opened 2026-09-03
ci(pieces): configurable heap thresholds, drop absolute default to 5 MB

amrdb:
## Summary

Follow-up to #15252 (merged). Two changes:

- **Wire repo variables through the workflow.** \`vars.PIECE_HEAP_ABSOLUTE_MB\` and \`vars.PIECE_HEAP_DELTA_MB\` are now passed as env into the check step, so ops can tune the limits from Settings → Secrets and variables → Variables without a code change. If a variable is unset, the workflow expands it to empty string; the code fallback is switched from \`?? '<n>'\` to \`Number(...) || <n>\` so an empty string still resolves to the default (previously it resolved to \`NaN\`, silently disabling the rule).
- **Absolute default: 10 MB → 5 MB.** 10 MB was picked conservatively in the first pass. From the 2026 audit, actually-light pieces land at 1–5 MB; the tighter default is where a fresh piece should aim. Repo variable \`PIECE_HEAP_ABSOLUTE_MB\` overrides it centrally if 5 MB proves too tight for some new piece.

Bypass and semantics unchanged: existing pieces still run only under the delta rule; the absolute cap only applies to genuinely new pieces (no npm history); \`skip-heap-check\` label still bypasses.

## How was this tested?

- Diff verified: the two files only.
- Ran the check locally against \`piece-qrcode\` before change (limit=10) and after change (limit=5); qrcode is 3.17 MB current, 1.88 MB previous, delta +1.29 → passes both. Absolute rule doesn't apply (existing piece).

### DIFF
--- a/.github/workflows/validate-publishable-packages.yml
+++ b/.github/workflows/validate-publishable-packages.yml
@@ -39,3 +39,32 @@ jobs:
 
       - name: validate publishable packages
         run: npx ts-node -r tsconfig-paths/register -P packages/server/engine/tsconfig.lib.json tools/scripts/validate-publishable-packages.ts
+
+      - name: find changed pieces
+        id: changed
+        run: |
+          OUTPUT=$(npx ts-node -r tsconfig-paths/register -P packages/server/engine/tsconfig.lib.json tools/scripts/pieces/find-changed-pieces.ts)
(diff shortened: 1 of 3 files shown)


amrdb:
Superseded by #15254, which shows the actual 4-line diff. This branch was based on a pre-squash commit history so GitHub was rendering it as a full-file re-add.

PR #15256, opened 2026-09-03
feat: human input UI

Talaljaber:
## Description

UI pass over the **Human Input** piece (`packages/pieces/core/forms`, shown as "Human Input"), the two public pages it generates, and one shared renderer bug the piece made visible.

### Hosted form (`/forms/:flowId`)

`ap-form.tsx` never rendered `FormMessage`, so no zod error ever reached the visitor: pressing **Submit** with an empty required field did nothing at all. It is now rendered under every field, and `FormControl` wraps exactly one input per branch (it used to wrap a Fragment, which meant Radix `Slot` dropped the `aria-describedby` / `aria-invalid` it injects, so the message would have been visual-only).

Also on that page:

- **A required File input was not validated** — `createPropertySchema` returned a bare `z.unknown()` and an empty upload submitted as `""`. It now requires a `File` when the author marked the field required.
- **Clicking a toggle's label did nothing** — no `id` on the `Checkbox` to match `htmlFor`. Added.
- **A toggle prefilled from the query string** arrived as the string `'true'`; it is now coerced (`true`, `1`, `yes`, `on`, any case).
- **Fixed `w-[500px]` card** → `w-full max-w-[500px]` for phones.
- Required marker via `showRequiredIndicator` instead of a literal `'*'`; the toggle branch now shows one too.
- `ReadMoreDescription` is only rendered when a description exists (it emitted an empty `<p>` before). Dead `placeholder` on the `type="file"` input removed.
- **Draft/published cache key**: `formsKeys.form` and the chat `useQuery` keyed on `flowId` alone with `staleTime: Infinity`, so opening the published URL then the draft URL in one session served the published version from cache. Both keys now include the draft flag.

### Shared renderer: ```` ```text ```` blocks in `ApMarkdown`

The block rendered a single-line `<input disabled>`, which clipped anything wider than the step panel and could not be selected. **122 piece files** show a URL this way. Human Input showed two URLs that differ only in a query string past the clipping point, so both boxes read as identical. The value now wraps (`whitespace-pre-wrap`, needed because react-markdown places it inside a `<pre>`) and is selectable; the copy button is unchanged. This is the one product-wide visual change in the PR.

### Config panels

- **Chat UI trigger** documented only the published URL, although the chat page has accepted `?useDraft=true` all along and the builder has an **Open Chat** drawer. The panel now shows the Draft URL and points at the drawer, mirroring the Web Form trigger.
- **Web Form trigger**: the note about Respond on UI sat *above* the `Wait for Response` toggle it describes; it now sits directly under it.
- Both notes were **WARNING** variants phrased as instructions ("use **Respond on UI** in your flow…" / "Ensure that **Respond on UI** is used in your flow"), which read as an error even when the step was already in the flow. They are now **INFO** notes that state what the trigger waits for, plus the consequence with the configured timeout via `{{webhookTimeoutSeconds}}` — the same pattern as the core Webhook trigger. Every line of panel copy is under 70 characters.
- `waitForResponse` and the nested `Required` checkbox were `required: true` with a boolean default — an asterisk on a field that is always answered. Now `required: false, defaultValue: false`. Stored value is unchanged: the builder writes `defaultValue ?? false` for every checkbox regardless of `required`.
- Descriptions and placeholders added where missing: `Inputs`, `Field Name`, `Field Description`, `Bot Name` (replacing "The name of the chatbot"), Respond on UI `Text` and `Attachment`. `Text (Markdown)` → `Text` with the format in the description. All under the 70-character `ReadMoreDescription` fold.
- Raw `<br>` spacing in markdown replaced with blank lines.

### Scope

UI Improvement lane only. No `aiMetadata`, `outputSchema`, `run()`, `classification`, `audience`, or action-level `description` was changed.

### i18n

- Renaming `Text (Markdown)` → `Text` and rewriting the panel copy orphans those Crowdin keys in `packages/pieces/core/forms/src/i18n/*.json` until the nightly `generate-translations.yml` regenerates. Fallback is English. **Locale files are intentionally not edited** — running `i18n:extract` locally produced ~4.5k lines of unrelated drift and was reverted.
- `t('Type your message here...')` (chat input placeholder, previously hardcoded) has no `en` entry yet; same workflow picks it up.
- Required-field message uses `t('This field is required')` to match the existing TEXT branch rather than `formErrors.required`, whose bare key `required` would render worse until translated.

### Deliberately not changed

- `required` for **TOGGLE** is still not enforced — an unticked toggle is a legitimate "no"; enforcing would reject submissions valid today. Product call.
- The `Inputs` label rendering red when empty is correct behaviour (`data-[error=true]`), not a bug.
- `putBackQuotesForInputNames` / payload shape — back-compat, out of scope.
- `waitForResponse` is now `required: false` in the piece while `FormProps` in `core-execution` still says `z.boolean()`; behaviour is identical because the builder always writes a boolean. Tightening the schema belongs to a different lane.
- Hardcoded English in `ChatNotFound`, `ChatIntro`, `ErrorBubble`, and the form's 404 — separate i18n pass.

## How was this tested?

- `npx turbo run build lint --filter=@activepieces/piece-forms --force` — passes, zero lint errors. Piece bumped `0.4.21` → `0.4.22`.
- `bun run typecheck` in `packages/web` — clean.
- ESLint on every changed web file — no new errors. The package-wide run is dominated by pre-existing `linebreak-style` errors from `core.autocrlf=true` on this Windows checkout; root `lint-dev` was not run because `--fix` would rewrite line endings repo-wide.
- Served metadata verified over `/api/v1/pieces/@activepieces/piece-forms` after an API restart: version, labels, descriptions, placeholders, both INFO variants, Draft Chat URL, no `<br>`.
- Manually in the builder: all three panels at 0.4.22; URL boxes wrap and show `?useDraft=true`.

- Reviewed three times by a separate model pass (verify → fix → re-verify); final verdict SHIP.
- Editions: all changes are edition-agnostic — piece metadata, public-page rendering, and one shared component. No `platform.plan` flags, no `ee/` imports.

Fixes # (issue)

Reviewer note: the required-File validation is the one behaviour change on already-published forms. It only rejects a submission that leaves a field the author explicitly marked **Required** empty — which every other field type already enforces — so it is treated as a bug fix.

### DIFF
--- a/packages/pieces/core/forms/package.json
+++ b/packages/pieces/core/forms/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-forms",
-  "version": "0.4.21",
+  "version": "0.4.22",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
--- a/packages/pieces/core/forms/src/lib/actions/return-response.ts
+++ b/packages/pieces/core/forms/src/lib/actions/return-response.ts
@@ -9,12 +9,13 @@ export const returnResponse = createAction({
   name: 'return_response',
   classification: 'WRITE',
   displayName: 'Respond on UI',
-  description: 'Return a file or text (markdown) as a response.',
+  description: 'Send text or a file back to the person waiting.',
   aiMetadata: { description: 'Sends markdown text and/or one file attachment back to the person waiting on the Web Form or Chat UI trigger that started the flow. Pick it as the replying step when that trigger has "Wait for Response" enabled; it produces nothing visible in flows started any other way. Not idempotent: each call emits a response to the waiting caller and stores a new file for any attachment.', idempotent: false },
   props: {
     markdown: Property.LongText({
       displayName: 'Text',
       description: 'Shown to the person waiting. Markdown is supported.',
+      placeholder: 'Thanks, we received your request.',
       required: false,
     }),
     file: Property.File({
--- a/packages/pieces/core/forms/src/lib/actions/return-response.ts
+++ b/packages/pieces/core/forms/src/lib/actions/return-response.ts
@@ -13,11 +13,13 @@ export const returnResponse = createAction({
   aiMetadata: { description: 'Sends markdown text and/or one file attachment back to the person waiting on the Web Form or Chat UI trigger that started the flow. Pick it as the replying step when that trigger has "Wait for Response" enabled; it produces nothing visible in flows started any other way. Not idempotent: each call emits a response to the waiting caller and stores a new file for any attachment.', idempotent: false },
   props: {
     markdown: Property.LongText({
-      displayName: 'Text (Markdown)',
+      displayName: 'Text',
+      description: 'Shown to the person waiting. Markdown is supported.',
       required: false,
     }),
     file: Property.File({
       displayName: 'Attachment',
+      description: 'A file the person can download with the response.',
       required: false,
     }),
   },
--- a/packages/pieces/core/forms/src/lib/actions/return-response.ts
+++ b/packages/pieces/core/forms/src/lib/actions/return-response.ts
@@ -9,7 +9,7 @@ export const returnResponse = createAction({
   name: 'return_response',
   classification: 'WRITE',
   displayName: 'Respond on UI',
-  description: 'Send text or a file back to the person waiting.',
+  description: 'Return a file or text (markdown) as a response.',
   aiMetadata: { description: 'Sends markdown text and/or one file attachment back to the person waiting on the Web Form or Chat UI trigger that started the flow. Pick it as the replying step when that trigger has "Wait for Response" enabled; it produces nothing visible in flows started any other way. Not idempotent: each call emits a response to the waiting caller and stores a new file for any attachment.', idempotent: false },
   props: {
     markdown: Property.LongText({
--- a/packages/pieces/core/forms/src/lib/triggers/chat-trigger.ts
+++ b/packages/pieces/core/forms/src/lib/triggers/chat-trigger.ts
@@ -10,7 +10,8 @@ import {
 } from '@activepieces/pieces-framework';
 import { chatSubmissionTriggerOutputSchema } from '../output-schemas';
 
-const responseMarkdown = `The sender waits for a **Respond on UI** step in this flow.`;
+const responseMarkdown = `The sender waits for a **Respond on UI** step in this flow.
+Without one, they see an error after {{webhookTimeoutSeconds}} seconds.`;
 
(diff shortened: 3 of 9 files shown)


ibrahim-abuznaid:
Pulled this locally and ran it on 0.4.22. The empty-required-field fix is the big one for me — before, hitting Submit on an empty form did nothing at all, now it surfaces the errors properly. Also checked the required-file validation, the toggle label click, the truthy query prefill (true/yes/1/on), and the URL boxes wrapping in the builder. All working, no console errors. Nice cleanup 👍

One heads-up unrelated to your code: the renamed copy leans on the nightly translations job to backfill the Crowdin keys, but that workflow's been failing since Aug 16, so those strings will sit on the English fallback until it's fixed.

ibrahim-abuznaid (approved):

ibrahim-abuznaid (approved):

AhmadTash:
The change in `markdown.tsx` removes the `<input class="grow bg-background">` that three E2E specs read the webhook URL from:

- `packages/tests-e2e/scenarios/ce/flows/webhook-should-return-response.spec.ts`
- `packages/tests-e2e/scenarios/ce/pieces/piece-isolation.spec.ts`
- `packages/tests-e2e/scenarios/ee/pieces/piece-isolation.spec.ts`

They call `page.locator('input.grow.bg-background').inputValue()`, which matches nothing after this PR, so all three will time out. E2E only runs with the `ready-for-e2e` label, which is why it shows as skipped here.

Please update those locators (for example `page.locator('pre code').first().textContent()`) and add the label so E2E runs.

Talaljaber:
> The change in `markdown.tsx` removes the `<input class="grow bg-background">` that three E2E specs read the webhook URL from:
> 
> * `packages/tests-e2e/scenarios/ce/flows/webhook-should-return-response.spec.ts`
> * `packages/tests-e2e/scenarios/ce/pieces/piece-isolation.spec.ts`
> * `packages/tests-e2e/scenarios/ee/pieces/piece-isolation.spec.ts`
> 
> They call `page.locator('input.grow.bg-background').inputValue()`, which matches nothing after this PR, so all three will time out. E2E only runs with the `ready-for-e2e` label, which is why it shows as skipped here.
> 
> Please update those locators (for example `page.locator('pre code').first().textContent()`) and add the label so E2E runs.

The E2E jobs fail in global-setup at sign-up, before any spec runs, in both editions. The suite has not passed since 2025-11-17; this PR is the only run in the last 400 that executed at all. The api (ce) tests on this head cover fresh-instance sign-up and pass. The pipeline issue (turbo strips the workflow env, dev piece dists are never built) is tracked separately.

PR #15264, opened 2026-09-04
fix(agents): agent steps on Bedrock no longer fail with "provided model identifier is invalid"

### LINKED ISSUE #15260: [BUG] Agent runs on Bedrock fail on step 1: fast tier resolves to invalid model id
## Problem

Every agent run on a Bedrock provider fails on its first step with `The provided model identifier is invalid`, before the configured model handles anything. Present in 0.89.0; still present on main. Azure and Mistral hit the same path.

Ref: Pylon 5836

## Root cause

Fast tier hardcoded to `anthropic/claude-haiku-4.5` (`packages/core/piece-types/src/lib/ai-providers.ts`, `ACTIVEPIECES_CHAT_TIERS`). `resolveModelIdForProvider` (`packages/server/api/src/app/ee/agent/agent-helpers.ts`) has no curated list and no manual catalog for bedrock/azure/mistral, so it fabricates `claude-haiku-4-5` (strip vendor prefix, dots to dashes) and returns it unchecked. Worker runs step 1 of every agent turn on it (`run-agent-turn.ts`, `isFirstStep && fastModel`). Bedrock rejects it: not a model id or inference profile.

Wire-verified on the 0.89.0 image: step 1 = `POST /model/claude-haiku-4-5/converse-stream` -> 400 ValidationException; run FAILED with the exact reported error. Surfaced error names the CONFIGURED model, not the failing request's model.

## Fix

When the provider has no curated chat models and no manual catalog, fall back to the run's resolved configured model instead of the fabricated tier id (`resolveFastModelId` call sites: `agent-rpc-handlers` getAgentConfig, chat-personalization). Keeps the fast-tier optimization where ids are real (anthropic/openai/google/openrouter); fixes bedrock/azure/mistral in one guard.

## Out of scope

* Chat-surface tier resolution on the same providers (smart tier fabricates `claude-sonnet-4-6` the same way)
* Bedrock `sessionToken` support; unattended web-tool gating (separate reports)

majewskibartosz:
Fixes [GIT-1840](https://linear.app/activepieces/issue/GIT-1840)
Closes #15260

## Problem
1. Every agent run on a Bedrock provider fails on step 1 with `The provided model identifier is invalid`: the hardcoded fast tier (`anthropic/claude-haiku-4.5`) is transformed by stripping the vendor prefix and swapping dots for dashes into `claude-haiku-4-5`, which Bedrock rejects. Azure and Mistral hit the same path.
2. The run dies before the configured model (valid, works via direct Converse) handles a single step, and the surfaced error names the configured model instead of the failing request's model.

## Fix
- `agent-helpers.ts` — `resolveModelIdForProvider` gains an optional `fallbackModelId`, used only when the provider has no curated list and no admin catalog; the fallback still goes through `pickAllowedModel` (modelScope/modelIds allow-list enforced, fail-closed) and tier ids / empty strings are never used as model ids. `resolveFastModelId` requires it; `resolveTierModel`/`resolveFastModel` thread it optionally.
- `agent-rpc-handlers.ts` — getAgentConfig passes the run's resolved model as the fast-model fallback; executePieceTool passes the conversation's stored model; `configuredToolConversationOrThrow` now selects only the four columns it reads.
- `chat-personalization-service.ts` — hoists the resolved modelId and reuses it as the fallback (drops a duplicate resolution).

## Verification
- Regression test red-first: the BEDROCK fallback test fails on pre-fix code (returns `claude-haiku-4-5`), passes with the fix. 170 agent unit tests green.
- Live end-to-end on this branch (EE stack, real worker + engine, AWS stubbed at the socket): agent step 1 request is `POST /model/us.anthropic.claude-sonnet-5/converse-stream`, `claude-haiku-4-5` appears nowhere on the wire. Same scenario on the official 0.89.0 image reproduces the reported failure byte-for-byte.
- Allow-list inverse: a fallback outside a modelScope=selected key's modelIds still throws (pinned by test); curated providers keep their cheap fast tier (pinned by test).
- `npx turbo run build --filter=api` and `npm run lint-dev` clean. Full `test/unit` carries 47 pre-existing env-dependent failures (redis-memory-server), identical on clean upstream/main, none in agent files.
- Visual: none - backend-only, no user-visible change.
- Other affected paths: chat-surface smart-tier resolution (CHAT source), `agent-memory-ai.ts`, `agent-draft-ai.ts`, `agent-service.ts#withDefaultModel` still fabricate ids on catalog-less providers — no configured model exists to fall back to at those sites; tracked as follow-ups on [GIT-1840](https://linear.app/activepieces/issue/GIT-1840).
- Repro: reproduced on the 0.89.0 image and fix wire-verified live — full steps in the Reproduction block.

<details>

## Plan
- One guard at the shared choke point (`resolveModelIdForProvider`) instead of per-caller patches; the three call sites with a real configured model supply it as the fallback.
- Footprint estimate: 3 files, ~20 product lines; actual ~30 after review hardening (allow-list routing, tier-id/empty-string guards).

- modelScope=selected key, fallback outside modelIds → ENTITY_NOT_FOUND, fail-closed (same as pre-fix).
- Stored `conversation.modelName` is a tier id (`smart`, persisted when a step names no model) or empty → fallback ignored, behavior unchanged.
- AGENT-source conversation with null modelName → fallback dropped by `spreadIfDefined`, unchanged.
- Curated/catalog providers → fallback never consulted, fast tier keeps the cheap model.

## Alternatives rejected
- Optional `fastModelId` through the worker contract: 4+ files and a core package version bump for identical behavior.
- Curated Bedrock model list: a hand-maintained catalog duplicating what `ListFoundationModels` already serves, goes stale.
- Returning the fallback without `pickAllowedModel`: bypasses the admin model allow-list (flagged in review, fixed).
- Throwing when no catalog exists: breaks chat surfaces that currently degrade instead of failing setup.
</details>

<details>

## Environment
- Bug: official image `activepieces/activepieces:0.89.0`, `AP_EDITION=ee`, `AP_EXECUTION_MODE=SANDBOX_CODE_ONLY`, fresh Postgres DB, separate Redis DB index, fake AWS credentials.
- Fix verification: this branch via `npx turbo run serve --filter=api --filter=@activepieces/engine --filter=worker`, `AP_EDITION=ee`, `AP_DEV_PIECES=webhook,ai`.
- Both lanes: outbound AWS intercepted in-process with `NODE_OPTIONS="--require shim.cjs"` — logs every Bedrock request (URL + toolConfig names) and answers with an AWS-faithful 400 ValidationException for any model except the configured profile; the Bedrock control plane (ListFoundationModels/ListInferenceProfiles over `https.request`) is stubbed so provider creation passes without real credentials.

## Harness
Public gist: https://gist.github.com/majewskibartosz/22a29958a9ff223929e652c20b539c62 — `shim.cjs` (interceptor), `scenario.mjs` (drives sign-up → Bedrock provider → flow build → webhook fire → run state), `README.md` (exact commands), plus both captured wire logs. Run: `git clone https://gist.github.com/22a29958a9ff223929e652c20b539c62.git git-1840-repro && cd git-1840-repro` and follow README.md.

## Trigger
Published flow: Catch Webhook (authType none) → Run Agent (provider `bedrock`, model `us.anthropic.claude-sonnet-5`, no tools, max steps 3, prompt "Return exactly: LOCAL_BEDROCK_SONNET_5_OK"); fire the webhook.

## Results
- 0.89.0: conversation-title generation hits the configured model, then agent step 1 = `POST /model/claude-haiku-4-5/converse-stream` (toolConfig `["ap_fetch_url"]`) → 400; flow run FAILED with `bedrock (us.anthropic.claude-sonnet-5): undefined: The provided model identifier is invalid.` — the error names the configured model while the wire shows the haiku request, which is why logs alone mislead.
- Branch: step 1 = `POST /model/us.anthropic.claude-sonnet-5/converse-stream`; `claude-haiku-4-5` absent from the entire capture; the run reaches the provider and fails only against the reject-everything stub.
</details>

### DIFF
--- a/packages/server/api/src/app/ee/agent/agent-helpers.ts
+++ b/packages/server/api/src/app/ee/agent/agent-helpers.ts
@@ -176,7 +176,7 @@ function pickAllowedModel({ provider, selectedModel, candidates, modelScope, mod
     return selectedModel && allowed.includes(selectedModel) ? selectedModel : allowed[0]
 }
 
-function resolveModelIdForProvider({ provider, selectedModel, config, modelScope, modelIds }: { provider: AIProviderName, selectedModel: string | null, config?: AIProviderConfig, modelScope?: AiProviderModelScope, modelIds?: string[] }): string {
+function resolveModelIdForProvider({ provider, selectedModel, config, modelScope, modelIds, fallbackModelId }: { provider: AIProviderName, selectedModel: string | null, config?: AIProviderConfig, modelScope?: AiProviderModelScope, modelIds?: string[], fallbackModelId?: string }): string {
     const catalog = manualTextModelCatalog({ config })
     if (!isNil(catalog)) {
         return pickAllowedModel({ provider, selectedModel, candidates: catalog, modelScope, modelIds })
@@ -185,6 +185,9 @@ function resolveModelIdForProvider({ provider, selectedModel, config, modelScope
     const tierModelId = resolveTier({ tierId: selectedModel }).modelId
     if (provider === AIProviderName.ACTIVEPIECES || provider === AIProviderName.OPENROUTER) {
         return tierModelId
+    }
+    if (isNil(curatedModels) && !isNil(fallbackModelId) && fallbackModelId.length > 0 && isNil(findTier({ tierId: fallbackModelId }))) {
+        return pickAllowedModel({ provider, selectedModel: fallbackModelId, candidates: [fallbackModelId], modelScope, modelIds })
     }
     const nativeModelId = tierModelId.replace(/^[^/]+\//, '').replace(/\./g, '-')
     const candidates = isNil(curatedModels) ? [nativeModelId] : curatedModels.map((model) => model.id)
@@ -218,9 +221,9 @@ function reportKeyOutcome({ platformId, providerId, log }: { platformId: string,
     }
 }
 
-async function resolveTierModel({ platformId, tierId, provider, providerConfigId, scope, log }: { platformId: string, tierId: string, provider?: AIProviderName, providerConfigId?: string, scope: ProviderScope, log: FastifyBaseLogger }): Promise<{ model: LanguageModel, modelId: string, provider: AIProviderName }> {
+async function resolveTierModel({ platformId, tierId, provider, providerConfigId, scope, log, fallbackModelId }: { platformId: string, tierId: string, provider?: AIProviderName, providerConfigId?: string, scope: ProviderScope, log: FastifyBaseLogger, fallbackModelId?: string }): Promise<{ model: LanguageModel, modelId: string, provider: AIProviderName }> {
     const providerConfig = await resolveRunProvider({ platformId, scope, log, ...spreadIfDefined('provider', provider), ...spreadIfDefined('providerConfigId', providerConfigId) })
-    const modelId = resolveModelIdForProvider({ provider: providerConfig.provider, selectedModel: tierId, config: providerConfig.config, modelScope: providerConfig.modelScope, modelIds: providerConfig.modelIds })
+    const modelId = resolveModelIdForProvider({ provider: providerConfig.provider, selectedModel: tierId, config: providerConfig.config, modelScope: providerConfig.modelScope, modelIds: providerConfig.modelIds, ...spreadIfDefined('fallbackModelId', fallbackModelId) })
     return {
         model: agentAiUtils.createChatModel({
             provider: providerConfig.provider,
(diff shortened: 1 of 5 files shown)


hazemadelkhalel:
Superseded by #15355, which is merged. Your diagnosis was right: the resolver was deriving a model id from the tier by string surgery, and Bedrock has no catalogue to check it against. Credited you with Co-authored-by.

Two things pushed it to a different shape. This branch had gone CONFLICTING after #15346 moved the model-resolution code out of agent-helpers. And threading the run's model in does not hold, because withDefaultModel and the draft builder resolve a default with no catalogue and persist the result as the agent's modelName, so on a catalogue-less provider the stored model is itself a fabrication and reading it back trusts it. #15355 deletes the derivation instead: the Anthropic translation is now data on the tier, and a provider whose models we cannot list refuses rather than inventing, which covers all nine such providers rather than the two call sites.

PR #15297, opened 2026-09-06
fix(flows): reject flow operations computed against a stale version

AhmadTash:
## Description

The builder and the server each keep their own copy of a flow and stay in sync by mailing one-way, **name-keyed** operations — "add a step after `step_3`", "update `step_5`". Nothing on either side can tell that the two copies have diverged: the request carries no version, and the client never adopts the server's tree back. So a second editor session — a tab left open, another browser window — goes on editing a tree that no longer exists, and the server applies it.

That is how a flow ends up with **two steps sharing one internal step name**, which is the root of GIT-1805: the corrupted flow that grew to 632 steps and 1.9 MB and left the editor spinning forever.

### The sequence

The stale session allocates a name against its *own* tree, believing it free, while the server already has one. `_addAction` then trusts the client-supplied name verbatim — nothing has ever enforced uniqueness on write (`assertImportedStepNamesAreSafe` validates the name *pattern*, not uniqueness). Reproduced against a real server built from `main`:

```
both sessions load: trigger, step_1, step_2, step_3

session A pastes -> allocates step_4 -> HTTP 200
server now: trigger, step_1, step_2, step_3, step_4

session B pastes -> allocates step_4 -> HTTP 200   <- same name, stale tree

server now: trigger, step_1, step_2, step_4, step_3, step_4
duplicates: step_4 x2
```

The gap between the two pastes is irrelevant — session B's tree is stale from the moment A writes, whether that is 200 ms or the 29 minutes seen in the customer's audit trail. From there every later edit matches *both* nodes and the flow grows triangularly (a 35-step paste onto a collided tree adds T₃₅ = 630 nodes).

### What changed

**Operations now carry the version they were computed against**, as an `If-Match` header holding `${versionId}:${updatedMs}`:

- **Server** — the controller already loads the flow, so the check costs nothing. A mismatch throws `FLOW_VERSION_CONFLICT`, mapped to **412 Precondition Failed**. Including the version *id* alongside the timestamp also catches the draft being swapped out from under a session (`USE_AS_DRAFT`).
- **Client** — sends the token, and now **adopts `updated` from each response**. It was keeping its own tree and taking only `id` and `state`, which is precisely why it could never notice it had fallen behind. Within one tab the serialized queue keeps the token current, so there are no false conflicts; a second tab goes stale immediately.
- **Recovery** — on 412 the client discards the queued operations, reloads the flow, and tells the user their changes were not saved. `PromiseQueue.halt()` was terminal with no way back, so one failed request silently stranded every later edit in the session while the browser went on showing them as saved; `discardPending()` gives the conflict path a way to resume.

`applyOperation` already sets `updated` explicitly on every operation, so the token changes on every write — no new column and no migration.

### Deliberately opt-in

An API consumer that sends no `If-Match` behaves **exactly** as before. This is defence for the interactive editor, not a new requirement on the public API, which is why it is non-breaking.

### Relationship to #15291

Complementary, and independent — this branch is off `main` and does not depend on it.

- **#15291** rejects an `ADD_ACTION` whose name is already taken. That stops the *symptom* at the one chokepoint every gesture routes through, whatever seeded it — including a malformed `IMPORT_FLOW` that no version token would catch.
- **This PR** rejects any operation computed against a tree that no longer exists. That stops the *cause*, and covers divergence a name check cannot see: updating a step another session deleted, or a `DUPLICATE_ACTION` whose server-side name allocation no longer matches what the client optimistically applied.

Either one alone would have prevented GIT-1805. Both together close the class.

### Considered and rejected

- **Auto-renaming a colliding step server-side.** The client has already optimistically applied the name, so a silent rename makes the two sides disagree about a step's identity, and every later name-keyed operation targets the wrong node. Failing loudly is the safer half of "reject or auto-rename".
- **A per-flow lock.** It serializes writers but does not help here — the stale session's operation is well-formed and would simply acquire the lock and apply. Worth doing separately for a different reason: `LOCK_AND_PUBLISH`, `CHANGE_STATUS`, `USE_AS_DRAFT` and rename all call `flowsApi.update` *outside* the builder's serialized queue, so publish genuinely races in-flight edits against an unlocked read-modify-write.

## How was this tested?

Three new integration tests in `flow-operations.test.ts`, driving the real endpoint:

- a matching token applies the operation;
- a token from a superseded version returns **412** with `FLOW_VERSION_CONFLICT`;
- sending no token still works, pinning the backwards-compatible path.

**163/163 passing across all 16 CE flow integration test files** (`test/integration/ce/flows`), including the flow-run E2E suite. `turbo run lint` clean for `api`, `web`, `@activepieces/core-utils` and `@activepieces/core-execution` (pre-existing warnings only: 491 in api, 72 in web's `features/tables/`).

The two-session collision above was reproduced against a live stack built from `main` before the fix; that same script is the acceptance check for this branch, and re-running it is the one thing I have **not** yet done end-to-end — the local Docker engine needs a restart after an unrelated disk-space incident. The unit-level behaviour is covered by the integration tests above, which exercise the identical code path through the real controller.

Linear: [GIT-1805](https://linear.app/activepieces/issue/GIT-1805/bug-cloud-flow-editor-spins-indefinitely-after-publish-hung-on)

No env var, migration, column or field change. The `If-Match` header is optional and absent by default; a client that does not send it is unaffected. Only a caller that opts in can receive the new 412.

### DIFF
--- a/packages/core/execution/src/lib/flows/flow-version.ts
+++ b/packages/core/execution/src/lib/flows/flow-version.ts
@@ -41,11 +41,11 @@ export const FlowVersionMetadata = z.object({
 
 export type FlowVersionMetadata = z.infer<typeof FlowVersionMetadata>
 
-function tokenOf({ id, updated }: Pick<FlowVersion, 'id' | 'updated'>): string {
-    return `${id}:${new Date(updated).getTime()}`
-}
-
 export const flowVersionToken = { of: tokenOf }
 
 export const FLOW_VERSION_TOKEN_HEADER = 'if-match'
+
+function tokenOf({ id, updated }: Pick<FlowVersion, 'id' | 'updated'>): string {
+    return `${id}:${new Date(updated).getTime()}`
+}
 
--- a/packages/core/execution/src/lib/flows/flow-version.ts
+++ b/packages/core/execution/src/lib/flows/flow-version.ts
@@ -41,3 +41,11 @@ export const FlowVersionMetadata = z.object({
 
 export type FlowVersionMetadata = z.infer<typeof FlowVersionMetadata>
 
+function tokenOf({ id, updated }: Pick<FlowVersion, 'id' | 'updated'>): string {
+    return `${id}:${new Date(updated).getTime()}`
+}
+
+export const flowVersionToken = { of: tokenOf }
+
+export const FLOW_VERSION_TOKEN_HEADER = 'if-match'
+
--- a/packages/core/utils/src/lib/activepieces-error.ts
+++ b/packages/core/utils/src/lib/activepieces-error.ts
@@ -25,6 +25,7 @@ export type ApErrorParams =
     | ExistingUserErrorParams
     | FlowOperationErrorParams
     | FlowOperationInProgressErrorParams
+    | FlowVersionConflictErrorParams
     | FlowRunRetryOutsideRetentionErrorParams
     | InvalidApiKeyParams
     | InvalidAppConnectionParams
@@ -273,6 +274,14 @@ export type FlowOperationInProgressErrorParams = BaseErrorParams<
 ErrorCode.FLOW_OPERATION_IN_PROGRESS, {
     message: string
 }>
+
+export type FlowVersionConflictErrorParams = BaseErrorParams<
+ErrorCode.FLOW_VERSION_CONFLICT,
+{
+    expected: string
+    actual: string
+}
+>
 
 export type InvalidJwtTokenErrorParams = BaseErrorParams<
 ErrorCode.INVALID_OR_EXPIRED_JWT_TOKEN,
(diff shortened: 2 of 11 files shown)


AhmadTash:
@greptile review

All three P1s addressed in 5c9f531862:

**Version check is not atomic** — the read, check and write now run inside a per-flow `distributedLock` (`flow-operation-${flowId}`), the same primitive `applyStatusChange` already uses. It is taken only when the caller sent an `If-Match` token, so a client that opts out of the guarantee pays nothing for it.

**Stale edits get new tokens** — operations capture a `syncGeneration` at creation and skip themselves if a conflict reload has happened since, which reaches debounced work that is not yet in the queue. Worth noting why I did not take the other option you offered: capturing the *token* at creation is wrong in the opposite direction, because two edits queued back to back would both carry the pre-first-response token and the second would falsely conflict. The generation guard cancels stale work without breaking rapid sequential editing.

**Failed reload strands builder** — the queue now stays halted until the reload succeeds. On failure `saving` is cleared and the user is told to reload, rather than the builder silently resuming with a stale token.

One caveat I want to flag rather than let the new test imply more than it proves: `applies only one of two operations sharing a token` passes with *and* without the lock, because Fastify's in-process `inject` serializes the requests. I verified that by temporarily bypassing the lock. It pins the outcome, not the mechanism — the lock is there for genuinely concurrent processes, which this harness cannot reproduce. Commented as such in the test.

169/169 passing across all 17 CE flow integration test files; `api` and `web` lint clean.

AhmadTash:
@greptile review

Both addressed in 789d900abc.

**Private helper precedes export** — `tokenOf` now sits below `flowVersionToken` and `FLOW_VERSION_TOKEN_HEADER`, matching Imports → Exported functions/constants → Helpers → Types. (Function declarations hoist, so the const initializer is fine referencing it.)

**Lifecycle test claimed publish** — you were right, the name over-claimed. I first tried adding a real `LOCK_AND_PUBLISH` step, but publishing needs a piece trigger and the smoke flow uses an EMPTY one, so it returns `ENTITY_NOT_FOUND: Trigger not found for piece undefined@undefined`. Making it pass needs piece-metadata scaffolding that has nothing to do with this change, so I dropped the claim from the test name rather than half-build it here.

Worth flagging as a real gap though: **no integration test anywhere covers `LOCK_AND_PUBLISH`** — I grepped `test/integration/` and there are zero occurrences. That deserves its own change with the piece-metadata setup done properly.

169/169 across all 17 CE flow integration test files; `api` and `core-execution` lint clean.

PR #15298, opened 2026-09-06
feat(google-docs): improve the step forms

Talaljaber:
## Description

A step-form pass over the Google Docs piece, in the same spirit as the Google Sheets (#15200) and Tables passes: shorter copy, consistent labels, examples moved into placeholders, optional fields tucked into Advanced, and one field that asked users to paste Drive ids replaced with a picker. No prop names, default values, `audience` or `classification` were changed. One prop changed shape (`images`, below); everything else keeps its saved configuration untouched.

**Documents are picked, not pasted.** Read Document, Append Text and Edit Template File each took a `Document ID` text box. They now share one `Document` dropdown that lists the account's Google Docs from Drive (shared drives included) and filters server-side as you type. The search text is escaped before it goes into the Drive query (`O'Brien` and a backslash both work) and the picker asks for one page of 1000 instead of paging through the whole account on every keystroke; when Drive reports more, the placeholder says to type to narrow. The stored value is the same Drive file id as before.

**Image replacements get a real editor.** Edit Template File's `Images` field was a bare key/value dictionary with no label, description or placeholder on either side, so it was not clear what was being mapped to what. It is now an `Image Replacements` row editor with two labelled columns, `Image Object ID` (with a `kix.…` placeholder and a note that the id is found under `inlineObjects` in the Read Document output) and `New Image URL`.

**Compatibility of `images`.** This is the one prop whose type changes (`Property.Object` → `Property.Array({ properties })`), which is a data shape change, and two things had to move with it:

- *Engine side.* Before `run()`, the engine passes every ARRAY prop that has `properties` through `arrayZipperProcessor`, which turns a plain object into a one-element array holding that object. So a step saved on 0.4.9 with `{ "kix.x": url }` reaches `run()` as `[ { "kix.x": url } ]`, never as the dictionary. `toImageReplacements` reads an array row that carries neither editor column as a wrapped dictionary; the unit test pins that exact shape (`toImageReplacements([{ 'kix.one': … }])`), which is the only legacy shape production sends. A row with only one of the two columns is still skipped so a half-filled editor row is not misread as an image id. `typeof` / `in` narrowing, no casts, no `any`.
- *Builder side.* Every existing step carries `images: {}` (the web saves `{}` for any `Object` prop) and a patch upgrade keeps the saved input, so the new row editor opened on `{}` and `ArrayPieceProperty` threw into the field's error boundary. That is a web assumption in `array-property.tsx`, not something the piece can fix, so this PR carries the same one-file guard as #15189 (`toArrayFields()` returns `[]` for anything that is not an array), byte-identical — whichever PR merges first, the other rebases clean.

Flows pin exact piece versions, so nothing changes on merge; both paths matter only when a step is upgraded to 0.4.10 in the version picker.

**Placeholder format is a card choice.** The three formats (`{{KEY}}`, `[[KEY]]`, `{KEY}`) render as cards so the user sees the literal syntax instead of reading a description of it. Default unchanged.

**Copy.** Every description is now one sentence under the 70-character fold. `Append Text to Document` is `Append Text`; `Edit Template File` keeps its name but its Advanced section now holds only the optional image rows. `Find Document` says what "find" means (name contains) and `Create If Not Found` lost its question mark. The Service Account connection description is now a numbered setup guide with links to the IAM service-accounts page and domain-wide delegation, and the key field says to paste the whole downloaded file. Examples (`Weekly sync notes`, `Q3 report`, `https://example.com/logo.png`) moved into placeholders where the input renders one.

**Framework.** `minimumSupportedRelease` goes from `0.86.4` to `0.88.2` because the forms now use `advanced`, `placeholder` and `display: 'cards'`. Piece version `0.4.9` → `0.4.10`.

**Skill notes.** `.agents/skills/piece-builder/step-form-review.md` records what this review taught (a prop type change is a data migration: OBJECT saves `{}`, patch upgrades keep the saved input, the engine zips a plain object into `[obj]` before `run()`), and `check-step-form.mjs` gains a `--before <saved.json>` mode that errors on any prop whose type changed or that disappeared. Docs only; no runtime code.

**Known, accepted**
- **Old Document IDs in the new picker.** Steps saved before this PR hold a plain ID or a `{{…}}` expression in what is now a dropdown. The dropdown shows a label only for a value in its options and the `ƒ` toggle flips only on click, so an upgraded step that takes its ID from a previous step shows the placeholder while the saved value still runs. Clicking `ƒ` reveals it. Same class as Greptile's earlier note.
- **A filled legacy `images` dictionary opens as an empty row editor.** The guard renders a non-array value as no rows, and the generic editor cannot know which column a dictionary key maps to. The saved mappings still run until the user adds or removes a row, which writes the row array back and replaces the dictionary; re-enter the rows at that point. Inherent to the `[]` guard both reviewers proposed (and #15189); needs a step that used images on 0.4.9, an upgrade, and an edit to hit.

**Deliberately left alone**
- **Custom API Call** still shows the old form here. It is shared code in `pieces-common` and is being fixed in #15248; this PR does not touch it.
- The `Variables` field on Edit Template File stays a dictionary (people map whole objects into it) and stays `required`, which on an object property only paints the asterisk. Framework limitation, same as on Sheets.
- The folder picker (`folderIdProp`) still pages through every folder with an `any` cast, and Find Document's own `name contains` is still unescaped. Both are on PIE-493; both files change in this PR, so they follow once it merges rather than conflict with it. The Google Sheets half of that ticket is already up as its own PR (`fix/google-sheets-drive-query-escaping`).
- `translation.json` was regenerated with the CLI. Renamed labels orphan their existing translations until Crowdin catches up. The CLI also does not extract labels or placeholders nested inside an `Array` property's columns, so `Image Object ID` / `New Image URL` will show in English on every locale until the extractor grows that path. No locale files were hand-edited.

## How was this tested?

Manually on Windows 11 against a local CE dev server with `AP_DEV_PIECES=google-docs`, clicking through every changed form at the default panel width with a fresh step and a connected Google account.

- `.agents/skills/piece-builder/scripts/check-step-form.mjs` on the served metadata: **2 errors / 11 warnings / 1 info → 0 / 3 / 0**. The three remaining warnings are all on Custom API Call (shared code, #15248) or the cross-action `body` label, which the scanner cannot tell is shared code.
- Served metadata confirmed after an API restart: version `0.4.10`, `minimumSupportedRelease` `0.88.2`, `Image Replacements` is `ARRAY` + `advanced` with the two named columns.
- Document picker: disabled with "Connect your account first" before a connection is chosen, lists the account's documents once connected, narrows the list as you type, returns the full list for a whitespace-only search, and survives `O'Brien` and a backslash.
- `vitest run` in `packages/pieces/community/google-docs`: 5/5, covering the row-editor shape, the engine-wrapped legacy dictionary, the bare dictionary, half-filled rows and empty inputs. The engine-wrapped case is the shape the reviewer reproduced against a real document; the run-time behaviour was confirmed there rather than re-run here.
- Web: `tsc --noEmit -p tsconfig.app.json` clean; `array-property.tsx` is byte-identical to #15189's branch.
- `npx turbo run build lint --filter=@activepieces/piece-google-docs --force`: build green, lint 0 errors, 2 warnings, both pre-existing `any`s in the folder picker (PIE-493).

Linear: PIE-492
Follow-up: PIE-493

Flows pin piece versions, so nothing changes on merge. On upgrade to 0.4.10 the saved `images` value keeps working through the wrapped-dictionary path; the builder renders the old value instead of erroring.

The Document picker's search text is now escaped before it is interpolated into the Drive `q` literal, closing a query-shape injection rather than opening one; the query runs under the user's own OAuth token against their own Drive.


ibrahim-abuznaid (changes requested):
Thanks Talal. I ran this on the local build and clicked through Read Document, Append Text, Edit Template File (Advanced closed and open) and Find Document. The forms match the description: the Document picker with "Connect your account first", the placeholder-format cards with `[[KEY]]` preselected, Variables, then "Advanced · 1 option" holding Image Replacements with the two labelled columns and placeholders, Find Document with the `Q3 report` placeholder and the Create If Not Found toggle. With a Google connection the Document picker lists the account's documents (newest first), narrows on search (`PR15298` → 2), returns the full list for a whitespace-only search, and survives `O'Brien` and a backslash without an error, so the escaping fix works. Served metadata is 0.4.10 / `minimumSupportedRelease` 0.88.2, every changed description is under 70 characters, lint is 0 errors (the two `any` warnings are pre-existing), vitest 4/4. I also ran Edit Template File for real with the new row shape: two `[[…]]` placeholders replaced and the inline image swapped in one `batchUpdate` (three replies). Good pass on the copy and the picker.

**Blocking: the legacy `images` dictionary never reaches `toImageReplacements` in the shape you tested.** Before `run()`, the engine passes every ARRAY prop that has `properties` through `arrayZipperProcessor` (`packages/server/engine/src/lib/variables/props-processor.ts:66`, `processors/array-zipper.ts`). A plain object is not passed through: it becomes a one-element array containing that object. So a step saved with `{ "kix.one": "https://…/one.png" }` arrives in `run()` as `[ { "kix.one": "https://…/one.png" } ]`. Your array branch drops that row because it has no `imageObjectId` / `url`, and the dictionary branch is never reached. The step succeeds and replaces no images. I ran it for real on a test document with one inline image (`kix.7cgcw4kby480`), once per shape:

```
NEW    images=[{imageObjectId:"kix.7cgcw4kby480", url:…/google-sheets.png}]
       batchUpdate replies = [{replaceAllText:{occurrencesChanged:1}}, {replaceAllText:{occurrencesChanged:1}}, {}]   → image contentUri changed
LEGACY images={"kix.7cgcw4kby480": …/google-drive.png}
       batchUpdate replies = [{replaceAllText:{}}]                                                                     → image contentUri unchanged, step SUCCEEDED
```

The legacy run sent one request (the text one) and no `replaceImage`; the run log still shows the dictionary as the step input, so nothing in the UI tells the user the images were ignored. Same result when I fed the action's `props` straight through `propsProcessor.applyProcessorsAndValidators`: the engine hands `run()` `[ { "kix.…": "…" } ]`.

Fix in the helper: a row that is an object without the two columns is a wrapped legacy dictionary.

```ts
const fromDictionary = (dict: object): ImageReplacement[] =>
  Object.entries(dict).map(([imageObjectId, url]) => ({ imageObjectId, url: String(url) }));

export function toImageReplacements(images: unknown): ImageReplacement[] {
  if (Array.isArray(images)) {
    return images.flatMap((row) => {
      if (typeof row !== 'object' || row === null) return [];
      if ('imageObjectId' in row && 'url' in row) {
        return [{ imageObjectId: String(row.imageObjectId), url: String(row.url) }];
      }
      // The engine wraps a legacy `{ objectId: url }` dictionary in a one-element array before run().
      return fromDictionary(row);
    });
  }
  if (typeof images === 'object' && images !== null) return fromDictionary(images);
  return [];
}
```

Please add the wrapped shape to the test file too (`toImageReplacements([{ 'kix.one': '…' }])`). It is the only legacy shape production will ever send.

**Second thing, and the reason I would not write "existing flows keep working" in the body yet: the builder shows an error on every upgraded Edit Template File step.** The old `Property.Object` saved `{}` by default (the web gives OBJECT props `{}` in `form-utils.tsx`), so every existing step carries `images: {}` whether or not anyone used images. A patch upgrade in the version picker keeps the input as-is (`getInputAfterVersionChange` returns `currentInput`). The web's `ArrayPieceProperty` then calls `.map` on that object, throws, and the field's error boundary renders "input value is invalid, please contact support" where Image Replacements should be. Reproduced with `{}` and with a filled dictionary, with and without a connection. With the connection the Document dropdown resolves the document name fine, and the error sits right below it. I have the screenshots if you want them. The piece cannot fix this one, it is a web assumption in `array-property.tsx:75-83`. Two ways out, your call: (a) a companion one-line web fix (`Array.isArray(formValues) ? formValues.map(…) : []`, plus the same guard in `append`/`remove`) and a sentence in the PR body, or (b) keep `images` as `Property.Object` and drop the row editor. I lean (a); the row editor is the best change in this PR.

For scope: flows pin exact piece versions, so nothing changes on merge. Both issues appear only when someone upgrades a step to 0.4.10, which the version picker offers by default. Not urgent, but the body's compatibility claim is not true today, so please fix before merge rather than after.

**Non-blocking**

- Old steps in the new Document picker. Steps saved before this PR hold a plain ID or a `{{…}}` expression in what is now a dropdown. The dropdown shows a label only when the value is in the fetched options, otherwise it shows the placeholder, and it does not switch to dynamic mode on its own (the toggle changes only on click). An upgraded Append Text step that takes its ID from a previous step therefore shows a placeholder while the saved expression still runs. Same class as Greptile's last note. Fine to accept, but say it in the PR body.
- `folderIdProp` still pages through every folder with an unescaped query and an `any` cast, and the Sheets spreadsheet picker has the same unescaped `name contains`. You said follow-up: please open the PIE ticket and link it here.
- "Fixes # (issue)" is still the template text. We track pieces work in Linear; please link a PIE ticket.

ibrahim-abuznaid (approved):

PR #15299, opened 2026-09-06
feat(ninety): create and update EOS to-dos, issues and rocks, and trigger on new ones

### LINKED ISSUE #15270: [Feature]:[Pieces] - New ninety.io piece (EOS platform)
## Context

New piece: [ninety.io](<http://ninety.io>) (EOS business-operating-system app: To-Dos, Issues, Rocks, Milestones, Scorecard). Requested by a paid customer via sales. No Pylon ticket.

Ninety Public API v1, base `https://api.public.ninety.io/v1`. Interactive spec: `https://api.public.ninety.io/v1/swagger`.

Requester has not given a concrete workflow yet, so v1 below is the minimum useful surface, not the whole API. Deferred items at the bottom.

## Auth (verified in Ninety docs)

* Personal Access Token, `Authorization: Bearer <token>`. Generated in Ninety under Settings > Developer Settings.
* No OAuth2. One `SecretText` connection prop.
* API access gated to Ninety's Thrive plan. Requests run with the token owner's permissions, so a teams dropdown only ever sees that user's teams.
* Rate limit 25 req/s per user, `429` on overage, resets per second.

## Constraint that shapes the design (verified)

**No webhooks.** Ninety docs: "No webhooks or push notifications (polling required)". Every trigger polls.

Query endpoints are `POST .../query` taking `sortField` / `sortDirection` / `pageIndex` / `pageSize`. `createdDate DESC` + dedupe on id is the pattern for all three triggers.

### Triggers (3, polling)

* New To-Do: `POST /v1/todos/query`, optional `teamId`, `isPersonal`
* New Issue: `POST /v1/issues/query`, optional `teamId`, `intervalCode` (SHORT_TERM | LONG_TERM)
* New Rock: `POST /v1/rocks/query`, optional `teamId`, `userId`

### Actions (13)

* To-Dos: Create `POST /v1/todos`, Update `PATCH /v1/todos/{id}`, Find `POST /v1/todos/query`
* Issues: Create `POST /v1/issues`, Update `PATCH /v1/issues/{issueId}`, Find `POST /v1/issues/query`
* Rocks: Create `POST /v1/rocks`, Update `PATCH /v1/rocks/{id}`, Find `POST /v1/rocks/query`
* Milestones: Create `POST /v1/milestones`
* Scorecard: Find Measurables `POST /v1/scorecard/kpis/query`, Set Score `POST /v1/scorecard/kpis/{kpiId}/scores`
* Teams: List Teams `GET /v1/teams`
* Custom API Call

### Props

* `teamId`: dynamic dropdown from `GET /v1/teams` (returns `id` + `name`). Shared across triggers and actions.
* `userId`: free text, NOT a dropdown. Users API is get-by-id only, no list endpoint exists.
* Rock create body is nested: `{ rock: { teamId, title, dueDate, statusCode, levelCode, quarter, description, additionalTeamIds, futureScope }, addCreatorToFollowersList }`.
* Issue create requires `teamId`. To-Do create makes a personal to-do when `teamId` is omitted or empty.
* Issue `priority` is 0 to 5.

## Deferred (add on request)

Custom API Call already covers these as single HTTP calls, so they are not worth dedicated actions until someone asks:

* Get by ID for To-Dos, Issues, Rocks, Milestones, Users (`Find` covers the list case)
* Delete To-Do / Issue / Rock (soft) / Measurable Score / Measurable Note
* Update Milestone, Set Measurable Note

Two triggers deliberately left out on reliability grounds, not size:

* **Issue Completed**: the issues query DTO has no `completed` filter, so it means filtering a `createdDate`-sorted page client side. Misses anything completed outside the current page window.
* **Updated Rock**: `updatedAt DESC` polling refires on every field touch and does not dedupe cleanly.

## Blocked by the API (not piece gaps)

* Webhook triggers: none exist.
* New Milestone trigger: no milestone query/list endpoint. Milestones only come back nested on `GET /v1/rocks/{id}`.
* New Score trigger: scorecard is write and delete only, no score read endpoint.
* Delete Milestone: no endpoint.
* User directory / assignment endpoints: none.
* Bulk writes: none.

OdaiAhmed99:
## Description

Ninety is where teams run EOS — to-dos, issues, rocks, milestones and the scorecard. A paid
customer asked for it through sales. This adds 13 actions plus Custom API Call, and 3 triggers,
against Ninety's Public API v1.

Ninety has no webhooks; their docs say so outright ("No webhooks or push notifications, polling
required"). So all three triggers poll, sharing one factory, and dedupe on `createdDate` rather
than on an id. An id-keyed dedupe cannot work here: rocks expose `_id`, not `id`, so the key
would be `undefined` for every rock, `pollingHelper` would delete the checkpoint instead of
storing it, and every poll would re-emit the whole page forever. There's a regression test
holding that down.

The awkward part is that the three query endpoints don't share a dialect, and only one of them
can sort by creation date. Issues take `sortField`/`sortDirection` in uppercase with a
zero-based `pageIndex`, and `createdDate` is a valid sort field, so New Issue really does get a
newest-first page. To-dos take `sort`/`order` with a one-based `page`, but `sort` only accepts
`title`, `dueDate`, `ordinal`, `userOrdinal` and `completedDate` — ask for `createdDate` and you
get a 400. Rocks reject `createdDate` as a sort field too. So New To-Do and New Rock read a page
and let the timestamp checkpoint work out what's new.

Two field-name rules worth knowing if you review this: every Ninety record uses `_id`, except
users, which use `id`; and a user's address is `primaryEmail`, not `email`.

One limitation to flag rather than hide: Create Milestone can't create a completed milestone.
Ninety accepts `isDone` and `completedDate` on the create endpoint according to its own spec,
then ignores both and stores the milestone as not done. Rather than leave two inputs that
quietly do nothing, the action doesn't offer them and says so in its description.

**Out of scope:** get-by-id and delete actions, Update Milestone, Set Measurable Note — Custom
API Call covers each in one HTTP call. Issue Completed and Updated Rock triggers were left out
for reliability, not effort: the issues query has no completed filter, and `updatedAt` polling
refires on every field touch. New Milestone and New Score are blocked by the API outright.

**Limits we can't engineer around:** since neither rocks nor to-dos can be ordered by creation,
New Rock reads the team's non-archived rocks at Ninety's maximum page size of 200, and New To-Do
walks non-archived to-dos 100 at a time up to 5 pages, stopping at the first short page — so a
normal team costs one request. A team sitting on more than 500 open to-dos, or 200 open rocks,
could miss one.

The public API is gated to Ninety's Thrive plan; a token from a lower plan gets a 403. Nothing
else is needed to run it — one Personal Access Token pasted into a connection, no env var and no
self-hosted setup step.

This also adds a few gotchas to the E2E wiki page, picked up while testing this against a real
account.

## How was this tested?

Against a live Ninety Thrive account, and then through the builder — which was worth doing,
because between them they turned up ten defects that the unit tests could never have caught.
Every one was the piece faithfully implementing what Ninety's OpenAPI document said.

The three that meant it simply didn't work:

- The New To-Do trigger 400ed on every poll, because it asked to sort by `createdDate`. It could
  never have fired once.
- Find To-Dos 400ed on every call, for the same reason — its Sort By defaulted to `createdDate`.
- Every Owner dropdown stored `value: undefined`, because `GET /v1/users` returns `id` and the
  code read `_id`.

The rest were schemas promising fields that never arrive: to-dos and issues come back as `_id`
not `id`; a to-do has no `updatedDate`; a user has `primaryEmail`, not `email`; the rock schemas
offered `milestones`, `updatedAt` and `additionalTeamIds` on endpoints that don't return them;
and Create Milestone declared a `completedDate` it can't produce. Each is fixed and pinned by a
test.

Verification ran in three passes. First the piece's own code against the live API, one action at
a time. Then all three triggers through the real builder with Playwright — pick the connection,
Test Trigger, and check every key the output schema declares actually exists on the payload that
comes back. Finally every action's output audited the same way, nested list fields included.

136 tests in the piece (`npx vitest run`) cover request shaping, envelope unwrapping, date
normalisation, the nested rock body, which fields a PATCH sends and which it omits, the polling
checkpoint and the New To-Do page walk. The polling suite was checked for discrimination by
reverting the factory to an id-keyed dedupe: 8 of 19 fail. Build and lint are green.

CE only — this is a community piece with no edition-gated behaviour.

Fixes GIT-1845


OdaiAhmed99:
Thanks — the New Issue finding is real, and the fix is in 8af3cb1.

Issues are the one Ninety resource that can be ordered by creation date, so rather than raise the page size the trigger now keeps reading pages until one reaches past the checkpoint, capped at five. The shared factory passes `lastFetchEpochMS` through to make that possible. A quiet team still costs one request, and the Test button takes a single page instead of walking the account. Three tests cover it, including the 150-in-one-burst case; reverting the walk to a single page fails two of them.

The to-do and rock triggers cannot do the same, because `/todos/query` and `/rocks/query/paged` both reject `createdDate` as a sort field — verified against a live account, with a bogus-value control to confirm the rejection is real validation rather than a silently ignored key. They instead read the whole open set behind a cap, which is why their limits are stated in the description rather than fixed.

On the second point about `index.ts` — I could not work out which requirement is meant. There is no `displayName` validation in the CLI or in CI that I can find, and the `authors` handle already appears in both forms across merged pieces by the same author. If you can name the specific check, I will fix it.

OdaiAhmed99:
Both fixed.

**The cap.** You were right that five pages was a business limit wearing a safety guard's clothes, and I had the trade-off backwards. The walk only runs as far as there are issues newer than the last poll, so its length tracks how fast a team raises issues, not how many exist — a quiet team still costs one request. Fifty pages now stands where five did, and its job is to stop a stale checkpoint reading an entire account in one poll, not to bound what a real team can produce between two checks. `567374f`.

Worth naming why the to-do and rock triggers keep a five-page cap: theirs is a different shape. Neither endpoint can sort by creation date, so those triggers read the whole open set on every poll rather than walking back to a checkpoint. Raising their cap would make every poll read up to fifty pages forever, not just during a burst. That one really is a coverage-versus-cost trade, and the description states the ceiling.

**The title.** Also right, and the convention was sitting in the history where I should have looked — `feat(clay): send rows to a table and trigger on rows Clay sends back`, `feat(flows): sort the automations list by name`. Retitled to `feat(ninety): create and update EOS to-dos, issues and rocks, and trigger on new ones`.

ibrahim-abuznaid:
Reviewed locally: ran the suite (139/139), eslint (clean), and mutation-tested rather than trusting the pass count. Strong piece — honest description, clean commits, and `classification` / `audience` / `aiMetadata` / `outputSchema` complete and correct across all 13 actions and 3 triggers. Schemas match the response types field for field; i18n has zero missing keys and zero orphans.

**Three one-liners before merge**

1. `src/index.ts:30` — `authors: ['odaithalji']` is a dead handle (404). Yours is `OdaiAhmed99`, as in clay / mssql / qbd-conductor. Same typo is in formio, worth a follow-up.
2. `.eslintrc.json:6` — the `["*.ts","*.tsx"]` override is `"rules": {}`. 729/731 pieces carry the `no-restricted-imports` block; copy formio's. Lint stays green.
3. `package.json:7-10` — missing `@activepieces/core-piece-types` and `@activepieces/core-utils`, declared by 727/731 pieces.

**The tests are good but oversold**

The polling tests are the best thing here and they hold up — I threw seven mutations at them and all seven were caught (poisoned `createdEpoch`, zero-based page walk, ASC sort, dropped date truncation, un-nested rock body, tri-state returning `false`, `primaryEmail`→`email`). The boundary cases are genuinely pinned.

But the suite asserts only on the request *body*, so each of these leaves **139/139 green**:

- every `PATCH` → `POST` (no test asserts `request.method` — `grep -n method test/*.ts` is empty)
- drop `statusCode` from Update Rock (omission is pinned, inclusion isn't)
- delete `findRocks`, `createMilestone`, `findMeasurables` from the actions array (nothing imports `src/index.ts`)
- revert the owner dropdown to `user._id` — the exact bug you describe as "pinned by a test". Nothing imports `props.ts` or `auth.ts`, so all six dropdown loaders and `validate` are dark.

Cheapest fixes: assert the verb in the update/query tests, one "everything set" body assertion per write action, one test importing the piece. Also `client.test.ts:284` is a tautology (`userLabel` is typed `: string`), and `actions.test.ts` / `action-edge-cases.test.ts` stub `pieces-common` wholesale including a fake `HttpError`, so `describeError`'s `instanceof` can never be true there — the polling files' `importOriginal()` pattern is the right one.

Not your doing: **CI never runs these tests.** `ci.yml:152` filters `turbo run test` to an allowlist with no piece target; 18 pieces ship a vitest config and none run on any PR. Tracked in PIE-490.

**One to confirm:** `find-rocks.ts:61` sends `futureScope: 'all'` verbatim. That value appears nowhere else in the piece, has no test, and leaving the dropdown empty already means all horizons. Did you check it live?

**Minor:** `new-rock.ts:13` aiDescription still promises `milestones`, which the trigger schema drops · `client.ts:496,548-549` dead type fields · `create-rock.ts:45` follow-flag Checkbox has no `defaultValue`, so the builder always sends `false` and overrides Ninety's default · `update-issue.ts` is the only update action with no Owner and no Archived setter · page/pageSize unbounded on the find actions · a few `=== undefined` that should be `isNil` · description says 136 tests, it's 139.

Greptile's two comments are both stale — the pagination one was fixed by your two follow-up commits.

ibrahim-abuznaid (approved):

PR #15307, opened 2026-09-07
fix(auth): guard the last sign-in method and validate SSO domains

lina-alomari:
## Description

Four fixes to Platform → Security → **Single Sign On**, all in the same surface: a destructive action with no confirmation, a lockout footgun, domain fields that accepted values the server would never honour, and an error message that fired mid-keystroke.

**Confirm before disabling SAML SSO.** The `Disable` button in the SAML wizard called `platformApi.update({ federatedAuthProviders: { saml: null } })` on the first click. That drops the IdP metadata and certificate server-side, so re-enabling means fetching both from the identity provider again — an irreversible action behind an unlabelled button, with a success toast as the only feedback. It now routes through `ConfirmationDeleteDialog`, which names exactly what is lost and, when an `ssoDomain` is set, that the domain survives the disable. `ConfirmationDeleteDialog` gained an `onSuccess` callback for this: the dialog owned the mutation but exposed only `onError`, so the wizard had no hook to close itself and toast on success.

**Refuse to disable the last usable sign-in method.** Nothing stopped an admin from turning off email auth, Google auth and SAML in sequence and leaving the platform with no way to sign in. Guarded on both sides — `assertSignInMethodRemains` in `platform.service.ts` rejects the update with `ErrorCode.VALIDATION`, and the UI disables the switch (and the SAML `Disable` button) with a tooltip pointing at the cause, so the API refusal is the backstop rather than the first thing an admin meets.

The subtle part is what counts as *usable*. `platform.googleAuthEnabled` can be `true` on a server that has no Google credentials, in which case Google sign-in does not work and cannot be the method that remains. Google therefore counts only when `AP_GOOGLE_CLIENT_ID` **and** `AP_GOOGLE_CLIENT_SECRET` hold non-blank values, via a single `signInMethodUtils.isGoogleConfigured()` shared by the guard and by the `THIRD_PARTY_AUTH_PROVIDERS_TO_SHOW_MAP` flag that the sign-in page and the new `useSignInMethods` hook read — one definition, so the guard and the button can never disagree.

Blank is the case worth spelling out. `system.get` returns `process.env[...]` verbatim and its `?? default` only catches `null`/`undefined`, so `AP_GOOGLE_CLIENT_ID=""` arrives as `''` and `isNil('')` is `false`. Both the old flag and the first version of this guard therefore read an empty credential as "configured" — the flag has been rendering a Google button against an empty client id, which `federated-authn-service.ts` hands straight to the OAuth client. Treating blank and whitespace-only as unset fixes the guard and that pre-existing button in one place. (Raised by Greptile on the first pass, as a P1 against the guard; the shared predicate takes it further than the suggested one-line change so the flag is fixed too.)

That mismatch is also surfaced directly now: a self-hoster with the toggle on and no credentials configured gets an inline hint naming the two env vars. The hint is suppressed on Cloud, where both vars are provisioned and are not an admin's to set.

**Validate auth domains against real domain rules.** Allowed auth domains accepted any non-empty string (`z.string().min(1)`) on the client and were stored unvalidated; `POST /sso-domain` accepted any string up to 253 chars; and the SSO domain form used `z.hostname()` plus a `.includes('.')` refine, which accepts labels the server would then persist. All three sites now agree on `z.regexes.domain`, and a fourth validator joins them: `AllowedAuthDomain` in `platform.service.ts`, so allowed domains are checked server-side rather than being stored on trust. Allowed domains are also trimmed and lowercased before they are saved, and `normalizeAllowedAuthDomains` does that server-side too, so an API client gets the same treatment as the form rather than the form being the only thing standing between a typo and the database. `ssoDomain` is the narrower case: the form trims and lowercases it on the way out, but the server only validates the shape and stores what it is given.

**Stop flagging a half-typed SSO domain as invalid.** `FormMessage` rendered on every keystroke, so typing `acme.` put "Enter a valid domain, like acme.com" under the field mid-word. Both domain fields now defer the message until the field has been touched or a submit has been attempted.

Two smaller things ride along, both consequences of the above:

- `ssoMutations.useUpdatePlatformSso` and the allowed-domains dialog now surface server errors with `api.extractServerErrorMessage` instead of failing silently — necessary once the server can refuse an update the UI thought was fine.
- The three mutations that can change which sign-in methods exist — `ssoMutations.useUpdatePlatformSso`, the SAML disable dialog and the SAML save step — invalidate the flags query, so `THIRD_PARTY_AUTH_PROVIDERS_TO_SHOW_MAP` reflects a SAML change without a page reload. Without it the guard reads stale state right after the mutation that changed it. The allowed-domains, save-domain and verify-domain mutations deliberately do not: none of them touches a sign-in method.

**Postponed: the Google toggle's edition scope.** An earlier commit on this branch hid the Google row on Cloud; it has been reverted, so the row renders on every edition again, exactly as on `main`. The reasoning that motivated it still holds and is worth recording here, because it is why the control needs rethinking rather than relocating. The Cloud sign-in page is an anonymous request, so `getPlatformIdForRequest` short-circuits to `null` there and the flags hook takes its env-only branch, in which `platform.googleAuthEnabled` is not part of the expression. It *is* read once a principal resolves (`enterprise-flags.hooks.ts` builds the map as `isGoogleConfigured() && platform.googleAuthEnabled`), so flipping it on Cloud moves the admin's own authenticated flag view and nothing a signing-in user would ever see — which is exactly why it looks like it works.

And even where the toggle does reach the sign-in page, it controls **whether the button renders**, not whether Google sign-in works. `THIRD_PARTY_AUTH_PROVIDERS_TO_SHOW_MAP` is the only consumer of `platform.googleAuthEnabled` in the product. `GET /v1/authn/federated/login` and `POST /v1/authn/federated/claim` are both `securityAccess.public()` and read neither it nor anything derived from it, and `assertEmailAuthIsEnabled` returns early for every non-`EMAIL` provider, so there is no Google counterpart to the email check. Turning Google off hides the button and leaves the endpoint open.

All of that is pre-existing and stays untouched: a real way to control the Google auth flow is being worked out separately, and hiding the row was the wrong shape for it. Nothing in this PR changes where the Google toggle appears or what it does — the only Google-related change that remains is the shared `isGoogleConfigured()` predicate above, which the guard needs and which fixes the blank-credential button as a side effect.

## How was this tested?

**Automated** — the full CI-equivalent suite locally, all green:

| Check | Result |
|---|---|
| `npm run lint-core` | 0 errors, 23/23 tasks (498 pre-existing warnings, none in a file this PR touches) |
| `turbo run build` (web, worker, api, engine, core pieces) | 26/26 tasks |
| Unit tests (28/28 tasks) | web **725** (78 files), engine 471 (39), sandbox 265 (20), worker 271 +1 skipped (17), shared 508 (16), server-utils 180 (13), core-execution 139 (11), core-utils 28 (2), pieces-framework 15 (3) |
| `test-ce` | 1,060 passed, 3 skipped (88 files) |
| `test-ee` | 332 passed (32 files) |
| `test-cloud` | 565 passed (53 files) |
| `check-migrations` | no schema drift |
| `check-migration-rollback` | no new migration files |
| `PR size` | server/api 282/600 · web 511/1,200 |

New coverage, both in suites CI actually runs:

- `test/integration/cloud/platform/platform.test.ts` — rejects disabling the last remaining sign-in method, rejects it when the Google credentials are blank or whitespace-only, allows the disable while another method remains, and normalizes allowed auth domains before saving.
- `test/app/routes/platform/security/sso/saml-dialog.test.tsx` — the domain field stays quiet while a partial domain is typed, shows the error on blur, and clears it once the domain is valid.

**Editions** — the guard lives in `platform.service.ts` on the CE path, so it applies to CE, EE and Cloud alike; the SSO page's sidebar entry is locked when `platform.plan.ssoEnabled` is false, though the route itself carries no guard. The Google credentials hint is suppressed on Cloud only; the Google row itself renders on every edition, unchanged from `main`. No new env var, secret or setup step is introduced — the hint documents config that already exists.

Every number above was re-measured on the current base, after the Google-row revert and after the second `main` merge. One flake worth naming: `server-utils`' `model-catalog.test.ts` failed once on `ENOTFOUND cdn.activepieces.com` — it fetches the live AI model catalog, is in a package this PR does not touch, and passed 180/180 on re-run.

**Known nit:** the disable-confirmation message branches on `platform.ssoDomain` being set, but its wording says the domain "stays verified" — so a domain that is set and still pending verification is told it stays verified. Worth branching on `ssoDomainVerification.status` instead, or softening the copy; say which and I'll change it.

**Not done:** no manual click-through of the flows in a running app, and no screen recording. The web tests cover the deferred-error behaviour; the confirmation dialog, the disabled switches and the tooltips are asserted only indirectly through types and lint.

## Fixes

No linked issue.

No field, endpoint or column changed, and nothing is newly required. `POST /v1/platforms/:id` does gain two ways to return `409` (`ErrorCode.VALIDATION`, per `error-handler.ts`): disabling the last usable sign-in method, and an `allowedAuthDomains` entry that is not a valid domain. Both previously "succeeded" into a state the product cannot honour — a platform nobody can sign in to, or a domain that matches no email address — so no caller doing something meaningful has to change. Flagging for reviewer judgement rather than assuming: if we count a newly-rejected call as a functional break regardless of whether it worked, this needs the `⛓️‍💥 breaking-change` label and an entry in `docs/install/reference/breaking-changes.mdx`.

Authentication surface: it changes which sign-in methods a platform can end up with, and what a SAML domain may contain. The direction is restrictive — it removes a lockout path and tightens domain validation on both the form and the API — and it grants nothing new. Two things worth a reviewer's eye: the guard must not be bypassable by a caller that sends `federatedAuthProviders` and `emailAuthEnabled` in one request (covered by the "allows disabling while another remains" test), and `ssoDomain` matching is what routes a user to an IdP, so `z.regexes.domain` replacing `z.hostname()` narrows, never widens, what is accepted.


othmanemad:
Reviewed at `5a3275e178`. I checked the form behaviour by actually running it against the repo's `react-hook-form` setup, not just reading the diff.

**1. The guard counts Google. The doc in this same PR says it shouldn't.** (blocking)

Code:
```ts
const googleRemains = googleConfigured && (params.googleAuthEnabled ?? platform.googleAuthEnabled)
```
The brain page you added in `a0b23d542d` says:
> An anti-lockout guard must therefore count only email and SAML […] counting `googleAuthEnabled` lets an admin turn off email and SAML and strand every member who has no Google account.

`use-sign-in-methods.ts` follows the code, not the doc. Pick one — either stop counting Google, or delete that bullet.

**2. The allowed-domains dialog can lock the admin out of editing their own list.** (blocking)

Save is disabled when the form is invalid, and the form is checked as soon as it opens. The old rule was `min(1)` and the server didn't check at all, so people may already have entries like `@acme.com` saved. What I measured:

```
stored 'acme.com'   -> Save enabled
stored '@acme.com'  -> Save disabled     <- legacy row
after "Add Domain"  -> Save disabled
clicking Save       -> nothing happens
```

And nothing tells them why. This form has no `mode`, so react-hook-form doesn't fill in error messages until a submit happens — and submit can't happen because the button is off:

```
typed "acme" -> no error shown
then blurred -> still no error shown
```

So someone opening the dialog just to delete a bad entry sees a dead Save button and a blank form. The new `invalidAuthDomain` message never appears here.

**3. Same dead end on the SAML domain field, but milder.**

That one uses `mode: 'onChange'`, so the error does exist — it's just hidden until the field is blurred, while Save stays off. Type `acme` and you get a greyed-out button with no reason given until you click away. The three new tests always blur first, so they don't catch it.

Either show the message once the field isn't empty, or leave the button enabled and let submit show the error.

**4. The guard doesn't check `plan.ssoEnabled`.**

It only asks "is a SAML config saved". If a platform's SSO entitlement lapses while email auth is already off, they're locked out — the exact thing this PR is meant to prevent.

**5. Domains aren't de-duplicated after lowercasing.**

Neither the form nor `normalizeAllowedAuthDomains` drops duplicates. This PR makes it easier to hit: `Acme.com` and `acme.com` used to be two different-looking entries, now they both save as `acme.com` and you get the same domain listed twice.

PR #15357, opened 2026-09-08
fix(web): show friendly trigger errors

### LINKED ISSUE #15324: [BUG]: the flow status-change error dialog prints the raw piece error at customers
Split out of activepieces/activepieces#15277, which is scoped to the fix only.

**Problem**

When a flow status change fails, the web client opens a "Status update failed" dialog and renders the piece's error verbatim:

`packages/web/src/features/flows/hooks/flow-hooks.tsx`, the `ErrorCode.TRIGGER_UPDATE_STATUS` branch, passes `params.standardError` straight into the dialog.

What the customer sees is the raw serialized piece error: a minified stack trace pointing at a bundle path inside our container, an `errorName` like `"_"`, the full GraphQL mutation text, and the third party's own wording.

**Why it matters**

This is not theoretical. In Pylon 5833 a Linear cleanup call returned 401 and the dialog printed Linear's own sentence, "Authentication required, not authenticated. You need to authenticate to access this operation." The customer reasonably read that as their Activepieces session having expired, checked that they were signed in and an admin, and reported it as a false-logout bug. The real cause was a rejected Linear API key on the connection, which the dialog never said.

So the raw dump actively produced a misdiagnosis, and cost a support round trip on a trial account.

**What is needed**

The dialog should lead with a human sentence naming the piece and what failed, and keep the raw text behind a disclosure for anyone who wants it.

A formatter for exactly this already exists and is unused on this path: `formatPieceError` / `tryParseFriendlyPieceError` in `packages/core/utils/src/lib/friendly-piece-error.ts`. It already extracts a plain message, an HTTP status and the API's own message from a thrown piece error. The `standardError` in this dialog is already a serialized `FriendlyPieceError` (`__apErrorVersion: 1`), so it can be parsed rather than printed.

jabrailkhalil:
## Description

When a flow status change failed, the dialog displayed the trigger piece's serialized error directly. Customers could see stack traces, request details, and provider-specific raw payloads instead of the actionable error message.

The dialog now parses the existing FriendlyPieceError format, displays its readable message, and keeps the raw payload under collapsed Technical Details. The status-change hook passes the parsed error through without changing the fallback behavior for non-friendly errors.

## How was this tested?

- Added a regression test covering a serialized FriendlyPieceError: the friendly message is shown and technical details are collapsed.
- Targeted web Vitest test passed.
- Web TypeScript typecheck passed.
- Changed source files pass ESLint.
- `bun run lint-dev` passed.
- The change is UI-only and does not vary by CE / EE / Cloud edition path.

Fixes #15324

### DIFF
--- a/packages/web/src/components/custom/ap-error-dialog/ap-error-dialog.tsx
+++ b/packages/web/src/components/custom/ap-error-dialog/ap-error-dialog.tsx
@@ -1,4 +1,4 @@
-import { isNil } from '@activepieces/core-utils';
+import { isNil, tryParseFriendlyPieceError } from '@activepieces/core-utils';
 import { t } from 'i18next';
 import { AlertCircleIcon } from 'lucide-react';
 
@@ -21,6 +21,8 @@ const ApErrorDialog = () => {
 
   if (isNil(params)) return null;
 
+  const friendlyError = tryParseFriendlyPieceError(params.error);
+
   return (
     <Dialog open={!!params} onOpenChange={closeDialog}>
       <DialogContent>
@@ -41,6 +43,11 @@ const ApErrorDialog = () => {
                   {params.description}
                 </DialogDescription>
               )}
+              {friendlyError && (
+                <p className="text-sm text-foreground">
+                  {friendlyError.message}
+                </p>
+              )}
             </div>
           </div>
         </DialogHeader>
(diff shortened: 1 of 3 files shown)


jabrailkhalil:
Per CONTRIBUTING.md, external PRs require the `keep-open` label to be exempt from the close-external-prs workflow. Please add that label if this change is accepted for review.

PR #15360, opened 2026-09-08
feat(slack): send a message to multiple users in parallel

OdaiAhmed99:
## Description

Travelopia DMs each team member from a loop of single **Send Message To A User** steps. The loop runs one iteration at a time, so the Slack part is about 35% of their run time, and they already split the flow with Call Flow to stay under the 10 minute limit. The engine has no parallel loop or branch, so the fix belongs in the Slack piece.

This adds **Send Message To Multiple Users**. One step replaces the whole loop.

### What it does

Two modes, chosen with a **Message Mode** dropdown:

- **Same message for everyone** — pick users from the multi-select, write one message
- **Personal message per user** — one entry per user, each with its own message

DMs are sent in parallel, **5 at a time by default and up to 20** (`Parallel Sends`). Slack's ~1 message per second is *per channel* and every DM is its own channel, so parallel sends across users are the supported shape.

### Output

```json
{
  "sent":   [{ "userId": "U0…", "channel": "D0…", "ts": "1788868050.404719" }],
  "failed": [{ "userId": "U0…", "error": "channel_not_found" }],
  "summary": { "total": 2, "sentCount": 1, "failedCount": 1 }
}
```

A per-user failure lands in `failed[]` and **does not stop the other sends**, so the step succeeds with a mixed result and a downstream Branch can inspect `summary` without a Code step.

The one exception is a connection-level failure: if nothing sent and every error is auth-shaped (`invalid_auth`, `not_authed`, `token_revoked`, `missing_scope`, …), the step **throws**. A revoked token is not a per-recipient outcome, and reporting it as 20 successful failures would hide a misconfiguration the user has to fix.

### Bounded by design

`@slack/web-api` defaults to `tenRetriesInAboutThirtyMinutes` and `timeout: 0` — a ~30 minute retry window with no HTTP timeout, inside a 600s flow budget. Left alone, one recipient stuck behind sustained 429s could consume the whole step and kill it with no output, losing the record of every DM that already succeeded.

So the client gets an explicit bounded `retryConfig` (with `randomize`) and a request timeout, sized so the worst case fits the budget: 33s per recipient × `ceil(80 / 5)` rounds = **528s**, inside 600s. That is also where the 80-recipient cap comes from — concurrency alone does not make a large list safe, because the per-recipient retry budget is unchanged.

No custom retry layer: the SDK already retries 429 and transient 5xx, and stacking another would double-retry and multiply duplicate-delivery risk.

### Notes for reviewers

- `failed[]` means *"no successful response received"*, not *"Slack did not deliver"*. A request can time out after Slack has already sent the message, so this action is at-least-once, not exactly-once — a re-run re-sends to everyone.
- Duplicate user IDs are deduped before sending, last write wins, so `summary.total` is the deduped count.
- Everything detectable is validated **before the first request** — recipient count, user ID shape, message presence, and Slack's 50-blocks-per-message limit — so a bad configuration cannot half-deliver.
- `slackSendMessage` gains an optional `clientOptions` so the bounded retry and timeout can be passed in. It is additive and defaulted, so the other 64 call sites are untouched.
- Message is optional when Block Kit blocks are provided, matching #15195.

Piece version `0.17.10` → `0.17.11`.

## How was this tested?

**CE, local dev stack, against a real Slack workspace.**

**142 unit tests** across 6 files (`vitest run` in the piece), covering:

- plan building for both modes, dedupe and last-write-wins
- the concurrency pool: empty list, limit 1, limit > items, limit = items, default 5, max 20, over-max clamped, and that peak in-flight never exceeds the configured limit — including when every send fails
- 19 Slack error codes: 11 per-user codes land in `failed[]`, 8 auth-shaped codes throw
- timeouts: one timeout → others still complete; all timeouts → still returns a result
- recipient cap: exactly 80 sends, 81 rejected with zero requests issued
- that a bounded `retryConfig` and non-zero `timeout` actually reach the client constructor
- that Username, Profile Picture and Icon Emoji reach every recipient
- that neither the bot token nor the message body appears in `failed[]` or logs

**Live Slack** (real workspace, real DMs):

- single and multiple recipients delivered, real `channel` / `ts` returned
- **failure isolation**: one good ID + one nonexistent ID → good delivered, bad in `failed[]` as `channel_not_found`, step **SUCCEEDED**
- an existing Slack connection created before this change worked unmodified, including one pinned to the previous piece version
- `outputSchema` verified by walking all 11 declared paths against the real output of a mixed success/failure run, so `failed[]` is populated rather than empty

**Playwright** (CE): the new `slack-bulk-dm.spec.ts` passes, and the two pre-existing specs still pass — `3 passed` with `--workers=1`. The spec stays on the builder form and asserts the recipient fields swap when Message Mode changes; it never sends a real DM.

### ⚠️ The speedup number still needs a real workspace

The concurrency is verified, but **only against simulated latency**: 20 recipients at 120ms per DM gives 2495ms serial → 497ms at limit 5 (**5.0×**) → 125ms at limit 20 (**20.0×**), with peak in-flight exactly equal to the limit and ~5% overhead over the theoretical floor.

The **real** loop-versus-parallel comparison is **not yet measured**, because the workspace available for testing has only **3 users**, and dedupe collapses repeats by design so the list cannot be padded. At that size the signal is inside the noise — parallelizing 3 sends saves ~400ms while per-run engine overhead varies by ~2400ms, and the serial and parallel samples overlap completely.

**Please re-run the comparison on a workspace with 10–20+ real users before relying on the speedup figure.** Observed per-send latency there was roughly 90–300ms, which is consistent with the bounded values but is an estimate rather than a p50/p95.

Tracked in Linear: PIE-488


ibrahim-abuznaid:
I reviewed this and also ran it end to end: the unit suite and eslint locally (both green, CI runs neither for pieces), and the action itself through the engine against a real Slack workspace. Same message, personal messages, one bad ID isolated as `channel_not_found`, malformed ID and empty list rejected before any send, blocks-only, and 10 recipients at Parallel Sends 1 / 5 / 20 (step duration 3.46 s → 0.92 s → 0.61 s) all behave exactly as the description says. Nice work on the validation and the bounded client.

One thing blocks it, and it is invisible to every check in the PR:

### Same-message mode cannot be configured in the builder

`userIds` is a `Property.MultiSelectDropdown` with an `options()` function, and it sits inside the `recipients` `DynamicProperties`. The builder cannot render that combination. The Users field shows **`Error: dynamicPropsInfo is required`** instead of a picker, so a user has no way to choose recipients in the default mode.

**How to reproduce**

1. Run the PR locally with `slack` in `AP_DEV_PIECES`, open any flow, add **Slack → Send Message To Multiple Users**, pick a Slack connection.
2. Leave Message Mode on *Same message for everyone*.
3. Look at the **Users** field: it renders the error text above. In DevTools you will see the `/pieces/options` call for `recipients` (type `DYNAMIC`) but never one for `userIds`.
4. Optional, to see the engine side: `POST /api/v1/pieces/options` with `propertyName: "userIds"` or `"recipients.userIds"` returns HTTP 200 with an empty body, and the worker logs `ENTITY_NOT_FOUND` from `getPropOrThrow`.

**Why**

- Web: the children of a DYNAMIC prop are rendered without `dynamicPropsInfo`, and `properties-utils.tsx` falls back to that error div for `DROPDOWN` / `MULTI_SELECT_DROPDOWN` children.
- Engine: `piece-loader.ts` → `getPropOrThrow` does a flat `props[propertyName]` lookup, so a nested `options()` can never be executed.
- The DYNAMIC response does not embed options for a nested dynamic dropdown either.

Static metadata, `tsc`, vitest and the Playwright spec (which asserts labels only) all stay green, which is why this got through.

**Fix**

Same pattern the catalog already uses (baserow, attio, …): fetch inside the `props` function and return a static dropdown with the options embedded. `getUsers` in `common/props.ts` needs an `export`.

```ts
props: async (propsValue): Promise<InputPropertyMap> => {
  const mode = propsValue['mode'];
  if (mode === 'personal_message') { /* unchanged */ }

  const auth = propsValue['auth'] as SlackAuthValue | undefined;
  const users = auth ? await getUsers(getBotToken(auth)) : [];
  return {
    userIds: Property.StaticMultiSelectDropdown({
      displayName: 'Users',
      required: true,
      options: {
        disabled: users.length === 0,
        placeholder: users.length === 0 ? 'Connect Slack first' : 'Select Users',
        options: users,
      },
    }),
    text: Property.LongText({ /* unchanged */ }),
  };
},
```

**How to verify the fix**

Rebuild the piece and restart the dev server (the engine caches a piece by name@version, so a rebuild alone does not reload it). Repeat steps 1–3: the Users field is now a normal multi-select listing every workspace member, mode switching still works, and Test Step delivers to the picked users. I applied exactly this locally and confirmed it. A unit test for the `props` function (mock `getUsers`, assert the returned `userIds.type === 'STATIC_MULTI_SELECT_DROPDOWN'` and that the options are populated) would keep it from regressing.

### The new Playwright spec cannot pass in CI

`tests-e2e-ce.yml` runs with `AP_DEV_PIECES: store,webhook`, so the Slack piece is not loaded there and `getByTestId('Slack')` will time out. On this PR the e2e job was *skipped* (it only runs with the `ready-for-e2e` label), not green. Either add `slack` to that env and `.env.e2e` and prove it with the label, or drop the spec from this PR.

### Smaller notes, not blocking

- The "worst case fits in 600 s" bound holds only at Parallel Sends ≥ 5; at 1 with 80 recipients it is ~45 min. The SDK also sleeps for `Retry-After` on a 429 *before* retrying, outside the request `timeout`. Worth a sentence in the description.
- Timeouts land in `failed[].error` as the opaque `slack_webapi_request_error`; mapping that to `timeout` would help a downstream Branch.
- Bot/app members in the Users list fail with `messages_tab_disabled`, which is a good realistic entry for `failed[]`.

ibrahim-abuznaid (approved):

AhmadTash (approved):

PR #15365, opened 2026-09-08
fix(connections): decode the oauth2 code only when the redirect page posts it encoded

### LINKED ISSUE #14871: [BUG] OAuth2 authorization code URL-decoded twice before token exchange, corrupting codes with percent-encoded characters
## Problem

OAuth2 connect fails at token exchange (surfaces as INVALID_CLAIM) for any provider whose authorization codes contain percent-encoded sequences. A code with a stray `%` hangs the connect popup forever with no error. URL-safe codes are unaffected, which is why mainstream providers work.

Reported via a cloud customer email thread (HTTP (OAuth2) piece against a third-party API). No Pylon ticket.

## Root cause

The code is URL-decoded twice before the server exchanges it:

1. `packages/web/src/app/routes/redirect.tsx:29-30` — `new URLSearchParams(location.search).get('code')` performs the one correct decode, then posts via postMessage.
2. `packages/web/src/features/connections/utils/oauth2-utils.ts:79` — `getCode` applies `decodeURIComponent(event.data.code)` on top.

Backend encodes once, correctly: `credentials-oauth2-service.ts:65`.

## Steps to reproduce

1. Provider issues code `k1%2Fk2` → redirect URL carries `code=k1%252Fk2`
2. Redirect page extracts `k1%2Fk2` (still correct)
3. Popup listener double-decodes → `k1/k2` sent to the token endpoint → provider rejects
4. Stray-`%` variant: code `abc%zzdef` → `decodeURIComponent` throws URIError inside the message listener → promise never resolves → popup hangs

Claim-verify test, 3 passing repro tests on main (rides the fix PR): `packages/web/test/features/connections/utils/oauth2-code-double-decode-claim-verify.test.ts`

## Fix

Remove the `decodeURIComponent` wrapper at `oauth2-utils.ts:79`. No-op for URL-safe codes, so zero risk to currently-working providers. Invert the claim-verify tests to pin the fix.

othmanemad:
## Description

OAuth2 connections whose authorization code contains a literal `%` were corrupted before the token exchange, and a code with an invalid percent escape hung the connect popup forever with no error.

**Root cause.** Two redirect pages are live and they disagree on percent-decoding:

| Redirect page | Decodes? | Used by |
|---|---|---|
| `redirect.tsx` (`URLSearchParams.get`) | once | `OAUTH2`, `PLATFORM_OAUTH2` |
| `secrets.activepieces.com/redirect` (not in this repo) | **zero** | `CLOUD_OAUTH2` |

`getCode` applied one *unconditional* `decodeURIComponent` to both. Managed apps therefore landed on exactly one decode and were correct, while custom and platform OAuth2 double-decoded.

**Fix.** `resolveRedirectContract` pairs each redirect URL with whether that page posts an encoded code, in one place, so the URL and its decoding contract cannot drift apart again. `getCode` decodes only when the contract says to, and falls back to the raw value if `decodeURIComponent` throws rather than dying inside the listener.

**Why not just delete the decode.** That was #14879, reverted 21 hours later in #14926 (no reason recorded). It broke every managed OAuth app, self-hosted as much as Cloud — `CLOUD_OAUTH2` means "managed app", not "on Cloud", and `platform.cloudAuthEnabled` defaults to true everywhere. It also fired far more readily than the bug it fixed: a raw posted code breaks on any `/`, `+` or `=`, so Gmail stopped connecting at once. Fixing it at the secrets service is not possible either — one deployment answers every released version, all of which decode client-side. Reasoning recorded in `brain/knowledge/decisions/000035-…`.

**Known and deliberately out of scope:** `+` becomes a space on `redirect.tsx` (that path also feeds Google SSO login, so it deserves its own PR); `getCode` still never times out on an abandoned popup; the unreferenced `/api/redirect` page.

## How was this tested?

**Automated** — 14 tests in `oauth2-authorization-code-decode.test.ts`, parametrised over *both* sender contracts. #14879 shipped tests that CI really ran and they still missed the regression, because all of them modelled `redirect.tsx` only. Both regressions were simulated to confirm the tests actually catch them:

- delete the decode (what #14879 did) → 3 managed-app rows fail
- restore the unconditional decode (current `main`) → the custom-OAuth2 corruption row fails

Full web suite: 775 passing. `npm run lint-dev`: 0 errors.

**Manual, local EE (`AP_EDITION=ee`)** — against a stub OAuth2 provider issuing the code `k1%2Fk2`:

- **custom `OAUTH2`** via the HTTP (OAuth2) piece — token endpoint received `k1%2Fk2`, byte-identical to what was issued. Before the fix the same stub received `k1/k2`. Popup closed cleanly.
- **managed `CLOUD_OAUTH2`** — a real Google Sheets connect completed through `secrets.activepieces.com/claim` and resolved the account, confirming the path #14879 broke still works.

`PLATFORM_OAUTH2` shares the code path with `OAUTH2` and is covered by unit tests, not exercised manually. Cloud not tested directly; the managed path it uses is the one verified above.

Fixes #14871

No API, env var, or schema change, and no self-hoster action required. Flagging the one judgement call for the reviewer: a provider that *double-encodes* its redirect is currently propped up by the bug and will break. Such a provider violates RFC 6749 and already fails against curl, Postman and standard OAuth libraries; the blast radius is a self-hoster's own custom app, and it fails loudly at connect time. Happy to reclassify as functional-breaking if you disagree.

Touches the OAuth2 connect path, and **tightens an existing weakness**: the sender check was `redirectUrl.startsWith(event.origin)`, a prefix test in the wrong direction, so `"https://secrets.activepieces.com/redirect".startsWith("https://secrets.activepieces.co")` passed. Any page the provider redirects to inherits `window.opener` and can post to the builder, so that guard is the security boundary. It is now an exact origin comparison, with tests. Removing the unconditional decode also removes an unhandled `URIError` inside a message listener.


packages/web/src/features/connections/utils/oauth2-utils.ts
```diff
@@ -15,8 +15,26 @@ import {
...
+function resolveRedirectContract({
+  oauth2Type,
+  platformRedirectUrl,
+}: ResolveRedirectContractParams): RedirectContract {
+  if (oauth2Type === AppConnectionType.CLOUD_OAUTH2) {
+    return {
+      redirectUrl: CLOUD_OAUTH2_REDIRECT_URL,
+      postsPercentEncodedCode: true,
+    };
+  }
```
AbdulTheActivePiecer, line 28: postsPercentEncodedCode: is true only when oauth2Type === AppConnectionType.CLOUD_OAUTH2, then why do you need postsPercentEncodedCode ?

AbdulTheActivePiecer (approved):

PR #15381, opened 2026-09-08
fix(flows): importing a flow inside the builder reliably shows the imported version

### LINKED ISSUE #15378: [BUG] After-import cache purge matches the builder query key only by accident
## Problem

In-builder import bounces `/flows/:id → /flow-import-redirect/:id → /flows/:id` to force a remount and purges the flow query cache on the way. The purge filter matches the builder page's query key only by accident:

* filter (`after-import-flow-redirect.tsx`): `['flow', flowId, undefined]`
* page key 0.84.0–0.90.3: `['flow', flowId, projectId]` → index-2 mismatch → `removeQueries` no-ops → builder remounts from stale cache → import persists server-side, canvas keeps the pre-import flow. Reproduced live on the 0.86.3 image.
* page key 0.90.4+ (#15142 added `versionId`): `['flow', flowId, undefined, projectId]` → prefix matches again. Accidental fix.

Already regressed exactly this way once: #13178 fixed the symptom, #13086 broke the filter 6 days later inside the same 0.84.0 release. Nothing stops the next query-key reshape from silently re-breaking import refresh.

Ref: Pylon 5914

## Root cause

`packages/web/src/app/guards/after-import-flow-redirect.tsx:13` hardcodes the 3-element key shape; coupling to the page key in `packages/web/src/app/routes/flows/id/index.tsx:26` is implicit and unenforced.

## Fix

Purge by 2-element prefix:

```ts
queryClient.removeQueries({ queryKey: ['flow', flowId] })
```

Optionally derive the builder page's own query key from `flowHooks.createFlowQueryKeys` so redirect and page cannot drift.

majewskibartosz:
Fixes [GIT-1863](https://linear.app/activepieces/issue/GIT-1863)
Closes #15378

## Problem

1. In-builder import bounces `/flows/:id → /flow-import-redirect/:id → /flows/:id` and purges the flow's query cache in the redirect so the builder remounts fresh. The purge filter `['flow', flowId, undefined]` is coupled to the builder page's query key shape and matches it only by accident.
2. On 0.84.0–0.90.3 (page key `['flow', flowId, projectId]`) the filter matched nothing: the import persisted server-side while the canvas kept rendering the pre-import flow from the surviving cache entry ([Pylon 5914](https://app.usepylon.com/issues?issueNumber=5914)). #13086 broke #13178's fix exactly this way inside the same 0.84.0 release; #15142 re-fixed it by accident in 0.90.4.
3. On current main the filter matches the page key again, but still misses version-pinned entries (`['flow', flowId, '<versionId>']`), and the next key reshape silently re-breaks the purge with no test failing.

## Fix

- `packages/web/src/app/guards/after-import-flow-redirect.tsx`: purge by the 2-element prefix `['flow', flowId]`. Positions 0–1 are the only part of the key that has never changed (position 2 has been reshaped twice); react-query partial matching then removes every cached query for the flow regardless of future key shapes. Early-return when `flowId` is missing instead of navigating to `/flows/undefined`.
- New regression test renders the guard and asserts both real key shapes are purged while other flows' caches survive.

## Verification

- Regression test red-first: against the pre-fix filter, the version-pinned entry `['flow', id, 'version_1']` survives (test fails); with the fix both shapes purge (2/2 pass).
- `cd packages/web && npm test`: 763/763 pass. `npm run lint-dev`: 0 errors.
- Live drive on this branch (api + web): imported twice in one builder session, with and without a pending draft edit — the canvas refreshed to the imported flow both times, matching `GET /v1/flows/:id`.
- Repro: stale canvas reproduced live on the official 0.86.3 image (import persists server-side, canvas keeps the pre-import flow; a rename made seconds earlier even reverts on screen) — full steps in the Reproduction block.
- Visual: driven states — builder canvas before import (old draft step), import dialog, canvas after two consecutive imports (refreshed); no styling change.
- Other affected paths: checked — the only other `removeQueries` in web (`delete-agent-dialog.tsx`) filters by the exact key it sets; `['flow-versions', flowId]` is a different key root, intentionally untouched (refetches on mount, staleTime 0).

<details>

## Plan
- One-line hardening at the guard: purge by the stable 2-element prefix instead of a helper-built exact key. Footprint estimate: 1 file, ~4 lines; actual: 1 file −7/+4, plus the test.
- Key-history check: positions 0–1 of the page key never changed across the file's history; position 2 was reshaped twice (`498137420c` added projectId, #15142 inserted versionId). The literal prefix sits on the invariant.

- Missing `flowId` → early return (route cannot match without the param; previously navigated to `/flows/undefined`).
- Over-purging → element-wise matching cannot cross key roots: `['flow-versions', …]`, `['flows', …]`, `['flow-runs', …]` are distinct roots; the test asserts another flow's cache survives.
- Version-pinned cache entries (`?versionId=` views) → now purged too, which is the guard's intent: forget everything about this flow after an import.

## Alternatives rejected
- Reverting to `flowHooks.createFlowQueryKeys` → that exact coupling is the bug's origin (#13086 broke #13178 with it).
- A shared prefix helper / query-key factory → indirection over a 2-element invariant with a single consumer; deriving a prefix from the 3-element helper means `.slice(0, 2)`, strictly worse than the literal.
- Rewiring the builder page's own key through the helper → changes cache identity for three other consumers; out of scope.
- Also purging `['flow-versions', flowId]` → different key root, the list refetches on mount anyway; left out to keep the change scoped.
</details>

<details>

## Environment
- Official image `activepieces/activepieces:0.86.3` (WORKER_AND_APP) against a fresh Postgres database and a spare Redis DB index; default CE edition. No pieces and no publish needed: both flows are EMPTY-trigger + one CODE step drafts.
- Control: this branch's base (current main) with identical steps.

## Harness
- Public gist: https://gist.github.com/majewskibartosz/de6beb04a80a3dd426f779000c32cf6c — `scenario.mjs` (creates a TARGET flow with step `old_draft_marker` and a SOURCE flow with `imported_marker`, exports SOURCE as the import template, dumps server-side flow state), `repro5914-template.json`, README with exact env vars.
- Run: `git clone https://gist.github.com/de6beb04a80a3dd426f779000c32cf6c.git git-1863-repro && cd git-1863-repro` then follow README (`node scenario.mjs setup` / `node scenario.mjs state <flowId>`).

## Trigger
Open TARGET in the builder, optionally rename the step (any draft edit), flow-name menu → Import → select the template JSON → Import. The dialog POSTs `IMPORT_FLOW` and navigates through `/flow-import-redirect/:flowId`; on affected versions the purge filter mismatch leaves the page's cache entry alive and the remounted builder renders it.

## Results
| Version | Draft edit first | Server after import | Canvas after import |
|---|---|---|---|
| 0.86.3 | yes | `imported_marker` | stale: `old_draft_marker` (fresh rename reverts on screen) |
| 0.86.3 | no | `imported_marker_2` | stale: `imported_marker` |
| this branch | yes | `imported_marker` | refreshed |
| this branch | no | `imported_marker_2` | refreshed |

The matrix rules out an API failure (server state is always correct), pins the staleness to the client cache purge, and shows draft edits are irrelevant to the mechanism — they only make the stale canvas obvious.
</details>

### DIFF
--- a/packages/web/src/app/guards/after-import-flow-redirect.tsx
+++ b/packages/web/src/app/guards/after-import-flow-redirect.tsx
@@ -2,22 +2,18 @@ import { useQueryClient } from '@tanstack/react-query';
 import { useEffect } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
 
-import { flowHooks } from '@/features/flows';
-
 export const AfterImportFlowRedirect = () => {
   const { flowId } = useParams();
   const navigate = useNavigate();
   const queryClient = useQueryClient();
   useEffect(() => {
-    if (flowId) {
-      queryClient.removeQueries({
-        queryKey: flowHooks.createFlowQueryKeys({
-          flowId,
-          versionId: undefined,
-        }),
-      });
+    if (!flowId) {
+      return;
     }
+    queryClient.removeQueries({
+      queryKey: ['flow', flowId],
+    });
     navigate(`/flows/${flowId}`, { replace: true });
   }, []);
-  return <></>;
+  return null;
 };
--- a/packages/web/test/app/guards/after-import-flow-redirect.test.tsx
+++ b/packages/web/test/app/guards/after-import-flow-redirect.test.tsx
@@ -1,0 +1,57 @@
+// @vitest-environment jsdom
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { render } from '@testing-library/react';
+import { MemoryRouter, Route, Routes } from 'react-router-dom';
+import { describe, expect, it } from 'vitest';
+
+import { AfterImportFlowRedirect } from '@/app/guards/after-import-flow-redirect';
+
+const FLOW_ID = 'flow_1';
+const OTHER_FLOW_ID = 'flow_2';
+const builderPageKey = (flowId: string) => [
+  'flow',
+  flowId,
+  undefined,
+  'project_1',
+];
+const versionPinnedKey = ['flow', FLOW_ID, 'version_1'];
+
+const renderRedirect = (queryClient: QueryClient) =>
+  render(
+    <QueryClientProvider client={queryClient}>
+      <MemoryRouter initialEntries={[`/flow-import-redirect/${FLOW_ID}`]}>
+        <Routes>
+          <Route
+            path="/flow-import-redirect/:flowId"
+            element={<AfterImportFlowRedirect />}
+          />
+          <Route path="/flows/:flowId" element={<div>builder</div>} />
+        </Routes>
+      </MemoryRouter>
+    </QueryClientProvider>,
+  );
+
+describe('AfterImportFlowRedirect', () => {
+  it('purges every cached query of the imported flow whatever its key shape, and keeps other flows', () => {
+    const queryClient = new QueryClient();
+    queryClient.setQueryData(builderPageKey(FLOW_ID), { id: FLOW_ID });
+    queryClient.setQueryData(versionPinnedKey, { id: FLOW_ID });
+    queryClient.setQueryData(builderPageKey(OTHER_FLOW_ID), {
+      id: OTHER_FLOW_ID,
+    });
+
+    renderRedirect(queryClient);
+
+    expect(queryClient.getQueryData(builderPageKey(FLOW_ID))).toBeUndefined();
+    expect(queryClient.getQueryData(versionPinnedKey)).toBeUndefined();
+    expect(queryClient.getQueryData(builderPageKey(OTHER_FLOW_ID))).toEqual({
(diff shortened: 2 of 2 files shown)


majewskibartosz:
## Visual verification

Driven as a real user on this branch (api + web, real Postgres): opened the target flow in the builder, ran the in-builder import twice in one session (first with a pending draft edit, then without), verified the canvas refreshed to the imported flow both times and matched `GET /v1/flows/:id`.

| Before import (old draft step) | Import dialog | After import (canvas refreshed) |
|---|---|---|
| before | dialog | after |

Driven states: builder canvas pre-import, flow-name menu → Import dialog with file loaded, canvas after import 1 (marker step replaced), canvas after consecutive import 2 (replaced again). No styling change in this diff — evidence pass only.

PR #15390, opened 2026-09-09
feat(flow-helper): add Create Waitpoint + Wait for Resume actions

AdamSelene:
### What

Two new actions in the **Flow Helper** piece:

- **Create Waitpoint** — creates a resumable waitpoint and returns `{ waitpointId, resumeUrl }` immediately, **without pausing** the flow.
- **Wait for Resume** — takes a `waitpointId` (from Create Waitpoint), **pauses** the run on that waitpoint, and returns the **full** resume payload (`{ payload, queryParams }`) when its resume URL is called.

### Why

There's currently no way to pause a flow on a control **you post yourself** and read back who acted / what they sent:

- `approval/create_approval_links` hands out a `/confirm` link, and `approval/wait_for_approval` discards the payload (returns only `{ approved }` from the query string), and mints its own waitpoint — so the URL you hand out belongs to a *different* waitpoint than the one it parks on.
- `slack/request_action_message` returns the full interaction payload but insists on posting its own message.

This pair returns the **bare resume URL** and the **full payload**, and `Wait for Resume` parks on the **same** waitpoint whose URL you embedded. That lets you, for example, add a button as an accessory on a message you already manage (value = the resume URL), then read the interaction body (incl. the acting user) when it's clicked.

### Notes

- One waitpoint per pause, addressed by id: `Wait for Resume` calls `waitForWaitpoint(waitpointId)` without minting its own, so the pending waitpoint from `Create Waitpoint` is the one that resumes it. (`createForPause` is only invoked by the waitpoint controller, never by the engine on pause.)
- Validated + single-use: on resume, `handleResumeSignal` looks up `{ id, flowRunId }` (unknown/stale ids are ignored) and deletes the waitpoint, so a resume URL is one-shot.
- `minimumSupportedRelease` set to `0.82.0` to match the sibling pieces that use the same `createWaitpoint`/`waitForWaitpoint` hooks (`approval`, `delay`, `webhook`, `subflows`).
- Flow Helper version bumped 0.1.8 → 0.1.9. Follows the existing action conventions (`audience`, `aiMetadata`, `classification`, `errorHandlingOptions`, output schemas).

### DIFF
--- a/packages/pieces/community/flow-helper/package.json
+++ b/packages/pieces/community/flow-helper/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-flow-helper",
-  "version": "0.1.8",
+  "version": "0.1.9",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "scripts": {
--- a/packages/pieces/community/flow-helper/src/index.ts
+++ b/packages/pieces/community/flow-helper/src/index.ts
@@ -8,7 +8,7 @@ import { createWaitpoint } from "./lib/actions/create-waitpoint";
 export const flowHelper = createPiece({
   displayName: "Flow Helper",
   auth: PieceAuth.None(),
-  minimumSupportedRelease: '0.36.1',
+  minimumSupportedRelease: '0.82.0',
   logoUrl: "https://cdn.activepieces.com/pieces/flow-helper.svg",
   authors: ["AbdulTheActivePiecer","AnkitSharmaOnGithub"],
   actions: [getRunId, failFlow, stopFlow, createWaitpoint, waitForResume],
--- a/packages/pieces/community/flow-helper/src/index.ts
+++ b/packages/pieces/community/flow-helper/src/index.ts
@@ -2,6 +2,8 @@ import { createPiece, PieceAuth } from "@activepieces/pieces-framework";
 import { getRunId } from "./lib/actions/get-run-id";
 import { failFlow } from "./lib/actions/fail-flow";
 import { stopFlow } from "./lib/actions/stop-flow";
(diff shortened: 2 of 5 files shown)


AdamSelene:
Thanks — good catch on the stranded-run risk. Pushed `c31e49de`: **Wait for Resume** now throws at BEGIN if `waitpointId` is blank, instead of parking the run un-resumably. That covers the realistic misconfiguration (Create Waitpoint not wired into the step).

On rejecting *consumed / foreign* ids specifically: a piece can't do that here, and it's a framework limit rather than an oversight —

- `waitForWaitpoint(waitpointId)` **ignores its argument by design**: `piece-executor.ts` (`createWaitForWaitpointHook`) just sets `hookResponse.type = 'paused'`. It never looks the id up.
- Pausing is **run-level**: a run resumes when *any* of its live waitpoints is completed, regardless of which step is parked. So the run is resumable as long as Create Waitpoint minted a still-pending waitpoint on it — which is exactly what this pair does.
- The piece has no way to check a waitpoint's existence/state: there's no `getWaitpoint` hook on `ctx.run`, and `waitpointClient` needs the `engineToken` / `internalApiUrl` that only the engine handler holds, not the piece.
- The **validation and single-use** guarantee Greptile is after does happen — at resume time, on the Create Waitpoint waitpoint: `handleResumeSignal` looks up `{ id, flowRunId }`, ignores unknown/stale ids, and deletes the waitpoint (one-shot). So a stale or foreign *resume URL* can't resume the run; a stale id passed to the *pause* step is the only residual gap, and it's inherent to the id-agnostic `waitForWaitpoint` hook.

Happy to add a `getWaitpoint`/validate hook to the framework in a follow-up if you'd prefer real pre-pause validation — it'd need an engine-side endpoint the piece context can reach.

PR #15392, opened 2026-09-09
feat(notion): readable friendly-view output for page & row triggers

ibrahim-abuznaid:
## Problem

The **Updated Database Item**, **New Database Item**, and **Updated Page** triggers produce output that is hard to read in the Friendly View / data selector:

- **No title.** A Notion row's title lives at `properties.<AnyColumnName>.title[0].plain_text`, where the column can be named anything and sit at any position. `OutputSchemaField.value` is a static dot-path with no "the property whose type is `title`" selector, so no static path can reach it. The schema therefore had no title field at all.
- **Every value is 3–4 clicks deep.** `Properties` was the last of 12 fields, and each Notion property is a richly-typed object, so reading one value took `Properties ▸ Name ▸ title ▸ Item 1 ▸ plain_text`.
- **Hardcoded template columns.** Two triggers hardcoded one template database's column names (`Task name`, `Status.status.name`, `Due date.date.start`, `Assignee.people[0].id`), which resolve **empty** on every other database.

## Fix (piece-level, additive)

New `common/enrich-page.ts`. `enrichPage()` adds two computed top-level fields to each emitted page/row, so a static schema can point at them and the payload reads like Airtable's flat `fields` map:

- **`title`** — plain text of the title property, found by `type === 'title'` (any column name/position), all segments joined.
- **`property_values`** — `{ columnName: scalar }` for every property, flattened by type. All 24 Notion property types are handled; an unknown type maps to `null` and never throws, so a future Notion type can't break a live flow.

Applied in the three triggers. All three are now backed by one shared, content-first `notionPageFields` schema (Title, Properties, then metadata, **Raw Properties last**). The raw `properties` object is kept in the payload and schema, so existing flows keep working.

Also in this PR:
- Deleted the hardcoded `Task name`/`Status`/`Due date`/`Assignee` template paths from the three triggers and from the `create_database_item` / `update_database_item` / `notion-find-database-item` action schemas (replaced with a generic raw-properties passthrough).
- Refreshed the three triggers' stale `sampleData` from real payloads (now includes `is_archived`/`is_locked`/`in_trash` plus the two new fields).
- Added a 49-case Vitest covering all 24 property types and real-data edge cases (empty title, empty rich_text, mixed partial/full people, bot user, title-not-first, spaces in column names, forward-compat unknown type).

## Defaults chosen — please confirm in review

- **Empty title → `""`** (viewer shows empty; Notion itself displays "Untitled").
- **`relation` → count** (raw ids stay in `properties`).
- **`files` → names only** (internal Notion file URLs expire in ~1h).
- **`people` / `created_by` / `last_edited_by` → `name ?? id`** (partial id-only users fall back to id).
- **Flatten key is `property_values`** (deliberately not `fields`/`value`/`properties`/`data`, which the viewer treats as drill-down wrappers).
- **`icon` is not flattened** — kept strictly to `title` + `property_values`; raw `icon` stays in the payload.

## Scope / notes

- The pre-existing `notionPageFields` (raw page shape used by the `notion-*` AI-action variants) was renamed **`notionRawPageFields`** — pure rename, no behavior change — so the shared enriched schema could take the `notionPageFields` name. Those AI actions are untouched here.
- **Follow-up (separate ticket):** apply the same enrichment to the database-item **actions** and replace the runtime title guesses (`Object.values(properties)[0]` / hardcoded `Name`) with the new `getPageTitle` helper.
- `Notion-Version` stays pinned at `2022-02-22`; the flattener tolerates `data_source_id`/`custom_emoji` so a future bump can't break it.
- Additive and non-breaking → minor bump `0.6.10 → 0.7.0`.

## Verification

- `tsc` build + `eslint` clean (0 errors; only pre-existing `no-explicit-any` warnings, none in the new file).
- **Vitest: 49/49.**
- Trigger output-schema paths all resolve against the refreshed `sampleData` (0 unresolved).
- Tested live on two real databases: enriched `title` + `property_values` on every row, including an empty-title row → `""`, id-only participants → id, and a bot `created_by` → name.

### Breaking change?

### Security impact?

### DIFF
--- a/packages/pieces/community/notion/package.json
+++ b/packages/pieces/community/notion/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-notion",
-  "version": "0.6.10",
+  "version": "0.7.0",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
   "dependencies": {
@@ -22,6 +22,7 @@
   "scripts": {
     "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
     "bundle": "node ../../../../dist/packages/cli/src/index.js pieces bundle",
-    "lint": "eslint 'src/**/*.ts'"
+    "lint": "eslint 'src/**/*.ts'",
+    "test": "vitest run"
   }
 }
--- a/packages/pieces/community/notion/src/lib/common/enrich-page.test.ts
+++ b/packages/pieces/community/notion/src/lib/common/enrich-page.test.ts
@@ -1,0 +1,466 @@
+/// <reference types="vitest/globals" />
+import {
+  richTextToPlain,
+  getPageTitle,
+  getDatabaseTitle,
+  flattenPropertyValue,
+  flattenProperties,
+  enrichPage,
+} from './enrich-page';
+
+// A rich-text segment as Notion serves it.
+const rt = (text: string) => ({
+  type: 'text',
+  text: { content: text, link: null },
+  plain_text: text,
+  href: null,
+  annotations: {},
+});
+const prop = (type: string, extra: Record<string, unknown>) => ({
+  id: 'x',
+  type,
+  ...extra,
+});
+
+describe('richTextToPlain', () => {
+  it('joins all segments and trims', () => {
+    expect(richTextToPlain([rt('Take '), rt('Fig '), rt('on a walk')])).toBe(
+      'Take Fig on a walk'
+    );
+  });
+  it('handles a mention segment (still carries plain_text)', () => {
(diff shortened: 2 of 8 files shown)


ibrahim-abuznaid:
Closing for now — will re-open from a fresh branch after confirming the flagged defaults (empty-title, icon). The change is complete and verified locally; the failing `main` check was an unrelated flaky timeout in packages/server/utils (large-result-utils), not this PR.

ibrahim-abuznaid:
Re-opened as #15394 from a fresh branch off main, with the three Greptile findings addressed (relation `has_more` → `"N+"`, `parent.database_id` restored in the trigger schema, Find Database Item results expose a Raw Properties map again).

PR #15406, opened 2026-09-09
feat(google-calendar): improve the step forms

Talaljaber:
## Description

Step-form pass on the Google Calendar piece: the nine human-facing actions, the seven triggers, the shared dropdowns and the service-account auth guide. The forms were built up over several years by different hands, so the same field carried different labels and option names from one form to the next, ten descriptions folded behind "show more", and everything sat in one flat list. Four runtime bugs found while reading the forms are fixed in the same pass.

### Labels

Every label is now a Title Case noun. "Title of the event", "Start date time of the event" and "End date time of the event" become **Title**, **Start Time** and **End Time** on Create and Update Event; "Date from" / "Date to" on Get Events become **Start Time** / **End Time** to match Find Busy/Free Periods; the three guest-permission fields become **Guests Can Modify**, **Guests Can Invite Others** and **Guests Can See Other Guests**. "Expand Recurring Event?" loses the question mark and the misleading required asterisk on both Get Events and the New or Updated Event trigger. "Event types" / "Event Types to Monitor" / "Event Types" collapse to **Event Types**, with the same option names ("Default", "Birthday", "Focus Time", "Out of Office", "Working Location", "From Gmail") on every form that offers them.

### Descriptions

The builder folds anything over 70 characters behind "show more"; sixteen descriptions crossed that line, the longest at 126 characters on Get Event by ID. All of them now fit on one line and say what the field does rather than restating its name. Every bare **Event ID** field (Update, Delete, Add Attendees, Get Event by ID) gets the same hint, "Paste the ID from the event URL or a previous step.", because the field had no description at all and the console showed users hitting it blind. The Event ID stays a text field: it is almost always mapped from a trigger or a Get Events step, the triggers' upcoming-events dropdown would not fit that use, and a type change here would be a migration.

Three descriptions say something the old copy hid:

- **Attendees** on Update Event reads "Replaces the current guest list. One email per item." The run only falls back to the existing guests when the array is empty, so three emails on a ten-guest event uninvites seven. Add Attendees to Event, which appends, sits next to it with otherwise near-identical copy.
- **Search Term** on Get Events reads "Exact phrase matched against title, description, location and guests." The run wraps the term in quotes before sending it as `q`, so "budget review" does not match "Q3 budget and review". The New Event Matching Search trigger sends an unquoted `q` and keeps the plain "Matched against…" wording.
- **Max Attendees** on Get Event by ID reads "Above this count, only you are returned." Google's `events.get` returns the requesting participant, not the organizer, when the count is exceeded.

### One wording per concept

- Create Event and Create Quick Event asked the same question with two option sets ("Yes, to everyone" / "To non-Google Calendar guests only" / "To no one" versus "All" / "External Only" / "none"). Both now read **All guests** / **External guests only** / **No one** under the description "Who gets an email invitation for the new event." Create Event pre-fills **All guests** as before; Create Quick Event pre-fills **No one**, which keeps its previous behaviour of sending no notifications when the field is untouched. Values are unchanged.
- **Color** is a shared dropdown; its description is now passed per action. Create Event says "Leave empty for the calendar default."; Update Event leaves it blank because there an empty field keeps the current color, as the info block above it says.
- Every dropdown placeholder now reads "Connect your Google account first" or "Select a calendar first"; there were three variants before.
- Create Quick Event's "Summary" field becomes **Event Text** with the example `"Lunch with Sam tomorrow at 1pm"`, since the field takes a phrase for Google to parse, not a summary.

### Advanced sections

Fields that most flows never touch move behind the Advanced fold: the three guest-permission toggles on Create Event; Max Attendees and Time Zone on Get Event by ID; and the filter props on the New Event, New Event Matching Search and Event Cancelled triggers. This is what raises `minimumSupportedRelease` to 0.88.2.

**Event Types** on Get Events stays required and in the main body. It was folded and made optional in an earlier revision, but the AI-only List Events action reuses these props verbatim and the MCP path never applies a multi-select default, so an LLM that omitted the field would have started receiving birthdays, working-location and from-Gmail entries the required field always excluded. A default that restricts results is also a poor fit for Advanced.

### Update Event

An info block under Event ID now says "Empty fields keep their current value."

That sentence was not true before this PR. `events.update` is a full PUT and the request body was hand-built from the nine form fields, so every update cleared `recurrence`, `reminders`, `transparency`, `visibility`, `extendedProperties`, `attachments` and `conferenceData`: renaming a weekly standup collapsed it to a single event. The run now spreads the fetched event into the body and overlays the fields the user filled in, which also covers `colorId`, the one form field that was previously sent through unguarded. A reschedule also carries the event's `timeZone` across, which Google requires on recurring events once `recurrence` is preserved. The AI-only Update Event action shares this run and its `aiMetadata` already promised the merge; it now happens.

The three guest-permission toggles used to be checkboxes, and the builder stores `false` for a checkbox the user never touched, so the run could not tell "untouched" from "switched off" and always wrote all three. They are now optional dropdowns (**Yes** / **No**) whose empty state reads "Keep current setting"; the run applies a value only when one was chosen. Steps saved with the old checkboxes hold `true` or `false`, which match the new option values, so they load and behave exactly as before. The toggles stay in the main body of Update Event rather than moving to Advanced. The AI-only variant already used tri-state dropdowns and does not inherit the info block.

### Auth

The service-account guide was one paragraph with `<br>` tags, two bare URLs and a bold "Note:". It is now two short markdown lines with real links to the Service Accounts console and the domain-wide delegation article. The JSON key field becomes a multi-line text box with "Paste the whole JSON key file.", matching the Gmail and Google Drive pieces; the stored connection value is a string either way, so existing connections are unaffected.

### Runtime fixes

- **Update Event** dropped every event field the form does not expose on each update (see above).
- **Update Event** cleared the event color when Color was left empty.
- **Update Event** reset the three guest permissions on every run, because an untouched checkbox is stored as `false` (see above).
- **Update Event** dropped the event's `timeZone` on a reschedule. Google requires it on recurring events, so with `recurrence` now preserved the reschedule would have failed; the new time carries the current `timeZone` across.
- **New Event Matching Search** joined several event types into one query value (`default&eventTypes=birthday`), which the HTTP client then percent-encoded into a single bogus type. The trigger now appends one `eventTypes` parameter per type, the way Get Events already does.
- **New Event** test failed on an empty calendar: `getLatestEvent` did `eventList.pop()!` and the caller dereferenced `undefined`. It now returns `null` for an empty list and the test returns an empty sample. (New or Updated Event polls `getEvents` directly and never calls `getLatestEvent`.)

### Metadata

- Version 0.10.3 → 0.12.0 (the guest-permission props change type; 0.11.0 was taken by the catalog-wide bump in #15398, published 2026-09-14); `minimumSupportedRelease` 0.86.4 → 0.88.2 (needed for `advanced`).
- `src/i18n/translation.json` regenerated from the built piece. No locale files were hand-edited; rewritten strings fall back to English until Crowdin catches up.
- `aiMetadata` on every action is untouched.
- `test/update-event.test.ts` (11 cases, `vitest`) covers the Update Event request body; the piece's first unit tests.

### Deliberately left out

- Get Events silently ignores End Time unless Start Time is also set. The description now says so ("Requires Start Time.") rather than changing the query, which is a behaviour change for a UI pass.
- Get Events keeps quoting the search term. Dropping the quotes would change which events existing flows match; the description now says what the query does instead.
- The Custom API Call form comes from `pieces-common` and is not touched here.
- The 16 AI-only actions are hidden from the builder and keep their current copy.

No prop names change. The three guest-permission props change from checkbox to dropdown, but their stored `true` / `false` values are valid options of the new control, so steps saved on 0.10.3 keep working unchanged.

## How was this tested?

- Served the piece from `AP_DEV_PIECES` on a local CE build and opened a fresh step for every human-facing action and trigger at default panel width; checked labels, descriptions, hint blocks, Advanced folds and dropdown placeholders.
- Ran the step-form checker on the served metadata before and after: 16 errors / 22 warnings → 0 errors / 2 warnings. The two remaining warnings are the shared Custom API Call form and the intentional "Event ID" text field versus the triggers' "Event" dropdown. The `--before` comparison against the 0.10.3 metadata reports no prop type changed or removed.
- `tsc --noEmit` on the piece passes; `eslint` reports 0 errors (19 pre-existing warnings in untouched run code).
- Edition paths: CE checked locally. The change is piece metadata and three run/trigger code paths with no edition-specific code, so EE and Cloud render the same forms.


ibrahim-abuznaid (commented):
Reviewed the full diff plus the runtime paths it touches. The form work is careful and the description is unusually honest — documenting the guest-toggle limitation instead of hiding it is the right call. Most of what follows is copy that promises something the code doesn't do yet.

Three blockers and five smaller fixes, all inline.

**One correction to the description, not the code.** The `getLatestEvent` null fix is attributed to **New or Updated Event**, but that trigger (`calendar-event.ts`, `new_or_updated_event`) polls `getEvents` directly and never calls `getLatestEvent`. The only caller is **New Event** (`new-event.ts:274`). Worth correcting so the next reviewer tests the path that actually changed against an empty calendar.

**Verified and cleared**, so nobody re-checks them:
- The repeated-`eventTypes` URL fix is correct. `BaseHttpClient.getUrl()` re-parses the URL's query string into `URLSearchParams` and `FetchHttpClient` appends `request.queryParams` on top, so nothing is dropped and there's no double-encoding.
- `minimumSupportedRelease: 0.88.2` is right — `advanced` landed in #14501 and 0.88.2 is the first tag containing that commit. (Gmail, Telegram and QuickBooks Desktop Conductor are the ones under-declared at 0.87.0 while using `advanced`; I'll file that separately.)
- MCP tool names derive from `actionName`, not `displayName`, so `Get all Events` → `Get Events` breaks no agent references.
- No prop keys or types changed, so steps saved on 0.10.3 are safe.
- `getLatestEvent` has exactly one caller and it's guarded with `isNil`.

**Two things outside this PR.** The renamed labels orphan their keys across all 12 locale files, and `generate-translations.yml` has failed every day since 2026-08-16 — so "falls back to English until Crowdin catches up" currently has no end date. Something to weigh before the next piece in this series. Also, this PR has no labels; the root CLAUDE.md wants one of 🌟 feature / 🐛 bug / skip-changelog plus 🧩 area/third-party-pieces.

packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts
```diff
@@ -8,18 +8,23 @@ export const updateEventProps = {
   calendar_id: googleCalendarCommon.calendarDropdown('writer'),
   eventId: Property.ShortText({
     displayName: 'Event ID',
+    description: 'Paste the ID from the event URL or a previous step.',
     required: true,
   }),
+  hint: Property.MarkDown({
+    value: 'Empty fields keep their current value. The guest toggles below are always applied.',
```
ibrahim-abuznaid: **Blocker: this sentence isn't true yet.**

`calendar.events.update` is a full PUT, and the `requestBody` below is hand-built — it lists `summary`, `attendees`, `description`, `colorId`, `location`, `start`, `end` and the three `guestsCan*` flags, and nothing else. `recurrence`, `reminders`, `transparency`, `visibility`, `extendedProperties`, `attachments` and `conferenceData` aren't in it, so Google clears them on every update.

Concretely: change only the Title on a weekly standup and the RRULE is dropped — the series collapses to a single event. Custom reminders go the same way.

The `colorId ?? currentEvent.data.colorId` change on line 108 fixes one instance of this class. The hint documents the whole class as handled. Either fix it generally —

```ts
requestBody: { ...currentEvent.data, ...overrides }
```

— or drop the first sentence. Shipping the claim is worse than the vague copy it replaces, because a user who reads it has no reason to go re-check their recurring events.

The underlying bug predates this PR and I'm filing it separately; only the new claim needs to block.

packages/pieces/community/google-calendar/src/lib/common/index.ts
```diff
@@ -70,13 +71,14 @@ export const googleCalendarCommon = {
   colorId: Property.Dropdown({
     auth: googleCalendarAuth,
     displayName: 'Color',
+    description: 'Leave empty for the calendar default.',
```
ibrahim-abuznaid: **Blocker: wrong on Update Event.**

This prop is shared between Create Event and Update Event. "Leave empty for the calendar default" is right for Create, but `update-event.action.ts:108` now does `colorId ?? currentEvent.data.colorId` — on Update, empty keeps the *current* color.

It also contradicts the info block this PR adds three fields above it. Needs neutral wording here, or per-action descriptions.

packages/pieces/community/google-calendar/src/lib/actions/get-events.ts
```diff
@@ -12,18 +12,19 @@ import { getEventsActionOutputSchema } from '../output-schemas';
 export const getEventsProps = {
   calendar_id: googleCalendarCommon.calendarDropdown('writer'),
   event_types: Property.StaticMultiSelectDropdown({
-    displayName: 'Event types',
-    description: 'Select event types',
-    required: true,
+    displayName: 'Event Types',
+    description: 'Only events of these types are returned.',
```
ibrahim-abuznaid, line 16: **Blocker: this changes AI behaviour, not just the form.**

`ai-list-events.ts:19` reuses `getEventsProps` verbatim, and on the MCP path `defaultValue` is never applied — `fillDefaultsForMissingOptionalProps` in `packages/server/api/src/app/mcp/tools/mcp-utils.ts` backfills only `ARRAY`, `DYNAMIC` and `CHECKBOX`, never `STATIC_MULTI_SELECT_DROPDOWN` — and the engine's props-processor applies none either.

So an LLM that omits the field now hits `event_types ?? []`, sends no `eventTypes` param, and gets birthdays, working-location and from-Gmail entries that the required field always excluded. "Runs behave as before" holds in the builder, but not for the AI action sharing these props.

Separately, `advanced: true` on a field whose default *restricts* results is worth a second look — Advanced should be for fields that don't change what comes back.

packages/pieces/community/google-calendar/src/lib/actions/create-quick-event.ts
```diff
@@ -12,35 +12,34 @@ export const createQuickCalendarEvent = createAction({
...
-      description: 'The text describing the event to be created',
+      displayName: 'Event Text',
+      description: 'e.g. "Lunch with Sam tomorrow at 1pm".',
       required: true,
     }),
     send_updates: Property.StaticDropdown<string>({
       displayName: 'Send Updates',
-      description:
-        'Guests who should receive notifications about the creation of the new event.',
+      description: 'Who gets an email invitation for the new event.',
```
ibrahim-abuznaid, line 28: The harmonization is right, but it leaves two identical-looking forms with opposite defaults.

`create-event.ts:64` is `required: true, defaultValue: 'all'` → invitations go to all guests. Here there's no `defaultValue`, and line 56 resolves `propsValue['send_updates'] || 'none'` → an untouched dropdown emails nobody.

Same label, same options, same description, opposite outcome — which is exactly the trap this PR is closing everywhere else. `defaultValue: 'all'` here finishes the job.

packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts
```diff
@@ -28,27 +33,27 @@ export const updateEventProps = {
...
     displayName: 'Description',
-    description: 'Description of the event. You can use HTML tags here.',
+    description: 'HTML tags are allowed.',
     required: false,
   }),
   colorId: googleCalendarCommon.colorId,
   attendees: Property.Array({
     displayName: 'Attendees',
-    description: 'Emails of the attendees (guests)',
+    description: 'One guest email per item.',
```
ibrahim-abuznaid: Worth saying this **replaces** rather than adds. The block at lines 86-93 only falls back to `currentEvent.data.attendees` when the array is empty, so setting Attendees to 3 emails on a 10-guest event silently uninvites 7.

Suggest "Replaces the current guest list. One email per item." — the contrast with **Add Attendees to Event** (which appends) is otherwise invisible, and these two actions now sit next to each other with near-identical copy.

packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts
```diff
@@ -8,18 +8,23 @@ export const updateEventProps = {
   calendar_id: googleCalendarCommon.calendarDropdown('writer'),
   eventId: Property.ShortText({
     displayName: 'Event ID',
+    description: 'Paste the ID from the event URL or a previous step.',
     required: true,
   }),
+  hint: Property.MarkDown({
```
ibrahim-abuznaid, line 14: This prop leaks into the AI action. `ai-update-event.ts` destructures `updateEventProps` and spreads `...otherUpdateProps`, omitting `eventId`, `calendar_id` and the three `guests_can_*` — but not `hint`.

Harmless at runtime (MARKDOWN is in `NON_INPUT_PROP_TYPES`, so it never reaches a tool schema), but the text is the *opposite* of that action's behaviour: it replaces the checkboxes with tri-state dropdowns whose unset value keeps the event's current setting. Add `hint: _omitHint` to the destructure.

packages/pieces/community/google-calendar/src/lib/actions/get-event-by-id.ts
```diff
@@ -13,21 +13,20 @@ export const getEventByIdProps = {
...
-    description:
-      'The unique ID of the event (e.g., "abc123def456"). You can find this in the event URL or from other calendar actions.',
+    description: 'Paste the ID from the event URL or a previous step.',
     required: true,
   }),
   max_attendees: Property.Number({
     displayName: 'Max Attendees',
-    description:
-      'Maximum number of attendees to include in the response. If there are more attendees, only the participant is returned.',
+    description: 'If exceeded, only the organizer is returned.',
```
ibrahim-abuznaid: Google's events.get reference says "If there are more than the specified number of attendees, only **the participant** is returned" — the requesting user, not the organizer.

The old copy was vague but faithful; this makes a specific claim that points the wrong way, and someone will end up building downstream logic on `attendees[0]` being the organizer. Something like "Above this count, only you are returned." keeps the brevity without the inversion.

packages/pieces/community/google-calendar/src/lib/actions/get-events.ts
```diff
@@ -39,21 +40,23 @@ export const getEventsProps = {
   }),
   search: Property.ShortText({
     displayName: 'Search Term',
+    description: 'Matches title, description, location and guests.',
```
ibrahim-abuznaid: `runGetEvents` line 82 does ``queryParams['q'] = `"${search}"` `` — the term is force-quoted into an exact-phrase query, so "budget review" won't match an event titled "Q3 budget and review".

Two problems: this description promises plain matching, and `new-event-matching-search.ts:147` now carries near-identical wording ("Matched against title, description, location and guests") over an *unquoted* `q`. Same sentence, different semantics on two forms a user will compare side by side. Either mention the phrase match here, or drop the quoting in the query.

Talaljaber:
> ### Confidence Score: 3/5
> The PR is not yet safe to merge because two previously reported behavior changes can silently alter guest permissions or notification delivery.
> 
> The Update Event action still always submits the three checkbox defaults, so editing an unrelated field can reset existing guest permissions. Create Quick Event also still defaults Send Updates to `all`, changing an untouched field from the previous no-notification fallback to emailing every guest.
> 
> **Files Needing Attention:** packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts; packages/pieces/community/google-calendar/src/lib/actions/create-quick-event.ts
> 
> ### Important Files Changed
> Filename	Overview
> packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts	Preserves the fetched event during updates and enables conference-data handling, but the previously reported guest-permission reset remains.
> packages/pieces/community/google-calendar/src/lib/actions/create-quick-event.ts	Improves the quick-event form, but its new `all` default still changes untouched notification behavior from no notifications to emailing all guests.
> packages/pieces/community/google-calendar/src/lib/triggers/new-event-matching-search.ts	Serializes selected event types as distinct query parameters and reorganizes optional filters.
> packages/pieces/community/google-calendar/src/lib/triggers/new-event.ts	Handles an empty latest-event result safely and moves optional filters into the advanced section.
> packages/pieces/community/google-calendar/src/lib/common/helper.ts	Returns null rather than an invalid non-null event when a calendar contains no events.
> packages/pieces/community/google-calendar/src/lib/auth.ts	Reworks service-account instructions and changes the JSON key input to a multiline field.
> Reviews (4): Last reviewed commit: ["fix(google-calendar): keep conference da..."](https://github.com/activepieces/activepieces/commit/a0039026004bc247d03a42f8ad65e67e3b7efb2c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62163165)

Both findings are intentional and covered in the description. The Create Quick Event default was requested in review to match Create Event; it only pre-fills new steps, and saved steps and the AI path keep the none fallback. The Update Event guest toggles are a pre-existing limitation: the builder stores false for every untouched checkbox, so no prop default can distinguish untouched from off. The reviewer accepted documenting it here and is filing the tri-state fix separately.

ibrahim-abuznaid:
Heads-up on an overlap. @OdaiAhmed99 independently fixed the same Update Event bug in #15506, working off PIE-510, before I'd connected the two tickets. We're consolidating here rather than landing both, so #15506 is now closed and its content moves onto this branch.

Three things come over, and all three improve on what's on `feat/google-calendar-step-form` today:

1. **`test/update-event.test.ts` + `vitest.config.ts`** — 10 tests, the piece's first suite.
2. **`toEventDateTime`** — this branch still writes `{dateTime}` over an all-day event's `{date}`. Set only Start Time on an all-day event and Google returns 400 on the mixed `date`/`dateTime` pair. Pre-existing rather than something this PR introduced, but the incoming tests cover it.
3. **A conditional-spread `overrides` object** replacing the `x ?? currentEvent.data.x` chain — same behaviour, but "unset contributes no key" becomes explicit.

Keep `conferenceDataVersion: 1` exactly as it is. #15506 doesn't have it and needs it; without it Google ignores `conferenceData` in the body and drops the Meet link.

Nothing else in this PR changes and the version stays at 0.10.4.

Separately: CI never ran against a0039026 — the push landed inside the GitHub Actions outage on 09-13, so four workflows hit `startup_failure` and four more sat queued for 19 hours. The push that brings these commits over will re-trigger everything, so there's no need to restart the dead runs first.

The eight inline comments from my earlier review are all addressed — thanks for the quick turnaround on those, and for correcting the trigger attribution in the description.

ibrahim-abuznaid:
Ported @OdaiAhmed99's #15506 onto this branch as two cherry-picked commits (Odai stays the author): the `overrides` object with `toEventDateTime`, plus `test/update-event.test.ts` (10 tests, the piece's first suite) and `vitest.config.ts`.

One behaviour change against a0039026: a reschedule now carries the event's `timeZone`. Google requires it on recurring events, and since this PR now preserves `recurrence`, a reschedule without it would come back 400. `conferenceDataVersion: 1` stays. Version stays 0.10.4.

Verified locally: the suite passes 10/10 and eslint reports 0 errors. Setting only Start Time on an all-day event still sends a mixed date/dateTime pair on both versions; that is pre-existing and left out of scope here.

packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts
```diff
@@ -28,29 +33,32 @@ export const updateEventProps = {
...
-    description: 'Emails of the attendees (guests)',
+    description: 'One guest email per item.',
     required: false,
   }),
   guests_can_modify: Property.Checkbox({
-    displayName: 'Guests can modify',
+    displayName: 'Guests Can Modify',
     defaultValue: false,
     required: false,
+    advanced: true,
```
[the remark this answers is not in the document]
ibrahim-abuznaid: Intentional and documented in the form hint above these checkboxes. Dropping the `defaultValue: false` would change nothing: the builder saves `property.defaultValue ?? false` for every checkbox (`packages/web/src/features/pieces/utils/form-utils.tsx`), so an untouched step still stores `false` and the run cannot tell it apart from an explicit off. The real fix is a three-state control, which is a prop type change and belongs in its own PR. Resolving so the score reflects what this PR can change.

packages/pieces/community/google-calendar/src/lib/actions/create-quick-event.ts
```diff
@@ -12,35 +12,35 @@ export const createQuickCalendarEvent = createAction({
...
       required: false,
+      defaultValue: 'all',
```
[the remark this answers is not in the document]
ibrahim-abuznaid: Applied in a96fc16b: `defaultValue: 'none'`. New steps now pre-fill **No one**, matching the previous no-notification behaviour. Saved steps and the AI path were already on the `none` fallback.

packages/pieces/community/google-calendar/src/lib/actions/create-quick-event.ts
ibrahim-abuznaid: Reversed in a96fc16b after Greptile pointed out that pre-filling All guests changes what an untouched new Quick Event step does. Quick Event now pre-fills **No one**; Create Event keeps **All guests**. Same labels and options on both forms, so the wording harmonization stands.

packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts
```diff
@@ -28,32 +33,35 @@ export const updateEventProps = {
...
+    options: {
+      disabled: false,
+      placeholder: 'Keep current setting',
+      options: [
+        { label: 'Yes', value: true },
+        { label: 'No', value: false },
+      ],
+    },
+  });
+}
```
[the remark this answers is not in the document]
ibrahim-abuznaid: Moved in dc733439: `guestPermission` now sits below the exports next to `toEventDateTime`. Function declarations are hoisted, so the props object above still resolves it.

packages/pieces/community/google-calendar/src/lib/actions/update-event.action.ts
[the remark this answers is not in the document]
ibrahim-abuznaid: Update: this landed in this PR after all, in 92019153. The three checkboxes are now optional Yes / No dropdowns whose empty state reads "Keep current setting", and the run only writes a value when one was chosen.

ibrahim-abuznaid (approved):

PR #15425, opened 2026-09-10
feat(floqer): add the Floqer piece

### LINKED ISSUE #15364: [Feature]: New piece — Floqer (API key auth, core actions, segment-change trigger)
## Context

New piece request: Floqer ([floqer.com](<http://floqer.com>)) — AI CRM enrichment / GTM platform. Smallest useful v1; add endpoints on concrete asks.

## API (verified from [https://api.floqer.com/openapi.json](<https://api.floqer.com/openapi.json>))

* Base `https://api.floqer.com`, paths under `/api/v1/`
* Auth: bearer token, key format `floq_...` (`securitySchemes.apiKey` = `http`/`bearer`)
* 210 endpoints across User, Workflows, Sheets, Rows, Actions, Sources, Agent, Shortcuts, Campaigns, ackDB
* No webhooks (`webhooks: {}`, no `/webhook*` path). Only trigger surface is the segment change feed below, so the trigger must be polling.

## Scope — v1

Auth: `SecretTextAuth` API key, validated with `GET /api/v1/user/`.

Actions:

* Run Shortcut — `POST /api/v1/shortcuts/{id}/run`, body `{ input_data }`. Shortcut id from a dropdown (`GET /api/v1/shortcuts/`); input props from the shortcut's `input_schema` `reference` keys.
* Add Rows — `POST /api/v1/workflows/{workflow_id}/sheets/{sheet_id}/rows`. Workflow dropdown from `GET /api/v1/workflows/`, sheet dropdown from `GET /api/v1/workflows/{workflow_id}`.
* Run Rows — `POST /api/v1/workflows/{workflow_id}/sheets/{sheet_id}/run`
* Custom API Call

Trigger — Segment Membership Changed (polling, `POST /api/v1/ackdb/observe/segments/{id}/changes`):

* Cursor-based append-only log: store the returned cursor, pass it back as `since`. Re-reading the same `since` is idempotent, so a crash before the store write is safe.
* First call bootstraps and returns no rows. Seed initial members from `GET /api/v1/ackdb/observe/segments/{id}/member-ids` so they are not emitted as new.
* `410 CURSOR_EXPIRED` or a `200` with `reconcile: true` both mean re-seed from `/member-ids`, not empty-success. `409` means a concurrent poll holds the advisory lock: skip that tick.

## Out of scope

Agent, Campaigns, Sources, the rest of ackDB, and all workflow / sheet / action authoring endpoints.

OdaiAhmed99:
## Description

Adds **Floqer** as a community piece — API-key auth and four actions.

| Action | |
|---|---|
| **Run Shortcut** | Queues a published Floqer shortcut. Its inputs are built dynamically from that shortcut's own `input_schema`. |
| **Add Rows** | Writes rows into a workflow sheet, optionally queueing them to run. |
| **Run Rows** | Runs rows that already exist in a sheet. |
| **Custom API Call** | Any Floqer endpoint, authenticated. |

Dropdowns for Workflow, Sheet (gated on Workflow through its refresher) and Shortcut (published only).

**Add Rows returns on partial success rather than throwing.** Floqer still writes the valid rows in a
batch when others are rejected, so throwing would discard the `row_ids` that actually landed. `rejected`
and `warnings` are first-class in the output schema — a flow that cares branches on `rejected.length`.

**No trigger in v1.** A Segment Membership Changed trigger was built and then deliberately pulled. Floqer
answers `/changes` with `403 No ackDB connection found for the user` on every account available to us, so
its cursor lifecycle — bootstrap → drain → reconcile → dedupe — could only ever be exercised against a
mock built from the OpenAPI spec. That mattered more than usual here, because the spec had already
disagreed with the live API once: errors are documented as `{error: {code, message}}` and the API actually
returns `{status, error, message}` with `error` a string. A live call caught that; a mock never would
have, and the same drift could sit in `/changes`. The trigger comes back when ackDB is provisioned.

Also records two gotchas found while building, in `brain/knowledge/pieces-engine/building-pieces.md`.

**One unrelated one-line fix rides along**, in `packages/server/utils`. A budget check over a
200,000-key object was spending ~3.6s of a 5s default limit, so it fell over on any slower CI runner
and failed the job for whichever PR happened to land there — twice on this one. It now has an
explicit 20s limit, in line with the limits already set in the worker package. Nothing to do with
Floqer; happy to split it out if you would rather it travelled separately.

## How was this tested?

Manually, against a real Floqer account on a local CE stack. Everything below was driven through the
platform's own `/pieces/options` and `/sample-data/test-step` endpoints, so the piece ran the way the
engine runs it rather than through a side channel:

- Auth validates against `GET /api/v1/user/`, and Custom API Call returns real user JSON — which proves
  both the Bearer mapping and the `{status, data}` envelope
- Add Rows inserts rows and returns `row_ids`; a bad row lands in `rejected[]` while the good one is
  still written; unicode, emoji, quotes and newlines all survive the round trip
- The 1000-row cap and the empty-batch guard fire in-piece, with no wasted request
- An API error becomes a **FAILED** run, never a green one — Run Rows relays Floqer's 400 verbatim
- Run Shortcut queues a run, with 8 dynamic props built from a live `input_schema`
- All three dropdowns either resolve or degrade with a message that says why
- Output keys match `outputSchema` exactly, on both the builder and the MCP surface

`tsc --noEmit` and `eslint` are both clean on the piece.

**Edition paths.** CE only — a community piece with no edition-gated surface, no new environment
variable, and no self-hosted setup beyond entering an API key.


packages/pieces/community/floqer/src/lib/actions/run-shortcut.ts
```diff
@@ -0,0 +1,120 @@
...
+                });
+                if (shortcut === undefined) {
+                    return {};
+                }
+                return buildInputProps(shortcut.input_schema ?? []);
+            },
+        }),
+    },
+    async run({ auth, propsValue }) {
+        const inputData = stripEmpty(propsValue.inputData);
```
ibrahim-abuznaid, line 42: `propsValue.inputData` isn't guaranteed non-null at runtime (see point 1). `Object.entries(null)` throws.

```suggestion
        const inputData = stripEmpty(propsValue.inputData ?? {});
```

packages/pieces/community/floqer/src/lib/actions/run-shortcut.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+async function findShortcut({
+    apiKey,
+    shortcutId,
+}: {
+    apiKey: string;
+    shortcutId: string;
+}): Promise<FloqerShortcut | undefined> {
+    const response = await floqerApi.enveloped<FloqerShortcut[]>({
```
ibrahim-abuznaid, line 62: Only remote call in the piece without a `try/catch` — all three dropdowns in `props.ts` degrade to `disabled(floqerApi.describe(...))`. When this throws the whole Inputs panel errors with no explanation, which is also what sets up the null above. A `try/catch` plus an `Array.isArray(response.data)` guard returning `undefined` would match the rest of the file.

packages/pieces/community/floqer/src/lib/actions/add-rows.ts
```diff
@@ -0,0 +1,96 @@
...
+            apiKey: auth.secret_text,
+            method: HttpMethod.POST,
+            path: `/api/v1/workflows/${propsValue.workflowId}/sheets/${propsValue.sheetId}/rows`,
+            body: {
+                rows,
+                run_after_add: propsValue.runAfterAdd ?? 'none',
+            },
+        });
+
+        return {
```
ibrahim-abuznaid, line 64: Suggest defaulting the whole documented shape, not just `warnings`:

```ts
return {
    row_count: 0,
    row_ids: [],
    rows_queued_for_run: 0,
    rejected: [],
    ...(response.data ?? {}),
    warnings: response.warnings ?? [],
};
```

packages/pieces/community/floqer/src/lib/common/client.ts
```diff
@@ -0,0 +1,104 @@
...
+
+function readErrorField(error: unknown, field: 'code' | 'message'): string | undefined {
+    if (!(error instanceof HttpError)) {
+        return undefined;
+    }
+    const body = error.response.body;
+    if (!isRecord(body)) {
+        return undefined;
+    }
+    const nested = isRecord(body['error']) ? body['error'] : undefined;
```
ibrahim-abuznaid, line 71: `nested` is only consulted when `body.error` is a record — but as your own test at `test/actions.test.ts:187` documents, Floqer sends `error` as a **string**. So `{status: 403, error: 'Forbidden'}` with no `message` loses the reason entirely and `codeOf` can never read a code. Worth reading `body.error` as a string fallback.

packages/pieces/community/floqer/src/lib/actions/run-rows.ts
```diff
@@ -0,0 +1,71 @@
...
+    const ids = normalizeRowIds(rowIds);
+    if (ids.length === 0) {
+        throw new Error(
+            'Select at least one row ID, or switch "Rows To Run" to the first 10 rows on the sheet.',
+        );
+    }
+    return { row_ids: ids };
+}
+
+function normalizeRowIds(rowIds: unknown): string[] {
```
ibrahim-abuznaid, line 64: Non-strings are mapped to `''` and filtered out silently. The natural wiring `Row IDs = [ {{step_1.row_ids}} ]` resolves to a nested array `[['r1','r2']]` — every element is dropped and the user sees "Select at least one row ID" while looking at a populated field. A mixed list is worse: it runs a subset with no warning. Suggest flattening one level and throwing on anything still non-string, naming the index.

Also: no 1000 cap here, while `add_rows` has one — worth the symmetry.

packages/pieces/community/floqer/src/lib/actions/add-rows.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (rows.length > MAX_ROWS_PER_CALL) {
+            throw new Error(
+                `Floqer accepts at most ${MAX_ROWS_PER_CALL} rows per call, but ${rows.length} were provided. Split them across multiple steps.`,
+            );
+        }
+
+        const response = await floqerApi.enveloped<FloqerAddRowsResult>({
+            apiKey: auth.secret_text,
+            method: HttpMethod.POST,
+            path: `/api/v1/workflows/${propsValue.workflowId}/sheets/${propsValue.sheetId}/rows`,
```
ibrahim-abuznaid, line 57: Worth `encodeURIComponent` on each interpolated segment here and in `run-rows.ts:43`, `run-shortcut.ts:47`, `props.ts:60`. These IDs aren't guaranteed to come from the dropdown — `PropertyType.DROPDOWN` renders with `allowDynamicValues`, so an author can feed any expression from a prior step, and a value containing `?` or `#` silently rewrites the request. Host is fixed so nothing crosses a privilege boundary, but it's the catalog convention (114 pieces do it).

packages/pieces/community/floqer/test/actions.test.ts
```diff
@@ -0,0 +1,292 @@
...
+    });
+
+    it('reports the status for branching', () => {
+        expect(floqerApi.statusOf(new HttpError({}, { status: 409, responseBody: {} }))).toBe(409);
+        expect(floqerApi.statusOf(new Error('network'))).toBeUndefined();
+    });
+});
+
+describe('Run Shortcut — input_schema as Floqer really returns it', () => {
+    // Captured from a live published shortcut. Every field came back
```
ibrahim-abuznaid, line 210: AGENTS.md bans comments in code with no carve-out for tests — fixture provenance like this belongs in the commit message or PR body. Same at `:236`. (`src/` is clean, nice.)

packages/pieces/community/floqer/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "types": "./dist/src/index.d.ts",
+    "scripts": {
+        "build": "tsc -p tsconfig.lib.json && cp package.json dist/",
+        "lint": "eslint 'src/**/*.ts'",
+        "test": "vitest run"
+    },
+    "dependencies": {
+        "@activepieces/pieces-common": "workspace:*",
+        "@activepieces/pieces-framework": "workspace:*",
+        "tslib": "2.6.2"
```
ibrahim-abuznaid, line 14: `tslib` belongs in `devDependencies` — that's what the generator emits and what `migrate-piece-utils.ts` actively moves it to; 729 of 733 pieces have it there. This file is also missing the standard `bundle` script that 728 pieces carry.

packages/pieces/community/floqer/vitest.config.ts
```diff
@@ -0,0 +1,17 @@
+import path from 'path'
+import { defineConfig } from 'vitest/config'
+
+const repoRoot = path.resolve(__dirname, '../../../..')
+
+export default defineConfig({
+  test: {
```
ibrahim-abuznaid, line 7: No `include` and no `root`, so running this config from the repo root globs the whole monorepo. Compare `packages/server/utils/vitest.config.ts`: `include: [path.resolve(__dirname, 'test/**/*.test.ts')]`.

packages/pieces/community/floqer/src/lib/common/output-schemas.ts
```diff
@@ -0,0 +1,67 @@
...
+            description:
+                'Rows Floqer could not accept. The valid rows in the same batch were still written, so a non-empty list means the batch was only partly imported — branch on this if that matters.',
+            listItems: [
+                {
+                    key: 'row',
+                    label: 'Original Row',
+                    description: 'The row exactly as it was sent, so it can be fixed and resent.',
+                },
+                {
+                    key: 'errors',
```
ibrahim-abuznaid, line 36: Following up the Greptile thread that got resolved without a change: `rejected` did gain `listItems`, but `errors` inside it is still a bare field. `OutputSchemaField.listItems` is recursive, so the `field`/`code`/`message` shape from `FloqerRowError` is expressible here. (Its other half, `row_ids`, I don't think is expressible — plain strings have no sub-fields. Fine to leave.)

ibrahim-abuznaid (changes requested):
Nice piece — clean `src/` (no comments, no `any`, no casts), correct auth wiring, and `minimumSupportedRelease: '0.82.0'` is the honest floor rather than the `0.36.1` cargo most pieces carry. Four things before merge.

**1. `run-shortcut.ts:42` — `stripEmpty` can be handed `null` and throws a raw `TypeError`.** The engine doesn't required-validate DYNAMIC props (`validateProperty` in `props-processor.ts` has no `PropertyType.DYNAMIC` case, falls to `default: return []`), and the builder writes `null` into that field on refresher changes (`clearPropertyValue` in `dynamic-piece-property.tsx`). Shortcut props fetch fails → `inputData` stays `null` → the run dies with `Cannot convert undefined or null to object`.

**2. `add-rows.ts:64` — `rejected` isn't defaulted.** `warnings` gets `?? []`; `rejected` doesn't, yet `outputSchema` advertises it and the description tells users to branch on `rejected.length`. Every test mock hardcodes `rejected: []`, so the omission case is untested.

**3. Please drop the `packages/server/utils` hunk.** #15476 already shrank that same fixture 200,000 → 50,000 keys and the merge kept both, so HEAD reads `length: 50_000` … `}, 20_000)`. It runs in ~216 ms against a 5 s default. Also there's no `20_000` test timeout in `packages/server/worker` — that one is in `packages/server/engine/vitest.config.ts`.

**4. Missing `src/i18n/translation.json`** — 730 of 733 pieces have one; without it no displayName or description reaches Crowdin. `npm run cli pieces generate-translation-file floqer`.

Rest is inline and non-blocking. Two things I checked so they don't come back as churn: the piece does **not** need to set `Content-Type` (`BaseHttpClient.getHeaders()` defaults object bodies to `application/json`), and not catching in `run()` is fine — `extractApiMessage` checks `message` first, so Floqer's message already surfaces on a failed run.

Last one is on us, not you: this piece's 26 tests never run in CI — `ci.yml`'s `turbo run test` filter names no piece package, and both `tsconfig.lib.json` and the `lint` script only cover `src/**`. I ran them locally, 26/26 green, but the green check doesn't cover them. Filing that separately.

OdaiAhmed99:
Thanks — all four are fixed, and I verified each claim before changing anything.

**1. Null `inputData`.** Confirmed: `validateProperty` has no `PropertyType.DYNAMIC` case so it falls to `default: return []`, and `clearPropertyValue` does write `null` via `form.setValue`. Fixed with `?? {}`. I also took the inline suggestion on `findShortcut` — it now catches and guards with `Array.isArray`, so a failed fetch leaves Inputs empty instead of erroring the panel.

**2. `rejected` not defaulted.** Fixed — but the literal suggestion doesn't compile. `FloqerAddRowsResult` declares those four keys as required, so spreading `response.data` after the defaults gives `TS2783: 'row_count' is specified more than once`. The intent was right and the type was the real problem: the envelope is now `Partial<FloqerAddRowsResult>`, which both compiles and is the honest contract, since the whole premise is that Floqer can omit a field the schema advertises.

**3. `packages/server/utils` hunk.** Confirmed — `main` already carries `length: 50_000` from #15476, so the PR only added the timeout. Reverted to match `main` byte-for-byte; the PR now touches nothing outside the piece. I couldn't execute that suite locally to reconfirm your ~216 ms, for the reason in the next point.

**4. `translation.json`.** Generated with the CLI, 52 keys.

**On the vitest `include` — the suggested pattern breaks on Windows.** `path.resolve(__dirname, 'test/**/*.test.ts')` produces a backslash path, and globs only match forward slashes, so the suite silently resolves to **zero test files**. I hit exactly this against `packages/server/utils` as well, which is why I couldn't re-time the test above — that config has the same latent problem on Windows, worth a separate look since it currently means those tests quietly don't run there.

I used `root: __dirname` with a relative `include: ['test/**/*.test.ts']` instead. That still fixes what you raised — verified by running the config from the repo root, where it picks up only the piece's tests, not the monorepo — and it works on Windows: **26/26 green**.

Also done: string `body.error` read as the message (your 403 case); `run_rows` flattens one level and names the offending index rather than silently dropping non-strings, plus a 1000 cap for symmetry with `add_rows`; `encodeURIComponent` on all four interpolated segments; `listItems` on `rejected[].errors`; `tslib` moved to `devDependencies` with the standard `bundle` script; both test comments removed.

Build and lint clean. And noted on the CI gap — agreed the green check doesn't cover these 26 tests today.

ibrahim-abuznaid (approved):

PR #15438, opened 2026-09-10
fix(auth): refuse zerobounce mx_forward verdicts at sign-up

AhmadTash:
## Description

When `AP_ZEROBOUNCE_API_KEY` is set, a new sign-up address is checked against ZeroBounce and refused on `spamtrap`, `abuse`, or `do_not_mail` with one of four sub-statuses (`disposable`, `toxic`, `possible_trap`, `global_suppression`). That let throwaway-domain farms through: `y4ze3ebtt0@zkqjdkd.com` signed up, verified a code and burned free credits. The domain is a 16-month-old Gname registration with no website whose MX records point at Cloudflare Email Routing, which is exactly ZeroBounce's `mx_forward` verdict ("domains set up to forward their MX records into one domain. Their use case is somewhat similar to disposable domains").

This PR adds `mx_forward` to the refused `do_not_mail` sub-statuses and caches it by domain the same way `disposable` is, since forwarding is a property of the domain and a farm reuses one domain for many random mailboxes. Role addresses (`role_based`, `role_based_catch_all`) stay accepted, and the whole-status refusal is deliberately not taken. Everything else is unchanged: fail-open on any error, the invitation carve-out, silent refusal on both call sites.

### Why `mx_forward` and not the whole status

Checked against our own Cloud sign-ups (PostHog, 60 days to 2026-09-10):

- Of 14 custom domains behaving like farms (many accounts, flows created and run, nothing published), 9 are on Cloudflare Email Routing (`hnyjktv.com`, `ofaldgw.com`, `jqrvlhc.com`, `xiakdi.com`, `zkqjdkd.com`, `onlylicensedcasinos.com`, `kwkub.cyou`, `paytrust.cc`, `filorosso.shop`); three of them share one origin IP. The largest farm, `emalupe.com` (1,463 accounts), is mail.tm and is already caught by `disposable`. The rest run their own MX or Zoho, which no email check catches.
- 40 randomly sampled custom-domain accounts that went on to publish a flow resolve to 19 Google, 4 Microsoft, 2 Zoho and 15 other hosted MX. Zero forwarding services.

Forwarding is common in farms and rare among real members, so the false-positive exposure is small and the invitation carve-out covers it. Whether ZeroBounce labels every Cloudflare-routed domain `mx_forward` or only some is not verified.

Brain decision 000032, the `ce-authentication` gotcha, the env-var docs and the breaking-changes page are updated in the same PR.

## How was this tested?

- `packages/server/api/test/unit/app/authentication/zerobounce.test.ts` (37 tests pass): `mx_forward` refused and cached by domain; `role_based`, `role_based_catch_all` and a bare `do_not_mail` pinned as allowed.
- ESLint on the changed source and test files.
- The check only runs when the key is set (Cloud in practice); CE/EE without a key skip the call entirely.

Fixes # (issue)

The ZeroBounce check shipped in 0.89.0, so a self-hoster with the key set now sees addresses on mail-forwarding domains refused at sign-up. Entry added to `docs/install/reference/breaking-changes.mdx` (invite the member first, or unset the key).

Touches the sign-up authentication path. Risk: a legitimate custom domain forwarded to a personal mailbox (Cloudflare Email Routing, ImprovMX) is refused. Mitigation: refusal stays silent (no new oracle), fail-open on any ZeroBounce error is unchanged, refusals are logged with `subStatus` so the mix can be watched after deploy, and an accepted invitation still bypasses the check.


SimoneRoznowicz:
Hi Ahmad — I came across this PR and noticed you are tightening sign-up protection against mx_forward domains after seeing throwaway-domain farms pass verification and burn free credits.

We built Agents Verify for AI-agent and bulk email workflows. It offers comparable quality to MillionVerifier, NeverBounce, ZeroBounce and other leading providers, but costs at least 75% less than the cheapest.

You can try it with 500 free credits on a real sample and compare the classification on forwarding, disposable, and catch-all domains. Worth trying and letting me know what you think?

Best,
Simone, co-founder at agentsverify(dot)dev

PR #15442, opened 2026-09-10
feat(embed): load GTM and Clarity inside the embedded app via SDK config

### LINKED ISSUE #14796: [Feature]-[Embed]: Allow custom tracking scripts (GTM / Clarity) in the embedded app's head
## Problem

Embed loads `/embed` in an iframe and serves the same SPA `index.html` (`packages/web/index.html`) as the main app. Head is static, baked at build time by `packages/web/vite-plugins/html-plugin.js` (only `apTitle` / `apFavicon` substituted). No runtime script-injection mechanism exists anywhere in web/api/ee (verified by grep). PostHog telemetry is explicitly disabled for embed (`packages/web/src/components/providers/telemetry-provider.tsx:42-46`).

Consequence: a session-recording tool (Clarity, GTM) running on the parent page cannot see into the iframe, and nothing can be added inside it. Embed customers have zero user-behaviour analytics on the builder.

Ref: Pylon 5631

## Proposed

Setting for customer-supplied tracking script tags injected into the embedded app's `<head>` at runtime. Platform-level column (alongside existing customization fields on `packages/server/api/src/app/platform/platform.entity.ts`) or embed-SDK config, whichever the platform team prefers.

Precedent for the mechanism: `packages/web/src/app/components/embedding-font-loader.tsx:19-23` already appends a `<link>` to `document.head` from SDK config (`styling.fontUrl`).

Off by default. No change for existing users.

## Notes for technical review

* App CSP emits only `frame-ancestors` (`packages/server/api/src/app/helper/embed-security.ts:18-20`), no `script-src` — third-party scripts would not be blocked.
* Open: trust boundary / XSS surface of admin-supplied script strings.
* Open: editor performance impact of a session recorder on the builder canvas.

AhmadTash:
## Description

Embed customers can now run their own analytics inside the embedded builder. Scripts on the host page cannot see into the iframe, so until now embedded users were invisible to tools like Google Tag Manager and Microsoft Clarity. SDK 0.16.0 adds one option to `configure()`:

```js
activepieces.configure({
  instanceUrl,
  jwtToken,
  embedding: {
    containerId,
    analytics: {
      gtmContainerId: 'GTM-XXXXXXX',
      clarityProjectId: 'abcdefghij',
    },
  },
});
```

The ids ride the existing `VENDOR_INIT` handshake. Once the managed-auth token exchange succeeds, the embedded app loads the GTM container (after the standard `dataLayer` start event) and/or the Clarity tag (after installing Clarity's queue stub, which its tag requires) into the iframe's `<head>`. Off by default; nothing changes for embeds that don't pass ids.

### Security

Only ids are accepted, never script strings. They are regex-validated (`GTM-[A-Z0-9]+`, alphanumeric for Clarity), URL-encoded, and only ever point at `googletagmanager.com` and `clarity.ms`. Nothing loads before the JWT exchange succeeds, so only a vendor holding a valid signing key can enable it. The loaded tags run with the embedded user's session, which is the same trust the vendor already holds by minting that user's JWT; the docs now carry a warning about that, about what Clarity records, and about consent ownership. Loader failures are caught so a third-party bootstrap can never crash the builder, and a tag script that fails to load removes its own element and warns instead of being mistaken for loaded. Frontend error reporting is intentionally untouched.

### Also in this PR

- Docs: two configure-table rows, a privacy/trust warning, and a 0.16.0 SDK changelog entry.
- `bun.lock` workspace version synced with the SDK bump (the Dockerfile's frozen install fails otherwise).
- Brain: embed page gotchas (Clarity stub, GTM diagnosis, `VENDOR_INIT` flattening, no `script-src` in CSP, local EE testing recipe) and benchmark numbers.

## How was this tested?

- Unit: 4 new web tests for the loader (valid ids load once under StrictMode, Clarity stub installed, invalid ids rejected, nothing outside embed) and 1 new SDK test for the `VENDOR_INIT` passthrough. ESLint and TypeScript clean on all touched files.
- End to end on an `AP_EDITION=ee` Docker image built from this branch, with a vendor harness page driving the real SDK bundle:
  - Clarity, real project id: `tag` 200 → `clarity.js` 200 → `collect` 204, no console errors.
  - GTM, real container: `google_tag_manager[id]` present, `gtm.js` → `gtm.dom` → `gtm.load` lifecycle, published GA4 and Custom HTML tags fired, `google-analytics.com/g/collect` 204.
- Performance, headless Chromium, medians of 3, scripted zoom/pan/step-panel interaction:
  - 12-step flow: Clarity +6% main-thread, GTM ≈ 0, no change in long tasks.
  - 155-step flow at 1x and 4x CPU throttle with both enabled: frame rate and p95 input latency unchanged, main-thread +0.2% / +4%, heap +40 MB / +17 MB (Clarity's DOM mirror), 0.9 MB download once, ~5 KB/s upload while active.
- Editions: verified on EE. Cloud shares the identical client and SDK code path (the iframe origin is the embed subdomain or `cloud.activepieces.com`). CE has no embedding.

Fixes #14796


AhmadTash:
@greptile review

AhmadTash:
@greptile review

othmanemad (approved):

PR #15464, opened 2026-09-11
fix(mooninvoice): dropdown props send filters as GET body, which is dropped

kishanprmr:
## Description

Company/Contact/Project dropdown props (`packages/pieces/community/mooninvoice/src/lib/common/props.ts`) sent their list filters as a GET request `body`. The piece's fetch-based `httpClient` silently drops the body on GET/HEAD requests, so all three dropdowns never actually applied their filters and effectively broke.

`makeRequest` (`common/client.ts`) now accepts a `queryParams` option and forwards it to `httpClient.sendRequest`, which appends query params to the URL regardless of method — confirmed this is what Moon Invoice's `company_list` / `contact_list` / `project_list` endpoints accept (POST to those endpoints returns an encrypted, unusable payload instead).

Also fixes `contactIdProp` sending `CompanyId`, when the API only accepts `CompanyID` — found while testing against a live account.

Bumped piece version 0.0.8 → 0.0.9.

## How was this tested?

- Verified all three dropdown calls (`company_list`, `contact_list`, `project_list`) against a live Moon Invoice account — each now returns real data with the new query-param request shape.
- `npx turbo run lint --filter=@activepieces/piece-mooninvoice` — 0 errors (pre-existing `any`/unused-import warnings only, none new).
- `npx turbo run build --filter=@activepieces/piece-mooninvoice` — builds clean.
- Piece has no UI surface beyond the flow builder's dropdown/action forms; not edition-specific (CE/EE/Cloud all use the same piece code).

Fixes PIE-503

### DIFF
--- a/packages/pieces/community/mooninvoice/package.json
+++ b/packages/pieces/community/mooninvoice/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@activepieces/piece-mooninvoice",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "type": "commonjs",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
--- a/packages/pieces/community/mooninvoice/src/lib/actions/add-new-contact.ts
+++ b/packages/pieces/community/mooninvoice/src/lib/actions/add-new-contact.ts
@@ -269,12 +269,12 @@ export const addNewContact = createAction({
       context.auth.props.email,
       context.auth.props.secret_text
     );
-    const response = await makeRequest(
+    const response = await makeRequest({
       accessToken,
-      HttpMethod.POST,
-      '/add_contact',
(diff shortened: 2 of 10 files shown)


kishanprmr:
@greptile-apps re: the 25-record pagination cap on `contactIdProp`/`projectIdProp` — that limit is pre-existing (the code already hardcoded `PageSize: 25` before this PR); this fix only restores the filtering that was silently broken by the GET-body bug, it doesn't introduce the cap. Since the dropdowns previously returned nothing at all, this is a strict improvement, not a regression. Pagination support for these dropdowns is a separate, larger change and out of scope for this bug fix — worth its own follow-up ticket if accounts with >25 contacts/projects need it.

ibrahim-abuznaid (approved):

PR #15468, opened 2026-09-11
hotfix(0.87.1): backport prewarm scope-file + version-gate fixes as 0.87.2

MrChaker:
## Description

Backports the prewarm fixes from `feat/prewarm-scope-via-file` (PR #15426) onto the `0.87.1-hotfix.5` release line as **0.87.1-hotfix.6**. Base is the `0.87.1-hotfix.5` tag; this PR targets the hotfix.5 branch so the diff shows only the hotfix.6 delta.

**What's in it:**
- **Prewarm scope as a file** — `getPrewarmData` no longer returns the platform scope (distinct pieces + code-step sources) inline on the worker socket, where a platform-sized payload blocks the app event loop and drops every worker connection (the hotfix.5 incident). The scope is saved as a `PREWARM_SCOPE` file (deterministic id, S3 signed URL when configured) and workers fetch it via the new `getPrewarmScopeFile` RPC.
- **Version-gated prewarm RPCs** — the app fails closed on a missing/mismatched `workerVersion`, and the worker holds startup (health server down, 10s retry loop) until connected to a version-compatible app, resuming automatically after a rolling deploy completes.
- **Bundle download resilience** — each piece tarball download retries up to 3 attempts (404/410 treated as permanent), failed bundles are partitioned out, and `bun install` proceeds with what landed; prewarm treats skipped bundles as best-effort instead of aborting.
- **`package.json` → `0.87.1-hotfix.6`** — the version gate compares package.json versions, and all previous hotfixes shared `0.87.1`; the full suffixed version makes hotfix.6 distinguishable from hotfix.5, which matters because the prewarm response shape changed between them. The release workflow's tag-verify step now also accepts an exact tag == package.json match (still failing closed otherwise).

**Backport adaptations:** core package versions untouched (per the no-version-bump constraint); prewarm remains always-on (this line predates `AP_PREWARM_CACHE_ON_STARTUP`) with the version-hold loop in front; the permanent-download-failure break uses a local `PieceBundleNotAvailableError` since `ErrorCode.PIECE_BUNDLE_NOT_AVAILABLE` doesn't exist in this line's core-utils.

**Known caveat:** `0.87.1-hotfix.6` is a semver prerelease (sorts below `0.87.1`); anything semver-comparing the running release (e.g. piece `minimumSupportedRelease` filtering) may treat this build as older than 0.87.1 — verify before shipping.

## How was this tested?

All 17 turbo build tasks pass on the branch with its own-era lockfile (`bun install` shows no lock drift). The equivalent changes were verified end-to-end on main's line on a GKE cluster: 153 MB scope prewarmed via signed URL with zero disconnects (vs. a permanent connect/disconnect loop pre-fix), version-skew hold + self-heal on app upgrade with the same pods, and 12/12 prewarm completions at 1000 flows / 401 pieces / 100 code steps (vs. 3/13 aborts before the bundle retry).

Fixes # (issue)

- [x] yes — security-sensitive (call out the risk and mitigation in the description above — the new PREWARM_SCOPE file is served only over the token-authenticated worker RPC, mirroring the existing getFlowBundle signed-URL transport)

### DIFF
--- a/.github/workflows/release-self-hosted.yml
+++ b/.github/workflows/release-self-hosted.yml
@@ -38,8 +38,10 @@ jobs:
       # checked-out source. The `tag` input is free-text, so a mistyped tag would publish an
       # image whose real version disagrees with its tag/git-tag/GitHub-release — invisible until
       # it is mixed with a correctly-built peer and the worker↔app version gate silently
-      # withholds jobs. Fail closed here. Hotfix/rc suffixes (0.79.2-hotfix.1, 0.79.2-rc.1) are
-      # deploy-only and are not in package.json, so compare against the base X.Y.Z only.
+      # withholds jobs. Fail closed here. A hotfix/rc suffix (0.79.2-hotfix.1, 0.79.2-rc.1) is
+      # accepted two ways: an exact match when package.json carries the full suffixed version
+      # (needed when a hotfix must be version-gate-distinguishable from its predecessor), or a
+      # deploy-only suffix where package.json carries just the base X.Y.Z.
       - name: Verify tag matches package.json
         # TAG is passed via env (not inlined as ${{ inputs.tag }} in the script) to avoid shell
         # expression injection — a crafted tag would otherwise run as shell in the run block.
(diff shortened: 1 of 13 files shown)


MrChaker:
Renumbered from 0.87.1-hotfix.6 to **0.87.2**: a `-hotfix` suffix in `package.json` is a semver *prerelease* (sorts below its base), and on the released hotfix.6 image the piece registry filter (`minimumSupportedRelease` vs `getCurrentRelease`) hid 326 of 764 pieces — reproduced on the benchmark cluster. `0.87.2` sorts above `0.87.1`, keeps the worker/app version gate able to distinguish this release from hotfix.5, and leaves piece filtering intact. The published `0.87.1-hotfix.6` image/tag should be considered defective and not deployed.

PR #15471, opened 2026-09-11
feat(pinterest): improve the step forms

Talaljaber:
## Description

A step-form pass over the Pinterest piece: every human-visible action and trigger plus the connection dialog. `@activepieces/piece-pinterest` 0.1.8 → 0.1.9, `minimumSupportedRelease` 0.36.1 → 0.88.2 (needed for `advanced`, `propertyGroups`, cards and steppers), category set to Marketing.

### Forms

- Shared dropdowns relabelled **Board**, **Pin**, **Ad Account**, **Board Section** (were `Board Id`, `pin Id`, `Ad account Id`, `Board Section Id`). Without a search term each loads up to two pages of 250 (was the first 25 only), deduplicates by id and stops on a repeated bookmark. Typing in **Board** or **Pin** searches Pinterest (`/search/boards`, `/search/pins`) so records past the first 500 stay reachable; Ad Account and Board Section have no search endpoint, so they may follow up to ten pages and filter the loaded list. The Pin dropdowns refresh on **Ad Account** and pass it to Pinterest (0.1.8 always listed the user's own Pins, even when the action targeted an ads account). Cut-short and empty lists say so in the placeholder.
- **Ad Account** is in Advanced on every form with the line "Only for ads accounts. Leave empty to use your own account."
- **Create Pin**: two section cards, *Save To* (board, section) and *Pin* (title, description, media type, media, link). Media Type is a two-card picker. Eight rarely used fields are in Advanced. 15 fields above the fold → 7.
- **Create Board / Update Board**: Privacy is a card picker; Ads-Only Board is in Advanced.
- **Find Pin by Keyword** (was "Find Pin by Title/Keyword") and **Find Board by Name**: the pagination field is called *Bookmark* on both and lives in Advanced; Max Results is a 1–250 stepper.
- **New Board**: Privacy filter relabelled, Page Size is a 1–250 stepper in Advanced with the default 25 now declared on the prop.
- **New Pin on Board**: Pin Types in Advanced.
- Descriptions are one short sentence; examples moved into placeholders.
- The OAuth2 description is a three-step setup guide and warns that trial-access apps only see their own Pins and boards.

### Behaviour changes

Checked against Pinterest's published OpenAPI spec (github.com/pinterest/api-description, v5). These are intentional and listed here because the title does not suggest them.

- **Find Pin returns Max Results.** `GET /search/pins` has no `page_size`, so the action only ever returned the first page. It now follows the response bookmark until it has Max Results Pins (capped at ten requests) and returns exactly that many. Consequences: a run can make up to ten requests instead of one, and a flow that received twelve Pins from a single page now receives up to 25. Max Results is clamped to 1–250 at run time; `0` or a negative value behave as 1.
- **Find Pin's bookmark is one opaque cursor.** A Pinterest bookmark addresses a page, never a position inside it, so when the cap falls inside a page the `bookmark` output is an `apc1.`-prefixed cursor that encodes the page bookmark and how many Pins of it were already returned. Passing it back continues without repeats or gaps, the first page included. At a page boundary the output is the raw Pinterest bookmark as before, and a raw bookmark from a 0.1.8 flow is still accepted. `has_more` is never true without a usable `bookmark`. A cursor that does not fit the current search throws.
- **Find Pin fails loudly.** A response without an `items` array throws instead of returning zero results, so "found nothing" and "the call failed" are distinguishable. The old `try/catch` that rewrote a 404 into "No pins found matching your search criteria." is gone; `makeRequest` already normalises errors, so `{{step['error'].message}}` now carries Pinterest's message.
- **Create Pin sends real base64.** The API wants `{ content_type, data }`, not `{ url }`. The action accepts a base64 data URI or raw base64 (wrapped lines and URL-safe alphabets are normalised), requires the `;base64` marker on data URIs, detects JPEG or PNG from the header when no type is declared, and rejects anything else. `image/jpeg` and `image/png` are the spec's only `ContentType` values.
- **Create Pin no longer offers Video URL.** `video_url` is not a valid `source_type` (video needs `video_id` from a separate upload flow), so it always failed with a 400. A saved step that still stores `video_url` now fails with "Video Pins are not supported by this action" instead of a Pinterest 400.
- **Update Board treats an empty description as "keep".** 0.1.8 sent `description: ""` whenever the field was blank and wiped the board's description on every run, while the help text promised "leave empty to keep". Empty and whitespace now mean no change, matching the existing `name` handling. Clearing a description is therefore no longer possible from this action; that is the intended reading.
- **Update Board drops Protected.** `PATCH /boards/{id}` accepts `PUBLIC` and `SECRET` only. Create Board keeps all three.
- **Delete Pin** URL-encodes `ad_account_id`; the other actions already did.
- `product_tags`, `note` and `is_removable` are not fields of `POST /pins`. The props are kept for saved-step compatibility, moved to Advanced and relabelled so they no longer promise something the API does not do.

### Compatibility

- No prop was added, removed or renamed and no prop type changed (`check-step-form.mjs --before` against the 0.1.8 metadata is clean). Every stored value survives the upgrade.
- Renamed English labels orphan their translations. `src/i18n/translation.json` was regenerated; the six locale files were repaired by hand because the nightly translation job has been failing since August: 15 pure renames (`Board Id` → `Board`, `pin Id` → `Pin`, `Pagination Bookmark` → `Bookmark`, …) carry their existing translation over, and the orphaned entries for rewritten descriptions were removed so the files stay in sync with the source until Crowdin fills the new strings.
- Added `Talaljaber` to `authors`.

### Out of scope

New Board and New Follower can drop items that arrive faster than their `maxPages` cap between polls (New Pin on Board already persists a bookmark). Left for a follow-up.

## How was this tested?

- `tsc --noEmit` on the piece: clean. ESLint: only pre-existing `no-explicit-any` warnings remain.
- `node .agents/skills/piece-builder/scripts/check-step-form.mjs pinterest`: 6 errors / 10 warnings before → 0 errors / 1 warning after. The remaining warning is the expected "ungrouped props render below sections"; every ungrouped prop on Create Pin is in Advanced.
- Opened all 9 forms as fresh steps at the default 400 px panel on CE (`AP_DEV_PIECES=pinterest`), before and after. Screenshots: https://claude.ai/code/artifact/56f08cfa-4abe-45b7-9313-877109bd4632.
- Built Create Pin, Update Board and Find Pin steps on 0.1.8 with values in every touched prop, upgraded them to 0.1.9 in the builder: every field renders with its saved value, no error banners.
- Find Pin's page loop simulated with three 25-item pages: limits of 10, 25, 30, 50 and 250 return 10, 25, 30, 50 and 75 items. Chaining runs of 10 with only the returned `bookmark` yields Pins 1–75 exactly once, in order, including cuts inside the first page. `max_results` of 0 and -5 behave as 1. A raw Pinterest bookmark is accepted; a malformed cursor throws.
- Base64 parsing exercised with raw PNG, raw JPEG, plain, parameterised, upper-case-type, line-wrapped and URL-safe inputs (accepted with the right `content_type`) and with a percent-encoded data URI, a data URI without the `;base64` marker, a GIF data URI, corrupt base64 and a plain URL (all rejected with a clear message).
- Locale repair verified: every remaining key in the six locale files exists in `translation.json`, and each renamed key carries the value its old key had.
- Not run against a live Pinterest account (no OAuth app available locally), so the request bodies and the pagination loops are verified against the spec, not the API.

Fixes # (none)


ibrahim-abuznaid:
Heads-up on a version collision: this PR and #15539 both bump `@activepieces/piece-pinterest` from `0.1.8` to `0.1.9`, and npm currently has `0.1.8`.

Whichever of the two merges second will merge green and then **never republish** — the registry already has `0.1.9` at that point, so the publish step is a no-op and those changes silently never reach cloud users.

The two PRs also overlap on content: both edit the same six pre-existing action files (`create-board`, `create-pin`, `delete-pin`, `find-board-by-name`, `find-pin`, `update-board`) plus `common/index.ts`, this one rewrites `src/i18n/translation.json`, and the two set different floors (`minimumSupportedRelease` `0.88.2` here vs `0.87.0` in #15539).

So it needs sequencing rather than just a rebase: pick which one lands first, then re-bump the second one's version (and reconcile the `minimumSupportedRelease` value) before it merges. Same note left on #15539.

Talaljaber:
> Heads-up on a version collision: this PR and #15539 both bump `@activepieces/piece-pinterest` from `0.1.8` to `0.1.9`, and npm currently has `0.1.8`.
> 
> Whichever of the two merges second will merge green and then **never republish** — the registry already has `0.1.9` at that point, so the publish step is a no-op and those changes silently never reach cloud users.
> 
> The two PRs also overlap on content: both edit the same six pre-existing action files (`create-board`, `create-pin`, `delete-pin`, `find-board-by-name`, `find-pin`, `update-board`) plus `common/index.ts`, this one rewrites `src/i18n/translation.json`, and the two set different floors (`minimumSupportedRelease` `0.88.2` here vs `0.87.0` in #15539).
> 
> So it needs sequencing rather than just a rebase: pick which one lands first, then re-bump the second one's version (and reconcile the `minimumSupportedRelease` value) before it merges. Same note left on #15539.

Thanks for catching the collision. Proposal: merge #15471 first, then #15539 rebases onto main, bumps @activepieces/piece-pinterest to 0.1.10 and takes minimumSupportedRelease: '0.88.2'. This PR puts 15 props in Advanced, which needs that floor; on 0.87.0 they render inline.

Reasoning: #15471 is the smaller PR (15 files, one piece) with no human review pending, while #15539 is already in a revision round after ibrahim's review, so a rebase folds into work that is happening anyway. The overlap is the six shared actions: #15539 moves their run() bodies into common/operations.ts, and #15471 fixes bugs inside those bodies (Update Board no longer wipes the description on empty input, Delete Pin encodes ad_account_id, Create Pin builds a real base64 payload and drops the never-valid video_url source, Find Pin paginates to Max Results with a bookmark plus skip cursor). Those fixes should be carried into pinterestOperations during the rebase, and translation.json regenerated afterwards since the new actions add strings.

ibrahim-abuznaid (changes requested):
Reviewed at `f2dfaa05`, against Pinterest's official OpenAPI description (v5.28.0) and the framework at each release tag. The mechanical side of this is clean — prop keys, action names and trigger names are 100% stable, so no saved flow loses a value, and `minimumSupportedRelease: 0.88.2` is the exactly correct floor. What holds it up is that a few real behaviour changes are shipping under a "step forms" title, plus two things that need doing before merge.

Note the branch is **62 commits behind main**, so the green checks are from Sep 11 and will need a re-run after an update-branch.

## Must fix before merge

**1. 283 localized strings across 6 locales regress to English.**

Piece translations are keyed on the raw English string, so rewording a `displayName` or `description` orphans its translation. This PR rewords 12 prop labels, 22 descriptions and 5 option labels, and touches no locale file. Measured per locale (translated entries that no longer resolve):

| locale | before | after | orphaned |
|---|---|---|---|
| de | 84 | 37 | 47 |
| es | 86 | 39 | 47 |
| fr | 86 | 38 | 48 |
| ja | 87 | 39 | 48 |
| nl | 82 | 37 | 45 |
| pt | 87 | 39 | 48 |
| zh | 3 | 3 | 0 |

Coverage goes from ~95-99% to ~45-47% — a step form that reads half-English, half-localized, which is worse than either extreme. And there is no self-healing: `generate-translations.yml` has failed **40 consecutive times, last success 2026-08-16**, so nothing will regenerate these.

The `translation.json` edit itself is correct — I checked it against what `generate-translation-file-for-piece` would produce and it matches exactly, including ordering. Only the locale files were left behind.

The fix is cheap: ~13 of the orphans are pure renames (`Board Id` → `Board`, `pin Id` → `Pin`, `Pagination Bookmark` → `Bookmark`, the four privacy option labels…) where the existing translation carries over unchanged. The other ~35 are description rewrites whose English meaning genuinely changed, so those should just be deleted from the locale files rather than carried forward. That lands post-merge coverage around 60-65% and leaves the files clean for whenever Crowdin is unblocked.

**2. Every dropdown became a 10-request, up-to-2500-option catalog fetch.**

`fetchAllPages` hardcodes `page_size=250` and loops to `maxPages = 10`, with no early stop once enough options exist and no dedupe. Create Pin mounts four of these (`ad_account_id`, `board_id`, `board_section_id`, `product_tags`), so opening that step panel can fire **40 sequential Pinterest GETs**, and it refires on every refresher change.

`makeRequest` maps 429 to a thrown error with no retry — and `retries` in pieces-common only covers 5xx, so 429 is never retried anywhere — which means a busy account gets "Rate Limit Exceeded" instead of a board list. The loaders also ignore the `searchValue` the framework passes, so typing in the dropdown search refetches the same truncated 2500 and can't reach anything past the cap.

Suggest capping at 1-2 pages for the interactive dropdowns, deduping by `id`, and surfacing truncation in the placeholder ("Showing the first N — type to search") rather than presenting a truncated list as complete.

## Behaviour changes that need to be declared

These are all defensible, but none of them are mentioned in the PR body, and a reviewer reading "improve the step forms" would not look for them.

**3. `update-board` can no longer clear a description.** The base used `if (description !== undefined) body.description = description`, so an empty string was sent and cleared it. The head switches to truthiness on the trimmed value, so `''` is dropped — and when description is the only field supplied it now fails the "at least one field must be provided" guard and throws. This cuts both ways: a blank optional LongText stores as `''`, so the base was arguably wiping descriptions by accident on every run. Either reading is reasonable; please state which one you intended in the description.

**4. `find-pin` returns `has_more: true` with an empty `bookmark`.** Fresh run, `max_results: 10`, 25-item page → `bookmark: undefined`, `skip: 10`, `has_more: true`. The page is continuable, but *only* via the new `skip` output — and `aiMetadata.description` still says pagination works "via a bookmark token" and never mentions `skip`. A released flow (or an agent) looping on `bookmark` alone therefore goes from silently skipping Pins to refetching page 0 forever. Greptile raised this as "First Page Cannot Continue" and the thread is resolved, but the behaviour is still present.

The cleanest fix is one opaque cursor in the existing `bookmark` field (base64 of `{bookmark, skip}`, decoded on input) so a single field round-trips and the two values cannot desynchronise. Failing that, update `aiMetadata` and the output descriptions, and never emit `has_more: true` with an unrepresentable continuation.

To be clear about what is *not* wrong here: I traced three chained runs through the `skip` arithmetic and it is correct — no Pin is skipped or duplicated when both fields are wired. `lastPageStart`/`lastPageSkipped` always refer to the page the cursor names, and the boundary cases come out right.

**5. `max_results` is not clamped at runtime.** `min`/`max` on NumberProperty are UI hints; the engine validates only that the value is a number, so stored expressions and MCP/AI callers reach `run()` unclamped. `max_results: 0` breaks the loop immediately and returns zero items with `has_more: true` (the base treated `0` as falsy and returned the whole page); `-5` makes `slice(0, -5)` drop the last five Pins and emits `skip: -5`. `const limit = Math.max(1, Math.floor(max_results ?? DEFAULT_MAX_RESULTS))` covers it.

**6. `find-pin`'s cost and error contract changed.** One request became up to ten, and a flow that previously returned 12 items from one page now returns up to 25. The old `try/catch` that mapped 404 to "No pins found matching your search criteria." and wrapped everything else was removed, so `{{step['error'].message}}` changes for any flow using continue-on-failure.

**7. `skip` and `bookmark` have no cross-validation.** Wiring one without the other silently duplicates or drops Pins, and a stale `skip` past the end of the resumed page clamps to the whole page while the loop still advances — dropping a page with no error. Related: a non-envelope 2xx (empty body, HTML gateway page, shape change) becomes `items: []`, `has_more: false`, which reads as "the search found nothing" rather than "the call failed". A `Find Pin → if total_results == 0 then Create Pin` flow duplicates Pins on every degraded response.

**8. `video_url` removed from Create Pin.** The removal is right — the spec's `PinMediaSource` discriminator only allows `image_base64 | image_url | video_id | multiple_image_base64 | multiple_image_urls | pin_url`, so it never worked. But saved flows still pass it straight through `buildMediaSource`, and the step now renders with no card selected, so one click silently overwrites the stored value. Worth throwing an explicit "video Pins are not supported by this action" instead.

<details>
<summary><b>Minor — non-blocking</b></summary>

- `BASE64_PATTERN = /^[A-Za-z0-9+\/]+={0,2}$/` rejects line-wrapped base64 (what `base64` CLI and many encoders emit at 76 chars) and URL-safe base64, with a message that blames the user. `.replace(/\s+/g, '')` before the test is cheap insurance. Low impact — the base64 path could never have worked before this PR, so this is polish on net-new functionality.
- `isRecord` returns `true` for arrays (`typeof [] === 'object'`). Harmless at today's call sites, but it is a mislabelled guard exported from `common`.
- `fetchAllPages` has no cursor-progress guard — it only checks the bookmark is non-empty, never that it changed. A repeated bookmark refetches and appends the same page up to 10 times as duplicate options.
- `propertyGroups[].label`, option `description` and `placeholder` are **not** in `pathsToValuesToTranslate`, so none of them are translatable. `create-board` deleted a translatable `description` and replaced that guidance with untranslatable per-option descriptions, so non-English users lose it entirely. This is a framework gap from #14501 rather than anything in this PR — I'll open a separate ticket for it along with the broken nightly job.
- You are not in `authors` on the piece despite writing all five commits.
- The PR body says the Board dropdown now refreshes when the connection changes. That is a no-op — `dynamic-dropdown-piece-property.tsx:38` already appends `auth` to every refreshers array. The added entry is harmless, but the rationale is wrong.

</details>

<details>
<summary><b>Verified clean</b></summary>

Prop keys, action `name`s and trigger `name`s are 100% stable — the only addition is the optional `skip`, so no saved flow loses a stored value. No `required` transitions in either direction. Both new `defaultValue`s only affect newly added steps (existing input always wins in `getDefaultValueForProperties`), and New Board's `page_size: 25` matches the pre-existing `page_size || 25` call-site default.

`minimumSupportedRelease: '0.88.2'` is exactly right, not merely safe: `propertyGroups`, `display: 'stepper'`, `display: 'cards'`, `advanced`, `placeholder` and the option `icon`/`description` fields all arrive in the same commit and are absent at 0.87.0 and 0.88.1. Every framework option used is real and set on the property type that defines it. `propertyGroups` is internally consistent — all 7 referenced keys exist, and the 8 ungrouped props all carry `advanced: true`, so nothing is stranded or unreachable.

Dropping `PROTECTED` from Update Board is **correct** and worth keeping: the spec's `BoardUpdatePrivacy` is `PUBLIC | SECRET` only, while `BoardPrivacy` (used by Create Board, which rightly keeps all three) has all three. `page_size=250` is legal on all four endpoints used (the shared bookmark param maxes at 250). `/search/pins` genuinely has no `page_size`, so the premise of the pagination rewrite is right. `delete-pin` gaining `encodeURIComponent` is a fix. No comments added — it removes about ten. No new dependencies, and nothing touched outside the piece.

</details>

## Merge order against #15539

A real merge simulation (`git merge-tree`) gives **18 conflict hunks across 9 of the 10 shared files**. My call is **this PR merges first**, for two reasons beyond convenience:

- **Versions.** `0.1.9` then `0.2.0` publishes cleanly. The reverse silently breaks npm: `packagePrePublishChecks` compares only against the `latest` dist-tag (`currentVersion === latestPublishedVersion`), so a `0.1.9` published after `0.2.0` is already `latest` passes the guard, publishes with `--tag latest`, and drags `latest` **backwards** — cloud would resolve to the build without #15539's new actions, with green CI and no error anywhere.
- **#15539 rewrites the `run()` bodies this PR edits** into a shared `pinterestOperations`. If it lands first, the entire Find Pin pagination rewrite — the substance of your last three commits — has to be re-authored inside `operations.ts`, and its `searchPinsOperation` currently nulls `bookmark` on truncation, which reintroduces exactly the bug you set out to fix.

Your `props.ts`, `auth.ts`, trigger and `translation.json` changes don't overlap #15539 at all, so those survive either order intact.

Talaljaber:
@greptileai review adc77d8d1a. Both open findings are addressed there: the Pin dropdowns now refresh on Ad Account and pass it to /pins and /search/pins, and the Ad Account and Board Section loaders allow ten pages.

ibrahim-abuznaid (approved):

PR #15472, opened 2026-09-11
feat(slack): step form

Talaljaber:
## Description

Step-form pass on the Slack piece (0.17.10 → 0.20.0): the 28 human-facing actions, the 14 triggers and the shared props in `common/props.ts`. The forms grew over several years and eight authors, so the same field carried different labels from one form to the next, nine toggles ended in a question mark, thirty-odd descriptions folded behind "show more", every channel action opened with a 230 px instruction box, and Send Message To A Channel was a 2000 px flat list of twelve inputs. Four runtime defects found while reading the forms are fixed in the same pass.

Before/after of every form, fresh step, default 400 px panel, Advanced collapsed:
https://claude.ai/code/artifact/faa1ad38-d64e-4521-bb66-a3e6a8cbff44

### Labels

Every label is now a Title Case noun. "Send as a bot?", "React as user?", "Is Private?", "Include bots?", "Include disabled users?", "Mention flow of origin?", "Append to existing members?" and the seven "Ignore Bot Messages ?" / "Ignore Message from Yourself ?" toggles become **Send as Bot**, **React as User**, **Private Channel**, **Include Bots**, **Include Deactivated Users**, **Mention Origin Flow**, **Append to Existing Members**, **Ignore Bot Messages** and **Ignore My Own Messages**. "Reply to Thread (Thread Message Link/Timestamp)" becomes **Reply to Thread**, "Reaction (emoji) name" becomes **Emoji**, "Emojis (E.g fire, smile)" becomes **Emojis**, "Thread ts" becomes **Thread Timestamp**. Action names follow: **Update Message**, **Get Channel History**, **Search Messages**, **List Users**, **Upload File**, **Markdown to Slack Format**; the "Channel created" trigger becomes **New Channel**.

### Descriptions

The builder folds anything over 70 characters behind "show more"; the four Message Timestamp fields carried 233–236 characters with backticks in them, the thread field 346, the Channel dropdown 97. Every description now fits on one line, is plain text, and says what the field does rather than restating its name. Examples move into placeholders where the input renders one (`1710304378.475129`, `U012AB3CD`, `project-updates`, `:calendar:`, `invoice in:#finance`). Fields that had no description at all get one: the shared **Message** ("Slack mrkdwn formatting is supported."), **Send as Bot** ("Off posts as the connected user and needs a user token."), **Attachment** ("Sent with the message as its comment. Other options are skipped." — that is what `files.uploadV2` does with blocks and bot identity), **Search Query**, **Channel Name**, the **User** dropdown, and every trigger toggle. Actions that only work with a user token say so in their description (Delete Message, Search Messages, Set User Status, Update Profile), since Bot Token connections may not have one.

### One wording per concept

- **Message Timestamp** on Update, Delete, Get Message and Add Reaction reads "Timestamp of the target message, from its link or a trigger output." on all four; two of them used to say "the message you wish to retrieve" on actions that delete or react.
- The four **Channels** filters on New Command, New Mention, New Reaction and Reaction Removed shared one meaning and had four phrasings; all read "Empty means every channel the bot is in."
- **Unfurl Links**, **Also Post to Channel** (was "Broadcast reply to channel") and the bot identity fields have one description wherever they appear.
- The User dropdowns list `Real Name (@handle)` instead of the bare handle, so people can be found by name.
- Retrieve Thread Messages gets the same channel info block as every other channel action; it was the one form without it.

### Advanced sections

Fields most flows never touch move behind the Advanced fold: Username, Profile Picture, Icon Emoji and Mention Origin Flow (shared, so on every send/update/request action), Also Post to Channel and Unfurl Links on the send actions, Inclusive and Include All Metadata on Get Channel History, Remove Mention from Message on New Mention, and Use User Token on the Custom API Call. Send Message To A Channel goes from twelve visible inputs to six plus a six-option Advanced section, 2023 px → about 1200 px. Nothing required is hidden, and nothing whose default changes the result is hidden: Send as Bot (default on), Reply to Thread and Block Kit Blocks stay in the main body. Blocks stays visible on purpose: a blocks-only message is the case the Message field invites, and the Advanced section only auto-opens on validation errors, so a saved blocks-only step would otherwise open looking empty. This is what raises `minimumSupportedRelease` to 0.88.2.

### Banners

The two channel info blocks shrink from a three-step list plus a "Note" to one line: "The list shows the first 2000 channels. Can't find yours? Invite the bot: type /invite in the channel, choose Add apps and pick the bot. Or click ƒ and paste the channel ID." The single-select variant drops the `{{ ['id1', 'id2'] }}` array note that was copied from the multi-select one and never applied. New Mention in Channel now refuses to enable when both Users and User Groups are empty ("Pick at least one user or user group; with both empty this trigger never fires.") instead of showing a permanent banner, because with both filters empty the trigger genuinely never fires and nothing said so.

### Checkboxes

Nineteen checkboxes across actions and triggers were `required: true`, which only paints a red asterisk on a field that always has a value. They are `required: false` now; every `defaultValue` is unchanged, so runs behave as before.

### Shared props

Message Timestamp, Also Post to Channel and Unfurl Links are exported from `common/props.ts` and imported by the seven forms that use them, so the one-wording rule is structural rather than copy-pasted. `text` stays local on the three send/update actions: the shared one is required and those are optional with a blocks-only hint.

### Runtime fixes

- **New Team Custom Emoji** guarded with `type !== 'emoji_changed' && subtype !== 'add'`; with `&&` an `emoji_changed` event of subtype `remove` or `rename` passed through and fired as a new emoji. It is `||` now.
- **Get Channel History** made a bare `conversations.history` call and discarded the result before the paginated loop. Removed.
- **Retrieve Thread Messages** and the AI **Get Thread Replies** called `conversations.replies` once, so threads longer than one page were truncated. Both now use `fetchAllThreadReplies` in `common/utils.ts`: 200 per page, deduped on `ts`, stops on a repeated cursor or after 50 pages, and returns `has_more` and `response_metadata.next_cursor` that agree (empty cursor when everything was read, the real cursor with `has_more: true` if the cap stopped the walk). The output schema declares the cursor, and both actions take an optional **Continue From Cursor** input (Advanced) that starts the walk there, so a capped run can be resumed. The AI descriptions state the 10,000-per-call ceiling and how to continue instead of promising every reply.
- **Ignore My Own Messages** on a Bot Token connection compared against the bot's id, because `getUserId` called `auth.test` with the bot token. It now uses the user token when one is configured, and the three DM triggers say the toggle needs one.
- **Upload File** said "Upload file without sharing it to a channel or user" while passing the Channel prop as `channel_id`, which shares it. The description now says "Uploads a file, optionally into a channel."

### Metadata

- Version 0.17.10 → 0.20.0 (main shipped 0.19.0 with Send Message To Multiple Users while this was open; merged); `minimumSupportedRelease` 0.86.4 → 0.88.2 (needed for `advanced`).
- `src/i18n/translation.json` regenerated from the built piece with the same extractor the CLI uses, so every rewritten string has a source key. No locale files were edited; the 82 orphaned keys in the seven live locales fall back to English until the nightly translation job is fixed.
- `aiMetadata`, `audience` and `classification` on every action are untouched.

### Deliberately left out

- **Set User Status → Expires At** stays a Number taking Unix seconds; a DateTime would be a prop type change and therefore a migration. The description now names the unit.
- **Update Profile → User ID** stays a text field rather than the User dropdown used elsewhere, for the same reason. It is labelled **User ID** because it takes an ID, which the checker reports as a label difference against the dropdowns' **User**.
- **Send Message To A Channel** still drops blocks, bot identity, broadcast and unfurl when a file is attached; the Attachment description now says so instead of changing `slackSendMessage`.
- The Custom API Call form itself comes from `pieces-common`; only its Slack-specific Use User Token toggle is touched.
- The 44 AI-only actions are hidden from the builder and keep their current copy.

No prop names or prop types change and the one new prop (Continue From Cursor) is optional, so steps saved on 0.17.10 keep working unchanged.

## How was this tested?

- Served the piece from `AP_DEV_PIECES` on a local CE build and captured a fresh step for all 42 human-facing actions and triggers at the default panel width, before and after; opened every after-shot and compared it with its before-shot (the artifact above is that set).
- Ran the step-form checker on the served metadata before and after: 56 errors / 74 warnings → 0 errors / 6 warnings. The `--before` comparison against the 0.17.10 metadata reports no prop type changed or removed. The six remaining warnings are label differences where the field means something different (User ID vs User, Status Text vs Message, Thread Timestamp vs Reply to Thread, File vs Attachment, Group Handle vs Handle) plus `followRedirects` casing from the shared Custom API Call helper.
- `turbo run build lint --filter=@activepieces/piece-slack`: 0 errors, 21 warnings, the same 21 as on `main` (unused `context` params in trigger `onDisable` hooks).
- `vitest run` in the piece: 178/178 across 8 suites, including the new `test/step-form-fixes.test.ts` (emoji guard, thread pagination with dedupe, repeated-cursor stop, the 50-page boundary and continuation from a cursor on both thread actions, New Mention enable guard).
- Edition paths: CE checked locally. The change is piece metadata and three trigger/action code paths with no edition-specific code, so EE and Cloud render the same forms.


packages/pieces/community/slack/src/lib/actions/update-user-groups.ts
```diff
@@ -20,18 +20,19 @@ export const updateGroupUsersAction = createAction({
...
+      placeholder: 'engineering',
       required: true,
     }),
     userIds: Property.Array({
       displayName: 'User IDs',
-      description: 'The list of Slack User IDs to update this group with. Leave empty to clear the group (if not appending).',
+      description: 'Slack user IDs to add. Empty with append off clears the group.',
       required: false,
     }),
     appendUsers: Property.Checkbox({
```
ibrahim-abuznaid: **This one can wipe a user group.** `required: true → false` while keeping `defaultValue: true` is not behaviour-neutral on the agent path.

Chain, all on `main`:
- `ee/agent/tools/piece-input-plan.ts:133` — `property.required ? described : described.nullable()`. Optional props become nullable in the agent's input schema, so a model can legitimately emit `appendUsers: null`.
- `piece-tool-runner.ts` executes that input directly. `fillDefaultsForMissingOptionalProps` has exactly three call sites (`ap-add-step`, `ap-update-step`, `flow-run-utils`), all under `mcp/tools/` — **the agent path never backfills defaults**.
- `engine/src/lib/variables/props-processor.ts:89` maps nil on an optional prop to `undefined`; line 156 then skips validation.
- `agent-tools.ts:335` resolves actions by name out of `piece.actions` with no audience filter, so this action is addable as an agent tool.

Result: `appendUsers` is `undefined` at line 60, the `group.users` merge is skipped, and `usergroups.users.update` is called with only the supplied IDs — a 40-person group collapses to whatever was passed. With `required: true` the same input failed loudly with `Expected boolean, received: undefined`.

Fix is local — own the default in `run()` rather than trusting `defaultValue`:
```ts
const appendUsers = context.propsValue.appendUsers ?? true;
```

packages/pieces/community/slack/src/lib/actions/send-message-action.ts
```diff
@@ -30,34 +30,38 @@ export const slackSendMessageAction = createAction({
     channel: slackChannel(true),
     text: Property.LongText({
       displayName: 'Message',
-      description: 'The text of your message. When using Block Kit blocks, this is used as a fallback for notifications.',
+      description: 'Slack mrkdwn is supported. Empty sends blocks only.',
       required: false,
     }),
     sendAsBot:Property.Checkbox({
```
ibrahim-abuznaid: Same nil path as `update-user-groups.ts:32` (details there), and this is the most-used action in the piece.

`sendAsBot` absent/null → `undefined` → line 72 takes the else branch → `requireUserToken()` throws `Missing user token, please re-authenticate` for every Bot-Token / custom-auth connection, which is a connection type that worked before and that the error message sends the user to pointlessly re-auth.

```ts
const sendAsBot = context.propsValue.sendAsBot ?? true;
```

Worth noting the other 16 checkbox flips in this PR are fine — their defaults are `false`, which is the same branch nil takes. Only the three with `defaultValue: true` invert.

packages/pieces/community/slack/src/lib/triggers/new-command.ts
```diff
@@ -56,8 +54,9 @@ export const newCommand = createTrigger({
       },
     }),
     ignoreBots: Property.Checkbox({
```
ibrahim-abuznaid, line 56: Narrow note, not a defect in this PR — recording it because this is where I first went looking.

Relaxing `required` on a trigger checkbox is not symmetric with doing it on an action. `fillDefaultsForMissingOptionalProps` (`mcp-utils.ts:506`) backfills non-required `CHECKBOX → defaultValue ?? false`, but it reads `piece.actions[actionName]` only, and `ap-update-trigger.ts` never calls it (just `diagnoseMissingTriggerInputs` at `:118`). So **triggers get no default backfill on any path**. A trigger configured through `ap_update_trigger` with `ignoreBots` omitted reaches `run()` as `undefined` and line 97 stops filtering bot messages.

Builder-created triggers are fine — `web/src/features/pieces/utils/form-utils.tsx` seeds checkbox defaults without gating on `required` — so this only bites MCP-authored flows. It is a gap in the MCP trigger tooling that predates you, and I will track it separately; nothing to change here.

Separately, and worth stating because it reads like a behaviour change in the diff but is not: dropping `commands.includes(parsedCommand.command)` at line ~105 is safe. `parseCommand` (`common/utils.ts:172`) already does `if (!validCommands.includes(command)) return null` on the lowercased command, so the outer guard was strictly redundant. Verified including the mixed-case path.

packages/pieces/community/slack/src/lib/actions/retrieve-thread-messages.ts
```diff
@@ -16,23 +16,44 @@ export const retrieveThreadMessages = createAction({
...
+    const firstPage = await client.conversations.replies({
       channel: propsValue.channel,
       ts: messageTimestamp,
+      limit: 200,
     });
+
+    const messages = [...(firstPage.messages ?? [])];
+    let cursor = firstPage.response_metadata?.next_cursor;
+
+    while (cursor) {
```
ibrahim-abuznaid: Good fix, three things to tighten.

1. **Parent dedupe.** The loop appends each page's `messages` wholesale. The thread parent is `messages[0]` on page 1; whether Slack repeats it on cursor pages is not documented either way and I could not settle it from the API reference. Please test against a >200-reply thread. A `ts` dedupe is cheap insurance regardless.

2. **No page cap.** `while (cursor)` has no bound and no guard against a repeated cursor. `conversations.replies` is Tier 3 (50+/min), so a very long thread starts sleeping on 429s against the 600 s step cap, and the accumulated array pushes at the 5 MB payload limit.

3. **Two AI actions still make the un-paginated call this PR just fixed here.** `get-thread-replies.action.ts:39` (`slack_get_thread_replies`, `audience: 'ai'`) is the exact same single `conversations.replies` call and shares `threadRepliesActionOutputSchema` — after this merges, the human action returns every reply and the agent path returns page one. `search-all.action.ts:32` has the same shape via `client.search.all({ count: 100 })`, though it is milder since it returns `message_count`/`file_count` so the truncation is at least visible. Worth fixing the first in this PR so the two thread actions do not diverge.

packages/pieces/community/slack/src/lib/actions/retrieve-thread-messages.ts
```diff
@@ -16,23 +16,44 @@ export const retrieveThreadMessages = createAction({
...
+        limit: 200,
+        cursor,
+      });
+      if (page.messages) {
+        messages.push(...page.messages);
+      }
+      cursor = page.response_metadata?.next_cursor;
+    }
+
+    return { ...firstPage, messages, has_more: false };
```
ibrahim-abuznaid: `{ ...firstPage, ... }` carries page 1's `response_metadata` through, so on a multi-page thread the output advertises `has_more: false` next to a `next_cursor` that still points at page 2. The two keys disagree, and `next_cursor` is visible in the raw output the builder shows.

Strip `response_metadata`, or blank the cursor when collapsing the pages.

packages/pieces/community/slack/src/lib/common/props.ts
```diff
@@ -64,48 +55,55 @@ export const slackChannel = <R extends boolean>(required: R) =>
...
 
 export const blocks = Property.Json({
-  displayName: 'Block Kit blocks',
-  description: 'See https://api.slack.com/block-kit for specs',
+  displayName: 'Block Kit Blocks',
+  description: 'JSON array of blocks from the Block Kit Builder.',
   required: false,
-  defaultValue: []
+  defaultValue: [],
+  advanced: true,
```
ibrahim-abuznaid, line 121: Heads-up on `advanced: true` for props that existing steps have already configured — a platform gap rather than something you introduced, but `blocks` is where it bites hardest.

`web/src/app/builder/piece-properties/advanced-section.tsx:36` is `useState(false)`, and the effect at `:40-57` auto-opens the section **only** when `errors` contains a matching path. A prop holding a valid non-default value never opens it, and the collapsed header shows just `Advanced — 7 options`, where the count is the number of advanced props, not the number that are set.

So a live blocks-only message step (`text` empty, `blocks` populated — which the Message description explicitly invites) now opens as an empty Message field with the Block Kit JSON folded away, reading as a broken step. Same for anyone who set Username, Profile Picture, Icon Emoji, Mention Origin Flow, Also Post to Channel, or turned Unfurl Links off.

Not a blocker here, but worth deciding before the pattern rolls across the catalog — either `blocks` stays out of Advanced, or `AdvancedSection` learns to auto-open when a hidden prop differs from its default.

packages/pieces/community/slack/src/lib/common/props.ts
```diff
@@ -2,24 +2,16 @@ import { MarkdownVariant, Property } from '@activepieces/pieces-framework';
...
-	  1. Type /invite in the channel's chat.
-	  2. Click on Add apps to this channel.
-	  3. Search for and add the bot.
-  `;
-
 export const multiSelectChannelInfo = Property.MarkDown({
   value:
-    slackChannelBotInstruction +
-    `\n**Note**: If you can't find the channel in the dropdown list (which fetches up to 2000 channels), please click on the **(F)** and type the channel ID directly in an array like this: \`{\`{ ['your_channel_id_1', 'your_channel_id_2', ...] \`}\`}`,
+    "Can't find the channel? Invite the bot: type **/invite** in the channel, choose **Add apps** and pick the bot. Or click **ƒ** and paste IDs as `{`{ ['C012AB3CD', 'C045EF6GH'] `}`}.",
```
ibrahim-abuznaid: The rewritten banner drops the old `(which fetches up to 2000 channels)` note, but `getChannels` still hard-caps at `CHANNELS_LIMIT = 2000` (line 243, enforced by the `do/while` at 263).

On a workspace above that cap the dropdown silently truncates, and the new copy attributes a missing channel solely to the bot not being in it — so the user runs `/invite`, sees no change, and files a bug. The `ƒ` escape hatch is still mentioned, so it is recoverable; just a clause short.

packages/pieces/community/slack/src/index.ts
```diff
@@ -100,7 +100,7 @@ export { slackAuth, slackOAuth2Auth } from './lib/auth';
 export const slack = createPiece({
   displayName: 'Slack',
   description: 'Channel-based messaging platform',
-  minimumSupportedRelease: '0.86.4',
+  minimumSupportedRelease: '0.88.2',
```
ibrahim-abuznaid, line 104: Checked this one properly rather than assuming, and it comes out in your favour: **0.88.2 is the correct floor, not a guess.** Tag `0.88.1` has neither `advanced` on `BasePropertySchema` nor `AdvancedSection`/`splitProps` in `piece-settings/index.tsx`; `0.88.2` has both. It also matches the floor the airtable and open-router bumps already use.

One consequence worth being deliberate about: `piece-cache-utils.ts:88` filters piece versions by server release, so a self-hoster in the `[0.86.4, 0.88.2)` window is pinned to Slack 0.17.10 and stops receiving future Slack releases until they upgrade the server. An older web build would have ignored the unknown `advanced` key and simply rendered the form unsplit, so the floor buys the feature rather than preventing breakage.

No change requested — just flagging the trade so it is a choice rather than a side effect.

packages/pieces/community/slack/src/lib/triggers/new-direct-message.ts
```diff
@@ -19,13 +19,15 @@ export const newDirectMessageTrigger = createTrigger({
...
+			displayName: 'Ignore Bot Messages',
+			description: 'Skip messages posted by bots and apps.',
+			required: false,
 			defaultValue: false,
 		}),
 		ignoreSelfMessages: Property.Checkbox({
-			displayName: 'Ignore Message from Yourself ?',
-			required: true,
+			displayName: 'Ignore My Own Messages',
+			description: 'Skip messages sent by the connected user.',
```
ibrahim-abuznaid: This description is accurate on OAuth2 but not on custom auth, and the old label (`Ignore Message from Yourself ?`) was vague enough to dodge the issue.

`common/auth-helpers.ts:87-109`: for OAuth2 `getUserId` returns `data.authed_user.id`, but for `CUSTOM_AUTH` it calls `auth.test` with `props.botToken` and returns `body.user_id` — the **bot's** id. So on a Bot-Token connection this toggle skips the bot's own messages, not the connected human's.

Pre-existing behaviour, not something this PR breaks. But the new copy states it more confidently than the code earns. Either hedge the wording or fix `getUserId` for the custom-auth branch. Same text added at `new-mention-in-direct-message.ts:29` and `new-command-in-direct-message.ts:38`.

packages/pieces/community/slack/src/lib/actions/add-reaction-to-message.ts
```diff
@@ -26,19 +26,21 @@ export const addRectionToMessageAction = createAction({
     ts: Property.ShortText({
       displayName: 'Message Timestamp',
```
ibrahim-abuznaid: The unified Message Timestamp prop is copy-pasted byte-for-byte into four files — here, `delete-message.ts:24`, `get-message.ts:22`, `update-message.ts:26` — while this PR is already editing `common/props.ts`, which owns every other shared prop.

The stated goal is one wording per concept, but the next copy edit has to touch four files and will drift again, which is the state this pass is fixing. Same for `replyBroadcast` (send-message / request-action-message) and `unfurlLinks` (send-message / send-direct-message); and `common/props.ts:181` already exports a `text` LongText that three send/update actions redeclare inline.

Exporting `messageTs` from `common/props.ts` would be a couple of lines and make the invariant structural.

packages/pieces/community/slack/src/lib/triggers/new-mention.ts
```diff
@@ -23,11 +24,15 @@ export const newMention = createTrigger({
     info: multiSelectChannelInfo,
     users: userIds,
     usergroups: usergroupIds,
+    targetInfo: Property.MarkDown({
```
ibrahim-abuznaid: Correct diagnosis — I verified it: with `users` and `usergroups` both empty, `.some()` on empty arrays is false and `run` returns `[]` at line ~110, so the trigger genuinely is dead.

Two notes on the remedy. The banner renders on every render, including after the user has correctly picked three people, so it becomes permanent furniture on a form that already carries a markdown block — which cuts against this PR's own "a form does not stack coloured boxes" rationale.

An `onEnable` that throws when both lists are empty would actually prevent the dead trigger instead of describing it, and would not occupy the form in the happy path.

ibrahim-abuznaid (commented):
> **Edited.** I re-verified my own findings by test rather than by reading, and three were wrong — including one I had led with as a blocker. This review has been revised in place; the withdrawn comments are gone. Nothing here blocks.

Reviewed against `e9c6c19` in a local worktree. Ran `vitest run` (28/28) and `tsc -p tsconfig.lib.json --noEmit` (clean) in the piece.

This is a genuinely strong pass — the copy work is the best I have seen on a form cleanup here, the four runtime fixes are real, and the description made the review much faster.

### Worth acting on

**i18n.** The locale files are keyed on the exact English runtime string (`framework/src/lib/i18n.ts:94`). Re-running the real extractor (`cli/src/lib/commands/generate-translation-file-for-piece.ts`) against the loaded piece module at both revisions: **78 keys that have translations today are orphaned**, out of 203 in `de.json` — 121 stay live and 4 were already stale before this PR. The fallback is graceful (`i18n.ts:95` only assigns on a hit, so a miss leaves English), so nothing renders broken; it just goes untranslated.

Two things that make this smaller than it first looks: it affects **7 locales, not 12** — `initializeI18n` iterates `LocalesEnum` (`core/utils/src/lib/locale.ts`), so only `de, es, fr, ja, nl, pt, zh` are ever loaded, and the `ca/hi/id/ru/vi` files in this piece are never read at all. The 78 is identical across the 7.

The description is right that convention is to let the nightly job regenerate. The snag is that `generate-translations.yml` has failed every run since 2026-08-16, so this will not self-heal on its own. Either regenerate `translation.json` + the 7 live locales here, or land this and track the nightly fix.

**Test coverage.** Three behaviour fixes land with no test, and `test/blocks-only.test.ts` is the only suite in the piece — `ci.yml:257` runs only `test/integration/ce/tool-search`, so it never gates anything either way. The `&&` → `||` emoji guard and the replies pagination are exactly what a unit test pins down, and the harness already exists.

Everything else is inline.

### Checked and cleared

Noting these so nobody re-reviews them:

- **The 19 checkboxes that go `required: true → false` are fine.** I initially called this a blocker and was wrong. Actions are covered on every path: `fillDefaultsForMissingOptionalProps` (`mcp-utils.ts:531`) backfills non-required `CHECKBOX → defaultValue ?? false`, the EE agent path routes through it (`piece-tool-runner.ts:7` → `flow-run-utils.ts:214`) *and* strips nulls first (`piece-input-filler.ts:68-74`), and the builder seeds defaults without gating on `required`. Every `defaultValue` in the diff is unchanged — the only `defaultValue` lines are formatting. The one asymmetry is triggers, noted inline on `new-command.ts`.
- **The "eleven checkboxes" in the description is the trigger-side subset** — accurate as far as it goes; the whole-diff count is 19. A word either way would settle it. Not an error.
- `advanced` is on `BasePropertySchema`, so it is honored for every prop type this PR puts it on (ShortText, Checkbox, Json), and it survives metadata storage and retrieval intact — nothing parses or strips it. It also reaches `useUserToken`, since `pieces/common/src/lib/helpers/index.ts:359` merges `extraProps` into the action's props.
- `MarkdownVariant.INFO` is a true no-op: `framework/src/lib/property/input/index.ts:133` already defaults to `variant: request.variant ?? MarkdownVariant.INFO`. (The `WARNING` on New Mention is of course a real change — that one is intended.)
- The New Mention warning is factually correct: with `users` and `usergroups` both empty, `.some()` on empty arrays is false and `run` returns `[]`.
- No `audience: 'ai'` action imports from `common/props`, so the `advanced: true` additions do not touch the AI surface.
- The emoji trigger `&&` → `||` is a correct fix.

### CI

The `main` check needs more than a rerun. The cancelled job you see is **attempt 2** of run `34612415389` — it never got a runner (zero steps). Attempt 1 (job `103305922370`) ran 20 steps and genuinely failed:

```
FAIL packages/server/api/test/integration/cloud/project-role/project-role.test.ts
  > should fail to delete a project role if project role does not exist
  AssertionError: expected 200 to be 404
```

Separately, main was broken from 09-13 14:10 by #15500's `waitpoint-test-flow-cap.test.ts` hitting `127.0.0.1:3000`, until #15511 fixed it at 09-14 07:09 — that started after this run, so it is not the cause here, but it does mean a rerun would reuse a stale merge commit. Please update the branch so you get a fresh run on current main; that picks up #15511 and settles whether the project-role failure was a flake.

ibrahim-abuznaid (approved):

PR #15483, opened 2026-09-12
feat(airtable): clearer setup forms and working record lookups

Talaljaber:
Before / after gallery: https://claude.ai/code/artifact/1dea19e3-8287-448a-9d0d-a06fad4c0011 (every human-visible action and trigger, fresh step, default 400 px panel, Advanced collapsed)

## Description

Step-form pass on the Airtable piece (0.7.2 → 0.9.0): every human-visible action and trigger gets copy that fits the builder, one wording per shared prop, three secondary fields tucked into Advanced, and four dropdown/lookup bugs fixed along the way.

### Labels
- `Table` → `Fields` on the dynamic field editor (Create / Update / Clean Record); it was mislabeled.
- `Comment Text` → `Comment`, `Base Name or Keyword` → `Base Name`, `Trigger field` → `Trigger Field`.
- Action display names are unchanged.

### Descriptions
- Every description is now one plain sentence of at most 70 characters, so nothing folds behind "show more" and no markdown markers show literally (Upload File, Add Comment, Create Base, Create Table, the New or Updated Record trigger).
- Examples moved into placeholders where the input renders one: `recXXXXXXXXXXXXXX`, `comXXXXXXXXXXXXXX`, `image/png`, `e.g. jane@example.com`, `e.g. CRM`, `e.g. Contacts`, and `2024-03-05` on dynamic date fields.
- Descriptions that only restated the label (Base Name, Table Name) were removed.
- Action descriptions rewritten to say what the step does (`Adds a new record to a table.`, `Returns up to 1000 records whose field contains a value.` (with the Search Value described as case-sensitive, which `FIND()` is), `Updates a record and empties every field you leave blank.`, `Lists every table in a base with its fields and views.`).

### One wording per concept
- `Record ID` carries the same description and placeholder on every action that uses it.
- `View` explains its effect once: `Only records in this view are used. Empty: the whole table.`
- Every dropdown uses the same placeholders: `Connect your Airtable account first`, `Select a base first`, `Select a table first`, and `Could not load. Check the token's scopes.` when the fetch fails (Workspace, View, Record and Search Field dropdowns now catch errors instead of throwing).

### Advanced sections
- Add Comment to Record: `Parent Comment ID`.
- Create Table: `Description`.
- Upload File to Column: `File Name`.
- All three are optional with a neutral default; nothing required is hidden. `View` stays on the main form because it changes what a step reads.

### Banners
- None added. The piece had no `Property.MarkDown` and none of the forms needed one.

### Runtime fixes
- `src/lib/trigger/update-record.trigger.ts`: the Trigger Field dropdown sent the connection object as the bearer token, so it never loaded. It now sends the token and tells you when the table has no Last Modified Time field.
- `src/lib/actions/upload-file-to-column.ts`: the Attachment Column dropdown had the same broken token.
- `src/lib/common/index.ts` `findRecord`: follows Airtable's `offset` cursor so matches past the first page are returned (capped at 1000 records, 100 per request, 3 retries for 5xx and transport errors, plus a 30-second pause and retry on HTTP 429), and escapes the search value (`\` and `"`) before building the `FIND()` formula. A field name containing `{` or `}` is rejected with a clear error, the same rule Search Records already applies, because Airtable cannot escape braces inside a field reference.
- `src/lib/common/index.ts` `listRecords`: the Record dropdown pages up to 500 records instead of the first 50, with the same retries and a single 429 retry so the picker never blocks for more than 30 seconds.
- Non-2xx responses already throw `HttpError` from the shared http client, so the loops carry no status guard of their own.
- The empty `view=` query parameter is no longer sent when no view is selected, in `findRecord` and in the New or Updated Record poll alike. Hygiene for consistency, not a bug fix.
- Dynamic date and date-time fields show ISO placeholders (`2024-03-05`, `2024-03-05T10:30:00Z`), which Airtable accepts on Create, Update and Clean Record alike; the old human-readable hint only parsed on Create, which sends `typecast`.
- `src/lib/common/index.ts` `updateRecord`: Update Record and Clean Record now send `typecast: true` like Create Record, so dates typed as text, new select options and linked-record names are converted instead of failing with `INVALID_VALUE_FOR_COLUMN`. Called out in the breaking-changes entry.
- Output shapes are unchanged.

### Metadata
- `package.json` 0.7.2 → 0.9.0 (label renames; 0.8.0 was taken on main by a version-only bump while this PR was open, and a registered version never republishes).
- `minimumSupportedRelease` 0.86.4 → 0.88.2 because `advanced` is used.
- `src/i18n/translation.json` (the Crowdin source) regenerated from the built piece.

### Deliberately left out
- Action display names such as `Create Airtable Record` (renaming changes step titles and orphans Crowdin strings).
- `fetchTable`'s non-null assertion, the `DATETIME_PARSE` format string and the first-run 24 h lookback on the updated-record trigger: behaviour changes, not UI.
- `Property.Json` for tables/fields on Create Base and Create Table: a structured editor would be a type change that breaks saved steps.
- AI-only actions. Translated locale files under `src/i18n/*` are Crowdin-generated and untouched; `translation.json` (the Crowdin source) is regenerated from the built piece, so the new strings show in English until the next Crowdin sync.
- The Custom API Call card in the gallery differs before/after only because the served shared helper is newer than the database copy in the "before" run; this PR does not touch it.

## How was this tested?

- `check-step-form.mjs`: 8 errors / 9 warnings → 0 errors / 3 warnings, no `type-changed` or `prop-removed`. Remaining warnings: Custom API Call label casing (shared helper), `Record ID` vs `Record` (text input vs dropdown, different widgets), `Base Name` vs `Table Name` (different entities).
- `eslint src/**/*.ts`: 0 errors, 17 warnings before and after (all pre-existing `no-explicit-any`).
- `tsc --noEmit -p tsconfig.lib.json`: passes. The piece has no unit tests.
- Every form captured before and after on the local CE dev server at the default panel width with the piece served from source (gallery linked above). Tallest form, Upload File to Column: 1129 px → 1061 px. The last review round only reworded Search Value's description; that form was not recaptured.
- The final checker run used metadata dumped from the freshly compiled dist (same shape as the served JSON) because the dev server was down at the time.
- No prop names, types or behaviour-changing defaults changed; saved steps keep working.
- Functional change called out in `docs/install/reference/breaking-changes.mdx`: Find Record returns up to 1000 matches instead of the first 100, the Record dropdown lists up to 500 records instead of 50, and Update and Clean Record send typecast.

Fixes # (issue)


packages/pieces/community/airtable/src/lib/common/index.ts
```diff
@@ -614,12 +665,14 @@ export const airtableCommon = {
...
           const params = {
             displayName: field.name,
-            description: ['date', 'dateTime'].includes(field.type)
-              ? `${
-                  field.description ? field.description : ''
-                }Expected format: mmmm d,yyyy`
-              : field.description,
+            description: field.description,
             required: false,
+            ...(field.type === 'date'
```
ibrahim-abuznaid, line 705: **This placeholder is wrong on two of the three actions that show it.**

`createRecord` sends `typecast: true` (index.ts:214), so `March 5, 2024` parses there. `updateRecord` does **not** send typecast, and both **Update Record** and **Clean Record** route through it. `createNewFields` passes date values through untouched, so Airtable receives a human-readable string on a date field with no typecast and rejects it with `INVALID_VALUE_FOR_COLUMN`.

This shared `fields` property feeds all three actions, so one placeholder is serving three different backends.

Two ways out, either is fine:
- add `typecast: true` to the `updateRecord` body, which also makes Update/Clean behave like Create for select and linked-record fields, or
- make the placeholder ISO (`2024-03-05` / `2024-03-05T10:30:00Z`), which works everywhere without a behaviour change.

The old description carried the same wrong hint, so this is inherited rather than introduced. A form-clarity pass is the right place to close it.

packages/pieces/community/airtable/src/lib/trigger/update-record.trigger.ts
```diff
@@ -20,49 +21,70 @@ import { updatedRecordTriggerOutputSchema } from '../output-schemas';
...
+        console.debug(e);
+        return {
+          disabled: true,
+          options: [],
+          placeholder: "Could not load. Check the token's scopes.",
+        };
+      }
     },
   }),
   viewId: airtableCommon.views,
```
ibrahim-abuznaid, line 90: **The empty `view` parameter is still here.** (anchoring on the prop; the line is `update-record.trigger.ts:112`.)

The poll body still does:

```ts
view: propsValue.viewId ?? '',
```

That's the exact pattern removed from `findRecord`. The airtable.js serializer does not drop empty strings (`object_to_query_param_string.js` maps only `null`/`undefined` to `''` and always pushes the pair), so the request carries `view=` whenever no view is selected.

So one of two things is true, and it'd be good to say which in the description:
- the empty param is harmful, in which case the trigger has the same bug and this is a one-line fix in a file the PR already edits, or
- it's harmless, in which case the `findRecord` change is hygiene rather than the bug fix the description claims.

Suggested, to match `findRecord`:

```ts
...(propsValue.viewId ? { view: propsValue.viewId } : {}),
```

packages/pieces/community/airtable/src/lib/common/index.ts
```diff
@@ -214,28 +232,40 @@ async function findRecord({
...
+  do {
+    const request: HttpRequest = {
+      method: HttpMethod.GET,
+      url: `https://api.airtable.com/v0/${baseId}/${tableId}`,
+      authentication: {
+        type: AuthenticationType.BEARER_TOKEN,
+        token,
+      },
+      queryParams: {
+        filterByFormula: `FIND("${escapedSearchValue}",{${searchField}})`,
```
ibrahim-abuznaid, line 284: **The escaping is half done.** The search *value* is escaped, but `searchField` is interpolated raw into `{...}`.

`searchField` comes from the `fieldNames` dropdown, so its value is a real Airtable field name, and Airtable field names can contain `}` and `\`. Either one produces a malformed formula and a 422 that reads as "no records found" to the user.

Since this line is already being hardened, worth closing the other half:

```ts
const escapedSearchField = searchField.replace(/\\/g, '\\\\').replace(/\}/g, '\\}');
```

packages/pieces/community/airtable/src/lib/common/index.ts
```diff
@@ -102,28 +104,44 @@ async function listRecords({
...
+        ...(offset ? { offset } : {}),
+      },
+    };
+    const response = await httpClient.sendRequest<{
+      records: AirtableRecord[];
+      offset?: string;
+    }>(request);
+
+    if (response.status !== 200) {
+      throw new Error(`Airtable returned ${response.status} while listing records`);
```
ibrahim-abuznaid: **This guard is unreachable.** `httpClient` is `FetchHttpClient`, which throws `HttpError` for any status outside 200-299 *before* returning (`packages/pieces/common/src/lib/http/core/fetch-http-client.ts`, the `successCeiling` check). `sendRequest` never resolves with a non-2xx status.

So the description's "both record-listing loops throw on a non-200 page instead of returning the rows collected so far" was already true via `HttpError`, and the pre-existing `return []` fallbacks this replaces were dead too.

No harm in leaving it, but if it ever *did* fire, this generic message would replace Airtable's actual error body, which is the useful part. I'd drop both guards (here and at :262) and let `HttpError` through.

packages/pieces/community/airtable/src/lib/common/index.ts
```diff
@@ -26,6 +26,8 @@ import {
 import { isNil } from '@activepieces/pieces-framework';
 import { airtableAuth } from '../auth';
 
+const MAX_FIND_RECORDS = 1000;
```
ibrahim-abuznaid, line 30: **New 429 exposure worth a thought.** `findRecord` doesn't pass `pageSize`, so Airtable's default of 100 applies and 1000 records means up to 10 sequential requests. The Record dropdown is 1 `fetchTable` plus up to 5 `listRecords` calls every time it opens.

Airtable allows 5 requests per second per base, and `FetchHttpClient` defaults to `retries ?? 0` with no 429 handling. A rate limit now surfaces as a hard step failure where the old single-request version couldn't hit one.

Not a blocker, and the caps are sensible. But `pageSize: '100'` on the `findRecord` request would halve the round trips for free, and a `retries` value on these two loops would absorb the common case.

packages/pieces/community/airtable/src/lib/actions/find-record.ts
```diff
@@ -24,6 +24,8 @@ export const airtableFindRecordAction = createAction({
     searchValue: Property.ShortText({
       displayName: 'Search Value',
       required: true,
+      description: 'Records whose search field contains this text are returned.',
```
ibrahim-abuznaid: **`FIND()` is case-sensitive, and this wording doesn't say so.**

With `e.g. jane@example.com` as the placeholder right below, a user searching `jane@example.com` against a stored `Jane@Example.com` gets zero rows and no explanation. This is the thing people actually get wrong with this action.

The piece already documents it correctly on the AI side, which says the match is a "case-sensitive substring match". The human form is now less precise than the agent one on the same behaviour.

Something like: `Case-sensitive. Records whose search field contains this text are returned.`

ibrahim-abuznaid (commented):
Reviewed the full diff against the PR head and verified the claims locally. The two dropdown fixes are the real value here: on `main` both the Attachment Column and Trigger Field dropdowns passed the connection object as the bearer token, so they sent `Bearer [object Object]` and never loaded. A `SecretText` connection value is `{ secret_text: string }` (`packages/core/shared/src/lib/automation/app-connection/app-connection.ts:33`), so those two were broken for every user. Good catch.

Six inline notes below. Only the first one I'd want resolved before merge; the rest are cheap follow-ups or wording.

One thing that isn't line-anchored: **the result cap change is a functional change**, and the breaking-change box is ticked "no". `findRecord` goes from 100 records (Airtable's default `pageSize` when the param is omitted) to 1000, and `listRecords` from 50 to 500. The repo's own checkbox wording covers a "default/limit/behaviour change". A flow that iterates the Find Record output or reads `records.length` will see different results after this ships. Worth reclassifying, or at least calling out in the release notes.

### Verified, not taken on trust

- `tsc --noEmit -p tsconfig.lib.json` passes.
- `eslint` gives 0 errors and 17 warnings, all pre-existing `no-explicit-any`.
- All CI checks green.
- `translation.json` is a clean regeneration: 26 keys removed, 34 added, every one traceable to a source change. The `Response is Binary` / `Return Error as Output` / `Timeout` churn is drift from the shared Custom API Call helper on `main`, as the description says.
- `minimumSupportedRelease` at `0.88.2` matches `google-sheets`, the other piece using `advanced`.
- All three Advanced fields are `required: false`; nothing required is hidden.
- No code comments added, two removed.
- `find-table`'s "Case does not matter" is accurate (`toLowerCase()` on both sides).
- Upload's "Empty: the uploaded file's own name" is accurate (`filename ?? fileInput.filename`).

Minor process note: the before/after gallery is a `claude.ai` artifact link. External maintainers on this repo probably can't open it.

ibrahim-abuznaid:
Talal — three things, and two of them are corrections to my own review. Sorry for the noise.

### Two of my comments were wrong

1. **Brace escaping in the field reference.** I asked you to escape `}`. Airtable has no escape for braces inside `{Field}`, so there is nothing to escape with. Rejecting field names that contain braces — which is what you did — is the right call. Ignore my comment.
2. **`pageSize: '100'`.** I said it would halve the round trips. It does nothing: 100 is both Airtable's default and its maximum. Harmless to keep it explicit, but it is not the win I claimed.

Your commits fixed the other four findings correctly.

### One finding was real and still open, so I fixed it on your branch

`retries: 3` does not cover HTTP 429. `sendWithRetries` in `packages/pieces/common/src/lib/http/core/fetch-http-client.ts` retries only when the status is 500 or above, or when `fetch` itself throws. A 429 is neither, so it returns and the caller gets an `HttpError` on the first hit.

That matters more after this PR, not less. Find Record now reads up to ten pages where it used to read one, against Airtable's limit of 5 requests per second per base, and a 429 on page four discards the three pages already read.

I pushed a `sendPage` helper into `common/index.ts`. It retries only on 429, waits 30 seconds between attempts (the cool-down Airtable documents for this limit — `HttpError` carries no headers, so `Retry-After` cannot be read), and rethrows every other status untouched. `retries: 3` stays and still covers 5xx and transport errors. The action path retries twice, for at most 60 seconds inside a step's 600-second budget; the Record-picker dropdown retries once, because a dropdown that blocks for a minute is worse than one that reports a failure.

While there: the dropdown's catch returns `"Could not load. Check the token's scopes."` for every failure, which sends people to the wrong place when the real cause is a rate limit. Pre-existing, and I left it alone.

### Two housekeeping fixes, also pushed

- Your `breaking-changes.mdx` entry collided with the worker-group entry that landed on main. I merged main in and kept both, in that order. The PR is mergeable again.
- The breaking-change check was red because the docs entry, the label and the template box disagreed. Your entry describes a real limit change (100 → 1000 records returned, 50 → 500 in the picker), so I ticked `yes — functional` and applied the `⛓️‍💥 breaking-change` label. Rule R3 in `tools/scripts/breaking-change-check.ts` needs all three to agree, so correcting only one of them keeps it red. Say so if you disagree with the classification and I will revert it.

### One request

You resolved Greptile's threads but left all six of mine open with no reply. I could not tell which findings you had accepted and which you had rejected without diffing your five commits myself — and on the brace one you were right and I was wrong, which a one-line reply would have surfaced straight away. A short reply on the thread is enough; you do not have to resolve them.

### Separate from this PR

`createRecord` sends `typecast: true` and `updateRecord` does not, so Update Record and Clean Record treat date, select and linked-record fields differently from Create Record. Pre-existing, not something you introduced — flagging it so it does not get lost.

ibrahim-abuznaid (approved):

PR #15435, opened 2026-09-10
feat(linear): filter Updated Issue trigger by changed fields

AdamSelene:
## Description

The Linear **Updated Issue** trigger is `TriggerStrategy.WEBHOOK`, and Linear's "Issue" webhook fires on **every** field change (and, when no team is set, over `allPublicTeams` — the whole workspace). This makes a flow on that trigger run on any edit, most of them irrelevant (`updatedFrom` is often just `updatedAt`/`sortOrder`).

This adds an optional **"Only when these fields change"** multi-select to the trigger. When one or more fields are selected, `run()` keeps the event only if at least one of them appears in Linear's `updatedFrom`; otherwise it returns `[]` and no flow run is created. Leaving it empty preserves today's behaviour (fires on every update), so existing flows are unaffected until a user opts in.

Selectable fields: Status, Assignee, Priority, Title, Description, Labels, Estimate, Due date, Project, Cycle, Parent. Noise keys (`updatedAt`, `sortOrder`) are deliberately not offered.

Follows the existing `updatedFrom` filter precedent in `updated-project.ts`.

## How was this tested?

- `npx turbo run lint --filter=@activepieces/piece-linear` — passes (0 errors).
- Unit check of the `run()` filter logic: noise-only change with a Status filter → dropped; Status change with a Status filter → kept; empty selection → always kept on update; non-update action → dropped.
- Piece-only change; no server/edition-specific paths touched.

Closes GIT-1874

### Breaking change?  (required — CI fails if this is left unedited)

- [x] no — reviewed, not breaking
- [ ] yes — technical (removed/renamed API field or endpoint, dropped column, new required field, removed/required env var)
- [ ] yes — functional (default/limit/behaviour change, new self-hosted setup step)

### Security impact?  (required — CI fails if this is left unedited)

- [x] no — reviewed, no security impact
- [ ] yes — security-sensitive (call out the risk and mitigation in the description above)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15505, opened 2026-09-13
feat(linear): add Team to Updated Issue changed-fields filter

AdamSelene:
## Description

The changed-fields filter added to the **Updated Issue** trigger in #15435 lets a flow trigger only when selected fields change. It covered Status, Assignee, Priority, Title, Description, Labels, Estimate, Due date, Project, Cycle and Parent — but not **Team**.

Linear's webhook `updatedFrom` includes `teamId` when an issue is moved to another team, and the trigger's `run()` already matches selected values generically against the keys present in `updatedFrom` — so this only needs the option added to the dropdown (no logic change). It lets re-triage / team-routing automations fire only on team changes instead of on every issue update.

## How was this tested?

- Added a unit test in `updated-issue.test.ts` asserting a `teamId` change is kept and a non-team change is dropped when `Team` is selected; the full trigger suite (7 tests) passes with `vitest`.
- `lint` passes with no new warnings.
- Piece version bumped `0.6.0` → `0.6.1`.

Fixes # (n/a — small follow-up to #15435)


OdaiAhmed99 (commented):
Verified live against a real Linear workspace, not only the unit tests.

The claim holds. A real webhook delivery for an actual team move (issue moved from team e4 to team Test) carries teamId in updatedFrom:


"updatedFrom": {
  "updatedAt": "2026-09-15T17:08:30.151Z",
  "number": 4,
  "sortOrder": -73,
  "addedToTeamAt": "2026-09-15T17:08:30.203Z",
  "teamId": "f910d563-...",
  "previousIdentifiers": [],
  "stateId": "9bd0c6a6-..."
},
"data": { "teamId": "25a6a33a-...", "team": { "key": "TES" } }
updatedFrom.teamId is the team the issue left and data.teamId where it landed, which is the right shape for a changed-fields filter.

Both directions checked against real payloads. I captured two genuine deliveries — the team move above, and a title rename whose updatedFrom is { updatedAt, title } — and ran both through the shipped run(). With Team selected the move is kept and the rename is dropped. Piece suite 12/12.

One thing worth knowing, pre-existing and not introduced here: a team move also remaps the issue into the destination team's workflow, so stateId rides along in the same updatedFrom. A flow filtering on Status alone therefore already fires on team moves today — which is an argument in favour of this option, since people were reaching for Status and silently picking up moves. I've noted it in the piece-authoring wiki rather than asking for anything here.

Diff is one dropdown entry with a value matching the existing key style, no logic change needed, correct patch bump. LGTM.

OdaiAhmed99 (approved):

PR #15520, opened 2026-09-14
fix(mcp): unblock the OAuth consent flow for new and returning users

moaaz-ae:
## Description

We are now sending people from the website straight to "connect your MCP client", so for many new users the OAuth consent page is the first Activepieces screen they see. Walking that flow through every account state found several places where it dead-ends before the client ever gets a token:

- A visitor who started signing up but never entered a name (onboarding token) got the consent page with an empty project list and Authorize disabled forever. They are now sent to finish signing up and come back to consent.
- Deny was `history.back()`, which does nothing in the tab the client opened, so the client waited forever. `POST /v1/mcp-oauth/deny` now verifies the request and returns the client's redirect with `error=access_denied`.
- An expired request (10-minute window) showed "Authorization failed. Please try again", which can never succeed. It now says to go back to the client and start again, and disables Authorize.
- A browser already signed into another account authorized it silently. Consent now shows who is signed in, with a Switch account link.
- Password sign-up lost the `from=` on `/verify-email`, so after clicking the link the user landed on a bare sign-in and the client was orphaned. The redirect is remembered in browser storage and honoured after verification.
- A single project is preselected instead of requiring the dropdown.

## How was this tested?

Cloud edition locally, driving a simulated MCP client (dynamic registration, PKCE, loopback callback) through each state: no account via email code, half-finished signup, signed in, signed in as another account, expired request, deny. Every path ends with the client holding a token and `tools/list` answering, or receiving the right error. CE MCP OAuth API tests pass.

Touches the OAuth consent flow. The browser never trusts the `authRequestId` payload for anything but display text; Deny resolves the client redirect on the server after verifying the signed request, same as Approve.


packages/web/src/lib/navigation-utils.tsx
```diff
@@ -31,10 +33,27 @@ export const useDefaultRedirectPath = () => {
   return '/';
 };
 
+export const pendingRedirect = {
+  remember(from: string | null) {
+    if (from) {
+      ApStorage.getInstance().setItem(PENDING_REDIRECT_KEY, from);
+    }
+  },
```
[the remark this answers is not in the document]
moaaz-ae: Done in 0ba31ed: `remember` now clears the stored value when there is no `from`, so a later signup cannot inherit a redirect from an abandoned one.

packages/web/src/app/routes/mcp-authorize/index.tsx
```diff
@@ -23,21 +24,37 @@ import { Separator } from '@/components/ui/separator';
...
+    decodeJwtPayload(authRequestId);
   const [selectedProjectId, setSelectedProjectId] = useState<
     string | undefined
   >(undefined);
   const [searchValue, setSearchValue] = useState('');
   const [selectedTypes, setSelectedTypes] = useState<string[]>([]);
   const [authorized, setAuthorized] = useState(false);
+  const [requestExpired, setRequestExpired] = useState(
+    expiresAt !== null && expiresAt * 1000 <= Date.now(),
+  );
```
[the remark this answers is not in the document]
moaaz-ae: Done in 0ba31ed: a timer flips `requestExpired` when `exp` is reached while the page is open, and both Authorize and Deny are disabled once expired, so no request that can only 400 is sent.

packages/web/src/app/routes/mcp-authorize/index.tsx
```diff
@@ -76,11 +127,14 @@ function McpAuthorizePage() {
...
-    const returnUrl = `/mcp-authorize?${searchParams.toString()}`;
-    const loginParams = new URLSearchParams({ from: returnUrl });
-    return <Navigate to={`/sign-in?${loginParams.toString()}`} replace />;
+    return <Navigate to={signInPath} replace />;
   }
 
+  const switchAccount = () => {
+    authenticationSession.clearSession();
+    navigate(signInPath);
+  };
```
AhmadTash, line 139: Blocking: this leaves the previous account's data in the query cache.

`clearSession()` removes two `ApStorage` keys and nothing else; `queryClient` is a module singleton that survives a client-side `navigate`. `logOut()` gets away with the same call only because it assigns `window.location.href` and throws the page away.

Two consequences once B signs in and is redirected back here:

1. `['mcp-authorize-projects', searchValue, selectedTypes]` isn't scoped by user or platform, so A's project names render to whoever is now at the keyboard — which is the one flow where that person is deliberately not A.
2. Worse, the new preselect effect fires against A's cached data and sets `selectedProjectId` to A's project. The refetch replaces `options` but never clears that id (the effect is guarded by `!selectedProjectId`), so the select renders blank while Authorize stays enabled. Clicking it hits the server's `access_denied` 403 → "Authorization failed. Please try again." — the exact dead end this PR is closing.

```suggestion
  const switchAccount = () => {
    authenticationSession.clearSession();
    window.location.href = signInPath;
  };
```

That also drops the `useNavigate` import. Per-call-site `resetQueries` is the version that rots.

packages/web/src/features/authentication/components/sign-up-form.tsx
```diff
@@ -98,6 +102,10 @@ const SignUpForm = ({
   const redirectAfterLogin = useRedirectAfterLogin();
   const navigate = useNavigate();
   const { capture } = useTelemetry();
+  const askToCheckEmail = () => {
+    pendingRedirect.remember(searchParams.get(FROM_QUERY_PARAM));
+    setShowCheckYourEmailNote(true);
+  };
```
AhmadTash, line 108: Blocking: the sibling caller still drops the redirect.

`sign-in-form.tsx:101-103` sets the same note on `EMAIL_IS_NOT_VERIFIED` without remembering anything:

```ts
case ErrorCode.EMAIL_IS_NOT_VERIFIED: {
  setShowCheckYourEmailNote(true);
  break;
}
```

And `/sign-in?from=…` opens the drawer in `signin` mode, so the returning-but-unverified user — the likeliest person to arrive through a consent link, since they already started once — goes down the unpatched branch: verify → bare `/sign-in`, client orphaned. Same fix there (`pendingRedirect.remember(searchParams.get(FROM_QUERY_PARAM))` before the note), which is what makes this the root-cause fix rather than the sign-up half of it.

AhmadTash (commented):
Walked every state this touches against `main`. The shape is right and the server side is sound — deny resolving the redirect from the verified JWT, the `verifyAuthRequest` extraction keeping the type check on both paths, the ONBOARDING gate lining up with `auth-landing.tsx:19` / `auth-drawer-body.tsx:89` so the round trip really does land back on consent. Two things reintroduce the same dead end the PR is closing, both one-liners; inline.

Smaller ones, not blocking:

- **Client clock gates Authorize.** `requestExpired` is seeded from `expiresAt * 1000 <= Date.now()` and feeds `disabled` on both buttons, so a browser running more than 10 minutes fast gets a dead page on a request the server would have accepted — strictly worse than before, where the click reached the server. `markExpiredOnBadRequest` already turns the server's 400 into the right copy, so dropping `requestExpired` from the two `disabled` props keeps the message and loses the false positive.
- **No test for `/v1/mcp-oauth/deny`.** There are 14 `mcp-oauth-*.test.ts` files under `test/integration/ce/mcp/`, and `mcp-platform-project-selection.test.ts:37` already has the approve round-trip to copy. Valid request → redirect carries `error=access_denied` and `state`; tampered or expired → 400.
- **`userHooks.useUserById`** (`hooks/user-hooks.ts:39`) is the same plain `useQuery` with `enabled: !isNil(id)` and `staleTime: Infinity`, and swallows errors — drops the bespoke key, the `currentUserId ?? ''`, and keeps a `/v1/users/:id` failure out of the global `QueryCache.onError` reporter. (Avoiding `useCurrentUser` is right, it's `useSuspenseQuery`.)
- `markExpiredOnBadRequest` reads *every* 400 as expired. Correct today only because approve's sole 400 is the invalid/expired request — access checks return 403.

Nice catch on the `exp` claim actually being there (`AUTH_REQUEST_TTL_10_MINUTES_SECONDS`), and the brain gotchas landed in the right place.

packages/web/src/app/routes/mcp-authorize/index.tsx
moaaz-ae: Done in c93f0649dd: `switchAccount` now assigns `window.location.href`, so the query cache goes with the page. `useNavigate` is gone.

packages/web/src/features/authentication/components/sign-up-form.tsx
moaaz-ae: Done in c93f0649dd: `SignInForm` calls `pendingRedirect.remember(searchParams.get(FROM_QUERY_PARAM))` on `EMAIL_IS_NOT_VERIFIED`, same as sign-up.

AhmadTash (approved):

PR #15534, opened 2026-09-14
feat(analytics): stamp marketing attribution on the PostHog person at sign-up

moaaz-ae:
## Description

Carries the marketing attribution a visitor arrived with (`utm_*`, `gclid`, `fbclid`, `ref`, `ap_cta`, `ap_landing`, `ap_referrer`, `ap_sid`) into the sign-up request and stamps it once on the PostHog person from the server, together with the sign-up method (`password`, `email_code`, `google`).

### Why

The web client already stashes these parameters and sends them to PostHog with `$set_once` from the browser on every identify. That is lost when the Google redirect drops state, when an ad blocker eats the call, or when the tab closes early, and it repeats on every login for every user. The server knows the exact moment a user is created, so the stamp now happens there, once.

The same signal fixes an audit bug: the Google claim endpoint emitted `USER_SIGNED_UP` on every login, not only the first.

### How

- `@activepieces/shared`: `AttributionParams` contract and `SignUpMethod`; optional `attribution` on `SignUpRequest`, `VerifyEmailCodeRequest`, `CompleteSignUpRequest` and `ClaimTokenRequest`.
- Web: stash the parameters at boot (before PostHog initialises), keep them in `localStorage` for 30 days, send them with the four sign-up requests, and stop sending them from the browser identify. Identity reset on logout lives in `authenticationSession.logOut` instead of three call sites. The in-app Contact Sales link carries `ap_cta` and the PostHog session id so the sales form on the marketing site can join the two.
- Server: `getOrCreateWithProject` returns `{ user, created }`; `provisionOrOnboard`, `signUp`, `federatedAuthn`, `verifyCode` and `completeSignUp` return `{ response, signedUp }`. The password, email-code and Google controllers call `telemetry.identifySignUp` only when `signedUp` is true. It writes `signup_method` and the attribution as `$set_once`, so later logins never overwrite them. The Google `USER_SIGNED_UP` audit event fires only when the claim created the user.
- SAML users are not stamped: they are provisioned by their platform admin and never arrive through a campaign link.

No database changes.

## How was this tested?

- Typecheck for `shared`, `api` and `web`; eslint on every changed file.
- New unit test `telemetry-identify-sign-up.test.ts`: stamps method and attribution as `$set_once`, stamps only the method when no attribution was sent, stays silent when the platform has product telemetry off.
- `acquisition-utils.test.ts` (13 cases) covers the stash, TTL expiry, malformed storage and the new keys.
- Not exercised against a running Cloud instance in this round; the automated coverage above is what was run.

Fixes # (issue)

Risk: the change is inside the authentication services and controllers. Mitigation: no token issuance, verification or authorisation logic changes; the services return the same `AuthenticationResponse` wrapped with a boolean, the new request field is optional and zod-validated (trimmed strings, max 300 chars), and its only sink is a PostHog `$set_once` write.


AhmadTash:
The approach holds up — moving the stamp server-side, gating it on a real `created`/`provisioned` flag rather than on "a response came back", and writing `$set_once` are all the right calls, and gating the Google `USER_SIGNED_UP` audit event on `signedUp` is a genuine bug fix. Both new test suites pass locally (13 web + 3 server), `tsc --noEmit` is clean on `web`, and eslint is clean on every changed file.

Three things I think need fixing before this merges.

## 1. Password sign-ups on Cloud are never stamped

The attribution is attached to a request that discards it, and the request where the account is actually created never receives it. On Cloud production `signup_method: 'password'` effectively never appears.

On Cloud a password sign-up does not create the account at `/sign-up`. `authentication.service.ts:68` creates the identity with `verified: hasInvitations || isFederatedProvider` — `false` for a fresh email — so `provisionOrOnboard` takes the `!identity.verified` arm and returns `signedUp: false`. The gate at `authentication.controller.ts:37` is `signedUp && !isNil(response.platformId)`, so it skips `identifySignUp`, and `attribution` has already been destructured off the body.

The account is really created two steps later:

1. `POST /sign-up` → onboarding response, `signedUp: false`, attribution dropped
2. verification email link → `/verify-email` → redirects to `/sign-in`
3. `POST /sign-in` → `signInWithPassword` → `provisionOrOnboard` → `createPlatformWithProject` → `provisioned: true`

Step 3 is the real creation moment, but `authentication.service.ts:102` explicitly discards the flag (`const { response } = ...`), the `/sign-in` controller never calls `identifySignUp`, and the browser sends no attribution with a login.

The email-code path only works because the web sends attribution twice — at `/otp/verify` **and** again at `/complete-sign-up` — and the second call is the one that provisions. The password path has no equivalent second send.

Net: `SignUpMethod.PASSWORD` fires only for self-hosted, or for Cloud users who already hold an accepted invitation.

## 2. `signup.submitted` now fires on sign-in

`auth-drawer-body.tsx:534` adds a `SIGN_UP_SUBMITTED` capture to `EmailStep`'s `onSubmit`. `EmailStep` is the default landing step for `/sign-in` (and `/sign-up` just `Navigate`s to `/sign-in`), rendered at line 396 regardless of `effectiveMode` — it's the same box for sign-in and sign-up, and it isn't passed the mode, so as written it can't tell them apart.

So every returning user requesting a login code now emits a sign-up event, and the signup funnel fills with logins. There's already a separate `SIGN_IN_SUBMITTED` for that (`sign-in-form.tsx:140`). This event isn't mentioned in the description — was it intentional?

Either pass `effectiveMode` down and fire only on `'signup'`, or drop the capture.

## 3. `/complete-sign-up` hardcodes `EMAIL_CODE`

`authentication.controller.ts:113` assumes everyone reaching the name step came from the email-code route. Usually true, but that step is reached by any provider whose identity lands in onboarding, and `isPlaceholderName` fires when the last name is empty and the first name matches the email's local part — which a Google account with a thin name hits. Those get recorded as email-code sign-ups.

`completeSignUp` already fetches the identity a few lines earlier (`passwordless-auth.service.ts`, `getOneOrFail`), so `identity.provider` is in hand — deriving the method from it is strictly correct and no extra code.

---

Two smaller notes while I'm here: the `request-trial.tsx` rewrite drops the `flags` query param the old Contact Sales link sent (not mentioned in the description — does the sales form read it?), and `TelemetryContext.reset` now has zero consumers, so the function, its context-type entry and the no-op default can go.

Also worth a doc fix in this PR since it touches the area: `brain/knowledge/connections-auth/ce-authentication.md:19` says `provisionOrOnboard` falls back to onboarding "only … when the name is the placeholder derived from the email". The real condition is `!identity.verified || isPlaceholderName(identity)`, and on Cloud the unverified half is the branch a password sign-up always takes — which is exactly the trap in #1 above.

AhmadTash (approved):

PR #15535, opened 2026-09-14
feat(analytics): track onboarding, invitations and billing lifecycle in PostHog

moaaz-ae:
## Description

Adds server-side PostHog events for the lifecycle moments between sign-up and payment, a PostHog group per platform, and a `deployment` tag on every event so cloud and self-hosted traffic can be separated in the shared project.

### Why

PostHog currently sees sign-up, sign-in and flow events, then nothing until a support ticket. There is no way to answer "how many trials started this week", "who invited a teammate" or "which platforms upgraded", and self-hosted instances report into the same project as cloud with nothing to tell them apart.

### How

- `@activepieces/shared`: new events `onboarding.completed`, `checkout.started`, `plan.upgraded`, `plan.cancelled`, `plan.reactivated`, `trial.started`, `invite.sent`, `invite.accepted`, `sales.handoff.clicked` (all cloud-only except the two invite events) and a `DeploymentKind` enum.
- Server:
  - `platform-plan-telemetry.ts` owns billing events. The entitlement refresh in `autumn-utils.ts` compares the plan before and after and emits `trial.started` or `plan.upgraded` when it changed, skipping free plans. Checkout, cancel and reactivate are emitted from the plan controller.
  - `onboarding.completed` when the name step finishes a sign-up; `invite.sent` and `invite.accepted` from the invitation module and service; `flow.published` moved into the flow service so API and approval publishes count too.
  - Google and SAML logins emit `signed.in`, as password logins already do.
  - Pre-login email-code events are keyed by the identity id. `aliasIdentity` merges that person into the user once the user exists, and `trackForIdentity` switches to the user id as soon as there is one, so the funnel reads end to end on one person.
  - `identifyPlatformGroup` sets a PostHog group per platform (name, plan, creation date) on creation and on plan change. `onPlatformCreated` now returns the plan it created so no extra read is needed.
  - Every event, person and group carries `deployment` (`cloud`, `self_hosted`, `dev`).
- Web: `deployment` is registered on every event; in a development environment nothing is sent unless `localStorage.ap_posthog_dev` is `'1'`; `sales.handoff.clicked` is captured when a user clicks Contact Sales from a locked feature or the plan selector; the tracked-events catalog on the platform settings page lists the new events under a Billing group that is hidden on self-hosted.

No database changes. Independent of #15534 except that both bump the shared package version.

## How was this tested?

- Typecheck for `shared`, `api` and `web`; eslint on every changed file.
- `telemetry-cloud-only-events.test.ts` (21 cases): every new billing and onboarding event is dropped on CE and EE and sent on Cloud attributed to the platform owner, identity-level events keep the identity id, alias runs on Cloud only, every event carries `deployment`.
- `free-legacy-comp.test.ts` (10 cases) still passes with the entitlement refresh emitting telemetry.
- `tracked-events-catalog.test.ts` (5 cases): every event sits in exactly one group, the Billing group disappears on self-hosted, the accounts group keeps only the invite events.
- Not exercised against a running Cloud instance in this round; the automated coverage above is what was run.

Fixes # (issue)


AhmadTash:
Two things should land before this merges — both change what the new events actually mean, rather than how they're built.

### 1. `flow.published` is now double-counted

The description says the event was "moved into the flow service", but the browser-side capture is still there: `packages/web/src/features/flows/hooks/flow-hooks.tsx:114` fires `FLOW_PUBLISHED` on `!requiresApproval`, and `flow.service.ts:551` (`updatedPublishedVersionId`) now fires it for exactly the same publishes — the `route !== 'NEEDS_APPROVAL'` branch is the same condition. Every publish from the UI emits two events.

Either drop the web capture, or keep the server emit only for the paths the browser can't see. Worth noting the server emit also picks up publishes that aren't a user pressing Publish — the connection-swap republish in `app-connection-service/app-connection.handler.ts:365` and git-sync project-release import both dispatch `LOCK_AND_PUBLISH`.

### 2. `identifyPlatformGroup` sends platform names off Cloud

`telemetry.utils.ts:57` gates only on `isProductTelemetryEnabled`, which defaults to true on self-hosted — unlike its sibling `aliasIdentity`, which correctly returns early when `edition !== CLOUD`. It sends `properties: { name: platform.name, … }` from both `platform.service.ts:78` (all editions, on every platform creation) and `platform-plan-telemetry.ts:49`.

No existing PostHog call sends `platform.name`. `pickTelemetryPii` and decision [000034](brain/knowledge/decisions/000034-account-and-sign-in-telemetry-is-cloud-only.md) establish that identifying strings stay on Cloud, and a self-hosted platform name is usually the customer's company name. The same `ApEdition.CLOUD` guard `aliasIdentity` already uses fixes it; dropping `name` off Cloud works too. This also means the "Security impact: no" box needs revisiting.

### Two more worth fixing now, but not blockers

- **`plan.upgraded` fires on downgrades.** `trackPlanChange` (`platform-plan-telemetry.ts:62`) emits it for any non-free plan change that isn't trialing, so `business → plus` reads as an upgrade. It's really `plan.changed` — rename it, or compare plan rank.
- **Trial → paid conversion emits nothing.** A trial is the same `planId` with `trialEndsAt` set (`selectCurrentBaseSubscription`, `autumn-utils.ts:260`), so on conversion `plan === previousPlan` and `onEntitlementsRefreshed` returns at line 43. "Which platforms upgraded" — the question the description opens with — has no event behind it; `trial.started` is the only trial signal.

Both are cheaper to fix now than after dashboards are built on the names.

### Housekeeping

- The branch is `CONFLICTING` with `main` across 7 files.
- `packages/core/shared` is set to `0.174.0`, but `main` is **already** at `0.174.0`. Identical lines merge without conflict, so the bump gets eaten silently even though this PR adds new exports (`DeploymentKind`, 9 event types). It needs `0.175.0` after the rebase.
- `trackForIdentity` in `passwordless-auth.service.ts` — the actual identity→user distinct-id switch — has no test. The added case "should keep the identity id as distinct_id when a platform is known" exercises the unchanged `trackIdentity`.
- Minor: new Google/SAML users now emit `signed.in` on top of `signed.up`, which the email/password signup path doesn't — SSO sign-in counts will run high.

Verified locally at `d70fabfb0c`: the 21 + 10 server unit tests and 5 web catalog tests all pass, and eslint is clean on the changed files.

AhmadTash (approved):

PR #15539, opened 2026-09-14
feat(pinterest,trello): expand board, pin, card and custom field actions for agents

OdaiAhmed99:
## Description

Adds the `audience:'ai'` agent surface to two pieces.

**Pinterest — 29 new agent actions.** The piece had no agent surface at all: its six actions use dynamic dropdowns for board and pin IDs, which an agent cannot drive, yet they were advertised to agents as `both`. Adds atomics across boards (list/get/delete/search, sections), pins (list/get/save/delete, product tags, analytics) and the account (profile, analytics, top pins, followers, following, linked businesses, verified websites). All six originals are demoted to `human`; four gained an agent twin that shares their logic through a new `common/operations.ts` rather than duplicating it. The flow builder is unchanged.

`@activepieces/piece-pinterest` goes to 0.2.0 and `minimumSupportedRelease` to 0.88.2. The piece sat below the 0.86.3 release that filters agent actions out of the step picker, so without this every new atomic would have appeared to flow-builder users.

**Trello — 20 new agent actions.** A previous wave shipped 46 atomics but left real gaps: nothing for custom fields, no way to copy a card, and no delete for a board or a label. Adds custom field definitions, options and per-card values; card copy, activity history and voting; board and label deletion; moving a list between boards; and updating a checklist or one of its items.

`add_card_attachment` is split rather than duplicated. Its only input is a file, which an agent has no way to produce, so the action was advertised to agents while being impossible for them to call. The new `add_card_attachment_from_url` takes a link, and the file variant becomes human-only. `get_card_attachment` and `delete_card_attachment` stay `both`: their inputs are two plain IDs that serve either audience, so a twin would only duplicate them.

### Merged with #15471

#15471 ("improve the step forms") landed after this branch was opened and rewrote the same six human Pinterest actions, including real behaviour changes: a proper base64 `{content_type, data}` media source, `video_url` rejected up front, Find Pin following pages up to Max Results behind an opaque `apc1.` cursor, and Update Board treating a blank description as "keep".

Rather than take those bodies inline and leave the agent twins on the older code paths, the behaviour changes were folded into `common/operations.ts`, so the human action and its twin run exactly the same code. That was the point of extracting the shared operations in the first place, and it means the agent surface inherits the fixes instead of quietly disagreeing with the flow builder.

Three consequences for the agent surface, all deliberate:

- `searchPins` now follows further pages until it has Max Results, instead of truncating a single page.
- `updateBoardDetails` can no longer clear a description. Sending only a blank or whitespace description now fails with "At least one field (name, description, or privacy) must be provided" rather than no-opping.
- `createPinFromMedia` goes through the new media builder.

This also supersedes one of the two pre-existing bugs previously parked for a separate PR: `create-pin`'s `image_base64` option sent `media_source.url` where the spec requires `data` + `content_type`, so it could never have worked. #15471's builder fixes it, and the shared extraction carries the fix to the agent surface.

### Known gaps, deliberate

- **Pinterest `updatePin` and `getPinsAnalytics` are not included.** Both endpoints are beta and not enabled for every app; they return 401 on a connection where every sibling call succeeds. Editing a Pin is therefore not possible through this piece.
- **Trello custom fields (8 actions) and voting (2) ship without output schemas.** Both sit behind board Power-Ups the test account cannot enable, so writes 403 while the matching reads succeed and return empty. Schemas are deferred to a pass that can capture a real payload.
- **Pinterest cannot create a secret board**: `auth.ts` requests `boards:read_secret` but not `boards:write_secret`. Adding that scope forces every existing connection to re-consent, so it belongs in its own PR.
- **The new atomics have no i18n entries.** `src/i18n/translation.json` carries #15471's regenerated strings for the human surface only.

### Notes for reviewers

- **Secret boards are readable; only creating one is blocked.** List Boards offers the full privacy filter, including Secret and Public-and-Secret, because reading a secret board works with the `boards:read_secret` scope the piece already holds. Selecting All omits the parameter rather than forwarding it, matching how the New Board trigger treats the same choice — Pinterest does accept `ALL` as a filter value, but having two code paths in one piece disagree about it was not worth keeping.
- **The six shared Pinterest operations sit behind a single `pinterestOperations` object** rather than loose exports, and the human actions call exactly the same functions as their agent twins.
- The `both → human` demotion covers 7 actions. An agent currently holding one of those tools will stop seeing it, and should use the corresponding twin.

## How was this tested?

Every atomic was run against a **live connection** on a local dev instance via the `test-step` path, and each output schema was authored from the **captured payload** rather than the API reference. That caught five shapes the specs got wrong:

- Pinterest top-pins analytics is `{sort_by, pins[], date_availability}`, not `{items, bookmark}`, and `data_status` is an object keyed by metric, not a string
- Pinterest product tags return `{product_tags}`, and linked businesses a bare array
- A Pin's `pin_metrics` is a `90d`/`lifetime_metrics` tree that was previously flattened to one unexpandable line
- Trello's attachment endpoint returns a single object, not the array its API reference documents

Partial updates were proven by re-reading rather than assumed: setting one field on a Pinterest board and on a Trello checklist item left every other field untouched.

Destructive actions ran only against throwaway boards created by the test run and deleted afterwards, with final state verified back to the pre-run set, leaving nothing behind on either account.

### Re-verified after the #15471 merge

The merge changed behaviour on both audiences, so the affected paths were re-run live against the same account. The merge touched no Trello files, so Trello was not re-run.

| Path | Result |
|---|---|
| Update Board / `updateBoardDetails`, blank description | Keeps the old description — proven by re-reading the board, not by the response alone |
| Create Pin, `image_url` | Publishes |
| Create Pin, base64 data URI | Publishes — the shape that could never work before |
| Create Pin, raw base64 with sniffed content type | Publishes, and Pinterest returned the same pin id as the data URI form, so both produced an identical upload |
| Create Pin, `video_url` | Rejected by the action with its own message, not by a Pinterest 400 |
| Create Pin, malformed base64 | Rejected |
| `createPinFromMedia` | Publishes |
| Find Pin / `searchPins`, cursor round-trip | `apc1.` cursor returned, fed back, second page had no overlap and matched the wide run's second id exactly |
| Find Pin, malformed cursor | Throws rather than silently restarting from the first page |

The throwaway board and every Pin created were deleted, and the account was confirmed back to its pre-run 4 boards with no matching Pins left.

**Not proven, stated rather than implied:**

- **Multi-page following never executed.** The test account's search returns 3 results in a single page with no next bookmark, so the "follow up to ten pages until Max Results" branch was not exercised. The cut-inside-a-page cursor was.
- **Backward compatibility with a raw, non-`apc1.` bookmark is untested.** A synthetic value was forwarded to Pinterest rather than rejected locally, which shows the code path accepts it, but no real 0.1.8-era bookmark was replayed.
- Carried over from before the merge: the `Content-Type` fix on the three Trello custom-field writes (403 paid-plan wall), the six doc-authored custom-field item shapes (their wrappers are verified), and Pinterest's `listLinkedBusinesses` / `listVerifiedWebsites` / `listPinProductTags`, which returned empty and were corrected straight from the v5 spec.

Edition paths: piece-only change, no server or UI code, so behaviour is identical on CE, EE and Cloud. Verified locally that the flow builder is served only the human surface, with every `ai` atomic filtered out.

Fixes PIE-512

Two functional changes, both documented in `docs/install/reference/breaking-changes.mdx`:

1. **Seven actions move from `both` to `human`** — the six original Pinterest actions and Trello's `add_card_attachment`. An agent holding one of these tools stops seeing it. Each has a replacement among the new agent actions that takes plain IDs instead: Create Pin from Media, Create New Board, Update Board Details, Delete Pin by ID, Search Boards, Search Pins, and Add Card Attachment from URL. Flows and saved steps are unaffected.
2. **Update Board and Update Board Details can no longer clear a description** — a blank or whitespace description now means "keep", and a call supplying only a blank description fails instead of silently no-opping. Previously a blank field wiped the description on every run, contradicting the field's own help text.


ibrahim-abuznaid (changes requested):
Reviewed at `dfb3b3ba`, against the Pinterest OpenAPI description (v5.28.0) and the Trello REST reference. Great scope and the PR body is unusually honest about its gaps — thank you for that. Three things block merge, and a handful of schema/prop mismatches will mislead agents at runtime.

## Blockers

**1. Pinterest `0.1.9` is claimed by two open PRs.**

`pinterest/package.json` bumps `0.1.8 → 0.1.9`, and so does #15471. npm currently has `0.1.8`. Whichever of the two merges second merges green and **never republishes** — its actions silently never reach users.

#15471 also touches the same six pre-existing action files and `common/index.ts` that this PR refactors, rewrites `pinterest/src/i18n/translation.json`, and sets `minimumSupportedRelease: '0.88.2'` against this PR's `'0.87.0'`. The two need to be sequenced deliberately and the second one re-bumped. I've left the same note on #15471.

**2. `create-pin-from-media.ts` offers a `media_source.source_type` Pinterest does not accept.**

`{ label: 'Video URL', value: 'video_url' }`. The spec's discriminator mapping allows exactly `image_base64`, `image_url`, `video_id`, `multiple_image_base64`, `multiple_image_urls`, `pin_url`. An agent that picks "Video URL" gets a 400 every time — and this is a new AI-facing action, so it will get picked. Real video upload means `POST /media` to register the upload, then `source_type: 'video_id'` with `media_id` + `cover_image_url`.

The same bad option is pre-existing at `create-pin.ts:47`. Since the helper is now shared, worth fixing both here.

**3. Three Trello actions send a JSON body with no `Content-Type: application/json`.**

`create-board-custom-field.ts`, `create-custom-field-option.ts`, `set-card-custom-field-value.ts` set only `Accept`. `fetch-http-client.ts` `JSON.stringify`s an object body and returns `extraHeaders: {}` — it never sets the content type, so native `fetch` stamps `text/plain;charset=UTF-8`. The two sibling actions in this piece that send a JSON body both set it explicitly (`add-reaction-to-comment.ts`, `update-card-ai.ts`).

`set_card_custom_field_value` is the one I'd expect to fail outright: its body is nested (`{value:{text:…}}` / `{idValue:…}`), so it cannot degrade to query params — and it's the only write path for custom fields. Adding the header is free either way.

## Should fix before merge

**4. Three output schemas describe fields the API never returns.** All three checked against the spec:

- `listLinkedBusinessesActionOutputSchema` declares `id`, `business_name`, `profile_image`. `LinkedBusiness` has only `username` and `image_{small,medium,large,xlarge}_url`. The aiMetadata on `list-linked-businesses.ts` repeats it ("each business id and username") — an agent told to grab a business id gets `undefined` and passes it downstream.
- `listPinProductTagsActionOutputSchema` declares `x` / `y`. `ProductTagItem` has exactly one property: `pin_id`.
- `listVerifiedWebsitesActionOutputSchema` declares `verification_method`. `UserWebsite` has `website`, `status`, `verified_at` — `verification_method` lives on the *Create* model.

**5. The `bookmark` prop on `get-top-pins-analytics` / `get-top-video-pins-analytics` is invented.** Neither `/user_account/analytics/top_pins` nor `/top_video_pins` takes a `bookmark` parameter, and neither response returns one — your own output schema correctly omits it. The prop description tells the agent it's "an opaque cursor returned by a previous call" that it can never obtain, and the value is appended to every request. Paginate with `num_of_pins` only.

**6. `copy-card.ts` — `keepFromSource` is only sent when the multi-select is non-empty, but Trello's default is `all`.** So the prop's "Leave empty to copy the card without them" is backwards: an agent that omits the field clones comments, members, attachments and due date. Worth one live `POST /1/cards?idCardSource=…` to confirm before changing anything. Separately, the option list omits `customFields` — which Trello now requires to be named explicitly — and `start`.

**7. `update-checklist.ts` / `update-checklist-item.ts` guard optional strings with `!== undefined`.** `name: ''` populates `params`, which means the "Provide a name or a position" pre-check passes and `name=` reaches Trello, blanking the checklist name. Every sibling update in this piece uses truthiness for strings and reserves `!== undefined` for booleans (`update-label.ts`, `update-card-ai.ts`) — as do `copy-card.ts` and `add-card-attachment-from-url.ts` in this same PR. Empty string from a model filling a tool schema is the realistic path here.

**8. The missing-outputSchema split is internally inconsistent.** The body says custom fields and voting ship without schemas, which is fair — but `delete_board` and `delete_label` got `{success, id}` schemas in this PR, while `delete_custom_field` and `remove_card_vote` return the identical shape and got nothing. And these are the only `list_*` / `get_*` actions in the piece without one; all 14 pre-existing ones have them. Either add the trivial ones or name the follow-up ticket in the body.

**9. Version semantics.** Pinterest takes a PATCH bump for 29 new actions while Trello takes MINOR for 20 — `piece-versioning.mdx` says MINOR for added functionality. More substantive: 7 actions move `both → human` (6 Pinterest + `add_card_attachment`), which drops them out of the AI catalog via the `action.audience !== 'human'` filter in the metadata utils. Any agent already holding `createPin` or `add_card_attachment` as a tool loses it on upgrade. That's breaking by the doc's definition and neither bump reflects it — it at least needs calling out in the description.

<details>
<summary><b>Minor — non-blocking (15)</b></summary>

- **`searchPins` truncation poisons pagination** (`operations.ts`): items past `max_results` are dropped, but the returned `bookmark` is the cursor for the page *after* the untruncated page and `has_more` is just `!!response.bookmark` — the agent loses results per page while looping confidently. `total_results: items.length` compounds it ("how many Pins match X" answers with the truncated count). Inherited from `find-pin`, but `findPin` is now human-only and `searchPins` is the agent path.
- **`operations.ts` rewrites any 404 into "No pins found matching your search criteria."** An LLM reads that as an empty result rather than a failed call; it hides a bad `ad_account_id` or a scoped-out token. The discriminator is also a substring match on the piece's own error text, so rewording that string silently disables the branch.
- **Three Trello deletes synthesize `{success: true}`** (`delete-board`, `delete-custom-field`, `delete-label`), discarding the API response — a deviation from the five pre-existing Trello deletes, which all return the body. These are the irreversible ones, so the API's own confirmation is worth most here.
- **The body's "0.86.3 filters agent actions out of the step picker" is not right.** The filter commit (`208f500a7b3`) lands 2026-07-20, three days *after* the `0.86.3` tag; the first GA release containing it is `0.87.0`. Your `0.87.0` floor is correct — only the justification is off. Relatedly, `trello/src/index.ts` pins `minimumSupportedRelease: '0.86.4'`, which was never tagged; correct in effect, but it should read `0.87.0`.
- **AGENTS.md:** `buildPath`, `repeatedParam`, `assertMaxLength`, `assertUrl`, `paginatedSchema` and `buildValueBody` take positional arguments (rule is a single destructured object for any function with more than one parameter); `common/index.ts` exports its three helpers individually rather than grouping them into one named const (`operations.ts` does this correctly with `pinterestOperations`); `buildValueBody` sits above the exported action rather than below it.
- **Dead code:** `repeatedParam` in `common/index.ts` has zero callers.
- `assertUrl` early-returns on `''`, where the old `create-pin` ran `new URL()` unconditionally — an expression resolving to empty now reaches Pinterest as a generic 400 instead of failing fast with the piece's own message. Same shape in `buildPath`, which drops empty params entirely.
- `createPinOperation` still sends `note`, `is_removable` and `product_tags`, none of which `POST /pins` accepts (pre-existing, but the refactor makes it the shared path for two actions).
- **Base64 pin creation is gone from the agent surface** — `createPin` was demoted and `createPinFromMedia` is URL-only. The other five demotions have full AI twins; this one doesn't. Its aiMetadata says "raw image data is not accepted here", which reads as a Pinterest limitation rather than a piece one.
- `set-card-custom-field-value`: `checked: value === 'true'` silently writes `false` for `"True"`, `"1"`, `"yes"` — a wrong write rather than an error, on an agent-facing action.
- `list-card-activity`: `page` is an unbounded `Property.Number` with the "up to 19" ceiling only in the description, and `limit` is never sent — the documented "50 at a time" contract holds only because that's Trello's default.
- `listBoardMembershipsActionOutputSchema` sets `itemLabel`, which only applies when the step output is a top-level array; this action returns `{memberships, count}`, so rows render as "Item 1, Item 2…". `labelKey: 'memberType'` on the `memberships` field is what the other `{list, count}` schemas in the file do.
- `listCardActivityActionOutputSchema.activity` and `listCardVotesActionOutputSchema.voters` declare array fields with no `listItems`, so neither expands in the data selector — every other array field in the file supplies them.
- ~49 new displayName/description strings aren't in `translation.json`. That file is normally CI-generated, but the nightly translations job has been failing since 2026-08-16, so they'll ship untranslated in 7 Pinterest / 12 Trello locales until that's fixed.
- You're not in `authors` on either piece despite writing all 49 actions — worth adding.
- `add_card_attachment` keeps a now-dead `aiMetadata` block after the demotion.

</details>

<details>
<summary><b>Verified clean</b></summary>

The `pinterestOperations` refactor is faithful — same endpoints, methods, body construction, validation thresholds, synthesized returns and error wrapping for all six pre-existing actions; the empty-string edge cases above are the only behaviour deltas, so no silent break for live flows. No duplicate action `name` or `displayName` in either piece (35 and 74, all unique). Every new action is registered and nothing was dropped from either `actions` array. `add_card_attachment` keeps its `name`, so existing human flows are safe. No untagged `audience`, no `both` left on an action an agent can't drive. No GET sends a body. No `Dropdown` nested inside `DynamicProperties`. No dependency changes, nothing outside the two piece directories, no added code comments. Every OAuth scope the 29 new Pinterest endpoints need is already in `auth.ts`, and every enum I spot-checked (`privacy`, `pin_filter`, `feed_type`, both `sort_by` sets, both `metric_types` sets) matches the spec — including `explode: false` making `.join(',')` correct.

Two things worth stating because they look like bugs and aren't: `list-pin-product-tags` unwrapping `response.product_tags` and `list-linked-businesses` unwrapping a bare array are **both correct** per the spec (`ProductTagsResponse` requires `product_tags`; `/user_account/businesses` returns `type: array`). The two different unwrap styles in one PR read as an inconsistency, but the API really is inconsistent there.

</details>

OdaiAhmed99:
Thanks — this was a useful review. I verified every claim before acting: Pinterest against the official OpenAPI **v5.28.0**, and the body/header claim against `serializeBody` in `fetch-http-client.ts`. All held up. Where I could run fixes live against a throwaway board I did; where I couldn't, it's called out below rather than left implied.

## Blockers

**1. Version collision.** Pinterest is now **`0.2.0`** — the correct MINOR bump for 29 new actions (your #9), which also clears the direct clash with #15471 at `0.1.9`. Sequencing recommendation at the bottom.

**2. `video_url`.** Confirmed absent from the spec's `PinMediaSource` discriminator. Removed from `create-pin.ts`; on `create-pin-from-media.ts` I dropped the dropdown entirely and hardcoded `image_url` so an agent can't pick a losing value. **Verified live.**

**3. Missing `Content-Type`.** Confirmed — `serializeBody` returns `extraHeaders: {}`, so fetch stamps `text/plain`. Added to all three. **Not live-verified; see below.**

## Should-fix

- **#4** All three schemas confirmed wrong against the spec and corrected (`LinkedBusiness` has only `username` + `image_*_url`; `ProductTagItem` only `pin_id`; `verification_method` belongs to the Create model). The misleading aiMetadata on `list-linked-businesses` is fixed too.
- **#5** Confirmed neither analytics endpoint takes `bookmark`. Prop removed from both. **Verified live.**
- **#6** You were right — I tested both directions. Omitting `keepFromSource` carries checklist, due date **and** members; sending `keepFromSource=` yields `checklists 0 / comments 0 / due None / members []`. The action now always sends it. Added `customFields` and `start`; both accepted live.
- **#7** Switched to truthiness. **Verified live:** `name: ""` now errors instead of blanking the checklist.
- **#8** Added nine schemas — the two trivial deletes plus the custom-field and voting list/get actions (caveat below).

## Minor

Fixed: dead `repeatedParam`; `assertUrl` no longer passes `''`; the 404→"No pins found" rewrite removed; `parseCheckbox` rejects anything but `true`/`false`; `list_card_activity` sends `limit` and bounds `page` 0–19 (verified); `labelKey` on `listBoardMemberships`; `listItems` on the activity/voters arrays; `minimumSupportedRelease` → `0.87.0`; dead `aiMetadata` removed; authors added.

**One flag on the `searchPins` truncation fix:** it no longer returns a bookmark that would skip results (`bookmark: null` + `has_more: true` when truncated). That shared operation also backs the **released** `findPin`, which passes `max_results` with `defaultValue: 25` — so this changes a released action's output semantics on any page larger than the cap, not just the new agent action. It's a fix for silent result loss, but it is a behaviour change and I'd rather name it than bury it.

Two I'd push back on, happy to be overruled: the three deletes **keep** `{success, id}` (Trello returns `{"_value": null}`, a worse agent signal, and the schemas already describe the synthesized shape); and the `pinterestCommon` grouping + `buildPath` named-params touch ~30 pre-existing call sites, so I'd rather do them as a focused cleanup than inflate this diff. Not addressed: the untranslated strings (the nightly CI job is the real fix).

## Not verified

1. **The `Content-Type` fix.** Custom Fields is a paid Trello built-in, not an installable Power-Up — absent from all 557 entries of `plugins?filter=available`, and every `/1/customFields` write on a free workspace returns `403 "Custom Fields Power-Up disabled for board"`. All three calls fail before the body matters. The header is correct regardless, but there's no green run behind it.
2. **Six of the nine new schemas.** Wrappers are confirmed live; the per-item shapes are authored from Trello's documented `CustomField` model, not a captured payload — the same thing you flagged in #4, so I'd rather label it than let it read as verified.
3. **`searchPins` truncation.** The test account returns at most one search result, so that branch never executed. The untruncated path is verified.

Verified against real payloads: `listCardActivity` (field-for-field), `voters`, `removeCardVote`, `deleteBoard`, `createPin`.

## `both → human` demotion — breaking for agents

Exactly seven actions move `both → human` (Pinterest `createPin`, `createBoard`, `deletePin`, `findPin`, `findBoardByName`, `updateBoard`, plus Trello `add_card_attachment`). `filterActionsByAudience` resolves the AI catalog as `action.audience !== 'human'`, so all seven drop out of it. **Any agent already holding `createPin` or `add_card_attachment` as a tool loses it on upgrade.** Action `name`s are unchanged, so human flows are unaffected.

**My recommendation: this warrants the `⛓️‍💥 breaking-change` label and a `breaking-changes.mdx` entry.** Two independent vectors meet the handbook's "a consumer must take action" bar — the seven lost agent tools (users must repoint agents at `createPinFromMedia` and the Trello atomics), and the `findPin` output change above. I haven't applied the label or written the entry yet; say the word and I will.

Side effect worth naming: base64 Pin creation is now absent from the agent surface (`createPin` demoted, `createPinFromMedia` URL-only). That's a piece limitation, not a Pinterest one, and the aiMetadata now says so.

## Version sequencing against #15471

Verified directly: #15471 bumps `0.1.8 → 0.1.9`, npm latest is `0.1.8`, and both PRs branch from `0.1.8`.

**Recommendation: merge #15471 first, then rebase this PR keeping `0.2.0`** — giving a clean monotonic `0.1.8 → 0.1.9 → 0.2.0` where both publish. Beyond versioning, #15471 edits the same six action files and `common/index.ts` that this PR restructures into `pinterestOperations`; folding a focused step-forms change into the refactor is far safer than re-applying it onto files it has never seen. Whichever order is chosen, the second PR needs a rebase and a re-verify rather than a green auto-merge.

## Pre-existing bugs found, left for a separate PR

Both change what a valid input does:

- **`create-pin`'s `image_base64` option cannot ever have worked** — `createPinOperation` sends `media_source: { source_type, url }`, but the spec requires `data` + `content_type`. Same family as the `video_url` bug, one level deeper.
- **`createPinOperation` sends `note`, `is_removable` and `product_tags`** — none are in `PinCreate`.

ibrahim-abuznaid:
Follow-up on the sequencing, now that I've reviewed #15471 in depth. Two corrections to what I said earlier, both in your favour on version and against it on the floor:

**1. `minimumSupportedRelease` — keep 0.88.2, not 0.87.0.** I told you 0.87.0 was correct, and it is for this PR in isolation: 0.87.0 is the first GA release containing the `audience` filter, and nothing here needs more. But #15471 uses `propertyGroups`, `display: 'stepper'` and `display: 'cards'`, none of which exist in the framework before **0.88.2** (verified at the 0.87.0, 0.88.1 and 0.88.2 tags). Whichever of the two PRs merges second overwrites this line, so the combined piece must end up on `0.88.2` — please take Talal's value when you rebase rather than letting `0.87.0` win.

**2. Version order matters more than I realised.** `0.1.9` (#15471) then `0.2.0` (this PR) publishes cleanly. The reverse is actively dangerous: `packagePrePublishChecks` compares only against the `latest` dist-tag (`currentVersion === latestPublishedVersion`), so a `0.1.9` published after `0.2.0` is already `latest` passes the guard and publishes with `--tag latest`, moving `latest` **backwards** — cloud would then resolve to the build without your 29 new actions, with green CI and no error anywhere. So: #15471 first, and your `0.2.0` stays as is.

**3. When you rebase, one hunk needs real attention rather than a mechanical resolve.** #15471 rewrites Find Pin's `run()` into a capped multi-page loop with a `skip` continuation cursor. Your `searchPinsOperation` currently returns `bookmark: null` on truncation, which reintroduces the continuation bug that rewrite exists to fix. Porting the new body into `operations.ts` (and threading `skip` through `SearchParams`) is the part I'd not want lost in a conflict resolution. Same applies to his `create-pin` base64/data-URI handling and the `update-board` description change.

A merge simulation shows 18 conflict hunks across 9 shared files, so this is a deliberate rebase rather than a quick one — happy to walk it with you.

ibrahim-abuznaid (approved):

PR #15540, opened 2026-09-14
refactor(ai): name and type the AI billing path, end to end

AbdulTheActivePiecer:
## Description

The cost path grew a private vocabulary and a set of escape hatches while it was being built. This is the cleanup pass over the whole stack below it. No call is charged differently as a result — this changes names, shapes and types, not arithmetic.

**Naming.** Every string literal that decided how a call was charged is now an enum. `AiChargeBasis` says whether the charge came from what the provider reported or from a fixed credit. `BYOKBilling` says whether a customer-key call still owes its credit or the agent turn already paid it.

`ActivepiecesAiBillingScope` described where a call sat in the tenancy tree, which said nothing about who to charge. It becomes `ActivepiecesAiConsumerSource`, whose members name the two things that actually spend credits: `AI_STEP_IN_FLOW` and `CHAT`. The field is `source` to match. `PLATFORM` had no consumer and is gone, and `flowRun` is now required on the flow variant rather than optional, so a billed step cannot reach the reporter without the run it belongs to. `environment` leaves `AiUsageFlowRunContext` and `ActivepiecesAiFlowRun` — nothing read it.

**Shape.** No more passing functions as parameters and no more functions defined inside methods. The two report sinks — billed cost and provider key health — are registered once at boot, and every step of the cost path is a top-level function. The cost reporter itself existed twice, once in the api and once in the worker, differing only in how each reached its transport; one copy now lives in `server-utils` and takes `report` and `pageWebhookUrl` as parameters, so both processes install the same unbilled-call watch instead of two that drift.

**Types.** Provider auth and config schemas move to `core-utils` as one discriminated union, `AiProviderCredentials`, parsed in exactly one place when a row is read out of the `ai_provider` table. That deletes every `Record<string, unknown>` and all 26 casts from the model factories, and the worker contract now carries the real type.

An unbilled call now records what was lost — provider, model, platform, project, flow run and generation — instead of only counting.

Also in here: `ProjectAIProvider` moves off the deprecated `z.nativeEnum` to `z.enum`, and `CLAUDE.md` gains a rule against writing `void` in front of a promise. That last commit is self-contained and unrelated to billing; say the word and I will lift it out to its own PR.

## How was this tested?

- `npm run test-unit` — 28/28 tasks green: shared 16 files, server-utils 15, engine 39 files / 471 tests, sandbox 20 / 265, worker 20 / 304, web 80 / 739, core-execution 12, core-utils 3.
- `tsc` clean on `packages/server/api` and `packages/server/worker`.
- `turbo run lint` — 0 errors on api, worker, server-utils and core-utils.
- `activepieces-ai-cost.test.ts` grew by ~356 lines covering the renamed enums and the single cost reporter; `ai-utils.test.ts` and `create-language-model.test.ts` were rewritten against `AiProviderCredentials` rather than loose records.
- **Edition paths:** the reporter and the key-health sink are installed in `appPostBoot` outside the edition switch, so CE, EE and Cloud all take the same path. There is no UI in this PR and no new env var, so nothing is gated and nothing needs self-hoster setup.

- [x] yes — technical (removed/renamed API field or endpoint, dropped column, new required field, removed/required env var)

Everything renamed here is an internal TypeScript type or an enum member, never an API field, a column or an env var. `worker-contract.ts` changes are the api↔worker RPC, which the existing `versionsAreCompatible` gate already stops a skewed worker from crossing during a rolling deploy. `@activepieces/shared` was bumped 0.165.1 → 0.166.0 earlier in the stack, which covers the exports that moved to `core-utils`.

This changes how provider credentials are typed and parsed. Nothing changes about who can read them — `GET /:provider/config` is still `securityAccess.engine()`, keys are still decrypted only in `decryptRowAuth`, and plaintext still travels no further than the worker that calls the model.

Two things worth a reviewer's eye:

1. `aiProviderCredentials()` answers a failed parse by returning **empty** `auth` and `config` rather than throwing. That is deliberate — the update endpoint still validates against the loose union, so a row saved before this change must not start failing on read. The cost is that a genuinely malformed row now reaches the provider with no key instead of its stored one, and for `CUSTOM` / `CLOUDFLARE_GATEWAY` it loses `baseUrl` / `accountId` at the same time, since one parse covers the whole `config` object. Every field is `.partial()` and zod strips unknown keys, so this only fires on a type mismatch, not a missing or extra field.
2. `enrichWithKeysIfNeeded` now returns the auth config directly instead of a full `GetProviderConfigResponse`. Its single caller already discarded every other field, so this narrows what is returned rather than widening it.


packages/server/utils/src/activepieces-ai-cost.ts
```diff
@@ -1,123 +1,164 @@
...
-            }
-            if (chunk.type === 'finish' && !reported) {
-                reported = true
-                reportUsage({ billing, provider, modelId, call: toCall({ generationId, providerMetadata: chunk.providerMetadata }) })
-            }
-            controller.enqueue(chunk)
-        },
+        transform: (chunk, controller) => trackChunk({ chunk, controller, state, context }),
+        flush: () => settleStream({ state, context }),
+        cancel: () => settleStream({ state, context }),
```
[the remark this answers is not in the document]
AbdulTheActivePiecer: Both halves check out as fine.

**Typecheck.** `packages/server/utils/tsconfig.lib.json` sets `"types": ["node"]` and loads no DOM lib, so the `Transformer` in scope is Node's `stream/web` one, which has `cancel`. The `Transformer` without it is `lib.dom.d.ts`'s, which this package never pulls in. `tsc -p packages/server/utils/tsconfig.lib.json --noEmit` exits 0.

**Runtime.** It fires. On Node v22:

```js
const t = new TransformStream({
  transform(c, ctrl) { ctrl.enqueue(c) },
  flush() { console.log('flush ran') },
  cancel(r) { console.log('CANCEL RAN:', String(r)) },
})
t.readable.getReader().cancel('client aborted')
// → CANCEL RAN: client aborted
```

`flush` did **not** run on that path, which is exactly why both are wired: `flush` covers a close with no `finish` part, `cancel` covers a client abort or a provider error. Dropping `cancel` is what would leave the call unbilled.

packages/server/utils/src/activepieces-ai-cost.ts
```diff
@@ -1,123 +1,161 @@
...
+    const costUsd = openRouterCostOf(providerMetadata)
+    if (isNil(generationId) || isNil(costUsd)) {
         return undefined
     }
+    return { charge: AiChargeBasis.PROVIDER_REPORTED_COST, generationId, costUsd, ...tokens }
+}
+
+function createEmbeddingBillingInfo({ result, generationId }: CreateEmbeddingBillingInfoParams): ActivepiecesAiCall | undefined {
+    const costUsd = openRouterCostOf(result.providerMetadata)
+    if (isNil(generationId) || isNil(costUsd)) {
```
hazemadelkhalel: This drops a `costUsd` we already read just because `id` is missing, and OpenRouter declares it optional (`OpenRouterEmbeddingResponseSchema`, `id: z9.string().optional()`).

Ran it: without `id` the $0.0004 is discarded and the call lands in `unbilledCalls`, which pages on-call every 5 min. Suggest billing on cost and letting the id be optional.

packages/server/utils/src/activepieces-ai-cost.ts
```diff
@@ -1,123 +1,161 @@
...
-            if (chunk.type === 'finish' && !reported) {
-                reported = true
-                reportFlatCredits({ billing, provider, modelId, generationId })
-            }
-            controller.enqueue(chunk)
+function trackChunk({ chunk, controller, state, context }: TrackChunkParams): void {
+    if (chunk.type === 'response-metadata') {
+        state.generationId = chunk.id ?? state.generationId
+    }
+    if (chunk.type === 'finish' && !state.reported) {
```
hazemadelkhalel, line 83: `state.reported` is set before we know a charge could be built, so a `finish` with no cost metadata short-circuits `settleStream` and skips `recoverCostAndReport` — the exact case it exists for. Only set it when `createBillingInfo` returns a call.

packages/core/utils/src/lib/ai-provider-credentials.ts
```diff
@@ -0,0 +1,181 @@
...
+    z.object({ provider: z.literal(AIProviderName.QWEN), config: OpenAiCompatibleVendorConfig.partial(), auth: BaseAIProviderAuthConfig.partial() }),
+    z.object({ provider: z.literal(AIProviderName.MINIMAX), config: OpenAiCompatibleVendorConfig.partial(), auth: BaseAIProviderAuthConfig.partial() }),
+    z.object({ provider: z.literal(AIProviderName.MOONSHOT), config: OpenAiCompatibleVendorConfig.partial(), auth: BaseAIProviderAuthConfig.partial() }),
+])
+export type AiProviderCredentials = z.infer<typeof AiProviderCredentials>
+
+export function aiProviderCredentials({ provider, auth, config }: { provider: AIProviderName, auth: unknown, config: unknown }): AiProviderCredentials {
+    const parsed = AiProviderCredentials.safeParse({ provider, auth, config })
+    if (parsed.success) {
+        return parsed.data
```
hazemadelkhalel, line 178: One parse covers `auth` and `config`, so a row with a valid `apiKey` and one bad `modelType` comes back with `auth: {}`. Verified it then falls through to the host's `OPENAI_API_KEY` when one is set. Worth splitting the two parses, or at least logging on the fallback.

packages/core/ai-providers/src/lib/create-language-model.ts
```diff
@@ -1,4 +1,4 @@
-import { AIProviderName, observedProviderFetch, ProviderOutcomeReporter, spreadIfDefined } from '@activepieces/core-utils'
+import { AiProviderCredentials, AIProviderName, observedProviderFetch, ProviderOutcomeReporter, spreadIfDefined } from '@activepieces/core-utils'
 import { AzureProviderConfig, BaseAIProviderAuthConfig, BedrockProviderAuthConfig, BedrockProviderConfig, OPENAI_COMPATIBLE_VENDOR_BASE_URLS, OpenAICompatibleProviderConfig, VertexProviderAuthConfig, VertexProviderConfig } from '@activepieces/core-piece-types'
```
hazemadelkhalel: Seven of these eight are now unused (only `OPENAI_COMPATIBLE_VENDOR_BASE_URLS` survives) — they were the cast targets you removed. `turbo run lint` flags all seven.

packages/server/api/src/app/flows/flow-version/migrations/migrate-v26-ai-piece-cost-billing.ts
```diff
@@ -26,7 +26,15 @@ function withAgentMaxSteps({ actionName, input }: { actionName?: string, input:
     if (actionName !== AGENT_ACTION_NAME || typeof input[AgentPieceProps.MAX_STEPS] === 'number') {
         return input
     }
-    return { ...input, [AgentPieceProps.MAX_STEPS]: DEFAULT_MAX_STEPS }
+    return { ...input, [AgentPieceProps.MAX_STEPS]: storedMaxSteps(input[AgentPieceProps.MAX_STEPS]) ?? DEFAULT_MAX_STEPS }
```
hazemadelkhalel: `migrate-v22:22` runs first (index 66 vs 70) and already coerces a non-number to 20, so this only reaches flows at schemaVersion 23-26. Ran it: `"50"` → 50 at 26, still → 20 at 22.

packages/server/api/src/app/ai/ai-key-health-reporter.ts
```diff
@@ -0,0 +1,15 @@
+import { isNil, tryCatch } from '@activepieces/core-utils'
+import { aiProviderKeyHealth } from '@activepieces/server-utils'
+import { FastifyBaseLogger } from 'fastify'
+import { aiProviderService } from './ai-provider-service'
+
+export function installAiKeyHealthReporter(log: FastifyBaseLogger): void {
+    aiProviderKeyHealth.setReporter(({ platformId, providerConfigId, signal }) => {
+        tryCatch(() => aiProviderService(log).recordKeyObservation({ platformId, providerId: providerConfigId, signal })).then(({ error }) => {
```
hazemadelkhalel: `rejectedPromiseHandler` (`helper/promise-handler.ts`) is the helper your new CLAUDE.md rule names for exactly this, and this is `server/api` with logging only. Six lines become one.

packages/server/utils/src/ai-cost-reporter.ts
```diff
@@ -0,0 +1,50 @@
...
+import { aiUsageReportOf } from './ai-usage-report'
+import { onCallService } from './on-call.service'
+
+function install({ log, report, pageWebhookUrl }: InstallAiCostReporterParams): void {
+    activepiecesAiCost.setReporter((event) => {
+        tryCatch(() => report(aiUsageReportOf({ event, idempotencyKey: `ai:${apId()}` }))).then(({ error }) => {
+            if (isNil(error)) {
+                return
+            }
+            log.error({ error, provider: event.provider, model: event.modelId, generationId: event.call.generationId }, '[aiCostReporter] An AI call went unbilled')
```
hazemadelkhalel: Flat `generationId` and `model: <id>` break the server logging schema — `model: { id }` is the shape at `agent-draft-ai.ts:57` and `execute-agent-run.ts:141`. Inherited from the old worker copy, but `ai-key-health-reporter.ts` in this same PR gets it right with `aiProvider: { id }`.

packages/pieces/community/ai/src/index.ts
```diff
@@ -12,7 +12,7 @@ import { runAgent } from "./lib/actions/agents/run-agent";
 export const ai = createPiece({
   displayName: "AI",
   auth: PieceAuth.None(),
-  minimumSupportedRelease: '0.91.0',
+  minimumSupportedRelease: '0.92.0',
```
hazemadelkhalel, line 15: Root `package.json` is still 0.91.0 and v26 rewrites every AI step to piece 0.11.0 — worth confirming the release branch actually cuts 0.92.0, or a 0.91.x would leave migrated flows pointing at a piece the server won't resolve.

hazemadelkhalel (commented):
Verified each point by running or reading it in a worktree at the head. `tsc` clean on every touched package, `lint` 0 errors, unit suites green (server-utils 225, core-execution 149, ai-providers 48, core-utils 45, worker 323). Couldn't run `test-api` locally, `redis-memory-server` won't compile on my machine.

Two things that can't be inline because the files aren't in the diff.

**The eval harness is broken.** `worker/test/lib/agent-eval/core/llm-judge.ts:12` and `runner.ts:110` still call `createModel({ provider, auth, config, modelId })` and throw `Cannot destructure property 'provider' of 'credentials'`. Nothing catches it: worker `tsconfig.lib.json` and its lint glob are both `src/**/*.ts`, there is no `tsconfig.spec.json`, and the suite is `skipIf(!HAS_PROVIDER_KEY)`.

**"No call is charged differently" doesn't hold.** Four paths change, verified base vs head:

| | base | head |
|---|---|---|
| own-key stream cancelled before `finish` | 0 | 1 credit |
| managed stream with no `finish` | nothing | recovered cost |
| KB search embeddings | free | billed |
| `maxSteps` `"50"` at schemaVersion 26 | ceiling 20 | ceiling 50 |

Base's two stream observers had no `flush`/`cancel`, base's `observedEmbeddingFetch` had zero call sites, and `server/api` had zero billing producers. All four look deliberate, so I think it's the sentence plus a `breaking-changes.mdx` entry rather than the code.

Smaller description fixes: 25 of 26 casts (`create-language-model.ts:192` survives), one `Record<string, unknown>` left at `:159`, shared is 0.168.0 → 0.169.0 in this PR not 0.165.1 → 0.166.0 earlier, and `BYOKBilling` doesn't exist (it's `turnAlreadyCharged`).

---

**Not this PR, but they need owners.** All four verified at this branch's head. Recording them here so they don't get lost; happy to move each to its own PR or issue.

1. **A worker socket drop double-charges an AI call.** On disconnect the API runs `releaseConnectionJobs` → `returnJobToQueue` → `moveToDelayed`, which never consults `opts.attempts`, so the job is re-dispatched. The worker's `abortInFlightRuntime()` only shuts down `ctx.runtime` — `execute-ai.ts` never touches `runtime` and has no abort plumbing at all, and `execute-agent-run.ts`'s `AbortController` fires only on the wall-clock backstop or a `__cancel_check` that polls over the socket that just died. Its own comment says it exists to stop "the original keeps executing to completion and double-runs the requeued copy". An API rolling deploy mid-call bills twice and pays the provider twice. → #15493 / #15488.

2. **`attempts: 1` does not stop BullMQ stall-retries.** `attempts` appears zero times in bullmq 5.61.0's `moveStalledJobsToWait-8.js`, and `moveJobToWait` is called unconditionally, outside the `stalledCount > maxStalledJobCount` branch. With `lockDuration: 120_000` / `maxStalledCount: 3`, an evicted pod replays a whole agent turn up to 4x. The per-turn credit is deduped by `${conversation.id}:chatTurn:${runId}`, so this looks safe from that surface while every managed per-call charge duplicates underneath. Pre-dates the stack.

3. **Autumn is `failOpen: true`, and `track` is in the SDK's fail-open set.** Any 5xx, network error, or >5s (`DEFAULT_TIMEOUT_MS = 5e3`, `track` in `AUTO_TIMEOUT_OPERATION_IDS`) is rewritten by `FailOpenHook.afterError` into a 200 carrying `{value: 0, balance: null}` with `error: null`. `sendTrackEvent`'s `tryCatch` then sees success and its `log.error` never fires; `unbilledCalls` only records failures upstream of the reporter, so `AI_CALLS_WENT_UNBILLED` never pages; and `ai_usage` was dropped in `1766015156683-DropLegacyTables.ts`, so there's no row to reconcile from. That's 100% of AI credit revenue for the duration of an incident, silently and unrecoverably. Pre-dates the stack, and it's the largest of the four.

4. **The charge report is fire-and-forget and shutdown doesn't wait for it.** `billGenerate` calls `report(...)` without awaiting and returns the result, and `inFlightJobs` counts only `executeJob`/`completeJob`, so `drainInFlightJobs` sees zero and the process exits with the ack still on the wire. `socket.io-client`'s `onclose` then rejects it, it's logged once, and nothing retries. Recurs on every worker restart. → #15492.

packages/server/utils/src/activepieces-ai-cost.ts
AbdulTheActivePiecer: Changed in `033429c3` — the guard needs only the cost now, and `generationId` became optional on the cost variant of `ActivepiecesAiCall`. `ReportAiUsageRequest` already had it optional, so nothing else moved. Added a test for cost-present, id-absent.

One correction worth recording, because it changes how urgent this is: I fired a real embedding at OpenRouter through the same SDK and the same options to check, and the id is there — `gen-emb-…` alongside `usage.cost`, on both `text-embedding-3-small` and `openai/text-embedding-3-small`. Their docs example omits it, but the live response carries it, so nothing was going unbilled in practice.

The change stands on the schema rather than on live behaviour: requiring a field the contract marks optional, in order to record one it guarantees, is the wrong way round even if it never fires.

packages/server/utils/src/activepieces-ai-cost.ts
AbdulTheActivePiecer: Fixed in `033429c3` — the flag is only set when `createBillingInfo` returns a call, so a finish chunk with no cost falls through to `settleStream` and `recoverCostAndReport`.

Worth noting where it helps, since `createBillingInfo` returns undefined for two different reasons. With no generation id, `recoverCostAndReport` bails on its own first check and the call is unbilled either way. It is the other branch — id present, cost absent — that was losing a recoverable charge, and that is the case the lookup was written for.

The same commit splits the unbilled reason into `PROVIDER_REPORTED_NO_COST` and `NO_GENERATION_ID`, so the two stop being indistinguishable when it pages.

packages/core/utils/src/lib/ai-provider-credentials.ts
AbdulTheActivePiecer: Good catch, and it turned out to matter more than the corrupted-row case. Changed in `6f345f08`.

The asymmetry is the part I had missed: `auth` is validated on write and stored encrypted, so it cannot carry a legacy shape, while `config` is plain JSONB that predates these schemas. And a row whose `models` array has an older shape used to work perfectly at runtime — nothing reads `config.models` when building a model, only the provider listing that fills the dropdown. After this stack that row parsed to `{}` and took the key with it, and an absent key is what makes the SDK reach for the host's `OPENAI_API_KEY`, so it could run on the server's key rather than failing.

It degrades a step at a time now: the whole row, then with `models` blanked, then `auth` alone. The schemas stay strict deliberately — my first attempt put `.catch([])` on them, which quietly weakened `CreateAIProviderRequest` so a malformed `models` array would have been accepted and stored empty.

Two things I have not done: nothing logs when the fallback fires, which belongs at the call sites since `core-utils` has no logger; and the write path still stores `{}` for a config with a bad `modelType`, losing `baseUrl` and `apiKeyHeader`. I will raise that one separately.

packages/core/ai-providers/src/lib/create-language-model.ts
AbdulTheActivePiecer: Removed in `f4c98f59`.

On how they survived: lint does flag all seven, but as warnings, so the task still exits zero — and `tsc` never looks, since `noUnusedLocals` is not set anywhere in the repo. They are all `--fix`-able, so `npm run lint-dev` would have taken them out; I ran `turbo run lint` directly and read the exit code. My miss.

packages/server/api/src/app/flows/flow-version/migrations/migrate-v26-ai-piece-cost-billing.ts
AbdulTheActivePiecer: Confirmed — v22 at index 66, v26 at 70, and v22 already coerces a non-number to 20, so `withAgentMaxSteps` only reaches flows at schemaVersion 23-26.

Which also means the dynamic-expression case raised on this line is not introduced here: a `{{ }}` `maxSteps` on a flow below 23 is already flattened by v22 today. No change in this PR — thanks for pinning the range down.

packages/server/api/src/app/ai/ai-key-health-reporter.ts
AbdulTheActivePiecer: Half taken, in `f4c98f59`. The `tryCatch`/`isNil` dance is gone and it is two lines now.

I kept a plain `.catch` rather than the helper, because `rejectedPromiseHandler` logs a generic "Unhandled promise rejection" at `error` with only `{ error }` — so the swap would drop `aiProvider: { id }`, the field you would actually search on, and raise a background bookkeeping failure to error level. The `.catch` keeps both and still satisfies `no-floating-promises`. Happy to go the other way if the consistency is worth more than the context.

packages/server/utils/src/ai-cost-reporter.ts
AbdulTheActivePiecer: Fixed in `f4c98f59` — `model: { id }` and `generation: { id }`.

Nothing else in the repo logs a generation id, so I grouped it the same way rather than leaving it flat. Easy to change if you would rather it sat at the top level.

packages/pieces/community/ai/src/index.ts
AbdulTheActivePiecer: Deliberate, and the root version is meant to stay at 0.91.0 here — release numbers get set on the release branch rather than in a feature stack, so the breaking-change entries now sit under Unreleased too.

`minimumSupportedRelease` is the exception and stays at 0.92.0, because piece 0.11.0 calls `POST /v1/ai/execute` and no earlier release serves it; leaving it lower would let a 0.91.x server resolve a piece it cannot run. It is the only `0.92.0` left in the tree, so it is easy to move if the cut lands elsewhere — worth confirming at release time.

packages/server/api/src/app/flows/flow-version/migrations/migrate-v26-ai-piece-cost-billing.ts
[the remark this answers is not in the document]
AbdulTheActivePiecer: Confirmed, and changed.

The mechanism is as you describe: `{{trigger.maxSteps}}` is a non-numeric string, `storedMaxSteps` returns `undefined` for it, and the step came out pinned to `20`. The migration is forward-only, so the per-run limit was gone for good.

One bound on the blast radius, for the record. `migrateV22AgentStepToThinClient` runs earlier in the `migrations` array, targets the same `@activepieces/piece-ai` / `run_agent` pair, and already replaces any non-number `maxSteps` with `20`. So this only reaches flows already at schemaVersion 23 or later — anything below that had no expression left to lose by the time v26 ran.

`withAgentMaxSteps` now returns the input untouched when the stored value carries a mustache token, using `extractMustacheTokens` from `@activepieces/core-utils`. That is the same tokenizer the engine's props-resolver and the expression rewriter share, so what the migration treats as dynamic is exactly what resolves at run time. A numeric string such as `"3"` is still normalized to `3`, and `"abc"`, `"0"`, `"-5"`, `"3.7"` and `""` still fall back to `20`, since the engine's NUMBER cast would only turn those into `NaN`.

Pinned by a new case in `migrate-v26-ai-piece-cost-billing.test.ts`.

packages/core/utils/src/lib/ai-provider-credentials.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+    const withoutModels = AiProviderCredentials.safeParse({ provider, auth, config: configWithoutModels(config) })
+    if (withoutModels.success) {
+        return withoutModels.data
```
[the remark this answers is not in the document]
AbdulTheActivePiecer: Confirmed, and changed.

The chain is exactly as you set it out. `manualTextModelCatalog` returns `undefined` only when `models` is absent or nil, so `models: []` reads as a present-but-empty catalog; `resolveModelIdForProvider` then takes the manual-catalog branch and `pickAllowedModel` throws `ENTITY_NOT_FOUND` on an empty candidate list. The comment above `manualTextModelCatalog` spells out why an empty catalog is authoritative — an admin-listed catalog is the whole truth about what a key exposes — and that is what makes `[]` the wrong thing for a fallback to write: it states something about the key that the operator never stated.

`configWithoutModels` now drops the `models` key instead of emptying it, so a row whose catalog no longer parses falls back to curated resolution. That is what the "keep the key" fallback was for in the first place.

New unit tests in `packages/core/utils/test/ai-provider-credentials.test.ts` pin it: a malformed catalog keeps `apiKey`, `apiKeyHeader` and `baseUrl`, and comes back with no `models` key at all.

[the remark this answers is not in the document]
AbdulTheActivePiecer: Correcting my previous reply: the step is gone, not repaired.

I went looking for how a row could carry a `models` list we cannot parse, and through the API it cannot. All three configs that declare `models` — OpenAI-compatible, Cloudflare Gateway and Vertex — use the same `z.array(ProviderModelConfig)`, and `UpdateAIProviderRequest.config` either matches a member that demands that shape or falls through to an empty-object member that strips `models` outright. What lands in the column is zod's parsed output, so it is a valid catalog or none at all.

That left the step covering only rows edited outside the API — a hand-edited jsonb column, a restored backup. We are not going to defend against that: data put there by hand is the operator's to fix, not ours to guess at.

So `configWithoutModels` is deleted rather than corrected. `aiProviderCredentials` now falls straight from the full parse to auth-only, which is the step that covers the case that does happen in product: a row written before a rule tightened, such as a Vertex `region` saved before `3b86b2997f` added its regex.

Your finding stands either way — the empty catalog you flagged can no longer be written at all.

packages/server/api/src/app/flows/flow-version/migrations/migrate-v26-ai-piece-cost-billing.ts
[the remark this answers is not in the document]
AbdulTheActivePiecer: Pushed as `557886329e`.

packages/core/utils/src/lib/ai-provider-credentials.ts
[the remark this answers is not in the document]
AbdulTheActivePiecer: Pushed as `55eb08a2b9`, which deletes the step rather than correcting it.

AbdulTheActivePiecer:
@greptile-apps your last pass found nothing outstanding here. Please approve.

AbdulTheActivePiecer:
> I can’t approve yet. Rechecking the current head, `packages/server/api/src/app/ee/agent/agent-conversation-credits.ts:8-13` still rounds each call before accumulating it. That makes conversation totals diverge from the fractional credits charged by `aiUsageService`: two 0.3-credit calls record 0 instead of 0.6, while two 0.6-credit calls record 2 instead of 1.2. Please preserve fractional values or round only the aggregate; the existing P1 finding remains outstanding.

it is not a huge deal, don't worry about it we will fix it in an upcoming pr

othmanemad (approved):

othmanemad (approved):

PR #15550, opened 2026-09-15
feat(ai): read model tiers from a published file

Louai-Zokerburg:
## Description

The chat model tiers — Fast, Expert and Heavy — used to be a constant in the release. Moving Expert to a newer model meant a code change, a review, a deploy and a self-host upgrade. This moves the tier table to a file on the CDN that a separate console publishes, so a tier can point at a new model in minutes.

**How it works**

- The console (`activepieces/ap-analytics`) writes `ai/pricing.json` to the CDN. It holds the tiers — `id`, `label`, `modelId`, `nativeModelId`, `thinkingBudget` — and `defaultTierId`.
- The server reads that file on start-up and refreshes it every hour. With the CDN's own 5 minute cache, a published change reaches every server in about 65 minutes.
- The tiers shipped in the release stay as the fallback. A missing file, a malformed one, no tiers, or a `defaultTierId` that names no tier all fall back to the last good copy and then to `ACTIVEPIECES_CHAT_TIERS`, with a warning. Start-up never blocks on it: the warm-up races a 2 second timer and the timer is `unref`'d.
- `AP_AI_PRICING_URL` overrides the location. It is optional and self-hosters need to set nothing.
- The web app reads the tiers from a new `GET /v1/ai-providers/tiers` instead of the hardcoded list, so a tier renamed in the console shows its new name without a deploy.

**This PR no longer publishes prices**

It used to. While the branch was open, main shipped #15487–#15495 (release 0.92.0), which bills managed AI on the dollar cost the provider reports for each call and deletes every credit-weight table. There is no fixed price left to publish, so the merge commit drops that half: `creditWeight` on the tier, `modelWeights`, `unpricedModelCreditWeight`, `creditWeightForModel`, and `tierLabel` on `AIProviderModel`. Only the tier table survives.

The reader ignores keys it does not declare, so the console can keep publishing its credit fields until it is trimmed — there is a test for that. Trimming the console is follow-up work in the other repo.

**One bug fixed on the way in**

`resolveNamedModelId` looked a tier up in the published catalog and then checked the result against `aiProviderUtils.managedChatModelIds()`, which is built only from the constants in the release. The first tier we publish naming a model the release does not ship would have thrown *"not available on Activepieces AI credits"* on every agent step — while the builder was already offering that model. `managedModelCandidates` now also accepts the published tiers' own model ids. The guard is unchanged for everything else: the console still cannot admit an arbitrary model, only one it has put behind a tier.

## How was this tested?

- `packages/server/utils/test/ai-pricing-catalog.test.ts` — the published tiers, the fallback when the fetch fails, a rejected file (bad default tier, no tiers), the shared snapshot, start-up surviving an unreachable file, and that the credit fields an older console still publishes are ignored. It points `AP_AI_PRICING_URL` at an unreachable host so it never reads the live CDN.
- `agent-model-resolution.test.ts` — a new case mocks the catalog, points `smart` at a model id the release does not carry, and asserts both the tier id and the raw model id resolve.
- CI on this branch: lint, `tool-search (postgres)` and the deterministic gate all pass.
- **Not yet checked:** CE / EE / Cloud edition paths in a running app. Please exercise the chat tier picker before merging.

Fixes # (issue)

It was functionally breaking while it carried the credit-weight rescale. That is gone. What is left adds an endpoint, adds an optional env var, and changes nothing for a self-hoster: with no published file — or no outbound network — the server uses the tiers in the release exactly as it does today.

**Security notes.** The new outbound fetch goes through `safeHttp.retryingAxios`, so the SSRF filter applies to `AP_AI_PRICING_URL` as well. The published file is parsed with Zod and a file that fails the schema is rejected outright rather than partly applied. A published tier can only name a model, never grant one: `resolveNamedModelId` still refuses any model that is neither in the release nor behind a published tier. `GET /v1/ai-providers/tiers` is authenticated and returns no per-platform data.


Louai-Zokerburg:
@greptileai two notes on the 3/5 review.

**The flow-metadata finding is not in this PR's diff.**

`agent-conversation-service.ts`, `agent-run-controller.ts`, `piece-metadata-service.ts` and `web/src/app/routes/agents/id/runs/index.tsx` are not changed by this branch. They arrived here in a merge of `main` (#15548 and #15296). The diff against `main` is:

```
git diff --name-only origin/main HEAD | grep -E 'agent-conversation-service|agent-run-controller'
(no match)
```

The branch is currently 0 commits behind `main`, so the merge base is `main` itself. Those files read the same on both sides. Reviewing the merge commit's range is picking up code that is already shipped rather than code this PR proposes.

That said, the finding itself looks real, so we are reporting it separately against `main` instead of ignoring it. `flowReferencesFor` reads `FlowRunEntity` and `FlowVersionEntity` by id only, with no `projectId` in the `where`. The ids come from `agent_conversation` rows that are project-scoped, so it needs a conversation whose `flowRunId` points outside its current project to leak anything — which agent moves between projects can produce. It is also a plain violation of the repo rule that every query filters by `projectId` or `platformId`, independently of whether a path to it exists today.

**The pricing-snapshot finding is correct and still open.**

Agreed, and thanks for holding it. What changed so far: `load()` no longer awaits an in-flight refresh when something is cached, so the two entry points read the same cached object instead of different ones, and the cache is warmed before the server accepts traffic, so the first requests no longer select from the release table and bill from the published file.

What that does not fix is the case you are pointing at. If the hourly background refresh assigns a new snapshot between model selection and billing, the turn still selects from one and bills from the other. Closing that means recording the credit weight the turn was quoted and billing from the record. The assistant messages are built in the worker and arrive over RPC (`saveAgentMessages`), so it is a change to that contract, and we are doing it as its own PR rather than growing this one.

PR #15526, opened 2026-09-14
feat(web): simplify platform administration navigation and feature discovery

AhmadTash:
## Description

Rebuilds the platform admin area around one declarative page registry (`packages/web/src/app/routes/platform/admin-pages.ts`). Every admin page, route, redirect, nav entry, plan gate and tab derives from that single list instead of the three unconnected places that held them before.

What changes for an admin:

- **Shorter, grouped sidebar.** Three groups (Workspace, Build, Platform) with eleven parents and their sections drawn as indented sub-items under each parent. Groups are always expanded, so the active tab is always visible and highlighted.
- **Section parents are real pages.** Users & access, Connections, AI Center & MCP, Security, Billing & usage and Infrastructure each open an overview at their bare path: a card per child with its live count and plan tier, plus the few cross-cutting sections that only make sense for the whole area (sign-in policy and pending invitations, connection health and most-used pieces, AI credits and default model, a security checklist, infrastructure status).
- **Tabs address `?tab=<id>`.** Old page routes such as `/platform/security/sso`, `/platform/setup/mcp` or `/platform/infrastructure/workers` redirect to their new home, so bookmarks and docs links keep working.
- **Locked features show themselves.** A page or tab the plan does not include renders ghosted sample data under an upgrade card instead of a blurred empty page, with a docs link on Community.
- **Consolidations.** Branding folds into General. MCP Server becomes a tab of AI Center & MCP. Workers, Health, Triggers and Configurations become tabs of Infrastructure. Pieces becomes a two-tab page (Pieces, Piece Sets) with a plan banner.

The reasoning and the alternatives that were tried and rejected are recorded in `brain/knowledge/decisions/000035-platform-admin-is-fifteen-pages-derived-from-one-registry.md`.

## How was this tested?

- Scripted smoke test in headless Chromium over all 31 pages and tabs plus the 17 legacy redirects, recording every section's rendered state, failed requests and console errors. Run twice: against the EE dev backend, and against a matching 0.90.4 EE Docker backend with seeded data (pending invitations, six connections including one in `ERROR`) so the populated states of Pending invitations, Needs attention, Most used pieces, members and connections tables were exercised as well as the empty states.
- All pages render, no crashes, no `Trouble loading` states, every redirect lands on the right page.
- Found and fixed during the run: the embed page fired `GET /v1/signing-keys` and got 402 on every visit for plans without embedding, because sample mode mounts the real page. The query is now gated on `platform.plan.embeddingEnabled`.
[...]

PR #15562, opened 2026-09-15
feat(web): gate paid features at the action instead of the page

AhmadTash:
## Description

**Stack 1 of 3.** Base `main`. Followed by #15563 (registry) then #15564 (sample data). The three together are byte-identical to #15526, which stays open unchanged.

Moves the paywall from the door to the action. A locked feature used to replace its whole page, so an admin evaluating Activepieces could not see what they were being asked to pay for. Now the page opens, the data loads, and the primary control carries a crown and refuses with a dialog naming the feature and its plan.

- **`useFeatureGate`** returns the crown, a click interceptor and the dialog, so gating an action is three lines at the call site.
- **`PLATFORM_FEATURES`** holds each paid feature's title, description and bullets in one place instead of per-page copy.
- **`LockedFeatureGuard`** renders the new `FeatureTeaser` and drops the sales request form. `featureKey` stays accepted but unused so the other 21 callers keep compiling; #15564 removes the prop and the last callers together.
- **The crown replaces an action's own icon** rather than sitting beside it, since a plus and a crown read as two separate affordances.
- Projects, the team-project limit and the seat limit use the new gate, and the Users invite button crowns once seats are exhausted.
- **`DataTable`** spans its empty and error rows across the columns that actually render rather than the columns it was handed, so a table with hidden columns no longer leaves a ragged row.

No IA change, no route change, no server change. All of the stack's translation keys land here so that no intermediate branch renders a raw key.

## How was this tested?

- Cloud and EE dev stacks on a free plan: Projects, the Users invite button and the team-project dialog each show a crown and open the upgrade dialog rather than a dead control.
- EE stack with the plan flags set: no crowns, every action behaves as before.
- Community: the dialog offers a docs link instead of an upgrade button.
- `tsc` and `npm run lint-dev` clean on this branch on its own, not only at the top of the stack.


packages/web/src/features/projects/components/create-project-button.tsx
```diff
@@ -17,51 +23,58 @@ export function CreateProjectButton({
...
-    hasReachedLimit,
-    ensureTeamProjectAvailable,
-    teamProjectLimitDialog,
-  } = useTeamProjectLimitGuard({ projects });
+  const { hasReachedLimit, teamProjectLimitContent } = useTeamProjectLimitGuard(
+    { projects },
+  );
+  const { platform } = platformHooks.useCurrentPlatform();
+  const projectsGate = useFeatureGate({
+    locked: platform.plan.billedTeamProjectsLimit === 0,
```
hazemadelkhalel, line 31: The crown reads `billedTeamProjectsLimit === 0` but the actual block is `hasReachedLimit` (used >= limit). At limit 3 with 3 used the button looks completely normal and still refuses. Reproduced. Suggest driving the crown off `hasReachedLimit`.

packages/web/src/features/projects/components/create-project-button.tsx
```diff
@@ -17,51 +23,58 @@ export function CreateProjectButton({
...
         <Button
           variant="ghost"
           size="icon"
           className={cn('h-6 w-6 hover:bg-accent', className)}
-          onClick={onClick}
         >
           <Plus />
         </Button>
       );
     case 'full':
```
hazemadelkhalel, line 66: Only the `full` branch renders `crown`. The sidebar rail uses `variant="icon"` (primary-rail/index.tsx:468), so the most common entry point shows no lock signal at all. On main `onClick` reached all three variants.

packages/web/src/features/projects/components/new-project-dialog.tsx
```diff
@@ -271,9 +304,14 @@ const NewProjectForm = ({
               onClick={(e) => {
                 e.stopPropagation();
                 e.preventDefault();
+                if (gate?.locked === true) {
```
hazemadelkhalel, line 307: This is the only `gate.locked` check, and it is in the submit handler. A locked admin fills in the name, waits on the connections fetch, hits Create, and only then gets refused. On main the trigger was not wrapped in this dialog at all when `hasReachedLimit`, so it refused on the click. Please move the check back to the trigger.

packages/web/src/app/components/feature-teaser.tsx
```diff
@@ -0,0 +1,124 @@
...
+  description,
+  bullets,
+  tier,
+  documentationUrl,
+  videoUrl,
+}: FeatureTeaserProps) {
+  const { openDialog: openManagePlanDialog } = useManagePlanDialogStore();
+  const { data: edition } = flagsHooks.useFlag<ApEdition>(ApFlagId.EDITION);
+
+  if (edition === ApEdition.COMMUNITY) {
```
hazemadelkhalel, line 21: The CE branch returns above the video block, so the 5 pages passing `lockVideoUrl` (api-keys, project-role, projects, connections, templates) show nothing to Community users. Main rendered it for every edition.

packages/web/src/app/components/locked-feature-guard.tsx
```diff
@@ -26,68 +9,36 @@ export const LockedFeatureGuard = ({
...
+type LockedFeatureGuardProps = {
+  children: React.ReactNode;
+  locked: boolean;
+  lockTitle: string;
+  lockDescription: string;
+  lockVideoUrl?: string;
+  lockDocumentationUrl?: string;
+  lockBullets?: string[];
+  lockTier?: FeatureTier;
+  featureKey?: string;
```
hazemadelkhalel, line 42: `showContactSales` defaulted to true on main, so dropping `RequestTrial` removes the sales lead form from 16 of 17 callers in one go. That reads as a growth decision rather than a refactor. Worth keeping it wired until #15564 migrates these, or calling it out in the description.

packages/web/src/features/billing/hooks/use-feature-gate.tsx
```diff
@@ -0,0 +1,134 @@
...
+export function useFeatureGate({ locked, feature }: UseFeatureGateParams) {
+  const [open, setOpen] = useState(false);
+
+  return {
+    locked,
+    crown: locked ? (
+      <Crown className="size-3.5 shrink-0 text-primary-foreground/90" />
+    ) : null,
+    open: () => setOpen(true),
+    intercept: (handler?: () => void) =>
```
hazemadelkhalel, line 29: `intercept`, `open` and `dialog` have no readers. The single consumer touches only `.crown`, so `UpgradeFeatureDialog` never mounts anywhere and has not run yet. Either wire it at a call site or hold it for #15563.

packages/web/src/app/components/feature-banner.tsx
```diff
@@ -0,0 +1,54 @@
+import { ApEdition, ApFlagId } from '@activepieces/shared';
+import { t } from 'i18next';
+import { ExternalLink } from 'lucide-react';
+
+import { Alert, AlertDescription } from '@/components/ui/alert';
+import { useManagePlanDialogStore } from '@/features/billing';
+import { flagsHooks } from '@/hooks/flags-hooks';
+
+export function FeatureBanner({
```
hazemadelkhalel, line 9: `grep -rn "feature-banner" packages/` returns nothing. Dead on arrival, suggest dropping it until something needs it.

packages/web/src/lib/feature-tier.ts
```diff
@@ -0,0 +1,6 @@
+export const TIER_LABELS: Record<FeatureTier, string> = {
```
hazemadelkhalel, line 1: Zero references, and a third copy of `TIER_LABELS` alongside feature-teaser.tsx:110 and use-feature-gate.tsx:109. `ENTERPRISE_DOCUMENTATION_URL` is now in four places too.

packages/web/public/locales/en/translation.json
```diff
@@ -2887,5 +2887,160 @@
...
+  "{count, plural, =1 {# connection} other {# connections}}": "{count, plural, =1 {# connection} other {# connections}}",
+  "{count, plural, =1 {# domain} other {# domains}}": "{count, plural, =1 {# domain} other {# domains}}",
+  "{count, plural, =1 {# machine} other {# machines}} connected": "{count, plural, =1 {# machine} other {# machines}} connected",
+  "{count, plural, =1 {# worker is} other {# workers are}} on an incompatible version": "{count, plural, =1 {# worker is} other {# workers are}} on an incompatible version",
+  "{count} of 3": "{count} of 3",
+  "{credits} credits included": "{credits} credits included",
+  "{included} included": "{included} included",
+  "{provider} powers chat": "{provider} powers chat",
+  "{used} of {total} credits used this period": "{used} of {total} credits used this period",
+  "Sample data. {feature} requires a paid plan.": "Sample data. {feature} requires a paid plan.",
```
hazemadelkhalel, line 3025: These sample-data keys have no consumer in 15562, 15563 or 15564. Looks like leftovers from an earlier shape of the banner: this one plus `requires the {tier} plan.`, `Everything below is sample data...`, `Sample data · shows what you'll see once enabled`, `What you're seeing is sample data`, `Included with {tier}.` and the branding one.

packages/web/src/components/custom/data-table/index.tsx
```diff
@@ -333,6 +333,7 @@ export function DataTable<
 
   const scrollContainerRef = useRef<HTMLDivElement>(null);
   const rows = table.getRowModel().rows;
+  const visibleColumnCount = table.getVisibleLeafColumns().length;
```
hazemadelkhalel, line 336: Good catch. `columns` here already includes the injected select and actions columns, so the old count was wrong in both directions.

hazemadelkhalel (changes requested):
Reviewed by checking out the branch and running the components, not just reading the diff. The direction is right and the Users invite button is a clean example of it. Four behaviour issues reproduced, all in the projects path plus the CE teaser.

I wrote throwaway render tests to confirm these rather than eyeball them. Results:

```
✓ at limit 3 of 3 the New Project button shows NO crown even though it is blocked
✓ clicking a blocked button opens the create FORM, not the refusal
✓ the refusal only appears after submit, and the create mutation never fires
✓ the icon variant (sidebar rail) never renders a crown, even when locked
✓ CONTROL: full variant at limit 0 DOES render the crown
✓ Community edition renders NO showcase video
✓ CONTROL: Cloud edition DOES render the video
```

The regression is provable: I wrote one test describing main's behaviour (click a blocked button, expect the refusal immediately, expect no form fields, expect no create call). It passes on main and fails on this branch. Same file, same mocks.

Worth folding a trimmed version of that test into this PR. It pins the crown predicate, the refusal timing and the CE teaser branch in one file, and it is the kind of thing that silently rots across a 3 PR stack. Happy to hand over what I have.

One thing I want to correct in advance: the ~150 new translation keys are NOT stray. I attributed them and 94 are consumed by #15563, exactly as described. Only 7 have no consumer anywhere in the stack. The frontloading is fine.

hazemadelkhalel:
Here is the test, as promised. Drop it at `packages/web/test/app/billing/feature-gating.test.tsx` and run `npx vitest run test/app/billing/feature-gating.test.tsx` from `packages/web`.

It is written as the behaviour we want, so on this branch it fails on exactly the four points and passes on the three the PR already gets right:

```
✓ crowns the button when the plan includes no team projects
× crowns the button when the team project allowance is used up
× crowns every variant, including the sidebar rail
× refuses on the click instead of after the form is filled in
✓ still opens the form when the plan has room
× plays the showcase video on community, same as every other edition
✓ plays the showcase video on cloud
```

The three passing ones are load bearing. They are what tells you a failure is the component and not my mocks.

Two notes. `packages/web`'s lint script only globs `src/**`, so nothing here gets linted by CI: I ran `npx eslint 'test/app/billing/feature-gating.test.tsx'` by hand and it is clean. The two `eslint-disable` lines at the top are unavoidable, `@testing-library/jest-dom` is not a dependency of this workspace and `<video>` has no implicit ARIA role.

<details>
<summary>feature-gating.test.tsx</summary>

```tsx
/**
 * @vitest-environment jsdom
 */
/* eslint-disable jest-dom/prefer-in-document -- @testing-library/jest-dom is not a dependency of packages/web */
/* eslint-disable testing-library/no-container, testing-library/no-node-access -- <video> has no implicit ARIA role */
import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
import { render, screen, fireEvent } from '@testing-library/react';
import * as React from 'react';
import { beforeEach, describe, expect, it, vi } from 'vitest';

let teamProjectsLimit: number | null = 0;
let edition = 'cloud';
const createProject = vi.fn();

vi.mock('i18next', () => ({ t: (key: string) => key }));

vi.mock('lucide-react', async (importOriginal) => ({
  ...(await importOriginal<Record<string, unknown>>()),
  Crown: (props: Record<string, unknown>) =>
    React.createElement('span', { ...props, 'data-testid': 'crown' }),
}));

vi.mock('@/hooks/platform-hooks', () => ({
  platformHooks: {
    useCurrentPlatform: () => ({
      platform: {
        plan: {
          billedTeamProjectsLimit: teamProjectsLimit,
          globalConnectionsEnabled: false,
        },
      },
    }),
  },
}));
vi.mock('@/hooks/user-hooks', () => ({
  userHooks: { getCurrentUserPlatformRole: () => 'ADMIN' },
}));
vi.mock('@/hooks/authorization-hooks', () => ({
  useIsPlatformAdmin: () => true,
}));
vi.mock('@/hooks/flags-hooks', () => ({
  flagsHooks: { useFlag: () => ({ data: edition }) },
}));
vi.mock('@/features/billing/stores/manage-plan-dialog-state', () => ({
  useManagePlanDialogStore: () => ({ openDialog: vi.fn() }),
}));
vi.mock('@/features/billing', async (importOriginal) => ({
  ...(await importOriginal<Record<string, unknown>>()),
  useManagePlanDialogStore: () => ({ openDialog: vi.fn() }),
}));
vi.mock('@/features/connections', () => ({
  globalConnectionsQueries: {
    useGlobalConnections: () => ({ data: { data: [] }, isLoading: false }),
  },
}));
vi.mock('@/features/projects', () => ({
  projectCollectionUtils: {
    invalidate: vi.fn(),
    useCreateProject: () => ({ mutate: createProject, isPending: false }),
  },
}));
vi.mock('@/components/custom/global-connection-utils', () => ({
  DefaultTag: () => null,
}));
vi.mock('@/components/custom/multi-select-piece-property', () => ({
  MultiSelectPieceProperty: () => null,
}));
vi.mock('@/components/ui/sonner', () => ({ internalErrorToast: vi.fn() }));
vi.mock('@/components/custom/animated-icon-button', () => ({
  AnimatedIconButton: ({
    children,
    icon: _icon,
    iconSize: _iconSize,
    ...rest
  }: React.PropsWithChildren<Record<string, unknown>>) => (
    <button {...rest} data-testid="new-project-trigger">
      {children}
    </button>
  ),
}));
vi.mock('@/components/icons/plus', () => ({ PlusIcon: () => null }));
vi.mock('@/components/ui/sidebar-shadcn', () => ({
  SidebarMenuButton: ({
    children,
    ...rest
  }: React.PropsWithChildren<Record<string, unknown>>) => (
    <button {...rest}>{children}</button>
  ),
}));

import { FeatureTeaser } from '@/app/components/feature-teaser';
import { CreateProjectButton } from '@/features/projects/components/create-project-button';

const LIMIT_REACHED_COPY = "You've reached your team project limit";
const SHOWCASE_VIDEO =
  'https://cdn.activepieces.com/videos/showcase/api-keys.mp4';

const renderWithQueryClient = (ui: React.ReactElement) =>
  render(
    <QueryClientProvider
      client={
        new QueryClient({ defaultOptions: { queries: { retry: false } } })
      }
    >
      {ui}
    </QueryClientProvider>,
  );

const usedTeamProjects = (count: number) =>
  Array.from({ length: count }, () => ({ type: 'TEAM' })) as never;

beforeEach(() => {
  teamProjectsLimit = 0;
  edition = 'cloud';
  createProject.mockClear();
});

describe('create project gating', () => {
  it('crowns the button when the plan includes no team projects', () => {
    teamProjectsLimit = 0;
    renderWithQueryClient(
      <CreateProjectButton variant="full" projects={usedTeamProjects(0)} />,
    );
    expect(screen.getByTestId('crown')).toBeDefined();
  });

  it('crowns the button when the team project allowance is used up', () => {
    teamProjectsLimit = 3;
    renderWithQueryClient(
      <CreateProjectButton variant="full" projects={usedTeamProjects(3)} />,
    );
    expect(screen.getByTestId('crown')).toBeDefined();
  });

  it('crowns every variant, including the sidebar rail', () => {
    teamProjectsLimit = 0;
    renderWithQueryClient(
      <CreateProjectButton variant="icon" projects={usedTeamProjects(0)} />,
    );
    expect(screen.getByTestId('crown')).toBeDefined();
  });

  it('refuses on the click instead of after the form is filled in', () => {
    teamProjectsLimit = 3;
    renderWithQueryClient(
      <CreateProjectButton variant="full" projects={usedTeamProjects(3)} />,
    );
    fireEvent.click(screen.getByTestId('new-project-trigger'));
    expect(screen.getByText(LIMIT_REACHED_COPY)).toBeDefined();
    expect(screen.queryAllByRole('textbox')).toHaveLength(0);
    expect(createProject).not.toHaveBeenCalled();
  });

  it('still opens the form when the plan has room', () => {
    teamProjectsLimit = 3;
    renderWithQueryClient(
      <CreateProjectButton variant="full" projects={usedTeamProjects(1)} />,
    );
    fireEvent.click(screen.getByTestId('new-project-trigger'));
    expect(screen.queryByText(LIMIT_REACHED_COPY)).toBeNull();
    expect(screen.queryAllByRole('textbox').length).toBeGreaterThan(0);
  });
});

describe('feature teaser', () => {
  it('plays the showcase video on community, same as every other edition', () => {
    edition = 'ce';
    const { container } = render(
      <FeatureTeaser
        title="Enable API Keys"
        description=""
        videoUrl={SHOWCASE_VIDEO}
      />,
    );
    expect(container.querySelector('video')?.getAttribute('src')).toBe(
      SHOWCASE_VIDEO,
    );
  });

  it('plays the showcase video on cloud', () => {
    edition = 'cloud';
    const { container } = render(
      <FeatureTeaser
        title="Enable API Keys"
        description=""
        videoUrl={SHOWCASE_VIDEO}
      />,
    );
    expect(container.querySelector('video')?.getAttribute('src')).toBe(
      SHOWCASE_VIDEO,
    );
  });
});
```

</details>

hazemadelkhalel (approved):

PR #15563, opened 2026-09-15
feat(web): reorganise platform administration into fewer, clearer pages

AhmadTash:
## Description

**Stack 2 of 3.** Base #15562. Followed by #15564 (sample data).

The nav was a literal in `PlatformSidebar`, the routes a hand-written array, and each page's gate hardcoded in its own file. Nothing tied the three together, which is why Embedding sat under Setup while routing to `/platform/security/`. `ADMIN_PAGES` becomes the single source, and the sidebar, the routes, the crowns and the redirects all derive from it.

What changes for an admin:

- **Shorter, grouped sidebar.** Three groups (Workspace, Build, Platform), with each parent's sections drawn as indented sub-items rather than an in-page tab row, so the structure is visible without opening a page first.
- **Section parents are real pages.** Users & access, Connections, AI Center & MCP, Security, Billing & usage and Infrastructure each open an overview at their bare path: a card per child with its live count and plan tier, plus the cross-cutting sections that only make sense for the whole area.
- **Tabs address `?tab=<id>`,** and eleven redirects keep old deep links such as `/platform/security/sso` and `/platform/infrastructure/workers` working.
- **Consolidations.** Usage folds into Billing as a tab. AI Center loses its own wrapper now that the route renders its tabs directly. Branding moves back into General as one form with one Save. Pieces becomes a parent whose first tab is the catalog with Piece Sets beside it. Workers, Health, Triggers and Configurations become tabs of Infrastructure.
- Health and Workers kept their own inner tab rows in `?tab=`, which now belongs to the registry, so theirs moved to `?view=`. Without that they rendered an empty panel and ejected to the overview on click.

The registry declares the `sample` and `teaser` fields here but populates neither; #15564 turns the treatment on. Ticking those on in this PR would have rendered two upgrade prompts stacked on every locked page, since the pages still carry their own guard at this point.

## How was this tested?

- Every old `/platform/**` deep link redirects to its new home.
- Each section parent opens its overview; children open on `?tab=`; an unknown `?tab=` falls back to the parent rather than a blank page.
- Health and Workers inner tabs switch on `?view=` and no longer eject to the overview.
- Sidebar icons animate on hover again, and crowns appear only on fully gated pages.
- Checked on CE, EE and Cloud dev stacks: Community hides the AI Capabilities tab and Cloud hides Configurations, both through the registry's `isHidden`.
- `tsc` and `npm run lint-dev` clean on this branch on its own.


packages/web/src/app/routes/platform/billing/usage-tab.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <div className="flex w-full flex-col gap-4 p-6">
+      <div className="flex flex-col gap-1">
+        <div className="flex items-center gap-1.5">
+          <h1 className="text-xl font-medium">{t('Usage')}</h1>
+          <Tooltip>
+            <TooltipTrigger asChild>
+              <Info className="size-3.5 text-muted-foreground cursor-help" />
+            </TooltipTrigger>
+            <TooltipContent side="right" className="max-w-[240px]">
```
othmanemad: `max-w-[240px]` is exactly `max-w-60`, and this PR already writes `max-w-60` in `overview-shell.tsx` — two spellings of one value.

packages/web/src/app/components/sidebar/platform/platform-nav-item.tsx
```diff
@@ -0,0 +1,113 @@
...
+  label,
+  active,
+  crowned,
+}: PlatformNavSubItemProps) {
+  const navigate = useNavigate();
+
+  return (
+    <SidebarMenuItem>
+      <div className="flex items-stretch">
+        <div className="ml-[15.6px] mr-1 w-px shrink-0 bg-sidebar-border" />
```
othmanemad: `ml-[15.6px]` is an eyeballed decimal — `SidebarMenuSub` already ships this hairline, otherwise `ml-4` is 0.4px away.

packages/web/src/app/routes/platform/setup/mcp/index.tsx
```diff
@@ -141,6 +138,25 @@ export default function PlatformMcpPage() {
...
+
+function McpSection({
+  description,
+  children,
+}: {
+  description: React.ReactNode;
+  children: React.ReactNode;
+}) {
+  return (
+    <div className="w-full mx-auto max-w-[1198px] py-6">
```
othmanemad: `max-w-[1198px]` is the page-band width, also hardcoded in `mcp-server/page-band.tsx` — worth a `--container-page-band` token now that `CenteredPage` is gone.

packages/web/src/app/routes/platform/overview/security-overview.tsx
```diff
@@ -0,0 +1,154 @@
...
+                  )
+                : t('No allowed email domains set')
+            }
+          >
+            <OverviewRowLink to="/platform/users?tab=sso">
+              {t('Restrict domains')}
+            </OverviewRowLink>
+          </OverviewRow>
+          <OverviewRow
+            tone={keyCount === 0 ? 'ok' : 'warn'}
```
othmanemad: `keyCount === 0` renders `ok` on a failed fetch, while loading, and when `apiKeysEnabled` is false — so this row can read "No API keys issued" in green next to a card reading "Locked" or "Trouble loading API keys". The audit row below already handles the off-plan case with `off`.

packages/web/src/app/routes/platform/overview/overview-shell.tsx
```diff
@@ -0,0 +1,203 @@
...
+        <ChevronRight className="size-3.5 shrink-0 text-muted-foreground transition-transform group-hover:translate-x-0.5" />
+      </div>
+      {isLoading ? (
+        <Skeleton className="h-7 w-16" />
+      ) : (
+        <div className="text-2xl font-semibold leading-none tracking-tight">
+          {isError ? '—' : value}
+        </div>
+      )}
+      <div className="text-sm text-muted-foreground">
```
othmanemad, line 71: `description` renders while `isLoading`, so a card shows a skeleton next to "No workers connected" — branch it like `value` does above and all six overviews are fixed at once.

packages/web/src/app/routes/platform/overview/ai-overview.tsx
```diff
@@ -0,0 +1,176 @@
...
+    data: providers,
+    isLoading: isLoadingProviders,
+    isError: isProvidersError,
+    refetch: refetchProviders,
+  } = aiProviderQueries.useAiProviderConfigs();
+  const {
+    data: toolConfigs,
+    isLoading: isLoadingTools,
+    isError: isToolsError,
+  } = aiToolConfigQueries.useAiToolConfigs();
```
othmanemad, line 41: `useAiToolConfigs` has no `enabled` guard, but `aiToolConfigModule` is registered only for CLOUD and ENTERPRISE — on Community this 404s three times for data the hidden Capabilities card never reads. Needs `enabled: !isCommunity`.

packages/web/src/app/routes/platform/overview/ai-overview.tsx
```diff
@@ -0,0 +1,176 @@
...
+          <OverviewRow
+            tone={chatProvider === undefined ? 'off' : 'ok'}
+            label={
+              chatProvider === undefined
+                ? t('No provider enabled for chat')
+                : chatProviderName
+            }
+          >
+            <Button variant="outline" size="sm" asChild>
+              <a href="/platform/setup/ai?tab=providers">{t('Change')}</a>
```
othmanemad: Plain `<a href>` to an in-app path forces a full reload and drops the query cache — use `<Link to>` like the rows above.

packages/web/src/app/routes/platform/overview/infrastructure-overview.tsx
```diff
@@ -0,0 +1,193 @@
...
+    >
+      <OverviewCards>
+        <OverviewCard
+          to="/platform/infrastructure?tab=health"
+          title={t('Health')}
+          value={needsAttention ? t('Attention') : t('Healthy')}
+          isLoading={isLoadingHealth}
+          isError={isHealthError}
+          errorEntity={t('health checks')}
+          description={
```
othmanemad, line 91: `description` keys off `isUpToDate` while `value` keys off `needsAttention`, so latest release + failed `appRam` reads "Attention" and "All checks passed" at once.

packages/web/src/app/routes/platform/admin-pages.ts
```diff
@@ -0,0 +1,520 @@
...
+  return page.nav !== undefined;
+}
+
+function visibleNavPages(context: AdminPageContext) {
+  return ADMIN_PAGES.filter(hasNav).filter(
+    (page) => page.nav.isHidden?.(context) !== true,
+  );
+}
+
+function visibleTabs(page: AdminPage, context: AdminPageContext) {
```
othmanemad: These helpers take positional args — CLAUDE.md asks for one destructured object, and `activeTabId(page, context, requested)` is the easiest to mis-order at a call site.

packages/web/src/app/routes/platform/admin-pages.ts
```diff
@@ -0,0 +1,520 @@
...
+export type AdminPage = {
+  id: string;
+  path: string;
+  title: string;
+  description?: string;
+  component?: ComponentType;
+  overview?: ComponentType;
+  tabs?: AdminPageTabSpec[];
+  nav?: AdminPageNav;
+  sample?: boolean;
```
othmanemad, line 507: `teaser` and `sample` are never read anywhere, and `AdminPage.description` is set on 8 entries but read nowhere (each overview hardcodes its own) — the `FeatureTeaserProps` import exists only to type these.

othmanemad (commented):
Review pass over the registry rewrite. The gating audit came back clean — the `PlatformLayout` admin guard is intact, every flag-gated tab still carries its own guard, all 17 legacy paths redirect, and I found no missing i18n keys across the 33 files. No security regressions; the diff is frontend-only.

The inline notes below are the leftovers: three Tailwind arbitrary values, five loading/error states that render zeroed data as fact, and two convention items.

One finding that has no line to sit on: `app/components/global-search/static-pages.ts` still holds its own 19-entry list of platform-admin pages, untouched by this PR. Every path still resolves, but 11 of them only via `ADMIN_REDIRECTS`, the labels name pages that no longer exist ("Audit Logs", "Workers"), and there is no entry for the new overviews. That is the drift the registry exists to remove, so it is worth folding into `ADMIN_PAGES` while the rewrite is open.

othmanemad (approved):

PR #15598, opened 2026-09-17
fix(web): resolve audience:'ai' actions when rendering placed steps

ibrahim-abuznaid:
## Description

Steps that use an `audience: 'ai'` action render broken in the builder: no piece icon, no action name, an **empty settings panel with no props**, no referenceable output fields in the data selector, and a broken Test Step.

`GET /v1/pieces/:name` applies `filterActionsByAudience`, which defaults to `PieceAudienceFilter.HUMAN` when the caller sends no `audience` (`piece-metadata-service.ts`). That default is right for **discovery** — the step picker should not offer AI-only actions. But the same endpoint is also the **resolver** for steps already saved in a flow, and those callers never passed `audience`, so the action they were looking up had just been filtered out of the response.

Agents can add these actions through the MCP/API surface, and they execute correctly, so this only affects how an existing flow renders for a human.

Reproduced on a local instance — the exact request the step settings panel sends, against a Slack piece with 44 `audience: 'ai'` actions:

| Request | Response |
|---|---|
| `?version=0.18.0&locale=en` | 28 actions, `slack_search_messages` **absent** |
| `?version=0.18.0&locale=en&audience=all` | 72 actions, present, with its `props` and `outputSchema` |

One root cause, several symptoms:

| Call site | Symptom |
|---|---|
| `step-utils.tsx` `getMetadata` | `piece.actions[actionName].displayName` threw on the missing key, rejecting the react-query that backs the canvas node, minimap, drag overlay and run-log dialog — the step rendered with no icon and no action name |
| `pieces-hooks.ts` `usePiece` | the shared metadata hook behind step settings (via `usePieceModelForStepSettings` -> `StepSettingsProvider` -> `form-utils.buildInputSchemaForStep`), Test Step, the run-log input/output panel and the `outputSchema` hook — the panel had no action, so it built **no props** |
| `data-selector/index.tsx` | `outputSchema` resolved to `null`, so an AI action's output fields could not be referenced downstream |
| `update-piece-version-utils.tsx` | always threw *"The selected version does not include the current action or trigger"* for a step using an AI action |

`usePiece` now takes an optional `audience` and threads it into both the request and the **react-query key** — without the key change a HUMAN-filtered piece could be served from cache to a caller asking for `ALL`. `PieceAudienceFilter.ALL` is then passed at the sites that resolve an already-placed step, matching the existing precedent in `agent-tool-hooks.ts`:

- `usePieceModelForStepSettings` (step settings props form)
- `test-step/test-runner-context.tsx` and `test-step/test-action-section.tsx` (Test Step)
- `run-details/flow-step-input-output.tsx` (run log)
- `step-settings/piece-settings/connection-select.tsx` (connection dropdown; also keeps it on one cache entry with step settings)
- `features/pieces/hooks/use-piece-output-schema.ts` (`outputSchema` for the data mapper)

Discovery surfaces keep the default and are deliberately untouched, so AI-only actions stay hidden from the picker: `pieces-selector/index.tsx`, the AI/agents tab and the approvals tab. The platform piece-visibility sheet is also left as-is — whether admins should see AI-only actions there is a product decision, not part of this fix.

Two further sites came out of review (thanks @hazemadelkhalel):

- **`useMultiplePieces`** was the one `usePiece`-family hook without an `audience` param, and `mcp-server/activity/activity-feed.tsx` uses it to resolve the display name of an action an MCP agent actually ran. It degraded rather than crashed — `activity-utils.ts` falls back to `convertEnumToHumanReadable` — so the audit feed showed a *guessed* label for exactly the actions most likely to be AI-only. It now takes `audience` in both the request and the key, and the activity feed passes `ALL`. The approvals tab, the other caller, keeps the human-filtered default; threading `audience` into the key is what keeps those two off one cache entry.
- **`data-selector/index.tsx`** fetched with `ALL` but built its own key `['piece', name, version]`, recording neither `audience` nor `locale`. It and `usePiece` now build their keys through one exported `pieceQueryKey` helper, so every `piece` detail query has the same shape.

Also collapses the duplicated action/trigger lookup in `getMetadata` into one null-safe read, so an unknown action name degrades to an empty label instead of throwing.

Introduced in #13960, which added the audience filter to both the list endpoint (intended) and the single-piece endpoint (the resolver).

A follow-up worth considering separately: making `audience` required on `GetPieceRequestQuery`, or splitting the discovery and resolve endpoints, would turn this class of bug into a compile error rather than a silently filtered response.

## How was this tested?

Cloud, against a flow containing two `audience: 'ai'` actions (`gmail_search_email` -> `gmail_get_message`, chained via `{{step_1['output'].results.messages[0].id}}`):

- Reproduced the bug before the change: both steps rendered with blank icon placeholders and no action name, while the flow itself ran fine (`ap_validate_flow` valid, single-step and end-to-end runs both SUCCEEDED). Confirms this is presentation-only, not an execution regression.
- Verified the filter direction: searching Gmail in the step picker returns exactly the 11 actions tagged `human`/untagged and none of the 26 tagged `ai` — unchanged by this PR.

Local instance, for the settings panel: placed a Slack `audience: 'ai'` action in a flow and confirmed the empty-props panel, then the request/response table above pinning the cause to the missing `audience`.

Swept every file that indexes `piece.actions[...]` / `piece.triggers[...]` by step name to confirm no resolver was left behind: each one either passes `ALL` itself or receives its `pieceModel` from a caller that does, except the discovery surfaces listed above.

Gates, from `packages/web` so the `@/` alias resolves:

- `npx eslint` on the changed source and test files: clean.
- `npx tsc --noEmit -p tsconfig.app.json`: 132 errors, byte-identical to pristine `upstream/main` — pre-existing only, none in the changed files. `tsconfig.spec.json` reports no errors in the new test files.
- `npx vitest run test/features/pieces`: 7 files, 25 tests passing.

Unit tests, written by @hazemadelkhalel during review and landed here:

| Test | Covers |
|---|---|
| `packages/server/api/test/unit/piece-audience-filter.test.ts` | `filterActionsByAudience` drops `audience:'ai'` for both no-audience and `HUMAN`, keeps it for `ALL` |
| `packages/web/test/features/pieces/utils/step-utils-audience.test.ts` | `getMetadata` asks for `audience=all`, resolves the display name, and no longer throws when the action is absent from the response |
| `packages/web/test/features/pieces/pieces-hooks-audience.test.tsx` | `usePiece` cache isolation — discovery caller gets the filtered piece, the resolver gets the AI action, exactly two fetches on a shared `QueryClient` |
| `packages/web/test/features/pieces/pieces-hooks-multiple-audience.test.tsx` | `useMultiplePieces` carries `audience` in request and key, and the default caller stays human-filtered |

The `useMultiplePieces` file fails 3/3 with the hook change stashed, so it guards the behaviour rather than restating it. The server test is not reached by CI (`test/unit` is outside the `api` test targets) and was run in a separate fully-installed checkout: 3/3.

Fixes # (issue)


packages/web/src/features/pieces/hooks/pieces-hooks.ts
```diff
@@ -73,6 +74,7 @@ type UsePieceProps = {
   version?: string;
   enabled?: boolean;
   projectId?: string;
+  audience?: PieceAudienceFilter;
```
hazemadelkhalel, line 77: `useMultiplePieces` (line 163) never got this param, and `app/routes/mcp-server/activity/activity-feed.tsx:84` uses it to resolve MCP action names at line 113. The chain is reachable: `ap-research-pieces.ts:147` serves AI-only actions to agents with `ALL`, `ap-run-action.ts` runs them, `mcp-activity-service.ts:128` records the name. It degrades rather than crashes because `activity-utils.ts:25` falls back to `convertEnumToHumanReadable`, so the audit feed shows a guessed label for exactly the actions MCP is most likely to run.

When you fix it, thread it into the key too. `useMultiplePieces`'s key is `['piece', name, undefined, i18n.language]`, so adding it to the request alone would reintroduce the poisoning you just fixed here, with `approvals-tab-content.tsx:60` (default) and the activity feed (`ALL`) colliding on one entry.

packages/web/src/features/pieces/hooks/pieces-hooks.ts
```diff
@@ -100,16 +102,23 @@ type UsePiecesSearchProps = {
...
+    name,
+    version,
+    enabled = true,
+    projectId,
+    audience,
+  }: UsePieceProps) => {
     const { i18n } = useTranslation();
     const query = useQuery<PieceMetadataModel, Error>({
-      queryKey: ['piece', name, version, i18n.language, projectId],
+      queryKey: ['piece', name, version, i18n.language, projectId, audience],
```
hazemadelkhalel: Confirmed this is load bearing, not belt-and-braces. Rendering a discovery caller then a resolver caller against one QueryClient: on `main` the resolver reads the human-filtered entry and the AI action is missing; with this line it is 2 fetches and the resolver gets its own data.

packages/web/src/features/pieces/utils/step-utils.tsx
```diff
@@ -72,30 +73,28 @@ export const stepUtils = {
...
-            : piece.triggers[step.settings.triggerName!].description;
+            ? piece.actions[step.settings.actionName!]
+            : piece.triggers[step.settings.triggerName!];
         return {
           ...metadata,
           errorHandlingOptions: mapErrorHandlingOptions(piece, step),
-          actionOrTriggerOrAgentDescription: actionOrTriggerDescription,
-          actionOrTriggerOrAgentDisplayName: actionOrTriggerDisplayName,
+          actionOrTriggerOrAgentDescription: actionOrTrigger?.description ?? '',
+          actionOrTriggerOrAgentDisplayName: actionOrTrigger?.displayName ?? '',
```
hazemadelkhalel, line 97: Reproduced the old behaviour on `main` with a stubbed `piecesApi`: `TypeError: Cannot read properties of undefined (reading 'displayName')`, exactly as described. `mapErrorHandlingOptions` was already null-safe via `selectedAction?.`, so this was the only throw site.

packages/web/src/app/builder/data-selector/index.tsx
```diff
@@ -174,6 +175,7 @@ const DataSelector = ({ parentHeight, parentWidth }: DataSelectorProps) => {
           name: pieceName,
           version: pieceVersion,
           locale: i18n.language as LocalesEnum,
+          audience: PieceAudienceFilter.ALL,
```
hazemadelkhalel, line 183: Not a bug today, since nothing else writes this key shape, but the key on line 172 is `['piece', name, version]` and now does not record the audience it fetched with. It already omitted `locale` on `main`, so this widens an existing gap by one field, one file away from the same defect you are fixing. Reusing `usePiece`'s key would close both.

hazemadelkhalel (commented):
Checked this out and ran it rather than reading it. The fix is correct and the diagnosis is right. One instance of the same bug class is still open, one hook away.

What I ran, on both `main` and this branch:

```
filterActionsByAudience (new server unit test)
  main + branch  ✓ no audience drops audience:'ai'      ✓ ALL keeps it

stepUtils.getMetadata
  main    × asks the API for every audience   → expected undefined to be 'all'
          ✓ resolves display name when present            (control)
          × does not throw when action was filtered out
            → TypeError: Cannot read properties of undefined (reading 'displayName')
  branch  ✓ ✓ ✓

usePiece audience cache isolation
  main    ✓ discovery caller gets human-filtered piece    (control)
          × resolver caller gets the AI action
          × resolver is not served the cached human-filtered piece
  branch  ✓ ✓ ✓  (exactly 2 fetches, one per audience)
```

So the query-key half of the fix is load bearing, not defensive. Without it the second caller reads the first caller's filtered entry. Good catch.

I also checked the inverse risk: every `audience:` passed anywhere in `packages/web` is `ALL`, so nothing accidentally narrows a picker to AI-only. And I verified the transitive chains hold rather than taking them on trust: `usePieceModelForStepSettings` has exactly one caller (`builder/index.tsx:139`), `StepSettingsProvider` one instantiation (`:226`), `buildPieceSchema` one call site (`step-settings-context.tsx:77`).

**The gap: `useMultiplePieces` is the one `usePiece`-family hook that did not get `audience`,** and `activity-feed.tsx:84` uses it to resolve MCP action names. Details in the inline comment. It degrades rather than crashes, so it is a reasonable follow-up rather than a blocker, but it is worth folding in here because it is the same bug on the surface where AI actions are most likely to appear.

**Recommendation: fold in the `useMultiplePieces` fix, then ship.** Everything else here is good.

One correction on the description: "`packages/web` currently has only pure-utility tests and no API-mocking pattern" is not the case. 41 test files under `packages/web/test` use `vi.mock`, and several mock API or hooks modules directly, including `test/features/piece-sets/piece-sets-api.test.ts` (mocks `@/lib/api`) and `test/app/variables/variable-dialog.test.tsx` (mocks `@/features/variables/api/variables`). The paths in this PR stub cleanly. I wrote the four tests above while reviewing and will post them in a follow-up comment so you can drop them in if you want them.

hazemadelkhalel:
The four tests, as promised.

On the pattern question, the closest precedent is already sitting next to the hook you changed: `packages/web/test/features/pieces/pieces-hooks.test.tsx` mocks `@/features/pieces/api/pieces-api` and drives `piecesHooks.usePieces` through `renderHook` + `QueryClientProvider`. That is exactly the shape these needed, so stubbing `piecesApi` turned out to be a few lines.

Three of the files pass on your branch. The fourth is the `useMultiplePieces` gap, written as the behaviour we want, so it fails until that hook is threaded:

```
✓ utils/step-utils-audience.test.ts        (3)
✓ pieces-hooks-audience.test.tsx           (3)
× pieces-hooks-multiple-audience.test.tsx  (2)   <- the gap
✓ piece-audience-filter.test.ts            (3)   server, packages/server/api
```

The whole existing `test/features/pieces` suite still passes alongside them (22 passing, plus the 2 above). `npx eslint` is clean on all four. Note `packages/web`'s lint script only globs `src/**`, so nothing under `test/` is linted by CI, and there is one pre-existing prettier error in `pieces-api.test.ts` that is unrelated to this PR.

One caveat on the server test: `packages/server/api`'s `test-unit` target needed per-package `node_modules` present for bun's isolated linker to resolve `@activepieces/core-utils` from source. It runs fine in a normal checkout.

<details>
<summary>packages/server/api/test/unit/piece-audience-filter.test.ts</summary>

```ts
import { PieceAudienceFilter } from '@activepieces/shared'
import { describe, expect, it } from 'vitest'

import { filterActionsByAudience } from '../../src/app/pieces/metadata/utils'

const actions = {
    slack_send_message: { audience: 'human' },
    slack_list_channels: {},
    slack_search_messages: { audience: 'ai' },
} as never

const names = (audience: PieceAudienceFilter | undefined): string[] =>
    Object.keys(filterActionsByAudience(actions, audience)).sort()

describe('filterActionsByAudience', () => {
    it('drops audience:ai actions when the caller sends no audience', () => {
        expect(names(undefined)).toEqual(['slack_list_channels', 'slack_send_message'])
    })

    it('drops audience:ai actions for HUMAN', () => {
        expect(names(PieceAudienceFilter.HUMAN)).toEqual(['slack_list_channels', 'slack_send_message'])
    })

    it('keeps every action for ALL', () => {
        expect(names(PieceAudienceFilter.ALL)).toEqual(['slack_list_channels', 'slack_search_messages', 'slack_send_message'])
    })
})
```

</details>

<details>
<summary>packages/web/test/features/pieces/utils/step-utils-audience.test.ts</summary>

```ts
/**
 * @vitest-environment jsdom
 */
import { LocalesEnum } from '@activepieces/core-utils';
import { FlowActionType } from '@activepieces/shared';
import { beforeEach, describe, expect, it, vi } from 'vitest';

const { get } = vi.hoisted(() => ({ get: vi.fn() }));

vi.mock('i18next', () => ({ t: (k: string) => k }));
vi.mock('@/features/pieces/api/pieces-api', () => ({ piecesApi: { get } }));

import { stepUtils } from '@/features/pieces/utils/step-utils';

const AI_ACTION = 'slack_search_messages';

const pieceWith = (actions: Record<string, unknown>) => ({
  name: 'slack',
  displayName: 'Slack',
  version: '0.18.0',
  logoUrl: 'logo.png',
  actions,
  triggers: {},
  auth: undefined,
  categories: [],
});

const humanFilteredSlack = () =>
  pieceWith({
    slack_send_message: { displayName: 'Send Message', description: 'd' },
  });

const unfilteredSlack = () =>
  pieceWith({
    slack_send_message: { displayName: 'Send Message', description: 'd' },
    [AI_ACTION]: {
      displayName: 'Search messages',
      description: 'Search Slack',
    },
  });

const step = {
  name: 'step_1',
  type: FlowActionType.PIECE,
  displayName: 'Search messages',
  valid: true,
  settings: {
    pieceName: 'slack',
    pieceVersion: '0.18.0',
    actionName: AI_ACTION,
    input: {},
    inputUiInfo: {},
  },
} as never;

beforeEach(() => get.mockReset());

describe('stepUtils.getMetadata for an audience:ai action', () => {
  it('asks the API for every audience, not just human', async () => {
    get.mockResolvedValue(unfilteredSlack());
    await stepUtils.getMetadata(step, LocalesEnum.ENGLISH);
    expect(get).toHaveBeenCalled();
    for (const call of get.mock.calls) {
      expect(call[0].audience).toBe('all');
    }
  });

  it('resolves the display name when the action is present', async () => {
    get.mockResolvedValue(unfilteredSlack());
    const metadata = await stepUtils.getMetadata(step, LocalesEnum.ENGLISH);
    expect(metadata.actionOrTriggerOrAgentDisplayName).toBe('Search messages');
  });

  it('does not throw when the action was filtered out of the response', async () => {
    get.mockResolvedValue(humanFilteredSlack());
    await expect(
      stepUtils.getMetadata(step, LocalesEnum.ENGLISH),
    ).resolves.toBeDefined();
  });
});
```

</details>

<details>
<summary>packages/web/test/features/pieces/pieces-hooks-audience.test.tsx</summary>

```tsx
/**
 * @vitest-environment jsdom
 */
import { PieceAudienceFilter } from '@activepieces/shared';
import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
import { renderHook, waitFor } from '@testing-library/react';
import * as React from 'react';
import { beforeEach, describe, expect, it, vi } from 'vitest';

const { get } = vi.hoisted(() => ({ get: vi.fn() }));

vi.mock('react-i18next', () => ({
  useTranslation: () => ({ i18n: { language: 'en' } }),
}));
vi.mock('@/features/pieces/api/pieces-api', () => ({ piecesApi: { get } }));

import { piecesHooks } from '@/features/pieces/hooks/pieces-hooks';

const AI_ACTION = 'slack_search_messages';

const slack = (includeAiAction: boolean) => ({
  name: 'slack',
  displayName: 'Slack',
  version: '0.18.0',
  actions: {
    slack_send_message: { displayName: 'Send Message' },
    ...(includeAiAction
      ? { [AI_ACTION]: { displayName: 'Search messages' } }
      : {}),
  },
  triggers: {},
});

const wrapper = ({ children }: React.PropsWithChildren) => {
  const client = new QueryClient({
    defaultOptions: { queries: { retry: false } },
  });
  return <QueryClientProvider client={client}>{children}</QueryClientProvider>;
};

beforeEach(() => {
  get.mockReset();
  get.mockImplementation(({ audience }: { audience?: PieceAudienceFilter }) =>
    Promise.resolve(slack(audience === PieceAudienceFilter.ALL)),
  );
});

describe('usePiece audience cache isolation', () => {
  it('serves the discovery caller a human-filtered piece', async () => {
    const { result } = renderHook(
      () => piecesHooks.usePiece({ name: 'slack', version: '0.18.0' }),
      { wrapper },
    );
    await waitFor(() => expect(result.current.pieceModel).toBeDefined());
    expect(result.current.pieceModel?.actions[AI_ACTION]).toBeUndefined();
  });

  it('serves the resolver caller the AI action', async () => {
    const { result } = renderHook(
      () =>
        piecesHooks.usePiece({
          name: 'slack',
          version: '0.18.0',
          audience: PieceAudienceFilter.ALL,
        }),
      { wrapper },
    );
    await waitFor(() => expect(result.current.pieceModel).toBeDefined());
    expect(result.current.pieceModel?.actions[AI_ACTION]).toBeDefined();
  });

  it('does not serve a cached human-filtered piece to a resolver in the same client', async () => {
    const client = new QueryClient({
      defaultOptions: { queries: { retry: false } },
    });
    const sharedWrapper = ({ children }: React.PropsWithChildren) => (
      <QueryClientProvider client={client}>{children}</QueryClientProvider>
    );

    const { result: discoveryResult } = renderHook(
      () => piecesHooks.usePiece({ name: 'slack', version: '0.18.0' }),
      { wrapper: sharedWrapper },
    );
    await waitFor(() =>
      expect(discoveryResult.current.pieceModel).toBeDefined(),
    );

    const { result: resolverResult } = renderHook(
      () =>
        piecesHooks.usePiece({
          name: 'slack',
          version: '0.18.0',
          audience: PieceAudienceFilter.ALL,
        }),
      { wrapper: sharedWrapper },
    );
    await waitFor(() =>
      expect(resolverResult.current.pieceModel).toBeDefined(),
    );

    expect(resolverResult.current.pieceModel?.actions[AI_ACTION]).toBeDefined();
    expect(get).toHaveBeenCalledTimes(2);
  });
});
```

</details>

<details>
<summary>packages/web/test/features/pieces/pieces-hooks-multiple-audience.test.tsx</summary>

```tsx
/**
 * @vitest-environment jsdom
 */
import { PieceAudienceFilter } from '@activepieces/shared';
import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
import { renderHook, waitFor } from '@testing-library/react';
import * as React from 'react';
import { beforeEach, describe, expect, it, vi } from 'vitest';

const { get } = vi.hoisted(() => ({ get: vi.fn() }));

vi.mock('react-i18next', () => ({
  useTranslation: () => ({ i18n: { language: 'en' } }),
}));
vi.mock('@/features/pieces/api/pieces-api', () => ({ piecesApi: { get } }));

import { piecesHooks } from '@/features/pieces/hooks/pieces-hooks';

const AI_ACTION = 'slack_search_messages';
const AI_ACTION_DISPLAY_NAME = 'Search messages in a channel';

const slack = (includeAiAction: boolean) => ({
  name: 'slack',
  displayName: 'Slack',
  actions: {
    slack_send_message: { displayName: 'Send Message' },
    ...(includeAiAction
      ? { [AI_ACTION]: { displayName: AI_ACTION_DISPLAY_NAME } }
      : {}),
  },
  triggers: {},
});

const wrapper = ({ children }: React.PropsWithChildren) => (
  <QueryClientProvider
    client={new QueryClient({ defaultOptions: { queries: { retry: false } } })}
  >
    {children}
  </QueryClientProvider>
);

beforeEach(() => {
  get.mockReset();
  get.mockImplementation(({ audience }: { audience?: PieceAudienceFilter }) =>
    Promise.resolve(slack(audience === PieceAudienceFilter.ALL)),
  );
});

describe('useMultiplePieces, used by the MCP activity feed', () => {
  it('requests every audience, so a saved MCP action can be resolved', async () => {
    const { result } = renderHook(
      () => piecesHooks.useMultiplePieces({ names: ['slack'] }),
      { wrapper },
    );
    await waitFor(() => expect(result.current[0].data).toBeDefined());
    expect(get).toHaveBeenCalledWith(
      expect.objectContaining({ audience: PieceAudienceFilter.ALL }),
    );
  });

  it('resolves the display name of an action an MCP agent actually ran', async () => {
    const { result } = renderHook(
      () => piecesHooks.useMultiplePieces({ names: ['slack'] }),
      { wrapper },
    );
    await waitFor(() => expect(result.current[0].data).toBeDefined());
    const piecesByName = new Map(
      result.current
        .map((query) => query.data)
        .filter((piece) => piece !== undefined)
        .map((piece) => [piece.name, piece]),
    );
    expect(piecesByName.get('slack')?.actions?.[AI_ACTION]?.displayName).toBe(
      AI_ACTION_DISPLAY_NAME,
    );
  });
});
```

</details>

packages/web/src/features/pieces/hooks/pieces-hooks.ts
ibrahim-abuznaid: Folded in, in `a2547d03ab`.

`useMultiplePieces` now takes `audience` and threads it into the request **and** the key, via a shared `pieceQueryKey` helper. `activity-feed.tsx` passes `ALL`; `approvals-tab-content.tsx` keeps the default. Also split the props type so `usePieceSummariesByNames`, which shares it, does not inherit a param it ignores.

One deliberate divergence from your test file: it renders `useMultiplePieces({ names: ['slack'] })` and expects the request to carry `ALL`, which reads as the hook hardcoding `ALL` rather than taking a param. I went with the param, because the approvals tab is a discovery surface and hardcoding `ALL` would hand it unfiltered metadata — the leak this PR is otherwise careful to avoid. Your own note here is the argument for it: the collision you describe between `approvals-tab-content.tsx` (default) and the activity feed (`ALL`) only exists if the two callers can ask for different audiences. So I kept both your assertions and passed `audience: ALL` at the call site, and added a third case covering what your comment predicts — the picker caller stays human-filtered, the feed caller gets the AI action, two fetches, no collision.

Stashing the hook change fails all three, so it is guarding the behaviour and not restating it. Say the word if you'd rather the hook just hardcoded `ALL` and I'll flip it.

packages/web/src/app/builder/data-selector/index.tsx
ibrahim-abuznaid: Closed in `a2547d03ab`. Extracted the key into an exported `pieceQueryKey` helper and pointed this block and `usePiece` at it, so `audience` and `locale` are both recorded here now and there is one key shape for every `piece` detail query instead of three hand-written ones.

ibrahim-abuznaid:
Thanks for running it rather than reading it — and for the correction on the description, which was wrong: `test/features/pieces/pieces-hooks.test.tsx` is the precedent, sitting next to the hook I changed. The body no longer claims otherwise.

Both review points are folded into `a2547d03ab`, and all four of your test files are in:

| | |
|---|---|
| `useMultiplePieces` gap | `audience` in request + key; activity feed passes `ALL`, approvals tab keeps the default |
| `data-selector` key | both it and `usePiece` build keys through one exported `pieceQueryKey` |
| tests | 4 files, 12 tests |

Verified: `vitest run test/features/pieces` is 7 files / 25 tests green, including your three and the existing suite. `pieces-hooks-multiple-audience` fails 3/3 with the hook change stashed. The server test needed a fully-installed checkout to run (3/3 there) — worth flagging that `test/unit` is outside the `api` targets CI runs, so nothing in that directory is executed on a PR.

One divergence from your `useMultiplePieces` test, argued in the thread above: I made `audience` a param instead of hardcoding `ALL` in the hook, so the approvals tab stays human-filtered. Happy to flip it if you disagree.

ibrahim-abuznaid:
Verified end-to-end on a fresh instance built from `main` plus this PR.

Setup: `upstream/main` `12b1ff60ac` + the three commits (cherry-picked, no conflicts), fresh PGlite DB, dev pieces `gmail 0.16.0` and `slack 0.20.0`, real Gmail and Slack connections. The flow was built through the MCP, so it holds two `audience:'ai'` actions that cannot be added from the builder's picker: `gmail_search_email` and `slack_post_message`.

| Check | Result |
|---|---|
| Canvas renders both AI steps (icon + name) | pass |
| Step settings props panel populates | pass |
| Test Step - Gmail search | pass |
| Test Step - Slack post | pass |
| Run-detail panel renders both AI steps | pass |

Flow published and run end to end twice in PRODUCTION, both SUCCEEDED.

Server-side on the same build: `piece-gmail` returns 11 actions by default, 27 with `audience=ai`, 37 with `audience=all`; `piece-slack` 29 / 52 / 73.

PR #15612, opened 2026-09-17
fix(pieces): connection fields now translate for pieces with several sign-in methods

### LINKED ISSUE #15586: [BUG] Auth fields of pieces with multiple auth methods are never translated
## Problem

When a piece declares several auth methods (`auth: [authA, authB]`), none of its connection texts are translated: auth descriptions, field labels and field descriptions stay in English in every locale.

They are missing from the piece's `src/i18n/translation.json`, so they never reach Crowdin. Example: Umami (`umamiAuth = [selfHostedAuth, cloudAuth]`) still defines `Instance URL`, `Username`, `Password`, `API Key` and their descriptions in `lib/auth.ts`, but the generator dropped them from `translation.json` in #12530. Same for Slack, SavvyCal and Google Sheets: none of their auth texts are in their `translation.json`.

## Root cause

`pieceTranslation.pathsToValuesToTranslate` in `packages/pieces/framework/src/lib/i18n.ts` only walks a single auth object (`auth.props.*.displayName`, `auth.description`, …). When `auth` is an array, these paths resolve to `undefined`, so nothing is found.

Both sides rely on these paths:
- extraction: `generateTranslationFileFromPiece` in `packages/cli/src/lib/commands/generate-translation-file-for-piece.ts`
- display: `pieceTranslation.translatePiece`

Adding the keys to Crowdin by hand would therefore not help either.

## Affected pieces (30)

amazon-bedrock, amazon-s3, amazon-secrets-manager, aws-bedrock, baremetrics, clicdata, digital-ocean, fathom, getresponse, github, gmail, google-bigquery, google-calendar, google-docs, google-drive, google-forms, google-sheets, hubspot, intercom, klaviyo, mistral-ai, notion, savvycal, sendgrid, sign-now, slack, snowflake, teable, tenzo, umami

## Expected

Each entry of an auth array is extracted and translated like a single auth. There is also no path for `auth.displayName`, which is what labels each method when a piece offers several.

Related: #13530

majewskibartosz:
Fixes [GIT-1893](https://linear.app/activepieces/issue/GIT-1893)
Closes #15586

Investigation write-up: https://claude.ai/artifact/A7NU3FyrvvYWwMmtWNfCVe

## Problem
1. A piece that offers several ways to connect declares `auth: [authA, authB]`. `pieceTranslation.pathsToValuesToTranslate` describes a single auth object, so `auth.props.*.displayName` and friends resolve to `undefined` against an array and the walk stops.
2. Nothing is translated at serve time (auth descriptions, field labels, field descriptions stay English in every locale) **and** nothing is extracted at build time, so the strings never reach Crowdin. 30 pieces: slack, github, gmail, notion, google-sheets, google-drive, hubspot, snowflake, umami, and 21 more.

## Fix
- `pieces-framework/src/lib/i18n.ts` → `translateProperty` distributes the remaining path across an array's entries. Placed **after** the `*` branch: a path that already iterates a container (`options.options.*.label`) must keep consuming it there, or it would iterate each option's own values and look for `.label` on a string.
- `cli/.../generate-translation-file-for-piece.ts` → same rule in `getPropertyValue`, mirroring the `*` branch's `.map().filter(Boolean).flat()` so results stay flat. Exported for the regression test.
- `ci.yml` + root `test-unit` → add `@activepieces/cli`. It has had a test suite and a `test` script but sits in no CI filter, so none of it ran; the new test would have been inert too.
- `cli/package.json` → declare `vitest` (3 existing test files import it, resolving only via root hoisting, which bun's isolated linker does not give you).

No `translation.json` was hand-edited. The nightly `generate-translations` workflow regenerates all 30, then Crowdin returns the locale files.

## Verification
- Real extractor, real built piece bundles: **umami 84 → 92 keys** (+8, exactly the strings the ticket names); **airtable single-auth 201 → 201, byte-identical**.
- Real `translatePiece` on real umami metadata with a German map → `Instanz-URL` / `Benutzername` / `Passwort`, unmatched keys untouched.
- New regression tests proven **red-first** against pre-fix code (framework 2 of 5 failed, cli 3 of 6), green after. Framework 20 passed, cli 39 passed.
- `packages/server/api` `piece-metadata.test.ts` (CE integration, covers `/api/v1/pieces?locale=de`): 33 passed.
- `npm run lint-dev` clean, 33/33 tasks, 0 errors. engine 515, shared 508, sandbox 273, core-execution 149 all pass.
- UNVERIFIED: full `npm run test-api` was not run (no source touched under `packages/server/api`); only the directly-relevant integration file above.
- Visual: none - no rendered change on this commit. No locale file contains the new keys until the nightly generator and a Crowdin round trip complete.
- Other affected paths: checked - none. The path list has exactly two consumers and both are fixed; every other auth-array consumer (`getAuthPropertyForValue`, the 4 `packages/web` call sites) already branches on `Array.isArray`.

**Deliberately out of scope:** `auth.displayName`, which labels each method in the picker. Any auth without an explicit name falls back to `DEFAULT_CONNECTION_DISPLAY_NAME` (the literal `'Connection'`), and `multi-auth-list.tsx` decides what to render by string-comparing against it. Measured across the 30 pieces: 63 auth entries, 23 carry the sentinel, **3 of those are non-OAuth2** (`amazon-bedrock`, `amazon-secrets-manager`, `aws-bedrock`) and would show a translated "Connection" instead of their auth type. Making method labels translatable needs the picker to track whether a name was set explicitly, which is a UI change and a design call.

<details>

## Plan
- Root cause pinned before any edit: both the CLI extractor and the serve-time translator read the same static path list, so one rule fixes both halves.
- Chosen shape: one `Array.isArray` branch per walker, ordered after `*`. `Footprint:` estimated 2 files / ~8 lines; actual 2 files / 7 lines.
- Rejected: normalizing `auth` to an array inside `translatePiece` plus a duplicate set of array-specific paths (~30 lines, two walkers to keep in sync).
- Rejected: adding array handling at the top of each function, before the `*` check. Breaks `options.options.*.label`, traced by hand and confirmed against the option-label tests.

- Path already iterates the array (`options.options.*.label`) → `*` branch runs first and consumes it; covered by a test for both auth shapes.
- Empty auth array → `forEach` no-ops; extractor returns `[]`, contributing no keys. Covered.
- Array entry with no `props` (SecretText beside CustomAuth) → `filter(Boolean)` drops the undefined; covered by the umami-shaped fixture.
- Value with no locale entry → passes through unchanged; covered.
- Nested arrays flattening deeper than one level → each `*` and the array branch flatten at their own level, so results stay flat strings; verified by asserting every extracted key maps to itself.
- Self-referential metadata would recurse, but that is DoS-shaped and already contained by `translatePiece`'s try/catch.

## Alternatives rejected
- Hand-adding the keys to the 30 `translation.json` files → the extractor regenerates them nightly and would drop them again.
- Adding the keys to Crowdin directly → same, and the ticket notes it too.
- Fixing only the display side → the strings would still never reach Crowdin.
</details>

Purely additive: strings that were silently skipped are now extracted and translated. No API field, endpoint, column, required field or env var changes, and no action for self-hosters.

The change rewrites display strings on auth property **definitions** (labels and descriptions written in piece source), never connection **values** — no credential, token or `AppConnection` secret is in scope. The only write is `node[key] = translatedValue` where `key` comes solely from the hardcoded `pathsToValuesToTranslate`, so no attacker-controlled key reaches it.


majewskibartosz:
@greptileai review

majewskibartosz:
@greptile-apps review this PR

Issue #100, opened 2022-12-24
[Backend][Typescript] Instance Runs


abuaboud:
For the following instance run entity, we should support the following action.

1. List by project Id (Endpoint) 
2. Get by Id (Endpoint)
3. Create / Update Run ( Service layer only)

PR #15618, opened 2026-09-17
feat(instagram-business): revive on Graph v23.0 with 27 verified actions

OdaiAhmed99:
## Description

The Instagram for Business piece is effectively dead. It calls Graph **v17.0, which expired 2025-09-12**, so every request is served by whatever fallback version Meta picks. It ships 2 actions, 0 triggers and no output schemas, and `upload_reel` resolves with the boolean `false` after ~100s instead of throwing — a step that reports success while emitting garbage downstream.

This rebuilds it: Graph v23.0, the bugs fixed, and 25 new actions covering publishing, media, comments, insights and direct messages. All 27 actions are verified live against a real Instagram professional account, and each ships an `outputSchema` authored from its captured payload.

### What changed

- **Graph v17.0 → v23.0**, behind `INSTAGRAM_API_BASE`
- **Bugs fixed:** Upload Reel throws with Instagram's own status instead of returning `false`, and polls for five minutes per Meta's guidance rather than giving up after ~100s; Upload Photo waits on container status instead of publishing blind; the access token moves from the request body to the `Authorization` header; the `any` types are gone
- **25 new actions:** stories, carousels, publishing quota, media reads, carousel children, tagged media, delete media, comment enable/disable, 7 comment actions, account and media insights, 5 messaging actions
- **Output schemas for all 27 actions**, in a new `src/lib/output-schemas.ts`
- Actions reorganised into `account/`, `publishing/`, `media/`, `comments/`, `messaging/`
- `minimumSupportedRelease` 0.30.0 → 0.87.0

### OAuth scope expansion — requires re-consent

Adds `instagram_manage_comments`, `instagram_manage_insights`, `instagram_manage_contents`, `instagram_manage_messages`, `pages_read_engagement`, `pages_manage_metadata`, `pages_messaging`. **Existing Instagram connections must be reconnected.** `instagram_manage_messages` and `pages_messaging` need Meta App Review (Advanced Access) for production use; in Development mode they work only for users holding an Instagram Tester role on the app.

The Page dropdown now also stores the Facebook Page id, which messaging addresses. Steps saved before this release ask the user to re-select the Page rather than failing obscurely.

## How was this tested?

All 27 actions were run live against a real Instagram professional account through the Tier-2 harness (throwaway flow → `test-step` → assert on the captured payload), on CE with the piece hot-loaded via `AP_DEV_PIECES`. Payloads were captured to disk and every output schema is authored from a real capture, not from the docs.

Write actions were confirmed by reading the state back rather than trusting a `success` flag:

| Check | Evidence |
| --- | --- |
| `set_comment_visibility` | `get_comment` shows `hidden` flipping `true` → `false` |
| `set_media_comments_enabled` | `get_media` shows `is_comment_enabled` flipping |
| `delete_media` | `list_media` count drops 3 → 2 |
| `create_comment` | `get_media_insights` reports `comments: 1` |
| `reply_to_comment` | reply carries the original's `parent_id` |
| publishing | real photo, reel, story and carousel published; all three media types (IMAGE, VIDEO/REELS, CAROUSEL_ALBUM) captured |
| messaging | real DM sent and read back; conversation and message list verified |

Insights captures corrected the schema twice: a missing per-metric `id`, and account insights returning `total_value` where media insights return `values[]`. The conversations schema dropped `unread_count` and `message_count` — requested but never returned, so they would have been dead paths in the data selector.

## Known limitations

These are stated rather than hidden:

- **No webhook triggers.** Four were written and verified up to Meta's boundary — Meta's own verification GET is answered `200`, a wrong verify token gets `403`, and a replayed Meta payload produces a SUCCEEDED run with correct parsing — but **Meta never delivered a single event** across four real comments and a real DM. The ngrok request log shows zero POSTs ever reaching the callback, while Meta's verification GETs are present. Suspected Development-mode gating. The code is parked for a follow-up rather than shipped unproven.
- `list_tagged_media` only ever returned an empty list; its schema is reused from the (verified) `list_media`.
- Pagination (`next_cursor`) was never exercised — the test account has fewer than 25 items.
- `send_media_message` fails on some external image hosts (`gstatic.com` → `(#100) Upload failed`, `wikimedia.org` → `Upload attachment failure`) while Instagram's own CDN worked. Meta fetches the URL server-side and rejects hosts opaquely; noted in the prop description.
- Instagram exposes **no follower-list endpoint** — only `followers_count` / `follows_count`. Not a gap this PR can close.

## Dependency for Cloud users

**Instagram has no managed OAuth app at Activepieces**, so the piece cannot be connected on cloud at all, independently of this PR. `https://secrets.activepieces.com/apps?edition=cloud` lists 65 pieces with client ids; `facebook-pages` and `facebook-leads` share `1250436578920537` and Instagram is absent. Tracked separately — this PR is not blocked by it, but cloud users stay blocked until it is registered.

Fixes # (issue)

`upload_reel` no longer resolves `false` on timeout — it throws. Entry added to `docs/install/reference/breaking-changes.mdx`, and the new OAuth scopes require every existing Instagram connection to be reconnected.

Touches authentication: the piece requests additional OAuth scopes (comments, insights, content management, messaging, Page metadata), so users must re-consent. The access token also moves out of the request body into the `Authorization` header, keeping it out of any log that records request bodies. All outbound calls go through the piece's shared `graphRequest` helper against a single pinned Graph host.


ibrahim-abuznaid (commented):
Reviewed locally against a fresh worktree of this branch. The rebuild itself is strong: the `false`-on-timeout bug is real and genuinely fixed (old `createVideoPost` returned `false` on a failed poll), the token moved out of the request body into the header, the `any` types are gone, and every action carries `audience` / `aiMetadata` / `classification` / `outputSchema`. Build and lint both pass.

A few things I'd like fixed before this goes in.

### Blocking

**1. `bun.lock` was never regenerated.** `packages/pieces/community/instagram-business/package.json` says `1.0.0`; the committed `bun.lock` still records `0.2.7` for this workspace (`bun.lock:4556`). Run `bun install` and commit only this piece's lines — without it the version bump doesn't ship.

### Should fix

**2. `delete_media` and `delete_comment` are tagged `WRITE`; they're `DESTRUCTIVE`.** Both issue `HttpMethod.DELETE` and neither is reversible. Both actions' own `aiMetadata` already says "permanently … cannot be undone". Catalog precedent is 66 `DESTRUCTIVE` to 5 `WRITE` across `delete_*` actions.

`media/delete-media.ts:11` · `comments/delete-comment.ts:11`

**3. Eight `list_*` actions are tagged `READ`; they're `SEARCH`.** Enumerations returning zero-or-more results are SEARCH; `READ` is for a specific known resource addressed by a unique key. The sub-collection cases count too — github's `list_issue_comments` is SEARCH. Catalog precedent: 125 SEARCH vs 8 READ. `get_media`, `get_comment`, `get_profile` and both insights actions are correctly `READ`.

`media/list-media.ts:11` · `media/list-stories.ts:11` · `media/list-tagged-media.ts:11` · `media/list-carousel-children.ts:11` · `comments/list-media-comments.ts:11` · `comments/list-comment-replies.ts:11` · `messaging/list-conversations.ts:11` · `messaging/list-messages.ts:11`

**4. Three `Property.Array` props are used without a string guard.** There's no `ARRAY` entry in the engine's processor map (`packages/server/engine/src/lib/variables/processors/index.ts` — only `JSON` and `OBJECT` get parsed), so an array supplied through the dynamic-value toggle reaches `run()` as a raw string:

- `account/get-account-insights.ts:43` and `media/get-media-insights.ts:31` — `.map()` on a string throws `metrics.map is not a function`, which tells the user nothing.
- `publishing/publish-carousel.ts:43` — `items.length` counts characters, producing *"A carousel needs between 2 and 10 items, but 84 were supplied."*

The house pattern is `typeof x === 'string' ? JSON.parse(x) : x` (see `fireberry/create-record.action.ts:21`, `json/run-jsonata-query.ts:28`).

**5. `src/i18n/translation.json` is stale.** It still carries the old auth markdown and only `upload_photo` / `upload_reel` — all 25 new actions' display names and descriptions are absent, across 12 locale files. Build the piece, then `npm run cli -- pieces generate-translation-file instagram-business`.

**6. Two shared output schemas advertise keys the action never returns** — the same dead-path problem the description cites as the reason for dropping `unread_count` / `message_count`:

- `commentCreatedOutputSchema` (`output-schemas.ts:193`) declares both `media_id` and `parent_id`, but `create_comment` returns only `media_id` and `reply_to_comment` only `parent_id`.
- `listCommentsOutputSchema` (`output-schemas.ts:181`) declares `next_cursor`, which `list_comment_replies` never returns.

**7. PR body is missing a `Fixes #` line and a Linear link.**

<details>
<summary>Nits (7)</summary>

- `messaging/send-media-message.ts:38` — `url` description is 148 characters; keep it to one short sentence and move the host caveat to a placeholder.
- `common/index.ts:123` — `waitForContainer` treats an absent `status_code` as still-processing, so a container that never reports one burns the full 5 minutes before failing with "status unknown". Treating a missing code as terminal would fail fast. You verified all three media types live, so this is defensive only.
- `messaging/list-conversations.ts:37` — still requests `unread_count,message_count` in `fields` even though the schema (correctly) dropped them as never-returned.
- `common/index.ts:201` and `output-schemas.ts:35` — file order: imports → exported functions/constants → helpers → types.
- `media/list-tagged-media.ts:22` — `limit` has no description, while the identical `limit` on `list_media` does.
- `createCustomApiCallAction` is still missing from the piece. Pre-existing, but this PR is the natural moment.
- `getPages` doesn't paginate `/me/accounts`, so an account with more than 25 Pages loses entries from the dropdown. Pre-existing — the diff doesn't widen it.

</details>

### Looks right

Build and lint pass. `1.0.0` is the right segment for a behaviour change plus a scope change. Imports respect the core-package boundary (`PieceCategory` from `pieces-framework`, not `@activepieces/shared`), auth shape and dynamic props are fine, the output schemas are well-formed, and the labels line up with the breaking-changes docs entry. `minimumSupportedRelease: '0.87.0'` clears every framework floor the diff actually uses.

I did not re-run any of the live verification — taking the table in the description at face value.

PR #15658, opened 2026-09-20
feat(customer-io): add filtering Reporting Webhook trigger

AdamSelene:
## Description

The Customer.io piece has **no triggers** (`triggers: []`), so any flow reacting to Customer.io events must use a generic **Catch Webhook** — which creates one flow run per POST. A flow interested in a single campaign therefore ingests Customer.io's entire delivery firehose and discards almost all of it in-flow, wasting worker capacity.

This adds a **`New Reporting Event`** webhook trigger to `@activepieces/piece-customer-io`:

- **`onEnable`** registers a Customer.io [reporting webhook](https://docs.customer.io/integrations/api/webhooks/) (`POST /v1/reporting_webhooks`, App API, Bearer) scoped to the **event types** the user selects — so Customer.io only sends those at the source — and stores the returned id (webhook name is suffixed with the flow identifier so multiple flows stay distinguishable in Customer.io's UI).
- **`onDisable`** deletes that webhook (`DELETE /v1/reporting_webhooks/{id}`).
- **`run()`** optionally matches **Campaign ID / Action ID / recipient domain** and returns `[]` for non-matches, so unrelated events **never start a flow run** (the worker only creates a run when the trigger returns a non-empty array — the same mechanism the Linear `updated_issue` trigger uses).
- Region (US/EU) comes from the existing connection; auth reuses the piece's App-API Bearer token.

The full 70-entry event catalog (email/push/sms/whatsapp/slack/webhook/in_app/customer) is offered as a multi-select. Also adds vitest wiring to the piece (it had none) + a `run()` unit test, regenerates `src/i18n/translation.json`, and bumps the piece `0.3.8 → 0.3.9`.

Signature verification (`X-CIO-Signature`) is intentionally out of scope for this first trigger (parity with the Catch Webhook it replaces); the create API doesn't return the signing key, so it would be a connection field — a sensible follow-up.

## How was this tested?

- `npm run pieces generate-translation-file customer-io` (builds the piece; regenerates i18n) — succeeds.
- `vitest run` — 6/6 `run()` tests pass (matching campaign/action/domain returns the event; each non-match returns `[]`; domain matching is case-insensitive, tolerates a leading `@`, and rejects substring look-alikes like `notodoo.com`).
- `turbo run lint --filter=@activepieces/piece-customer-io` — 0 errors.

Fixes # (n/a)


AdamSelene:
Thanks @greptile-apps — both findings addressed in the latest commit:

- **P1 (webhook can become orphaned):** `onEnable` now rolls back — if `context.store.put` fails after the reporting webhook is created, it deletes the remote webhook (best-effort) before rethrowing, so a retry can't leave a duplicate active subscription.
- **P2 (cleanup failures hidden):** `onDisable` now suppresses only a confirmed **404** (already gone) and propagates auth/rate-limit/network/5xx errors, so the lifecycle framework can surface a failed teardown instead of leaving a stale subscription. Both paths go through a shared `deleteReportingWebhook` helper.

Build + `vitest` (6/6) + lint all green.

AdamSelene:
Thanks @greptile-apps. On the two new points:

**1. Recipient field — addressed (defensively).** The [reporting-webhook payload docs](https://docs.customer.io/integrations/data-out/connections/webhooks/) actually show the email under `data.recipient` (`"recipient": "test@example.com"`), not `data.email_address`. But to be safe across payload variants I now derive the recipient email from `data.recipient` → `data.email_address` → `data.identifiers.email` before domain-matching, with tests for the `email_address` and `identifiers.email` fallbacks. So `recipient_domain` works regardless of which field a given event carries.

**2. Event catalog — I believe this is a false positive; leaving the 70-entry list as-is.** Verified against Customer.io's official docs:
- `slack_undeliverable` **is** a valid Slack metric — it's listed under the Slack channel in both the [Webhooks API reference](https://docs.customer.io/integrations/api/webhooks/) and the [reporting-webhook docs](https://docs.customer.io/integrations/data-out/connections/webhooks/). Removing it would drop a real event.
- `email_deferred`, `slack_converted`, and `webhook_converted` are **not** Customer.io metrics — email has no `deferred`, and the Slack/Webhook channels have no `converted` (they only emit drafted/attempted/sent/clicked/failed/undeliverable). Adding them would put invalid values in the subscription.
- The per-channel catalog in the docs totals **70** (customer 3, email 13, push 11, sms 11, whatsapp 12, slack 6, webhook 6, in-app 8), which matches `CUSTOMER_IO_EVENTS`.

Happy to add any specific event if you can point at the doc entry, but I couldn't find those three in the reference. Build (piece `tsc`) clean, `vitest` 8/8, lint 0 errors.

AdamSelene:
Thanks — and good catch on the new P1.

- **Valid domain matches dropped:** fixed. The `??` chain took `data.recipient` first, which for non-email channels (sms/whatsapp/slack) is a phone/URL rather than an email — so the domain check missed the address in `identifiers.email`. Now the filter matches if **any** of `recipient` / `email_address` / `identifiers.email` is at the domain, so an SMS/WhatsApp event whose profile email is `…@odoo.com` still passes. Added a test with a phone-number `recipient` + matching `identifiers.email`.
- **72 → 70 in the description:** updated.

`tsc` clean, `vitest` 9/9, lint 0 errors.

OdaiAhmed99 (changes requested):
**Verdict: request changes**

Thanks @alan-eu, this is a solid trigger. I tested it end to end on a real Customer.io workspace. Creating the webhook, receiving events, all three filters, and deleting the webhook on disable all work. A few changes are needed before merge:

### P1
- **Version bump must be MINOR.** A new trigger is new functionality, so per `docs/build-pieces/piece-reference/piece-versioning.mdx` it's a minor bump. Main is at `0.4.0`, so this should be `0.5.0`, not `0.4.1`. The publish check fails on this.

### P2
- `new-reporting-event.ts:88`: **`aiMetadata` missing.** New triggers must ship `aiMetadata: { description }`, written for an agent: what fires it, what it needs, what it returns (`.agents/skills/piece-builder/ai-metadata.md`).
- `new-reporting-event.ts:88`: **`classification` missing.** Triggers are always `classification: 'READ'` (same file, and the team rubric).
- `new-reporting-event.ts:190`: `context.payload.body as CustomerIoReportingEvent`. Avoid the cast; use a type guard or a typed parse (AGENTS.md "No type casting").
- `new-reporting-event.ts:243`: `recipientMatchesDomain(recipient, domain)` takes positional params. Use one destructured object (AGENTS.md "Named parameters").
- `new-reporting-event.test.ts:17-18`: `buildContext(propsValue, body)` has the same two issues (positional params, `as` cast).

<details><summary>Nits (8)</summary>

- `new-reporting-event.ts:172`: remove the `// Rollback is best-effort…` comment (AGENTS.md: no comments in code). The reason can go in the PR description.
- Webhook name in **test mode**: during a test the webhook URL ends in `/test`, so `.pop()` gives `test` and the webhook is named `Activepieces (test)` for every flow (seen live). Taking the segment before `/test` would keep test webhooks distinguishable too.
- `new-reporting-event.ts:12`: file order. The `CUSTOMER_IO_EVENTS` constant sits above the exported trigger; move it below the logic (imports → exports → helpers → types).
- Prop descriptions for `events` (94 chars), `campaign_id` (106) and `recipient_domain` (102) are long. Keep them to one short sentence (~70 chars) and move the example (`odoo.com`) into a `placeholder`.
- `outputSchema` is missing on the new trigger. Fine if you state that the payload varies by `object_type`; otherwise a schema for the common envelope (`event_id`, `object_type`, `metric`, `timestamp`, `data.*`) would help the data selector.
- `X-CIO-Signature` isn't verified. The trigger rubric says signatures should be verified when the vendor offers them, but the create API doesn't return the signing key and it only exists after `onEnable` runs, so this can't be done cleanly here. Agree it's a follow-up ticket, not a blocker.
- Older code, not from this PR: `createPiece` typed `any` (`index.ts:36`), `index.ts:30` file order, `create_event` has no classification.
</details>

### PR metadata
- Needs labels: `🌟 feature` + `🧩 area/third-party-pieces` (a maintainer will add these).
- The body is out of date: it says `0.3.8 → 0.3.9` and "6/6 tests" (it's now 9). Please update it, add the Linear link (GIT-1915) and the Local verification note below. `Fixes #` could point to #15677.

### Unresolved review threads
None. 3 threads total, all resolved.

### Local verification
L0 build/lint/test/publish-check ❌ (publish failed) · L1 13 checks ❌ (0 P1, 6 P2 open)
L2 6/8 steps live ✅ · skipped: EU region (tested on one region only (the connection's own)), X-CIO-Signature verification (not implemented in the PR (out of scope))

AdamSelene:
Thanks for the thorough review @OdaiAhmed99 — addressed in 47c6d26365.

**P1**
- Version bump is now **0.5.0** (minor), from main's 0.4.0.

**P2**
- Added `aiMetadata.description` (agent-facing) and `classification: 'READ'` to the trigger.
- Dropped the `as CustomerIoReportingEvent` cast in `run()` — the body is now narrowed through an `isReportingEvent` type guard.
- `recipientMatchesDomain` takes a single destructured `{ recipient, domain }` object; `buildContext` in the test is now `{ propsValue, body }`.
- Test `as never`: kept, because it's the established context-shim idiom across piece tests (coupa, ai, fillout-forms all do the same) and there's no shared typed mock — building a full `TriggerContext` per test would diverge from every other piece. Happy to switch if you'd prefer a typed helper.

**Nits**
- Removed the `// Rollback is best-effort…` comment (rollback is now `deleteReportingWebhook(...).catch(() => undefined)`).
- Test-mode webhook name: now uses the segment before `/test`, so test webhooks stay named after the flow id instead of `Activepieces (test)`.
- Trimmed the long `events` / `campaign_id` / `recipient_domain` descriptions and moved the `odoo.com` example into a `placeholder`.
- `CUSTOMER_IO_EVENTS` file order: left at the top — it's referenced at module-eval time inside `createTrigger`, so moving it below the trigger would put it in the temporal dead zone. Left as-is intentionally.
- `outputSchema`: skipped for now since the payload shape varies by `object_type`; can add the common envelope in a follow-up.
- `X-CIO-Signature`: agreed, follow-up ticket.

**Metadata**
- Updated the PR body (version, 9/9 tests, Linear GIT-1915, `Fixes #15677`). Labels (`🌟 feature` + `🧩 area/third-party-pieces`) need a maintainer.

Local: lint 0 errors · build ✅ · 9/9 tests ✅.

AdamSelene:
Fixed — `bun.lock` now records `0.5.0` for the customer-io workspace, matching `package.json`. Kept the change to just that one line (I didn't sweep in unrelated lockfile drift for other pieces). `package.json` and `bun.lock` versions agree; piece is unchanged (vitest 9/9).

PR #15659, opened 2026-09-20
feat(router): let a branch condition ask a question instead of matching text

Louai-Zokerburg:
> **Stacked PR.** This branch sits on top of `feat/evaluation-models`, which is not pushed yet. Against `main` the diff therefore also carries that branch's two commits (`a51460662b`, `57359154d2` — evaluation models + Classify Text + the TypeSafe provider). Once the base branch is pushed, retarget this PR's base onto it and the diff becomes router-only: commits `d37225df7b`, `69d9e9be3a`, `d2d772a0e4`.

## Description

The router has 22 branch condition operators and every one of them is deterministic string, number, date or list matching. This adds a 23rd, `AI_MATCHES`, that takes a plain-English yes/no question about the first value plus a confidence level.

A condition row becomes:

```
First value   {{trigger.body.message}}
Condition     Answers yes to (AI)
Question      Is this refund request above our policy limit?
Take branch   Only when likely
```

Below the chosen confidence the branch is skipped. That is the point of the feature — a router that can decline to answer rather than guess. Zapier and n8n both ship an AI router; neither can do that.

### Who pays

The instance does. The condition calls Jev through **one Vercel AI Gateway key held by the deployment** (`AP_AI_GATEWAY_API_KEY`). There is no provider row, no project connection, no AI credit, and no provider/model picker in the builder.

The reasoning is the price. Jev costs $0.042 per 1M input tokens with free output, and a condition sends one sentence — a metering pipeline and a credit balance check would cost more to build and to run than the inference they would be accounting for. It also keeps the condition row to two inputs, which is most of what makes this usable.

The rejected alternative was resolving the project's own AI provider. It turns a one-line condition into a setup task and makes the feature invisible until someone connects a key. Written up in `brain/knowledge/decisions/000040-ai-router-conditions-run-on-an-instance-key-and-bill-nobody.md`.

### Zero setup

With `AP_AI_GATEWAY_API_KEY` unset, the `AI_ROUTER_ENABLED` flag is false and the operator is **absent from the condition dropdown**. A self-hoster who has not opted in sees nothing, rather than something that looks enabled and fails at run time.

### A latent bug fixed along the way

`evaluateConditions` looked its operator up in `CONDITION_EVALUATORS` and did `isNil(evaluate) → return true`. An engine older than the release that adds an operator therefore took **every branch using it, unconditionally** — on a refund flow, that approves every refund. Silent, and it bites on mixed-version fleets, a pinned self-hoster image, or a stale sandbox cache.

That default is now a throw (`UnknownOperatorError`). On a matched-version deploy the branch is unreachable, because the evaluator table is exhaustive over every synchronous operator, so the only behaviour this changes is the mixed-version case that was already wrong. Every future operator inherits the guard.

### Notes for review

- **`evaluateConditions` is now async and walks groups sequentially**, deliberately not `Promise.all`. An AND-group ordered `[TEXT_CONTAINS, AI_MATCHES]` never pays for the model call once the cheap condition has failed. There is a test pinning exactly that.
- **A failed or timed-out call fails the step**, it does not quietly take the fallback branch. Routing a refund down the wrong branch because a model timed out is a correctness bug that looks like normal operation.
- **No waitpoint.** Decision 000035 measures the pause/resume path at 150–400 ms warm, 1.1–1.6 s cold, which exceeds the 70–500 ms call it would be hiding. The cost accepted is that a flow worker thread blocks for the call, bounded by a 10 s abort.
- `ai-condition-api.ts` uses `global.fetch` with `AbortSignal.timeout(10_000)` rather than `retryFetch`, which retries 3× with a 3 s delay and would turn a 500 ms condition into a 9 s stall.
- **No migrations.** Additive enum, additive union arm, no existing flow references either, `LATEST_FLOW_SCHEMA_VERSION` unchanged.

## How was this tested?

**Automated** — `packages/server/engine/test/handler/conditions-evaluate.test.ts`, 69 passing. Six are new:

| Case | Asserts |
| --- | --- |
| above threshold | branch taken |
| below threshold | branch skipped |
| no threshold set | falls back to 0.5 |
| non-ok response | step fails, does not take the branch |
| unreachable / aborted | step fails |
| `[TEXT_CONTAINS(false), AI_MATCHES]` | **fetch is never called** — catches a silent latency regression |

Plus a new `unknown operator` block asserting the throw. `router-branching.test.ts` and `flow-branching.test.ts` (33 tests) still pass.

**Builds and lint** — `npx turbo run build` across api, worker, engine, server-utils and the core packages: 18/18. `tsc --noEmit` on the web app: clean. Lint: 0 errors in api, web and engine.

**Manual** — a two-branch router on a refund message. `"Please refund my £12 order"` takes the fallback; `"I want my £400 back"` takes branch 1; an ambiguous message drops below the threshold and takes the fallback on purpose. Removing the key makes the step fail visibly rather than silently routing.

**Editions** — CE with no key set: the operator does not appear in the dropdown and nothing else changes. Nothing here is EE-gated.

Fixes # (issue)

`AP_AI_GATEWAY_API_KEY` is **optional**. Unset, the feature is absent and every existing router behaves exactly as before. The unknown-operator change only alters a case that was already incorrect.

Three things for a security reviewer:

1. **New secret.** `AP_AI_GATEWAY_API_KEY` is read through `system` like `OPENROUTER_PROVISION_KEY` and never leaves the API process — it is not returned by the flag endpoint, which exposes only the boolean `AI_ROUTER_ENABLED`.
2. **Outbound HTTP, no SSRF surface.** The gateway URL is fixed in the SDK and is not user-controlled, so this is the same posture as the existing openai/anthropic provider calls and does not need `safeHttp`. The engine→API hop is internal and authenticated with the engine token on an existing `securityAccess.engine()` route.
3. **Flow data reaches a third party.** The condition's first value is sent to Vercel AI Gateway, so a flow can put customer data in front of a vendor the project did not configure itself. That is inherent to the instance-key model and is why the feature is off unless an operator sets the key.


Louai-Zokerburg:
Superseded — reopening cleanly. This PR was opened while the branch was still stacked on the evaluation-models work, so its description and history described a diff that no longer exists. The branch now carries the AI router condition only.

PR #15661, opened 2026-09-20
feat(dropbox): add AI-ready atomic actions for files and batch jobs

OdaiAhmed99:
## Description

Makes the Dropbox piece AI-ready: adds **26 `audience: 'ai'` atomic actions** so agents get a uniform surface, and removes the human-shaped duplicate pairs agents previously had to choose between.

Before this, Dropbox exposed 13 actions covering roughly a fifth of the account-scoped API, and agents had no way to read file metadata, revisions, previews or thumbnails at all.

**New capabilities the piece lacked**
- File metadata, revisions, and restore-to-revision
- Save-from-URL, zip folder download, document previews, thumbnails (single + batch)
- File locking (lock / unlock / read lock state)
- Copy references (cross-account copy), temporary upload links

**Async jobs are modelled as start + poll pairs.** Dropbox's `*_batch` endpoints return a job id rather than a result, so a single blocking action would time out on large batches and be indistinguishable from a failure. Both union branches are handled: these endpoints complete *inline* for small inputs and return `async_job_id` otherwise — both paths were exercised live.

**Duplicate-pair consolidation.** `copy` / `move` / `delete` were each split across separate file and folder actions over a single endpoint, a distinction that is meaningless to an agent. Each pair is replaced by one entry-based `ai` twin, and the six originals drop to `audience: 'human'`. That affects agent discovery only — `audience` is never read by the engine or `ap_run_action`, so existing flows are unaffected.

The other seven actions stay `'both'` so they remain in the step picker, which filters `audience: 'ai'` out by default.

Final split: **26 `ai` + 7 `both` + 7 `human`**. The human-visible surface is unchanged at 14.

Also raises `minimumSupportedRelease` from `0.30.0` to `0.86.4` — engines at or below `0.86.3` predate the canvas audience filter and would otherwise surface every atomic in the *human* step picker.

### Deliberately deferred

- **The `file_properties` namespace (9 actions) is not in this PR.** It is written and verified, but only against a custom OAuth app: the Activepieces cloud app (ID `12721888`) does not have `files.metadata.write` enabled, so the write operations return HTTP 400 on every `CLOUD_OAUTH2` connection. `auth.ts` already declares the scope — this needs an App Console change, not a code change. Follow-up PR once that lands.
- **`lock_dropbox_file` ships Tier-1 only.** Dropbox returns 403 on Basic plans. Confirmed a plan gate and not a scope gate by retrying on a second OAuth app carrying every scope — it failed identically. `unlock` and `get_file_lock` both pass, so only *taking* a lock is restricted.
- Out of scope entirely: the `sharing.*` namespace (42 live routes), `file_requests`, `account_info` and `contacts` — all need new OAuth scopes and a re-consent for every existing connection.

## How was this tested?

**Tier-1** — `turbo build` + `eslint` clean, 0 errors 0 warnings.

**Unit — 25 tests** (`vitest run`, following the existing per-piece pattern) over the shared transport and schema paths: `unwrapEntry` success / per-entry `failure` / empty / missing entries, the preview content-type branch, inline-vs-async union schemas, and per-entry failure reasons on batch status schemas.

**Tier-2 — every atomic in this PR run live against a real Dropbox account** on a local dev instance via the `test-step` path, with each `outputSchema` authored from the captured payload rather than from docs. Edition path: CE (`AP_EDITION=ce`), dev pieces hot-loaded. Only `lock_dropbox_file` is unverified, for the plan reason above.

Behavioural assertions, not just `SUCCEEDED`:
- `create_folder_batch` returned `complete` inline; copy/move/delete returned `async_job_id` and resolved on poll — both union branches exercised
- The three twins were each tested on **a file and a folder** — the whole point of consolidating them
- The lock trio was checked to throw on a per-entry `failure` rather than report HTTP 200 as success

Capture-driven corrections worth noting: `get_thumbnail_v2` returns metadata nested under `file_metadata` while `get_preview` returns it flat; batch failure entries are keyed `failure`, not `error`. Both would have shipped wrong from a docs-only reading.

Fixes PIE-528

Reasoning, so it can be challenged: the six demotions change agent *discovery* only and each intent stays covered by its twin; no existing flow changes behaviour and no action is required of self-hosters. The `minimumSupportedRelease` raise does mean instances on engines below `0.86.4` stop receiving new Dropbox versions until they upgrade — flagging that explicitly, since it is the one line here with upgrade impact.

No auth, secret or permission surface is added — `auth.ts` is untouched and no scope is requested. `save_url_to_dropbox` passes a user-supplied URL to Dropbox's API, but Dropbox performs that fetch from its own network, so it is not an SSRF vector on Activepieces infrastructure.


OdaiAhmed99:
Thanks — all five addressed. Summary of what changed and why.

**P1 · Property actions lack permission** — Agreed, and the nine `file_properties` actions are now **removed from this PR**. They were verified working, but only against a custom OAuth app; on standard cloud connections the five write operations return HTTP 400 because app `12721888` lacks `files.metadata.write`. Advertising them to every AI Dropbox connection while they cannot execute is the wrong trade, so they'll ship in a follow-up once the app-console change lands. I also removed the now-orphaned `property_template_ids` prop from Get File Metadata rather than leave a read path for a feature the PR no longer exposes.

**P1 · HTML previews become PDFs** — Real bug, fixed. `download` now returns the response `content-type`, and the extension follows it (`text/html` → `.html`, otherwise `.pdf`). The content type is also surfaced in the output so downstream steps can branch on it instead of guessing from the filename.

**P2 · Deletes classified as writes** — Fixed. `delete_dropbox_entry` and `start_dropbox_delete_batch` are now `DESTRUCTIVE`, matching the existing `delete_dropbox_file` / `delete_dropbox_folder`. Good catch: the inconsistency was within the same piece.

**P2 · Folder outputs look like files** — Fixed. The three consolidated twins now use a new `entryOperationOutputSchema` labelled *Entry*, carrying the `.tag` kind discriminator and noting that the file-only fields (size, rev, content hash) are present only for files.

**P2 · Shared paths lack tests** — Added, following the existing per-piece vitest pattern. **25 tests** across `common.test.ts` and `output-schemas.test.ts`, covering exactly the paths you named: `unwrapEntry` success, per-entry `failure`, empty and missing entry arrays; the preview content-type branch; the inline-vs-async union schemas (each launch schema exposes `async_job_id` *and* the inline fields); and per-entry failure reasons on the status schemas.

One of those tests is a regression guard worth calling out: Dropbox's union discriminator is literally named `.tag`, and the data-selector path parser splits on `.` while discarding empty segments — so a bare `.tag` key silently resolves to `obj.tag` and renders empty. The schemas use the escaped `['.tag']` form, and the test now fails if a bare one is reintroduced.

Piece is now **40 actions — 26 `ai`, 7 `both`, 7 `human`**, build and lint clean, 25 tests green. The human-visible surface is unchanged.

OdaiAhmed99:
@greptileai please re-review — all five findings addressed in 5d12f61077.

OdaiAhmed99:
Live verification of the two fixes that changed runtime behaviour, run against a real Dropbox account on `5d12f61077` (unit tests cover the helper in isolation, but not that the header actually arrives):

**P1 · HTML previews become PDFs** — confirmed fixed with a discriminating pair:

| Input | `contentType` returned | File written |
|---|---|---|
| `doc.rtf` | `application/pdf` | `doc.pdf` |
| `data.csv` | `text/html` | `data.html` |

A `.csv` preview really does come back as HTML, so before this it was written as `.pdf` with HTML bytes inside — exactly the failure described. The content type is now also surfaced in the output so a downstream step can branch on it rather than infer from the filename.

**P2 · Folder outputs look like files** — `copy_dropbox_entry` on a folder returns:

```json
{"metadata": {".tag": "folder", "id": "id:...", "name": "AP-PREVIEW-COPY", "path_display": "/AP-PREVIEW-COPY"}}
```

No `size`, `rev` or `content_hash` — the file-only fields the previous schema advertised. `entryOperationOutputSchema` now leads with the `.tag` kind discriminator and notes those fields are file-only.

The other three fixes are metadata-only (`DESTRUCTIVE` classification, schema wiring, removing the `file_properties` actions) and don't touch `run()`, so they're covered by the build plus the 25 unit tests rather than a live run.

Test artifacts cleaned up.

